PR #47, opened 2020-03-20
Fix/wa data

zjrosen1:


zjrosen1:
closing since igor fixed in master already

Issue #72, opened 2020-03-21
Convert to using highchart library


virsaviya:
Currently, our charting is implemented using `chart.js` which is good for simple needs, but quickly comes up short for our more complex and nuanced charting needs.  Convert to using `highcharts` instead.

PR #74, opened 2020-03-21
Converting to highcharts lib

virsaviya:
https://github.com/covid-projections/covid-projections/issues/72: Converting to highcharts 
- should have parity with the existing chart
- introduced 'today'
- some updates on copy, confirmed by @maxwhenderson 
- moved 'point of no return' above chart

chart 2

TODO: remove `charts.js` and separate data logic into util


zjrosen1:
closing as this was merged from the rebased branch

PR #130, opened 2020-03-22
Cleanup some of the warnings from CRA

gsoltis:
I removed a bunch of unused imports and one unused dependency. I also fixed a few link attributes and one iframe attribute according to recommendations.


gsoltis:
There's a little more to do to get completely clean, I will tackle the rest as I have time.

PR #151, opened 2020-03-23
[WIP - do not review] Add CTA button to directly take user to their state page

rachelrf:
This PR does a couple of things:

1. States pulse slightly when the user is not hovering, to make it clear that the map is interactive. This might conflict slightly with the changes to the map colors, so happy to merge these changes after.

2. Adds a "See my state" button. When tapped, it prompts the user for their browser location, then reverse geocodes for their US state using [this api](https://www.bigdatacloud.com/geocoding-apis/free-reverse-geocode-to-city-api) which is free and unlimited for client-side use. I think this is important functionality, because it is difficult to tap on the smaller states on mobile. It's also a nice and clear call to action for people that won't read the instructional text. 

Here is what it looks like on desktop and mobile:  

This change: desktop
Screen Shot 2020-03-22 at 10 30 30 PM
  

This change: mobile
Screen Shot 2020-03-22 at 10 30 23 PM
  

The style guide has black buttons, but I wanted to make the primary CTA inviting so I made it blue. Happy to change that if anyone has design guidance. 

Style guide button: desktop
Screen Shot 2020-03-22 at 10 31 28 PM

### DIFF
--- a/src/components/Map/Map.js
+++ b/src/components/Map/Map.js
@@ -3,7 +3,6 @@ import '../../App.css'; /* optional for styling like the :hover pseudo-class */
 import USAMap from 'react-usa-map';
 import { Redirect } from 'react-router-dom';
 import Grid from '@material-ui/core/Grid';
-import { LocationOn } from '@material-ui/icons';
 import {
   STATES,
   STATE_TO_INTERVENTION,
@@ -12,7 +11,11 @@ import {
   INTERVENTION_EFFICACY_ORDER_ASC,
 } from 'enums';
 import { Legend, LegendItem } from './Legend';
-import { CallToAction } from './Map.style';
+import {
+  LocatingText,
+  MapInstruction,
+  MapInstructionMobile,
+} from './Map.style';
 
 const ROULETTE_LOADING_INTERVAL = 200;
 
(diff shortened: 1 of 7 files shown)


src/components/Map/Map.js
```diff
@@ -1,31 +1,151 @@
...
+      clearTimeout(rouletteAnimationTimer);
+    }
+  }, [])
+
+  if (redirectTarget) {
+    return <Redirect push to={redirectTarget} />;
+  }
+
+  /* mandatory */
+  let mapHandler = event => {
```
zjrosen1: this can be removed

src/components/Map/Map.js
```diff
@@ -1,31 +1,151 @@
...
+          setGeolocatedStateCode(stateCode);
+          resolve(stateCode);
+        })
+        .catch((error) => {
+          setSeeMyStateButtonHidden(true);
+        });
+    });
+  }
+
+  //function getClientLocation(callback) {
```
zjrosen1: remove

src/components/Map/Map.js
zjrosen1: or it was removed and looks like you added it back?

rachelrf:
Some changes in this commit:

1. Only expose the "See your state" functionality on mobile.

2. "See your state" is a link rather than a button

3. Removed the functionality to make states pulse when the user is not interacting with the map. This is too busy with the additional colors.

Screen Shot 2020-03-24 at 11 17 30 PM

igorkofman:
Hey Rachel - this is really *nice* but we're in the process of rolling out
a search dropdown based interaction to solve the mobile navigation problem
while also allowing for navigating to a given city / county (once that's
available).

It could be really cool to integrate your idea by offering a "current
location" row in the dropdown.  Could you connect with @Jeremy Lubin
<lubin.jeremy@gmail.com> and see if that makes sense / design it out?

rachelrf:
@igorkofman sg, I'll talk to Jeremy

PR #173, opened 2020-03-24
Add ESLint and prettier

rachelrf:
run _yarn lint_ to lint and _yarn lint-fix_ to automatically fix lint issues


package.json
```diff
@@ -28,10 +30,21 @@
...
+    "lint": "eslint src || printf '\\n\\e[31mRun lint-fix to fix\\e[0m\\n\\n'",
+    "lint-fix": "eslint --fix src",
+    "prettier": "prettier --check '**/*.{css,scss,js}' || printf '\\n\\e[31mRun prettier-fix to fix\\e[0m\\n\\n'",
+    "prettier-fix": "prettier --write '**/*.{css,scss,js}'"
   },
-  "eslintConfig": {
-    "extends": "react-app"
+  "lint-staged": {
+    "src/**/*.{css,scss,js}": [
+      "eslint --fix src"
```
szhu: Thanks! Can you also include the `prettier --write` command?

rachelrf: It's not super obvious but I've configured eslint with a plugin[1] which will run `prettier --write` before linting, and running with `--fix` will run it in write mode.

[1] https://github.com/covid-projections/covid-projections/pull/173/files#diff-e4403a877d80de653400d88d85e4801aR7

rachelrf:
Good call -- added the pre-commit hook.

I don't have access to the CI server/Netlify, so it's hard to set up a lint check (I can't see the logs when checks are failing). Do you know who could give me access? 

If I can get CI access and verify the checks there, I think this will be good to go.

package.json
szhu: oh cool! thanks for adding and pointing that out.

szhu:
Honestly, no idea about the ci either. Maybe ask #webapp-eng? That can probably be added in another PR actually.

I'm fairly new to this project, which is why I didn't hit approve (not sure if there's a sign-off process), but it looks good to me now!

zjrosen1:
rebased develop and merged on another branch

PR #181, opened 2020-03-24
Changes color coding for conditions.

caseymhunt:
Typically in the United States, users associate green to mean "good" and
red to mean "bad."

As such, at first glance the map is misleading and seems to indicate
those most affected are relatively safe by color-coding "Shelter in
place" locations with green.

~This switches green for red to more clearly indicate, at a glance,
conditions within a particular state according to trained user patterns.~

This alters the design to use a monochromatic blue; darkness indicating severity in a given region. 

# Before
image

# After (edited)
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,15803 @@
+{
+  "name": "covid-projections",
+  "version": "0.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.8.3",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
+      "integrity": "sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g==",
+      "requires": {
+        "@babel/highlight": "^7.8.3"
+      }
+    },
+    "@babel/compat-data": {
+      "version": "7.8.6",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.8.6.tgz",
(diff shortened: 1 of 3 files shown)


zjrosen1:
Hey @caseymhunt, thanks for the PR, will have to discuss with the team, but from my perspective "Limited Action" originally Red is the bad case since by no action you are at the highest risk of overloading hospitals which would become the most affected states

caseymhunt:
Totally! I understand the intent.

Once a user reads the legend, it becomes clearer why the colors were chosen but at first glance this doesn't convey and confuses users when attempting to reconcile the difference between what they expect and what they are seeing. 

I've validated this suspicion with a few on-lookers. So it appears to not just be me!

jredding:
hhm.. this is interesting. What if we explored non Red, Yellow, Green colors.. and inside went with blues - so of like "Freeze" in place?

caseymhunt:
I like that idea. Let me play with it and I'll update the MR. 

I will be using the following color paradigm: less severity (light) -> more severity (dark)

caseymhunt:
Updated & rebased. Using a monochromatic blue. Great suggestion, I think it looks really nice!

caseymhunt:
It's also worth noting that this change actually increases accessibility for the map. I've simulated a few types of color blindness and attached screenshots of problematic ones of the current design vs. the same tests in the new design. 

# ⚠️ Before
## Tritanomaly
image

## Achromatopsy 
image

# 👍 After
## Tritanomaly
image

## Achromatopsy 
image

jredding:
There were tests for color blindness is the last colorset - I am going to defere to the design team on this - I'm just chiming in to provide some feedback and utilize your momentum why we have it.  This is definitely the design team's decision though.

caseymhunt:
> There were tests for color blindness is the last colorset - I am going to defere to the design team on this - I'm just chiming in to provide some feedback and utilize your momentum why we have it. This is definitely the design team's decision though.

No worries. The tests I ran weren't outright failures, more like contrast concerns. Glad to hear y'all are running a11y tests though- that's very reassuring!

zjrosen1:
we are going to revisit this with our upcoming county updates, closing for now

PR #183, opened 2020-03-24
Refactor / simplify dataloading

igorkofman:
Prepares us for switchover to python generated models

### DIFF
--- a/src/utils/model.js
+++ b/src/utils/model.js
@@ -14,17 +14,30 @@ async function fetchAll(urls) {
   }
 }
 
+export const ModelIds = {
+  baseline: 0,
+  strictDistancingNow: 1,
+  weakDistancingNow: 7,
+  containNow: 2,
+};
+
 export function useModelDatas(location) {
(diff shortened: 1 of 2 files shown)


src/utils/model.js
```diff
@@ -14,17 +14,30 @@ async function fetchAll(urls) {
...
+      let urls = [
+        ModelIds.baseline,
+        ModelIds.strictDistancingNow,
+        ModelIds.weakDistancingNow,
+        ModelIds.containNow,
+      ].map(i => `/data/${location}.${i}.json`);
       let loadedModelDatas = await fetchAll(urls);
-      setModelDatas(loadedModelDatas);
+      setModelDatas({
+        baseline: loadedModelDatas[0],
```
zjrosen1: should the indexes also use ModelIds.baseline instead of just [0]

igorkofman: no - these are serially sequential as returned from fetchAll

zjrosen1: ahh yes makes sense

PR #184, opened 2020-03-24
Rework "point-of-no-return" call to action.

mikelehen:
* We now have dynamic phrasing depending on current intervention and whether hospital overload is expected within 3 months.  See https://docs.google.com/document/d/1MA-hitauZ82XCRI_XZUY0G1fdsa9WkAsiWPOVaKNOJI/edit#
* Refactored entire Callout element into a dedicated CallToAction component (which subsumed the existing LastDatesToAct component).

New version for okay states:
image

New version for not-okay states:
image

As implemented it does not ever recommend "Social Distancing".  This is open for discussion.

### DIFF
--- a/src/screens/ModelPage/ModelPage.js
+++ b/src/screens/ModelPage/ModelPage.js
@@ -33,7 +33,7 @@ const LAST_DATES_CALLOUT_COLORS = {
 const DAYS = 1000 * 60 * 60 * 24;
 const ONE_HUNDRED_DAYS = 100 * DAYS;
 
-const LastDatesToAct = ({ model, outOfBoundsDate }) => {
+const CallToAction = ({ model, intervention }) => {
   function formatDate(date) {
     const month = new Intl.DateTimeFormat('en', { month: 'short' }).format(
       date,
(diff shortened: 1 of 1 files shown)


src/screens/ModelPage/ModelPage.js
```diff
@@ -42,19 +42,49 @@ const LastDatesToAct = ({ model, outOfBoundsDate }) => {
...
   }
 
+  function suggestedIntervention() {
+    switch (intervention) {
+      case INTERVENTIONS.NO_ACTION:
+        return INTERVENTIONS.SHELTER_IN_PLACE;
+      case INTERVENTIONS.SOCIAL_DISTANCING:
+        return INTERVENTIONS.SHELTER_IN_PLACE;
+      case INTERVENTIONS.SHELTER_IN_PLACE:
+        // TODO: Are we okay recommending lockdown?
```
crwilcox: $0.02 is to hold off on this. Shelter in place seems to give the needed room in capacity.

src/screens/ModelPage/ModelPage.js
```diff
@@ -42,19 +42,49 @@ const LastDatesToAct = ({ model, outOfBoundsDate }) => {
...
+  let actionText, actionDateRange;
   if (
     !model.dateOverwhelmed ||
     model.dateOverwhelmed - new Date() > ONE_HUNDRED_DAYS
-  )
-    return <b>Greater than 3 months</b>;
+  ) {
+    actionText = `${intervention} projected to successfully delay hospital overload by greater than 3 months.`;
+  } else {
+    actionText = `To prevent hospital overload, you must implement ${suggestedIntervention()} by:`;
```
crwilcox: Is 'you' the right choice here? Mostly thinking most of the viewers don't have that personal authority. They can individually do so.  Perhaps "To prevent hospital overload, ${suggestedIntervention()} must be implemented by:"?

src/screens/ModelPage/ModelPage.js
mikelehen: I dropped it.  NY is the closest to possibly hitting this state (by our model).  If they do, they'll see "stricter intervention" and we can revisit.

src/screens/ModelPage/ModelPage.js
mikelehen: Good point. Changed.

jamestamplin:
LGTM!

PR #185, opened 2020-03-24
Slim header

adpirz:
image

image

### DIFF
--- a/src/components/Header/Header.js
+++ b/src/components/Header/Header.js
@@ -8,10 +8,6 @@ import { colors } from '@material-ui/core';
 
 import { Wrapper, HeaderSubCopy, BlackBar, RedBar } from './Header.style';
 
-const SUBHEADING_1_COPY_PART_1 = '';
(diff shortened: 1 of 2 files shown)


zjrosen1:
where did we land with showing the intervention in the header?

zjrosen1:
Also Jermey Lubin is looking into redesigning it

adpirz:
@zjrosen1 Jumana mentioned Jeremy was working on placing it elsewhere

crwilcox:
https://github.com/covid-projections/covid-projections/pull/184 is modifying the call to action as well.

mikelehen:
I merged #184.  It adds a mention of the current intervention to the call-to-action.  I think we probably still want to highlight it in other places.

One suggestion that came up last night and I'm not sure if/where it's being tracked / worked-on is to change the chart so the trend for the current intervention is highlighted (as if you were hovering over it) automatically.  @adpirz Were you going to look into that?

adpirz:
@mikelehen Will look into that today

PR #211, opened 2020-03-26
Removing GC Job

jamestamplin:
We hired a lawyer!


jamestamplin:
@mikelehen / @zjrosen1 - PTAL

mikelehen:
Well that was quick! 😄

PR #227, opened 2020-03-27
Add lines for when hospitals are overwhelmed at each level of intervention

gsoltis:
Feedback on UI is especially welcome, as that is not my strong suit. 

CA

With an intervention disabled:
CA_one_disabled

### DIFF
--- a/src/components/Chart/Chart.js
+++ b/src/components/Chart/Chart.js
@@ -37,67 +37,90 @@ const Chart = ({ state, county, subtitle, interventions }) => {
     type: 'area',
     data: data[0].data,
   };
-  if (data[0].dateOverwhelmed !== null) {
-    extraLines.push({
-      value: data[0].dateOverwhelmed,
-      className: 'highcharts-series-0-line',
-      label: {
-        text: 'Hospitals Overloaded',
-        x: 10,
-        y: 20,
-      }
-    })
-  }
   const socialDistancing = {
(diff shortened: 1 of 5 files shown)


zjrosen1:
Jeremy is already got a pr for this

Screen Shot 2020-03-26 at 9 40 02 PM

zjrosen1:
@jugglerjer you should look at the plot lines toggle though

jugglerjer:
Oooh, is this the ability to turn certain plot lines on and off?

igorkofman:
We don't want to show multiple lines at the same time, do we?  Can we only
show the yellow line on yellow graph hover?

gsoltis:
I didn't see an easy way to do it based on hover. There are mouse over and mouse out events, but i think the logic for when an area is highlighted is a little more complex. You can have states where nothing is hovered but everything is 'inactive', as well as obviously states where everything is active.

I decided to go with showing multiple lines to try to communicate the difference an intervention could have, but it definitely does make it more cluttered.

gsoltis:
Code for toggling is here: https://github.com/covid-projections/covid-projections/pull/227/files#diff-50783417df68929923f38ce05707e08cR104

It's relatively easy to do as long as the plotlines have names. The hard part is knowing when to do it. Hiding and showing a series is easy to watch, other state transitions are more difficult.

zjrosen1:
@igorkofman we are only showing today and the intervention line, and Jeremy already has the correct intervention line showing based on states current intervention

Issue #191, opened 2020-03-25
Make endorsements grid responsive


arjun-ltse:
The endorsements grid is a fixed width and breaks on mobile.

arjun-ltse:
Duplicate of #149

PR #261, opened 2020-03-29
Make endorsements responsive

### LINKED ISSUE #149: Endorsements component is not responsive
Maybe related to #58?

Page width is 900px
The endorsements section is 964px

image

It's ok if it's wider, but these should be fixed:
- If there is space on the sides, center the endorsements
- If there is no space on the sides (for example on mobile), make sure the endorsements component is smaller

szhu:
Fixes #149 (and #191)

Before:

  
Screen Recording 2020-03-29 at 12 42 AM

After:

Screen Recording 2020-03-29 at 12 43 AM


zjrosen1:
This breaks the mobile padding and title alignment also I thin the endorsements should align with the title like everywhere else

Screen Shot 2020-03-29 at 4 35 14 AM
Screen Shot 2020-03-29 at 4 35 04 AM

szhu:
@zjrosen1 fixed!

current website vs this branch:

image

dopry:
nm  LGTM.

szhu:
@dopry jw, how were you able to reproduce that?

dopry:
@szhu it was a stuck buid/caching on my side. It works....

igorkofman:
I checked out this branch and it doesn't seem right.
1. the top padding for the initial box is too small vs master
2. we lose all side padding when resizing - maybe you didn't push some commit? 

https://www.dropbox.com/s/qvgz2h7gvle1eeh/Screenshot%202020-03-30%2010.06.31.png?dl=0

szhu:
Hey @igorkofman -- I'm not seeing what you're seeing. Here's what I see when I compare https://covid-projections-git-szhu-endorsements-responsive.covidactnow.now.sh vs https://covidactnow.org/:

Screen Recording 2020-03-30 at 10 53 AM

Observe that:
- The top padding is exactly the same
- We never lose side padding

One issue I'm aware of is that the sign-up component right above it is supposed to be auto-resizing but it fails sometimes. (It is untouched by this PR.) Try loading the two versions of the page under identical initial window size conditions.

PR #271, opened 2020-03-29
uses npm instead of yarn

zjrosen1:
- update github actions
- adds package-lock
- removes yarn.lock


zjrosen1:
we gonna yarn for now might revisit this guy

PR #276, opened 2020-03-29
Add a /compare page for testing / debugging model data.

mikelehen:
This is an experimental page for internal usage but needs to live in the
website so we can share all of the chart loading / rendering code with the rest
of the website.

The idea is to let us easily compare all of the (state) graphs for e.g. staging
vs. prod or a modeler can point it at their local model files and compare them
as well.

I think this will be useful, but for now it's experimental.  We can iterate and
or abandon it depending on future needs.

One specific improvement I'd like to make is adding logic to do a simple
algorithmic comparison of each model and then sort them based on how different
they seem to be.

image

### DIFF
--- a/src/components/Charts/ModelChart.js
+++ b/src/components/Charts/ModelChart.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React from 'react';
 import { dateFormat } from 'highcharts';
 import moment from 'moment';
 import { snakeCase } from 'lodash';
@@ -131,7 +131,7 @@ const ModelChart = ({
     },
   };
 
-  const [options] = useState({
+  const options = {
     chart: {
       styledMode: true,
       height: '600',
@@ -253,7 +253,7 @@ const ModelChart = ({
       wuhanStyle,
       availableBeds,
     ],
-  });
+  };
 
   return (
     <ChartContainer>
--- a/src/screens/CompareModels/CompareModels.js
+++ b/src/screens/CompareModels/CompareModels.js
@@ -1,0 +1,150 @@
+import React, { useState } from 'react';
+import LazyLoad from 'react-lazyload';
+import Button from '@material-ui/core/Button';
+import FormControl from '@material-ui/core/FormControl';
+import Grid from '@material-ui/core/Grid';
+import TextField from '@material-ui/core/TextField';
+
+import ModelChart from 'components/Charts/ModelChart';
+import { STATES, STATE_TO_INTERVENTION } from 'enums';
+import { useModelDatas } from 'utils/model';
+import { buildInterventionMap } from 'screens/ModelPage/ModelPage';
+
+import {
+  Wrapper,
+  ModelSelectorContainer,
+  ModelComparisonsContainer,
+} from './CompareModels.style';
+
+export function CompareModels() {
+  // NOTE: The actual website doesn't handle CORS requests so we have to hit the S3 buckets for now.
+  const [leftUrl, setLeftUrl] = useState(
+    'http://covidactnow.org.s3-website-us-west-1.amazonaws.com/data/',
+  );
+  const [rightUrl, setRightUrl] = useState('/data/');
+
+  // We have separate state for the input field text because we don't want to actually update our
+  // URLs (and reload all the charts) until the input field loses focus (onBlur).
+  const [leftText, setLeftText] = useState(leftUrl);
+  const [rightText, setRightText] = useState(rightUrl);
+
+  // We need to let you force a refresh because you may be pointing at model files on a localhost
+  // webserver that have changed since the page was loaded.
+  const [refreshing, setRefreshing] = useState(false);
+
+  function refresh() {
+    setLeftUrl(leftText);
+    setRightUrl(rightText);
+    setRefreshing(true);
+  }
+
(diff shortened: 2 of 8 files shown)


src/components/Charts/ModelChart.js
```diff
@@ -119,7 +119,7 @@ const ModelChart = ({
     },
   };
 
-  const [options] = useState({
```
mikelehen: I'm not sure if this was some kind of "trick" or just a bug...  but as written, the chart wouldn't re-render when you adjusted the params passed in, since this state value was static and didn't update after initial render.

yarn.lock
```diff
@@ -2099,11 +2099,6 @@ abab@^2.0.0:
   resolved "https://registry.yarnpkg.com/abab/-/abab-2.0.3.tgz#623e2075e02eb2d3f2475e49f99c91846467907a"
   integrity sha512-tsFzPpcttalNjFBCFMqsKYQcWxxen1pgJR56by//QwvJc4/OUS3kPOOttx2tSIfjsylB0pYu7f5D3K1RCxUnUg==
 
-abbrev@1:
```
mikelehen: I don't understand why yarn.lock has so many diffs.  I get something similar if I just run `yarn install` on the develop branch...  Is it not deterministic or something? :-/

src/components/Charts/ModelChart.js
zjrosen1: @Fiskell you are familiar with this change it might affect the county level data

yarn.lock
zjrosen1: probably everyone running different node versions

mikelehen:
@fridiculous 
1. You can enter any URL at the top, so the simple thing to do is just run `python -m http.server` or `npx http-server` from e.g. your covid-data-model/results/test/ directory and then point the compare page at that localhost server.  Then you can run run.py and it'll update.  Although...  I just realized I should probably add a [Refresh] button to facilitate that workflow. 😁 
2. Sure.  Once this is deployed, anybody can just go to https://covidactnow.org/compare/ to get started.  I can add a note about running a local http server directly on that page, to help people make best use of it quickly.

yarn.lock
mikelehen: I wondered if that might be it.  I kinda' hoped it wasn't though. 😄  I'm on node 12.16.0 and yarn 1.21.1.  If you know of a "better" one, I can switch and add it to the onboarding doc...

src/components/Charts/ModelChart.js
Fiskell: @mikelehen It's a bug, we have it fixed on the counties branch and, @zjrosen1 We fixed it on the branch Jeremy was working on.  I thought that was out, is this branch behind develop?

mikelehen: It's not fixed in develop: https://github.com/covid-projections/covid-projections/blob/adf258e017d64b3043f24ce92d605b38fbded3ab/src/components/Charts/ModelChart.js#L128

But looks like it is fixed in counties-model-page (https://github.com/covid-projections/covid-projections/blob/fe5497f5d0c89a50985771a36e6fcf07783dcf7c/src/components/Chart/Chart.js#L244).

I didn't know about that branch (I saw a bunch of county-related stuff in develop, so I thought the county work was [largely] merged).  Do we know when it's going to land?  I don't want to create merge conflicts for it...

src/screens/CompareModels/CompareModels.js
```diff
@@ -0,0 +1,108 @@
...
+import {
+  Wrapper,
+  ModelSelectorContainer,
+  ModelComparisonsContainer,
+} from './CompareModels.style';
+
+export function CompareModels() {
+  // NOTE: The actual website doesn't handle CORS requests so we have to hit the S3 buckets for now.
+  const [leftUrl, setLeftUrl] = useState(
+    'http://covidactnow.org.s3-website-us-west-1.amazonaws.com/data/',
```
fridiculous: could we make this equivalent to the current site instead, regardless of what domain it is?

fridiculous: (i understand this may add a challenge since staging and prod are referring buckets, not the domain itself)

mikelehen: Sure, we can make it whatever we want.  Just using `/data/` as the URL should work to point at the current site...  But I'm not sure if that ends up making sense.  Right now left always points to master (by default) and right always points to staging (by default).  If we make left point at the current site instead, that means:
* when you go to https://staging.covidactnow.org/compare/ you will be comparing the staging models against the staging models.
* when you go to https://covidactnow.org/compare/ you will be comparing the prod models against the staging models.

So the staging version ends up being a no-op by default, which seems unhelpful and confusion.

mikelehen: @dopry updated this so that "right" always points to `/data/ (thanks!), so on prod it'll always use prod as the left (baseline) and the right will be your local environment.  But you can also run a local http server (e.g. from covid-data-model/) and point at that, so you don't have to mess with running covid-projections locally or copying model files over repeatedly.

mikelehen:
I think this is good to go.  I'll rebase/merge tomorrow if nobody objects.

PR #282, opened 2020-03-30
Moved CTA to the top of the page, added nudge to states w/o strict intervention

asciimike:
Per conversation with James/Max/JKT on making the page less confusing:

> A reporter from has Arizona raised a concern about the model that I think you should be aware and perhaps act on it.

> On the website, currently for Arizona to prevent hospital overload, stricter intervention must be implemented by:Apr 21 to Apr 26. The reporter was concerned that this adds a sense of “nothing needs to be done now, as there’s no urgency”.

> I understand how the model works, but would be great to have a “canned response”. 

image

### DIFF
--- a/src/screens/ModelPage/CallToAction/CallToAction.js
+++ b/src/screens/ModelPage/CallToAction/CallToAction.js
@@ -40,7 +40,13 @@ const CallToAction = ({ interventions, currentIntervention }) => {
       </div>
     );
     nudgeText = (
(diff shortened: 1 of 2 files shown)


asciimike:
Jk the new UI solves this. All this PR does is move it to the top:

image

zjrosen1:
With the updates headers this is no longer a part of this page anymore

PR #287, opened 2020-03-30
Initial commit of the privacy policy, including links from the ToU

asciimike:
Not ready to be merged until we have final approval (edit: we have it now), but I anticipate it will look almost identical to this. Note that this is not intended to get a nav link, but only be reachable directly at the URL (`/privacy`) or via the Terms of Use.

### DIFF
--- a/src/screens/Terms/Privacy.js
+++ b/src/screens/Terms/Privacy.js
@@ -1,0 +1,528 @@
+import React from 'react';
+import Typography from '@material-ui/core/Typography';
+
+import { Wrapper, Content } from './Terms.style';
+
+const Privacy = ({ children }) => {
+  return (
+    <Wrapper>
+      <Content>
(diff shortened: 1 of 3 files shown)


src/screens/Terms/Privacy.js
```diff
@@ -0,0 +1,528 @@
...
+        <Typography variant="h5" component="h1">
+          Contact Information
+        </Typography>
+        <Typography variant="body1" component="p">
+          To ask questions or comment about this privacy policy and our privacy
+          practices, contact us at:{' '}
+          <a href="mailto:privacy@covidactnow.org">privacy@covidactnow.org</a>.
+        </Typography>
+        <Typography variant="body1" component="p">
+          To register a complaint or concern, please contact us as the email
```
zjrosen1: "at" not "as"

asciimike: Fixed

PR #307, opened 2020-03-31
add screenshot script

zjrosen1:


scripts/page_screenshots/enums/index.js
```diff
@@ -0,0 +1,61 @@
+module.exports = {
+  HOME_PAGE: {
+    SELECTOR: '.us-state-map',
+    DATASET: '/',
+  },
+  STATE_PAGES: {
```
dopry: Could this be dynamically built from
```
import STATES from enum;
Object.keys(STATES).foreach((stateAbbrev) => {... } so we don't have to maintain a states list in 2 places?
```

PR #311, opened 2020-03-31
Initial take on renaming to Stay at Home, per JKT's copy change.

asciimike:
TL;DR: it's more friendly, Shelter in Place is more of a disaster response term.

I left "Shelter in Place" in `interventions-data.js`, since it seems to be used in CSVs. Unsure if this data makes it to the front end and therefore needs to change, or if we can leave it around.

As a side note, I left all the variable names in place, I just changed the string values.


mikelehen:
This aligns with what most states are calling it.  They're almost all "Stay At Home" orders rather than "shelter in place".

PR #322, opened 2020-03-31
add GA tag for sharing on FB/Twitter

eusden:
Hacked this together to start getting some data. LMK if anything should adjust, thanks!


zjrosen1:
@eusden there is also the per state page share button, should probably add there as well

eusden:
@zjrosen1 thanks for the tip!  Didn't know that existed, will update.

PR #324, opened 2020-04-01
Renamed /models to /faq, added some FAQ to /faq

asciimike:


zjrosen1:
i forgot who asked but it used to be called FAQ and we renamed it

asciimike:
Max was OK renaming it, and we need a place to put the other FAQ info

PR #328, opened 2020-04-01
Refactor to load all models in CompareModels component to prep for diffing/sorting

mikelehen:
@rachelrf Not sure if we want to submit this on its own or you just want to pull it into your work.


mikelehen:
Note that this will likely conflict with work in the counties feature branch, so we should be careful when/how we merge.

mikelehen:
Obsolete via #464 .

Issue #296, opened 2020-03-30
/compare: Reorder states based on comparing the projected "hospital overload" date.


mikelehen:
The idea is to prioritize the "most different" projections.  Sorting by the delta between the "hospital overload" dates would be an easy first approximation.  We may want more of these in the future, so we might want a "Sort By" combo with "Alphabetical", "Hospital Overload", etc. in it.

mikelehen:
Merged via #464.  Thank you @rachelrf and @zjrosen1!

PR #330, opened 2020-04-01
Add sort and filter to the model comparison page

rachelrf:
Fixes https://github.com/covid-projections/covid-projections/issues/296

Builds on top of https://github.com/covid-projections/covid-projections/pull/328 (with a couple of small fixes), massive thank you to @mikelehen  :)

This PR adds sorting (alphabetical, hospital overload) as well as filtering by intervention. If anyone has more ideas for useful controls I'm happy to add them. 

Screen Shot 2020-04-01 at 1 29 46 AM


rachelrf:
I do not know why that test is failing 😒but I can take a look tomorrow

src/screens/CompareModels/CompareModels.js
```diff
@@ -59,17 +86,85 @@ export function CompareModels({ match, location }) {
...
+
+  function getDateOverwhelmed(stateAbbr, models) {
+    const currentIntervention = STATE_TO_INTERVENTION[stateAbbr];
+    const interventions = models[stateAbbr];
+    let interventionToModel = interventionToModelMap(interventions);
+    let model = interventionToModel[currentIntervention];
+
+    if (currentIntervention === INTERVENTIONS.SHELTER_IN_PLACE) {
+      model = interventionToModel[INTERVENTIONS.SOCIAL_DISTANCING];
+    }
```
mikelehen, line 145: This logic is a bit surprising.  Is this how we do the "poor compliance" version of the intervention, I guess?  It's probably worth a comment and/or centralizing this somewhere so it doesn't get out of sync with the chart component...

src/screens/CompareModels/CompareModels.js
```diff
@@ -59,17 +86,85 @@ export function CompareModels({ match, location }) {
     });
   }
 
+  const [sortType, setSortType] = useState(SORT_TYPES.ALPHABETICAL);
+  const [filterTypeIntervention, setFilterTypeIntervention] = useState(
+    DEFAULT_INTERVENTION_FILTER,
+  );
```
mikelehen, line 92: FWIW we're down to 1 state that's on "limited action" and ~10 that are on "social distancing", so the usefulness of this filtering may decline. :-)

src/screens/CompareModels/CompareModels.js
```diff
@@ -59,17 +86,85 @@ export function CompareModels({ match, location }) {
...
   }
 
+  function sortAlphabetical(a, b) {
+    return a < b ? -1 : 1;
+  }
+
+  function sortByDateOverwhelmed(a, b) {
+    const overwhelmedDifferenceA = getDifferenceInDateOverwhelmed(a);
+    const overwhelmedDifferenceB = getDifferenceInDateOverwhelmed(b);
+    return overwhelmedDifferenceA > overwhelmedDifferenceB ? -1 : 1;
```
mikelehen, line 113: If they're the same, should we be returning `0` instead of `1`?  Not sure, but as-is the order may be slightly non-deterministic.

src/screens/CompareModels/CompareModels.js
```diff
@@ -119,72 +214,121 @@ export function CompareModels({ match, location }) {
...
 
       <StateComparisonList
-        left={leftUrl}
-        right={rightUrl}
+        states={states}
+        leftModels={leftModels}
+        rightModels={rightModels}
         refreshing={refreshing}
+        sortFn={sortFunctionMap[sortType]}
+        filterFn={filterByIntervention}
```
mikelehen, line 261: nit: I don't care much, but I think you could get rid of sortFn / filterFn props and just apply them to states directly, i.e.:
```
<StateComparisonList
  states={states.filter(filterFn).sort(sortFn)}
  ...
  />
```
Seems a little simpler to me, but this is fine too if you prefer it.

src/screens/CompareModels/CompareModels.js
rachelrf: I think I'll leave it because it future we may want to add more filter functionality (open to hiding it for now though). Thoughts?

mikelehen: Sounds good to me!

mikelehen:
Sorry, can you fix the remaining merge conflicts again? 😬 I'll try to get it reviewed / merged quickly once that's done so it doesn't get out of sync again.  Thank you!

src/components/Charts/ModelChart.js
```diff
@@ -26,116 +30,140 @@ const ModelChart = ({
...
-    [INTERVENTIONS.SHELTER_IN_PLACE]: interventions.distancing.now,
-  };
-
-  let model = interventionToModel[currentIntervention];
+  let interventionsModelIsValid = true;
+  let model;
 
-  if (currentIntervention === INTERVENTIONS.SHELTER_IN_PLACE) {
-    model = interventionToModel[INTERVENTIONS.SOCIAL_DISTANCING];
+  // In future we can do further model shape/schema checking
```
mikelehen, line 36: So this file has a lot more diffs than I expected and it's hard to figure out what's changing / why.  Can you confirm these are intentional as opposed to a screwy merge or something? Thanks!

rachelrf: Sorry -- this looks like more changes than it is. Let me leave a couple of comments within the file explaining what each change is.

src/components/Charts/ModelChart.js
```diff
@@ -26,116 +30,140 @@ const ModelChart = ({
...
+  try {
+    let interventionToModel = interventionToModelMap(interventions);
+    let worstCaseIntervention = getWorstCaseIntervention(currentIntervention);
+    model = interventionToModel[worstCaseIntervention];
```
rachelrf, line 40: this is pulling the logic for "worst case" social distancing:
```
  if (currentIntervention === INTERVENTIONS.SHELTER_IN_PLACE) {
    model = interventionToModel[INTERVENTIONS.SOCIAL_DISTANCING];
  }
```
 that's used in like 3 places in the app into the one helper function to keep things from getting out of sync

src/components/Charts/ModelChart.js
```diff
@@ -26,116 +30,140 @@ const ModelChart = ({
...
+  } catch (e) {
+    interventionsModelIsValid = false;
```
rachelrf, line 42: This just adds some error handling for bad model data, which was an issue I kept running into when I was testing. This should help it fail gracefully.

The changes below look more significant than they are, really they're just only evaluating all of these variables if the model data is valid to avoid everything breaking when it isn't.

src/components/Charts/ModelChart.js
```diff
@@ -26,116 +30,140 @@ const ModelChart = ({
...
+      marker: {
+        symbol: 'circle',
+      },
+    };
+  }
 
   const options = useMemo(() => {
+    if (!interventionsModelIsValid) {
+      return {
+        modelHasBadData: true,
```
rachelrf, line 162: This just adds a little explanation if the model has bad data and the chart won't be displayed.

mikelehen:
This got merged via #464.  Thank you!

PR #356, opened 2020-04-02
Jt blank state update

jamestamplin:


jamestamplin:
Would love if someone could add a link back to the state page. On line 319

I can't run locally and don't want to guess at how to do this :)

PR #391, opened 2020-04-02
add sumo

eusden:
Tested locally.  

Note: The form disappears once you've subscribed and sumo cookies you.  Don't waste an hour of your life like me trying to figure out why the form wasn't showing up.

Also note: it doesn't seem like the form re-renders on js render.  Probably a sumo limitation, not going to deal with it for now.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -49,6 +49,9 @@
       gtag('config', 'UA-160622988-1');
     </script>
 
+    <!-- Sumo.com embed -->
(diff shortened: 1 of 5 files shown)


zjrosen1:
@eusden this form disappears even if i just change the page so it only shows on initial page load and as soon as you go to a new page it is gone, not seeing anything obvious yet as to why thats happening

eusden:
@zjrosen1 yes that's the behavior I saw.  I assume the script is watching the span but not after the initial render.  I think it's not a blocker since it was committed to move to sumo.

asciimike:
Looks like: https://help.sumo.com/hc/en-us/articles/219100867-Website-platforms-incompatible-with-Sumo

> If your website is a single-page app we don't guarantee compatibility with all Sumo features.

> Because pages are typically loaded via javascript in single-page apps, anytime there is a URL change, your browser doesn't actually reload the new page, and Sumo isn't aware that the page has actually changed. This will cause certain features to not work properly, as the events taking place on your page are essentially invisible to Sumo.

PR #396, opened 2020-04-02
Add "See county data" text

szhu:
Note: Easier to review if you disable whitespace diffs:
https://github.com/covid-projections/covid-projections/pull/396/files?diff=split&w=1

Shows "See county data" text next to the map toggle, unless the search is focused or unless the screen is really narrow.

Normally:

> image

When search is focused:

> image

When screen is narrow:

> image

### DIFF
--- a/src/components/Header/SearchHeader.js
+++ b/src/components/Header/SearchHeader.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import useMediaQuery from '@material-ui/core/useMediaQuery';
 import { GlobalSelector } from 'components/MapSelectors/MapSelectors';
@@ -25,6 +25,7 @@ const SearchHeader = ({
 }) => {
   const history = useHistory();
   const isMobile = useMediaQuery('(max-width:1350px)');
+  const [isGlobalSelectorFocused, setIsGlobalSelectorFocused] = useState(false);
 
   const handleSelectChange = option => {
     let route = `/state/${option.state_code}`;
@@ -52,6 +53,7 @@ const SearchHeader = ({
             <GlobalSelector
               extendRight={true}
               handleChange={handleSelectChange}
+              onIsFocusedChanged={setIsGlobalSelectorFocused}
             />
           </SelectorWrapper>
           {isMobile && (
@@ -59,6 +61,7 @@ const SearchHeader = ({
               onClick={() => toggleMobileMenu()}
               isActive={mobileMenuOpen}
             >
+              {!isGlobalSelectorFocused && <>See county data&nbsp;&nbsp;</>}
               <MapIcon />
             </MapToggle>
           )}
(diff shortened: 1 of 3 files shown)


src/components/MapSelectors/GlobalSelector.js
```diff
@@ -213,46 +213,49 @@ const GlobalSelector = ({ handleChange, extendRight }) => {
...
-            <StyledInputIcon>
-              <SearchIcon />
-            </StyledInputIcon>
-            <StyledInput
-              {...getInputProps({
-                isOpen,
-                onFocus: () => {
-                  openMenu();
+      }) => {
+        onIsFocusedChanged(isOpen);
```
szhu, line 217: I considered also doing this by hoisting `Downshift` up, but that seemed like it would produce a huge diff. This solution to work fine.

src/components/Header/SearchHeader.js
```diff
@@ -52,13 +54,17 @@ const SearchHeader = ({
...
+              onIsFocusedChanged={setIsGlobalSelectorFocused}
             />
           </SelectorWrapper>
           {isMobile && (
             <MapToggle
               onClick={() => toggleMobileMenu()}
               isActive={mobileMenuOpen}
             >
+              {!isGlobalSelectorFocused && !isNarrowMobile && (
+                <>See county data&nbsp;&nbsp;</>
```
mikelehen, line 66: Not sure about "See county data" (since it just opens the state/county map; doesn't show you data directly).  Maybe "Find on Map" or just "Map"?

src/components/MapSelectors/GlobalSelector.js
```diff
@@ -69,7 +69,7 @@ const CountyItem = ({ dataset }) => {
   );
 };
 
-const GlobalSelector = ({ handleChange, extendRight }) => {
+const GlobalSelector = ({ handleChange, extendRight, onIsFocusedChanged }) => {
```
mikelehen, line 72: nit: handleChange / onIsFocusedChanged isn't very consistent naming.  And you named it "focused" but it's actually driven by the `isOpen` state.  Maybe `handleIsOpen`?

Minor and maybe we don't care at this point. Feel free to punt.

src/components/Header/SearchHeader.js
```diff
@@ -25,6 +25,8 @@ const SearchHeader = ({
 }) => {
   const history = useHistory();
   const isMobile = useMediaQuery('(max-width:1350px)');
+  const isNarrowMobile = useMediaQuery('(max-width:500px)');
+  const [isGlobalSelectorFocused, setIsGlobalSelectorFocused] = useState(false);
```
mikelehen, line 29: If you rename the property to be "open"-based, rename this too.

src/components/MapSelectors/GlobalSelector.js
szhu: `handleIsOpen` is clear and consistent! I'll use it.

src/components/Header/SearchHeader.js
szhu: I like "Find on Map" because it mirrors the language of the search bar, make it more obvious that it's an alternative to text search.

image

PR #417, opened 2020-04-05
County Embeds

### LINKED ISSUE #418: All counties in state show same "must act by" dates.
It looks like every county in a state is going to show the same “Stay at Home order must be implemented between [dates]” header with the same dates because we use the state model rather than the county model. 
https://github.com/covid-projections/covid-projections/blob/d482b4ea5fcf086414c251a4d017f08824ef6fdc/src/screens/ModelPage/ModelPage.js#L167

@igorkofman and @zjrosen1 don't think this is intentional. So we should probably fix it.

mikelehen:
* Available at /embed/us/county/53033 or /embed/us/wa/county/king_county
* EMBED share link works on county pages.
* embed.js updated to accept data-fips-id widgets.
* Fixes #418.

### DIFF
--- a/src/screens/Embed/Embed.js
+++ b/src/screens/Embed/Embed.js
@@ -1,6 +1,8 @@
-import React, { useState, useEffect } from 'react';
+import _ from 'lodash';
+import React, { useState, useEffect, useMemo } from 'react';
 import { useParams } from 'react-router-dom';
 import { STATES, STATE_TO_INTERVENTION } from 'enums';
+import US_STATE_DATASET from 'components/MapSelectors/datasets/us_states_dataset_01_02_2020';
 import Tabs from '@material-ui/core/Tabs';
 import Tab from '@material-ui/core/Tab';
 import '../../App.css'; /* optional for styling like the :hover pseudo-class */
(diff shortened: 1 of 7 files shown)


src/screens/Embed/Embed.js
```diff
@@ -64,8 +123,9 @@ export default function Embed() {
         <StateHeader
           location={location}
           locationName={locationName}
+          countyName={selectedCounty?.county}
           intervention={intervention}
-          interventions={interventions}
+          interventions={stateInterventions}
```
mikelehen, line 128: Note that this matches ModelPage, but may not be the behavior we want.
https://github.com/covid-projections/covid-projections/issues/418

zjrosen1:
We should fix this rounding issue in this pr as well 

Screen Shot 2020-04-04 at 9 36 28 PM

zjrosen1:
County state header, state url links to non embed version. Should be updated to go to embed/us/state if state is clicked on

Screen Shot 2020-04-04 at 9 37 43 PM

mikelehen:
Hah, good call.  Will do.

src/screens/Embed/Embed.js
```diff
@@ -20,32 +22,89 @@ import ChartsTab from './ChartsTab';
...
+  const datasForView = selectedCounty
+    ? modelDatasMap.countyDatas
+    : modelDatasMap.stateDatas;
+  let interventions = null;
+  if (datasForView && !datasForView.error) {
+    interventions = buildInterventionMap(datasForView);
+  }
+  let stateInterventions = null;
+  if (modelDatasMap.stateDatas) {
+    stateInterventions = buildInterventionMap(modelDatasMap.stateDatas);
```
zjrosen1, line 72: reminder we are not going to do this and fix the state header by just passing in interventions instead of stateInterventions

src/screens/Embed/Embed.js
```diff
@@ -20,32 +22,89 @@ import ChartsTab from './ChartsTab';
...
+  }
 
   useEffect(() => {
```
zjrosen1, line 75: should be moved into utils/model to centralize data fetching

src/screens/Embed/Embed.js
```diff
@@ -20,32 +22,89 @@ import ChartsTab from './ChartsTab';
...
+    summaryData = _.find(summaryData.counties, [
+      'fips',
+      selectedCounty.full_fips_code,
+    ]);
+    if (!summaryData) {
+      setMissingCounty(true);
+    }
+  }
+
+  if (missingCounty || modelDatasMap?.countyDatas?.error) {
```
zjrosen1, line 97: oh didn't know this operator ?. was supported already cool!

src/screens/Embed/Embed.js
```diff
@@ -98,3 +158,20 @@ export default function Embed() {
...
+  // NYC HACK.
+  if (['36047', '36061', '36005', '36081', '36085'].includes(fips)) {
+    fips = '36061';
+  }
+
+  const statesData = US_STATE_DATASET.state_county_map_dataset;
+  for (const state in statesData) {
+    const countiesData = statesData[state].county_dataset;
+    for (const county of countiesData) {
+      if (county.full_fips_code === fips) {
```
zjrosen1, line 172: consdier updating to String(county.full_fips_code) === String(fips)

src/utils/hooks.js
```diff
@@ -3,15 +3,23 @@ import { useLocation, useParams } from 'react-router-dom';
...
   // Check if we're embedded in an iFrame
   const { pathname } = useLocation();
-  const { id } = useParams();
+  const { id, countyId } = useParams();
   const isEmbed = pathname.includes('/embed');
   const protocol = window.location.protocol;
   const hostname = window.location.hostname;
+
+  let path = `us/`;
+  if (id) {
```
zjrosen1, line 12: this is not handling the countyFipsId path

zjrosen1: ahh i see this is just for copying the embed url and we don't need to support both in this case

src/screens/Embed/Embed.js
mikelehen: Yup!

src/screens/Embed/Embed.js
mikelehen: Fixed.

src/screens/Embed/Embed.js
mikelehen: Sounds good. Done!

src/utils/hooks.js
mikelehen: Correct.

PR #425, opened 2020-04-06
Updating state data with state level aggregated beds data from definitive health

ghop02:


mikelehen:
Sorry, can you explain exactly what the underlying change here was? It looks like this ends up increasing the # of hospital beds in almost every state by ~10-50%. With at least one exception. Wyoming lost ~20%. Is that what we’re expecting?

ghop02:
Going to close this - pushing a global data update shortly.  Per discussion in slack we'll be updating the beds to use the aggregates

Issue #368, opened 2020-04-02
Make color palette colorblind-friendly


jamestamplin:
*summary:*
Color blind / accessibility issues

*expected result:*

*actual result:*

*environment (device, browser, mobile / web):*

*link:*

*screenshot (you can copy/paste images directly in):*

igorkofman:
I think we're pretty attached to the stoplight metaphor as a way of clearly communicating status, but I gather that there are ways of tweaking the colors to make them a lot more recognizable / seperable (by eg. adding blue to the green, perhaps in other ways).  Can someone make a suggestion on tweaks?

andrewringler:
The current selection of colors on the homepage map is pretty close to being accessible, you have now: Red #F03147, Orange #F88722, Green #4FE6C1, Gray #E3E3E3

The Red/Orange/Green combination is fine, but the gray (of predictions not available) is too close to the green to be easily readable (for those who can only distinguish contrast well, not color, and for B&W printing).
See https://vis4.net/palettes/#/4|s|f03147,f88722,4fe6c1,e3e3e3|ffffe0,ff005e,93003a|0|0
![existing_colors](https://user-images.githubusercontent.com/691187/79020928-324d2600-7b48-11ea-9f0a-00289c7da7c1.png)

I would recommend just changing the color of Predictions not available to a darker color perhaps to: #fbfbfb

So you then have:
- Hospital overload in 0-4 weeks, Red #F03147
- Hospital overload in 4-8 weeks, Orange #F88722
- Hospital overload not likely, Green #4FE6C1
- Predictions not available, Gray #FBFBFB

See the new color accessibility analysis here: https://vis4.net/palettes/#/4|s|f03147,f88722,4fe6c1,fbfbfb|ffffe0,ff005e,93003a|0|0
![ColorProposa1](https://user-images.githubusercontent.com/691187/79021064-940d9000-7b48-11ea-9c1c-0691c7c416f7.PNG)

And on the map, the new colors look like this:
![NewColors_OnMap](https://user-images.githubusercontent.com/691187/79021130-c1f2d480-7b48-11ea-8210-1ad537159c19.png)

thatneat:
The Chroma.js tool seems to be happy when I just change the gray color to a lighter gray `#fbfbfb` and leave everything else the same ( `#f03147`, `#f88722`, `#4be6bf` ).
[...]

PR #434, opened 2020-04-06
More helpful and cleaner tooltips

thatneat:
I was finding it hard to use the graphs at first glance. I think these items will help:

* consistent phrasing and capitalization
* include date of overload on the overload flag
* include series name on hover to remove need to cross-reference colors (relevant to #368 )

image

### DIFF
--- a/src/components/Charts/ModelChart.js
+++ b/src/components/Charts/ModelChart.js
@@ -44,13 +44,14 @@ const ModelChart = ({
     let plotLineText;
     switch (currentIntervention) {
       case INTERVENTIONS.SHELTER_IN_PLACE:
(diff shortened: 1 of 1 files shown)


igorkofman:
Thanks for the suggestion!  A lot of this stuff is going through re-design daily so it's hard to engage with this.

thatneat:
That makes sense. Happy to make an updated version with the latest redesign, would that be OK?

thatneat:
happy to update with the latest branch, just let me know.
This feels like a pretty significant usability issue to me - friends I
shared this with also thought more clarity in the labels would help them
understand what the graphs were trying to say.

Issue #439, opened 2020-04-07
Use north-south-rotated SVGs of states


thatneat:
*summary:*
State maps in the right-side inset as well as in the little circle emblems, appear oddly rotated compared with what you'd expect for that state.

This is especially noticeable with Washington and Oregon which have significant east-west and north-south borders.

*expected result:*
States are oriented north/south in their own maps. On the state-by-state level, the Web Mercator projection is probably fine for all states except Alaska.

*actual result:*
States are oriented according to the geoAlbers projection (as they are shown in the map with the entire USA), which causes states outside of the center to be at a jaunty angle.

I initially thought this was a kind of nice bit of character, but then I realized it was more likely accidental.

*environment (device, browser, mobile / web):*
any

*link:*
https://covidactnow.org/us/wa

*screenshot (you can copy/paste images directly in):*
![image](https://user-images.githubusercontent.com/3341011/78621814-5652f380-7838-11ea-8232-0e7d1b4d5baf.png)

thatneat:
I was thinking of doing another pass to make sure the states are nicely
centered, and also fix the little icons that have the states incorrectly
rotated.

PR #443, opened 2020-04-07
Derotate state maps

thatneat:
refs #439

I just changed the projection in the state-specific maps, except for Alaska, so that states at the edges of the USA wouldn't be jauntily rotated.

I then clicked through the states and made sure they still fit inside the box reasonably (thus the scale changes).

Washington state before:
image

Washington state after:
image


zjrosen1:
@thatneat this is awesome but i clicked on some states like CA and MI and they were being cut off

thatneat:
oops, you're right, I missed a few! I just did a more thorough check of all the states.
The page doesn't render at all with Puerto Rico and Guam, but I think that's a different problem.

thatneat:
Ok this is now a huge mess with me trying to keep up with merging upstream changes - my dealing-with-upstream-changes-fu is not the strongest. I'll make a fresh PR.

PR #452, opened 2020-04-07
Make AppBar state always reflect URL

szhu:
Previously, clicking the nav links in the footer did not change which link was highlighted in the AppBar (top bar). See this example -- we are on `/faq` but the AppBar still has **Map** highlighted.

image

The issue was by the fact that only the AppBar's own links told AppBar to update state. This PR changes AppBar to update state based on the current URL, so it's always up to date no matter how the page was navigated to.

### DIFF
--- a/src/components/AppBar/AppBar.js
+++ b/src/components/AppBar/AppBar.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { useHistory, useLocation, matchPath } from 'react-router-dom';
 import Logo from 'assets/images/logo';
 import { useEmbed } from 'utils/hooks';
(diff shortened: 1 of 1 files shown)


src/components/AppBar/AppBar.js
```diff
@@ -22,31 +22,30 @@ import {
...
-    }
-
-    // We are on the state page, don't highlight a tab
-    if (defaultPanelIndex === -1) {
-      return false;
-    }
+function getPanelIdxFromLocation(location) {
+  let idx = Panels.indexOf(location.pathname);
+  return idx === -1 ? false : idx;
+}
```
szhu, line 30: tldr: These items were moved up out of `App` because the linter told me to.

Putting both of them inside the function gives the following warning:
image

Adding `getPanelIdxFromLocation` as an effect dependency caused the effect hook to run on every re-render (and also created a warning saying that that would happen).

I realized the moving both `Panels` and `getPanelIdxFromLocation` statically enforces that they won't change between renders.

PR #456, opened 2020-04-07
add email subscribe event

eusden:
add GA event for conversion goal @Jtensminger


eusden:
CI seems to have problem trying to render the map, but that should be unrelated to my PR.

PR #430, opened 2020-04-06
Introduce typescript to the codebase, begin porting some of the _easy_ files

ajpalkovic:

PR #480, opened 2020-04-08
Orient states in a more familiar manner in minimap (refs #439)

zjrosen1:


thatneat:
looks like the test failure here is fixed in #430.

thatneat:
I also submitted #491 in case we just want that fix, which also looks like it unblocks my other four open PRs.

hmm, when I say "four open PRs" for a repo that I'm not officially a contributor to, it raises some alarm bells. Is there a more constructive way I can work with your team? Happy to just keep doing community-style contributions but it also seems like we're tripping over each other a bit. I intend to be working on COVID-related stuff a bunch for the next few weeks, and this seems like a worthwhile project so I've been focusing on it. Also happy to focus my time elsewhere though if you think that'd be better.

thatneat:
This version is missing some key changes; #512 is more correct. 
My comment above still stands though! Would appreciate some more direction :smile:

Issue #487, opened 2020-04-09
Share buttons on county pages are nonspecific


thatneat:
## summary
share buttons only provide specific text and links on state pages, not county pages

found while working on #398

## Steps to reproduce

1. go to https://covidactnow.org/us/wa/county/king_county
2. hit one of the share buttons

### expected result

* link on the share button goes back to the page you were on
* link text talks about the county

### actual result

* link on the share button goes to covidactnow.org
* link text is generic

## environment (web)

web

all

## link

https://covidactnow.org/us/wa/county/king_county

Issue #398, opened 2020-04-03
Bring back the "back arrow" next to the logo on state / county pages


schlosser:
*summary:* The back arrow is good, but we should have our branding somewhere there

*expected result:*

Something like this (see upper left):

 ![with-logo](https://user-images.githubusercontent.com/2433509/78315193-5e561f00-752a-11ea-9d23-4783dd4d2fec.png)

thatneat:
Is the back arrow still around? I'm not seeing it anywhere.
![image](https://user-images.githubusercontent.com/3341011/78792398-eb073f80-7965-11ea-9e31-a0b94eacc46c.png)

schlosser:
Might have gotten replaced instead of added on. This ticket can become "add arrow back"

igorkofman:
I'm pretty sure it was a standard material ui back icon ala our search icon

PR #488, opened 2020-04-09
Bring back the back arrow on state pages. Refs #398

thatneat:
This doesn't fix county pages currently because `match` isn't set for county pages, but that's because of #487 which I will deal with separately.


thatneat:
test is failing on an upstream lint issue

PR #491, opened 2020-04-09
Lint fix for a bunch of open PRs

thatneat:
It looks like #430 includes this and other fixes, but in case we just want this so that the tests can stop failing on all the other PRs, feel free to use it.


thatneat:
aaand the tests are failing because the data is updated and there are apparently data-dependent e2e tests failing. Should I file an issue for that?

PR #492, opened 2020-04-09
Add pyseir state projections with hospitalization

erccarls:
Release of PySEIR models for states with hospitalization data to set initial conditions.


zjrosen1:
NaN is not considered valid JSON so any dataset that contains NaN is failing to load the app, consider updating to just be null or converting NaN to the string "NaN"

zjrosen1:
closing since this out of date already

PR #508, opened 2020-04-10
Models update every day, make the UI reflect that

asciimike:


src/components/Charts/ModelChart.js
```diff
@@ -346,7 +346,7 @@ const ModelChart = ({
               placement="bottom"
             >
               <span>
-                <strong>Last updated April 6th</strong>.{' '}
+                <strong>Last updated {new Date().toLocaleDateString()}</strong>.{' '}
```
zjrosen1, line 349: probably should keep the same date format this would be MM/DD/YYYY

asciimike: Well, it would be MM/DD/YYYY for countries that do that, and DD/MM/YYYY for others. I'd prefer to not do a bunch of nonsense to get it in the same format, so long as its updated properly every day.

PR #512, opened 2020-04-11
Orient states in a more familiar manner in minimap (refs #439)

thatneat:
Let's give this one more try. All 50 states should now be visible and correctly rotated.

Washington state before:
image

Washington state after:
image


zjrosen1:
Nice, I was going to get to this tonight, I'll click through all the states in a few!

thatneat:
Thank you!

PR #529, opened 2020-04-14
Model/4 14 2020/ship state inference

erccarls:
This release contains state level max a-posteriori estimates for state level inferred covid trajectories as the "Model 2" slot. We also bundle an update to the static scenarios for states.

Counties should not be shown yet and thus are not included.


erccarls:
Do not merge until final review by @maxwhenderson or @erccarls please.

PR #545, opened 2020-04-17
Move website to API snapshot 77.

mikelehen:
* Point website at https://data.covidactnow.org/snapshot/77/
* Regenerate calculated interventions files (based on snapshot).
* Update "Last Updated" date to automatically be based on version.json.
* Tweak calculated interventions script to download files in parallel (downloading serially from the snapshot was slow).


fridiculous:
yes 👏

PR #549, opened 2020-04-18
Fix county embeds (they were trying to show inference projections).

mikelehen:
We were overloading the countyName property of ModelChart as a signal for two things:
1. Should we show the county disclaimer?
2. Should we try to show inference projections?

The embeds were not setting countyName because they didn't want a disclaimer. But this resulted in them hitting the inference projections codepath, which doesn't work for counties.

I fixed this by:
1. Removing `countyName` since it's not necessary (we can find out if it's a county from the projections).
2. Using `interventions.isCounty && !isEmbed` to decide whether to show the county disclaimer or not.

### DIFF
--- a/src/components/Charts/ModelChart.js
+++ b/src/components/Charts/ModelChart.js
@@ -6,6 +6,7 @@ import { INTERVENTIONS } from 'enums';
 import LightTooltip from 'components/LightTooltip/LightTooltip';
 import Chart from './Chart';
 import { Typography } from '@material-ui/core';
+import { useEmbed } from 'utils/hooks';
 
 import {
   ChartContainer,
@@ -45,7 +46,7 @@ const condensedFormatIntervention = (intervention, optCase) =>
 const ModelChart = ({
   height,
   condensed,
-  countyName,
+  isCounty,
   interventions,
   currentIntervention,
   showDisclaimer,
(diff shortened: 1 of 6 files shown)


src/screens/Embed/ChartsTab.js
```diff
@@ -6,17 +6,16 @@ import { EmbedChartContainer } from './Embed.style';
 
 export default function ChartsTab({
   state,
```
zjrosen1, line 8: since state is no longer used this can be removed as well

src/components/Charts/ModelChart.js
```diff
@@ -45,7 +46,7 @@ const condensedFormatIntervention = (intervention, optCase) =>
 const ModelChart = ({
   height,
   condensed,
-  countyName,
+  isCounty,
```
zjrosen1, line 49: the Projections model as a concept of county, and we should probably just add a isCounty method on the projections model instead of passing this everywhere

zjrosen1: so would just be interventions.isCounty()

mikelehen: Ooh, nice!  Done.  I made it a field instead of a method to be consistent with the existing `hasProjections` field I noticed.  We could make them both methods if we want to avoid storing redundant state (or we could make them properties with getters!). 🤷

src/screens/Embed/ChartsTab.js
mikelehen: woops, good catch!

PR #569, opened 2020-04-21
Pull React code out of the model; use model in script

igorkofman:


scripts/calculated_interventions.js
```diff
@@ -1,102 +1,55 @@
...
+    stateSummaryData.counties_with_data.map(async fipsCode => {
+      try {
+        const countyProjections = await fetchProjections(stateAbbr, {
+          full_fips_code: fipsCode,
+        });
+        countyFipsData[fipsCode] = countyProjections;
+      } catch (ex) {
+        console.error(ex);
+      }
+    }),
```
goldblatt, line 25: huhh i guess would this be slow?

igorkofman: slower than reading from the filesystem, but all parallelized

PR #572, opened 2020-04-22
Limit concurrency of calculated-interventions script to prevent file descriptor issues.

mikelehen:
Forgot to send this out earlier.

I don't love using bluebird, but it's easy.


mikelehen:
@igorkofman worked around this differently...

PR #580, opened 2020-04-22
Tweak base green color to match logo green.

mikelehen:
Josh approves (and probably wishes I would have done more to consolidate):

> Yes! let’s consolidate all instances of green into one, as long as text readability is maintained.


igorkofman:
Needs to recalc the map coloring!

mikelehen:
@igorkofman Thanks!  Map is updated.

PR #582, opened 2020-04-23
Add and configure storybook

pnavarrc:
This PR adds and configure Storybook, and create some stories.

[Storybook](https://storybook.js.org) is a development tool that allows implementing components in isolation without adding any production dependencies. It generates a component catalog of _stories_, which are examples of the use of the components with hardcoded props.

This will make it easier for us to iterate the design of the charts and get early feedback as we implement new components.

image

## Example
I created 3 stories in this PR. Each example is just an exported instance of the component with hardcoded props. I usually put the stories in the same directory than the components, but in this case, a separate folder might be less disruptive.

**Example Story**: `ClaimStateBlock.stories.js`

```jsx
import React from 'react';
import ClaimStateBlock from '../components/ClaimStateBlock/ClaimStateBlock';

export default {
  title: 'ClaimStateBlock',
  component: ClaimStateBlock,
};

export const Nevada = () => (
  <ClaimStateBlock 
    stateId="NV"
    countyName=""
  />
);
```

## Running
During development,  you can run Storybook as follows. It has hot reloading 

```shell
$ yarn storybook
```

If we want to deploy Storybook, we can build a static version by running the following command, and then deploy the folder `storybook-static` 

```shell
$ yarn build-storybook
```

## Testing
- Pull the branch and run `yarn storybook`
- Storybook should open in your browser
- Check that the stories render the components correctly

## Notes
- Learn more about Storybook here https://storybook.js.org/docs/basics/introduction/


pnavarrc:
I'm going to fix the conflict and ask someone to merge, thanks @goldblatt !

PR #585, opened 2020-04-23
Use API v1 to load state / county timeseries projections.

mikelehen:
Column numbers are gone and accessing properties from the API schema is all type-checked. 😁 I _think_ everything is working other than county inference.

### DIFF
--- a/src/models/Projection.ts
+++ b/src/models/Projection.ts
@@ -1,27 +1,6 @@
-const COLUMNS = {
-  hospitalizations: 8 + 1,
-  beds: 11 + 1,
-  deaths: 10 + 1,
-  infected: 9 + 1,
-  totalPopulation: 16 + 1,
-  rt: 13 + 1,
-  rtStdev: 14 + 1,
-  date: 0 + 1,
-};
+import { RegionSummaryTimeseries } from 'api';
 
 const MS_PER_DAY = 1000 * 60 * 60 * 24;
-
-//intersection between lines connect points [a, b] and [c, d]
-function intersection(a: any, b: any, c: any, d: any) {
-  const det = (a.x - b.x) * (c.y - d.y) - (a.y - b.y) * (c.x - d.x),
-    l = a.x * b.y - a.y * b.x,
-    m = c.x * d.y - c.y * d.x,
-    ix = (l * (c.x - d.x) - m * (a.x - b.x)) / det,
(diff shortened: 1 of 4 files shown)


igorkofman:
This is great!

src/models/Projection.ts
```diff
@@ -49,98 +28,47 @@ export class Projection {
   daysSinceDayZero: number;
   hospitalizations: number[];
   beds: number[];
-  deaths: number[];
```
mikelehen, line 52: deaths and infected are only available as cumulative in the API.  We could calculate the deltas if we need them, but right now we don't.

src/models/Projection.ts
```diff
@@ -49,98 +28,47 @@ export class Projection {
...
-      // create the new date by adding the delta to the day prior to overwhelming.
-      const startDate = this.dates[startIdx];
-      const deltaSecs = dayDelta * MS_PER_DAY;
-      const estimatedTimeStamp = startDate.getTime() + deltaSecs;
-      this.dateOverwhelmed = new Date(estimatedTimeStamp);
-    }
+    const shortageStart =
+      summaryTimeseries.projections.totalHospitalBeds.shortageStartDate;
+    this.dateOverwhelmed =
+      shortageStart === null ? null : new Date(shortageStart);
```
igorkofman, line 71: this will be funky until we switch to datapoint per day (the old code was interpolating between the points) but since we're switching, seems fine!

mikelehen: Yep, agreed.  That's next on my list.

src/models/Projection.ts
```diff
@@ -49,53 +46,39 @@ export class Projection {
...
     this.isInferred = parameters.isInferred;
     this.durationDays = parameters.durationDays || null /* permanent */;
     this.delayDays = parameters.delayDays || 0;
     this.rt = null;
     if (this.isInferred) {
-      this.rt = data[data.length - 1][COLUMNS.rt];
-      this.rtStdev = data[data.length - 1][COLUMNS.rtStdev];
+      // TODO(michael): Remove 'any' cast once schema is updated.
+      this.rt = (locationTimeseries.projections as any).Rt;
+      this.rtStdev = (locationTimeseries.projections as any).RtCI90;
```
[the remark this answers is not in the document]
mikelehen: Removed for now!

PR #586, opened 2020-04-23
Prework to start using API v1.

mikelehen:
cc/ @goldblatt 

This PR should have no behavioral changes. It just lays groundwork, including converting some code into TypeScript. For the best reviewing experience, you can review the commits individually, in order.

* Generate TypeScript .d.ts files from the JSON schema files generated in the covid-data-model repo.  This will give us statement completion and type-safety when reading from the API.
* Add `src/api/index.ts` with `fetch` methods for fetching API endpoints and returning the typed interfaces.
* Minimally convert Projections to TypeScript so we can benefit from the types.
* Convert calculated-interventions.js to TypeScript so it's not broken by the Projections conversion.

To see where this is headed, take a look at #585


src/api/index.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/**
+ * Fetches the county timeseries for each available intervention and returns it
+ * as a map.
+ *
+ * NOTE: Counties don't always support the projected/observed intervention, so
+ * it can be null.
+ */
+export async function fetchCountyTimeseriesMap(
```
igorkofman, line 80: I'm not totally sold on having the interface be state and county aware (vs eg taking an AreaDescriptor).  We might start modeling specific multi-county areas to model cities better and it introduces some code duplication.  We might also have a US level model.

src/api/index.ts
```diff
@@ -0,0 +1,122 @@
+/**
```
igorkofman, line 1: not urgent, but we should chat about how we want to use index.ts files

igorkofman:
This is so great!

src/api/index.ts
mikelehen: Very open to adjusting.

In my last project we'd end up with things like `api/api.ts` which was a bit ugly.  I saw this project uses `index.ts` and tried to follow that pattern, but I may have done it wrong or it may not be a precedent we want to continue.

If you let me know specifically what you'd expect this file to be named, I can just change it.

src/api/index.ts
mikelehen: Thanks for the nudge.  I reworked this and it came out much cleaner.

src/api/index.ts
igorkofman: let's keep it and make chat in high bandwidth about a plan

PR #590, opened 2020-04-24
Implement Rt Chart

pnavarrc:
Implement the first 3 charts from phase 3. See the designs on [Figma](https://www.figma.com/file/UUCr3ZpwaPV4o1KuGdSCNl/Covid)

**Growth Rate**
image

**Positive Tests**
image

**Hospital Usage**
image

## Notes

- The functions `optionsRt`, `optionsPositiveTests` and `optionsHospitalUsage` calculate options for Highcharts, taking care of formatting and internal calculations.
- The style `ZoneChart.style` has classes for the 3 charts above. Note that for some chart styles, is not possible to use CSS, and for others is not possible to pass a hardcoded style, that's why they are mixed.

## Questions
- Where should we put the "today" point (70% of the width of the chart)? For now, I'm passing the end of the chart as a parameter, but we need to adjust it.
- We need to decide where to put the boundaries for low, medium, and high for each type of chart. It's likely that this will come from outside the charts (the entire page will be styled according to these boundaries), so we need to sync on that. They are parameters for now.

## TODO
- [ ] Update the tooltip - is not in the designs yet, what do we want in it?
- [ ] Disable fading the Rt range when hovering the Rt line
- [ ] Update the styles of the tick labels (font-family, font-size)
- [ ] Data format: refactor the input so it's closer to the API (epoch ms, the structure of the data) https://github.com/covid-projections/covid-projections/pull/590#issuecomment-619509965
- [x] Annotations: shouldn't be draggable (fixed) https://github.com/covid-projections/covid-projections/pull/590#issuecomment-619509965


adpirz:
🔥 🔥 🔥

mikelehen:
FYI- Merged #582.

src/components/Charts/GrowthRateChart.js
```diff
@@ -0,0 +1,138 @@
+import React from 'react';
+import Highcharts, { dateFormat } from 'highcharts';
+import HighchartsReact from 'highcharts-react-official';
+import { GrowthRateChartWrapper } from './GrowthRateChart.style';
+
+// Required for the area-range chart
+require('highcharts/highcharts-more')(Highcharts);
```
pnavarrc, line 7: **NOTE**: We should take a look at the bundle size before merging, if `highcharts/highcharts-more` is too heavy we can re-write the area range chart as stacked area charts and make the lower area invisible (a bit hacky, but should work)

src/components/Charts/GrowthRateChart.js
```diff
@@ -0,0 +1,138 @@
...
+    name: 'Rt',
+    type: 'spline',
+    data: data.map(formatRtData),
+    animation: false,
+    marker: {
+      enabled: false,
+    },
+    lineWidth: 4,
+    zones: RT_ZONES,
+  };
```
pnavarrc, line 105: **Question**: Do we prefer to pass visual parameters (colors, line width) as options to Highcharts or to use styled-components only? We could enable `styledMode` for Highcharts, but it has to be for the entire chart.

src/stories/data/rt-live-md.js
```diff
@@ -0,0 +1,17 @@
...
+export const data = [
+  { date: '2020-03-05', rt: 1.61, rt_c90: 0.5 },
+  { date: '2020-03-12', rt: 1.62, rt_c90: 0.4 },
+  { date: '2020-03-19', rt: 1.47, rt_c90: 0.3 },
+  { date: '2020-03-26', rt: 1.25, rt_c90: 0.2 },
+  { date: '2020-04-02', rt: 1.12, rt_c90: 0.16 },
+  { date: '2020-04-09', rt: 0.8, rt_c90: 0.09 },
+  { date: '2020-04-16', rt: 1.0, rt_c90: 0.12 },
+  { date: '2020-04-23', rt: 1.04, rt_c90: 0.8 },
+].map(formatDataItem);
```
pnavarrc, line 17: **Note:** Mocked data to have the 3 zones (low, medium, high) - taken from rt.live - Maryland and modified a bit to have a dip below 1.0.

pnavarrc:
I'm still working on it but would love some feedback on the coding style from @mikelehen, @igorkofman or @adpirz

src/components/Charts/GrowthRateChart.js
```diff
@@ -0,0 +1,138 @@
...
+
+const formatRtData = d => ({
+  x: getDateMs(d.date),
+  y: d.rt,
+});
+
+const formatRtRangeData = d => [getDateMs(d.date), d.rt_l, d.rt_u];
+
+const GrowthRateChart = ({ data }) => {
+  const serieRt = {
```
mikelehen, line 94: nit: `seriesRt` (and `seriesRtRange` below)?

src/components/Charts/GrowthRateChart.js
adpirz: Previously we've used styled-components (actually think it was just raw css). I prefer styles living next to their components, especially given how quickly our styling can change, so this works.

src/components/Charts/zoneUtils.js
```diff
@@ -0,0 +1,156 @@
...
+    },
+    yAxis: {
+      ...baseOptions.yAxis,
+      tickPositions: getTickPositions(minYAxis, maxYAxis, ZONES_RT),
+    },
+    series: [
+      {
+        className: 'ZoneChart__AreaRange',
+        type: 'arearange',
+        fillColor: '#efefef',
```
pnavarrc, line 75: **TODO**: Where can we get that color (or similar from?)

src/components/Charts/zoneUtils.js
```diff
@@ -0,0 +1,156 @@
...
+    },
+    yAxis: {
+      ...baseOptions.yAxis,
+      labels: {
+        formatter: function () {
+          return formatPercent(this.value);
+        },
+      },
+      tickPositions: getTickPositions(minYAxis, maxYAxis, ZONES_HOSPITAL_USAGE),
+    },
```
pnavarrc, line 143: **TODO**: Simplify this function a bit, it shares a lot of code with `optionsPositiveTests`

mikelehen:
Is there a trick to getting the Rt range to show up?  I just ran 'yarn storybook' and clicked ZoneChart=>Growth Rate Rt and I see:
image

No confidence interval range. :-/

pnavarrc:
> Is there a trick to getting the Rt range to show up?

I broke it briefly while I was cleaning up the code 😅 , is working again (it was a typo in the story) cc @mikelehen

src/components/Charts/GrowthRateChart.js
igorkofman: +1 to inline stiles

mikelehen:
I did a hacky integration of this with the backend API here: [PR](https://github.com/covid-projections/covid-projections/pull/600), [demo](https://covid-projections-git-mikelehen-test-charts-and-api.covidactnow.now.sh/us/wa)

In general everything was super smooth, and I think we should probably try to get your PR checked in soon so we can party on everything together, but FYI here are a few minor notes from integrating and playing with the graphs:
1. I noticed a funny "bug."  It looks like I can click on the "Overloaded" / "Sufficient", "Shrinking" / "Growing", etc. labels and drag them around on the graph.  Hopefully there's an easy way to make them stay-in-place? 😬 
2. I could _almost_ plug our dataset data directly into your graphs, but [not quite](https://github.com/covid-projections/covid-projections/pull/600/files#diff-b9eba8aa2c8a8b80c2ddfd8e6fbe4552R36).  There are two issues.  1. You seem to expect "ms since epoch" for the x values, whereas we are producing `Date` values. 2. I had to kinda' flatten our data from `{ x, y: { rt, low, high}` into `{x, y, low, hi}` for the rt-range data. Neither of these are a big deal but we might want to try to unify if it's easy.
3. I'm not sure how big of an issue this will be in practice, but right now, for some states, the projected hospital usage in the future goes way over 100% and all of the labels get kinda' squashed.  E.g. see the graph for [MA](https://covid-projections-git-mikelehen-test-charts-and-api.covidactnow.now.sh/us/ma) where it goes to 1600%.
4. I know you're planning to rework the tooltips, but right now they show e.g. "0.154" instead of "15.4%"

src/components/Charts/zoneUtils.js
```diff
@@ -0,0 +1,156 @@
+import { COLOR_MAP } from '../../enums/interventions';
```
goldblatt, line 1: I think this files is A+ should we move these to be a common enum and share also with the status indciator component?

src/components/Charts/zoneUtils.js
```diff
@@ -0,0 +1,156 @@
...
+    annotations: [
+      currentValueAnnotation(x, y, formatPercent(y)),
+      ...zoneAnnotations(endDate, minYAxis, maxYAxis, y, ZONES_HOSPITAL_USAGE),
+    ],
+    series: [
+      {
+        name: 'Hospital Usage',
+        data,
+        zones: ZONES_HOSPITAL_USAGE,
+      },
```
goldblatt, line 153: this data is available as const rtData = projection.getDataset('rt'); will get this data now

src/components/Charts/zoneUtils.js
pnavarrc: yes - I moved the definitions to enums, following your example from https://github.com/covid-projections/covid-projections/pull/599

src/components/Charts/utils.ts
```diff
@@ -0,0 +1,172 @@
...
+  name: string;
+  value?: number;
+}
+
+const getZoneLabels = (
+  endDate: Date,
+  minYAxis: number,
+  maxYAxis: number,
+  value: number,
+  zones: any[],
```
goldblatt, line 135: Zone[]?

src/components/Charts/zoneUtils.js
```diff
@@ -0,0 +1,145 @@
...
+const toHighchartZone = (zone, level) => ({
+  color: zone.color,
+  name: zone.label,
+  value: isFinite(zone.upperLimit) ? zone.upperLimit : undefined,
+  className: 'ZoneChart__Line',
+  labelClassName: `ZoneAnnotation ZoneAnnotation--${level}`,
+});
+
+const ZONES_RT = [
+  toHighchartZone(CASE_GROWTH_RATE.LOW, 'Low'),
```
goldblatt, line 30: I think you can use the enum values here? not sure the exact syntax

src/components/Charts/zoneUtils.js
```diff
@@ -0,0 +1,145 @@
...
+      },
+      tickPositions: getTickPositions(minYAxis, maxYAxis, ZONES_POSITIVE_RATE),
+    },
+    annotations: [
+      currentValueAnnotation(x, y, formatPercent(y)),
+      ...zoneAnnotations(endDate, minYAxis, maxYAxis, y, ZONES_POSITIVE_RATE),
+    ],
+    series: [
+      {
+        name: 'Positive Tests',
```
goldblatt, line 107: eventually think this should be a constant somewhere as well

src/components/Charts/utils.ts
```diff
@@ -0,0 +1,172 @@
...
+};
+
+interface Zone {
+  className: string;
+  color: string;
+  labelClassName: string;
+  name: string;
+  value?: number;
+}
+
```
goldblatt, line 129: oh what is the difference for this and the one in the enum

src/components/Charts/ZoneChart.style.js
```diff
@@ -0,0 +1,45 @@
...
+      fill: ${COLOR_MAP.RED.BASE};
+      stroke: none;
+    }
+    &.ZoneAnnotation--Medium rect {
+      fill: ${COLOR_MAP.ORANGE.BASE};
+      stroke: none;
+    }
+    &.ZoneAnnotation--Low rect {
+      fill: ${COLOR_MAP.GREEN.BASE};
+      stroke: none;
```
goldblatt, line 42: I wonder if these can use the enums

src/components/Charts/utils.ts
```diff
@@ -0,0 +1,172 @@
...
+  _.ceil(1.2 * axisMax, 1),
+];
+
+export const baseOptions: Highcharts.Options = {
+  title: {
+    text: undefined,
+  },
+  subtitle: {
+    text: undefined,
+  },
```
goldblatt, line 70: i am confused why these aren't just the title and subtitle of the chart, but that's moot

src/components/Charts/utils.ts
```diff
@@ -0,0 +1,172 @@
...
+const getZoneLabels = (
+  endDate: Date,
+  minYAxis: number,
+  maxYAxis: number,
+  value: number,
+  zones: any[],
+): Highcharts.AnnotationsLabelOptions[] =>
+  zones.map((zone: Zone, i: number) => {
+    const from = i === 0 ? minYAxis : zones[i - 1].value;
+    const to = zone.value || maxYAxis;
```
goldblatt, line 139: hmm fromValue and toValue?

src/components/Charts/zoneUtils.js
```diff
@@ -0,0 +1,145 @@
...
+const ZONES_RT = [
+  toHighchartZone(CASE_GROWTH_RATE.LOW, 'Low'),
+  toHighchartZone(CASE_GROWTH_RATE.MEDIUM, 'Medium'),
+  toHighchartZone(CASE_GROWTH_RATE.HIGH, 'High'),
+];
+
+const ZONES_POSITIVE_RATE = [
+  toHighchartZone(POSITIVE_TESTS.LOW, 'Low'),
+  toHighchartZone(POSITIVE_TESTS.MEDIUM, 'Medium'),
+  toHighchartZone(POSITIVE_TESTS.HIGH, 'High'),
```
goldblatt, line 38: same as above comment about the enums

src/components/Charts/zoneUtils.js
```diff
@@ -0,0 +1,145 @@
...
+];
+
+const ZONES_HOSPITAL_USAGE = [
+  toHighchartZone(HOSPITAL_USAGE.LOW, 'Low'),
+  toHighchartZone(HOSPITAL_USAGE.MEDIUM, 'Medium'),
+  toHighchartZone(HOSPITAL_USAGE.HIGH, 'High'),
```
goldblatt, line 44: same as above comment about the enums

src/components/Charts/utils.ts
pnavarrc: That's a good question, I feel like having that outside the chart gives us more flexibility in terms of styling, but we can reconsider if just using the title and subtitle of the chart would be easier.

src/components/Charts/utils.ts
pnavarrc: This one is what the chart needs to color the line by zones, we could rename `label` → `name` in the enum to avoid an unnecessary transformation, but the rest is necessary (we could rename `upperLimit` → `value`, but I think that's confusing, so I would prefer to keep it as `upperLimit`).

src/components/Charts/zoneUtils.js
pnavarrc: yes, let's fix it once we centralize the enums - this is only used for the tooltips (which need fixing too)

PR #593, opened 2020-04-24
Fix legend tooltip and other minor changes

### LINKED ISSUE #579: When you select one of the dark gray counties (have data but no model) the selected state is confusing
## summary
When you select one of the dark gray counties (have data but no model) the selected state is confusing.  In particular because we just make opacity 0.5 in CountyMap and it makes the dark gray look like the light gray.  

Prolly the thing to do is either outline it or make it even darker gray in this case.  
I'd take any fix that makes the currently selected county clear!

## expected result
You can tell which county you selected 

## actual result
https://covidactnow.org/us/or/county/douglas_county 

## environment
*(device, browser, mobile / web)*

## screenshot
*(you can copy/paste images directly in)*

### LINKED ISSUE #581: Remove "Predictions not available" from the map view
## summary

The legend "Predictions not available" is no longer relevant, as every state/county has predictions. Having the legend there might be confusing for the users (they might think it applies to the entire map).

## expected result

The legend "Predictions not available" shouldn't be visible if predictions are available for all the states in the map.

## actual result

The "Predictions not available" is showing, even if the predictions are available for all the states.

## environment

- Web, Chrome

## link

https://covidactnow.org/

## screenshot

image

jwparktom:
* Tooltip now shows on hover for map labels on the homepage; remove "Predictions not available" if all states have predictions (fix #581); change hover color for county maps (fix #579)

### DIFF
--- a/src/components/CountyMap/CountyMap.js
+++ b/src/components/CountyMap/CountyMap.js
@@ -63,12 +63,6 @@ const CountyMap = ({
                   <Geography
                     key={geo.rsmKey}
                     geography={geo}
-                    fillOpacity={
-                      selectedCounty &&
-                      selectedCounty.full_fips_code === geo.properties.GEOID
-                        ? 0.5
-                        : 1
-                    }
                     fill={fillColor}
                     stroke={'white'}
                     onMouseEnter={() => {
--- a/src/components/CountyMap/CountyMap.js
+++ b/src/components/CountyMap/CountyMap.js
@@ -48,6 +48,17 @@ const CountyMap = ({
           <Geographies geography={counties}>
             {({ geographies }) =>
               geographies.map(geo => {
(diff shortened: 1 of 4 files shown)


jwparktom:
cc @igorkofman

igorkofman:
The tooltips are great but the content in them is no longer relevant in the current version :/

igorkofman:
Could you restrict this change to just dropping the un-needed legend item?

jwparktom:
@igorkofman  Removed the tooltip and dropped the no-pred legend.

Hovering over Inyo County (selected):
Screen Shot 2020-04-26 at 6 13 01 PM

mikelehen:
I think this is obsolete now. Correct me if I'm wrong (and sorry it never got merged).

PR #594, opened 2020-04-24
Light blue selector for county

adpirz:
image
image

### DIFF
--- a/src/components/CountyMap/CountyMap.js
+++ b/src/components/CountyMap/CountyMap.js
@@ -48,18 +48,18 @@ const CountyMap = ({
           <Geographies geography={counties}>
             {({ geographies }) =>
               geographies.map(geo => {
+                const isSelected =
+                  selectedCounty?.full_fips_code === geo.properties.GEOID;
+
                 return (
                   <Geography
                     key={geo.rsmKey}
                     geography={geo}
-                    fillOpacity={
-                      selectedCounty &&
-                      selectedCounty.full_fips_code === geo.properties.GEOID
-                        ? 0.5
-                        : 1
-                    }
+                    fillOpacity={isSelected ? 0.5 : 1}
                     fill={
-                      countiesWithData.includes(geo.properties.GEOID)
+                      isSelected
+                        ? 'blue'
+                        : countiesWithData.includes(geo.properties.GEOID)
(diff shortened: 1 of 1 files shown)


mikelehen:
Nice!  Though I'm a little loathe to introduce a new color.  When I look at the examples you pasted, it's not at all obvious that blue represents selection and the other two colors convey information about the counties.  Iowa is kind of extra bad because so few counties have colors, but even with WA, the blue doesn't jump out at me as "the selected county."

image

I suck at UX though so I'm not sure exactly what I'd want to see.  Out of curiosity, did you play around with any other approaches? Like just use opacity to convey selection?  But maybe the two different tones of grey are problematic.  Can we just outline the selected county in solid black or something?

igorkofman:
@yvonne - our new designer who owns this area is +1 on outline.  can we try that?

adpirz:
So outline was my first option, but the problem with outline is there's no clean way to keep the outline on top, see results:

image
image

jwparktom:
What about darker gray color for counties with gray color and 0.5 opacity for counties with non-gray color? I tried it in #593

igorkofman:
@jwparktom - in our version a selected county with not enough data looks just like de-selected counties with no data at all:
https://www.dropbox.com/s/1rndkd9cav5vp4d/Screenshot%202020-04-26%2000.30.27.png?dl=0

Which is also true in prod

igorkofman:
@adpirz oh yea, that doesn't look so hot.  What about making selected county a dark gray either always (regardless of initial color) or if it's dark gray.

jwparktom:
@igorkofman Working for me... that's weird. Wrong link, maybe?
https://ibb.co/rtTVqf6
https://ibb.co/Np6LYBs

igorkofman:
https://covid-projections-git-fork-jwparktom-aesthetic-changes.covidactnow.now.sh/us/ca/county/inyo_county is where i was looking, but must be the wrong place!

jwparktom:
@igorkofman Oops, my bad :) Should be good now.

igorkofman:
Still not working for me on your branch, but I think visually it's probably right (hover state becomes permanent)

mikelehen:
I think this is obsolete now (we implemented a black outline).

PR #595, opened 2020-04-25
Allow 1-day-interval datasets.

mikelehen:
I have a [dummy draft PR](https://github.com/covid-projections/covid-projections/pull/596) that can be used to see what this actually looks like pointed at a 1-day-interval snapshot: https://covid-projecti-git-mikelehen-allow-1-day-intervals-use-1-a5a45e.covidactnow.now.sh/us/sd/county/meade_county (note the hospital overload lines up as expected now).

* We had logic to extract a 200-day duration by dividing it by 4 (to compensate for 4-day-interval), but in practice the API gives us way less than 200 days of data (it's ~4mo == 120 days), so I just removed the duration logic for now.
* Actually testing with a snapshot with 1-day-interval (e.g. snapshot 146) exposed a bug (https://github.com/covid-projections/covid-data-model/issues/315) which I have temporarily worked around (see fixZeros() function) so that it's not blocking.

Note: We may want to reintroduce some sort of duration logic, but I think for phase 3 we're mostly going to be looking at _historical_ durations instead of future-facing, so it would likely need to be reworked anyway.

### DIFF
--- a/src/models/Projection.ts
+++ b/src/models/Projection.ts
@@ -57,6 +57,9 @@ export class Projection {
     this.cumulativeDeaths = timeseries.map(row => row.cumulativeDeaths);
     this.cumulativeInfected = timeseries.map(row => row.cumulativeInfected);
 
+    this.fixZeros(this.hospitalizations);
+    this.fixZeros(this.cumulativeDeaths);
+
     this.cumulativeDead = this.cumulativeDeaths[
       this.cumulativeDeaths.length - 1
     ];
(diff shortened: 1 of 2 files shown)


src/models/Projection.ts
```diff
@@ -57,6 +57,9 @@ export class Projection {
     this.cumulativeDeaths = timeseries.map(row => row.cumulativeDeaths);
     this.cumulativeInfected = timeseries.map(row => row.cumulativeInfected);
 
+    this.fixZeros(this.hospitalizations);
```
igorkofman, line 60: a comment?

PR #600, opened 2020-04-26
Add Pablo's charts to ChartsHolder, plumbing in the API datasets.

mikelehen:
This is minimal, but a starting point for iteration (we need to deal with missing data, tweak the end date for charts, etc.).

### DIFF
--- a/src/components/LocationPage/ChartsHolder.tsx
+++ b/src/components/LocationPage/ChartsHolder.tsx
@@ -10,6 +10,13 @@ import NoCountyDetail from './NoCountyDetail';
 import ModelChart from 'components/Charts/ModelChart';
 import { Projections } from 'models/Projections';
 import { Projection } from 'models/Projection';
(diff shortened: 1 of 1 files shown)


src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -22,6 +29,27 @@ const ChartsHolder = (props: {
...
+      y: d.y?.rt,
+      low: d?.y.low,
+      hi: d.y?.high,
+    }));
+
+  const testPositiveData =
+    projection &&
+    projection
+      .getDataset('testPositiveRate')
+      .data.map(d => ({ x: d.x, y: d.y }));
```
pnavarrc, line 47: You probably don't need this map here, the data is not really being transformed

mikelehen: Ooops, haha.  Yeah, that was me being sloppy when removing the date transformation I had before..  Thanks!

igorkofman:
merged

PR #601, opened 2020-04-26
Plumb Phase 3 data from API through getDataset().

mikelehen:
I've created PR #600 to demonstrate this in action by combining with @pnavarrc's chart work and doing a quick/dirty integration into the model page. Demo here: https://covid-projections-git-mikelehen-test-charts-and-api.covidactnow.now.sh/

### DIFF
--- a/src/models/Projection.ts
+++ b/src/models/Projection.ts
@@ -26,7 +26,7 @@ export type DatasetId =
   | 'cumulativeDeaths'
   | 'cumulativeInfected'
   | 'rtRange'
-  | 'icuUsage'
+  | 'icuUtilization'
   | 'testPositiveRate';
 
 export interface RtRange {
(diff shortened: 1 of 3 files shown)


src/models/Projection.ts
```diff
@@ -92,22 +123,77 @@ export class Projection {
...
+  private deltasFromCumulatives(cumulatives: Array<number | null>) {
+    let previous = 0;
+    return cumulatives.map(value => {
+      const delta = (value || 0) - previous;
+      previous = value || 0;
+      return delta;
+    });
+  }
+
+  private calcIcuUsage(timeseries: Timeseries): number[] {
```
igorkofman, line 177: "Utilization" vs "Usage"?

src/models/Projection.ts
```diff
@@ -92,22 +123,77 @@ export class Projection {
...
+    return dailyPositives.map((positive, idx) => {
+      const negative = dailyNegatives[idx];
+      const total = positive + negative;
+      return total > 0 ? round(positive / total, 3) : null;
+    });
+  }
+
+  // Given a series of cumulative values, convert to a series of deltas.
+  // Treats null values as 0s.
+  private deltasFromCumulatives(cumulatives: Array<number | null>) {
```
igorkofman, line 168: Will this end up with a large first delta? Maybe we chop it off - thought would have to figure out how to keep alignment with dates?

src/models/Projection.ts
mikelehen: Sure. Changed.

src/models/Projection.ts
mikelehen: Yeah...  I was debating a little about that but decided to keep it simple for now since it doesn't really matter for test positive rate where we use this (since we calculate a percentage, the large initial delta is fine).

But this will break down as soon as we decide to chart daily cases or anything like that, so I went ahead and reworked it.  While I was at it, I added smarter null-handling just to make it more robust for when we do start throwing more data at it.

PR #607, opened 2020-04-27
Add copy text above charts.

mikelehen:
A few misc. TODOs left, but otherwise done.  Can be merged as-is to keep things moving, or I can make revisions tomorrow.

### DIFF
--- a/src/components/LocationPage/ChartsHolder.tsx
+++ b/src/components/LocationPage/ChartsHolder.tsx
@@ -17,6 +17,20 @@ import {
   optionsHospitalUsage,
   optionsPositiveTests,
 } from 'components/Charts/zoneUtils';
+import {
+  Level,
+  CASE_GROWTH_RATE,
+  POSITIVE_TESTS,
+  HOSPITAL_USAGE,
+  determineZone,
+} from 'enums/zones';
+// TODO(michael): These format helpers should probably live in a more
+// general-purpose location, not just for charts.
+import {
+  formatDecimal,
+  formatPercent,
+  formatInteger,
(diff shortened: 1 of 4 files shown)


mikelehen:
Note that I'm very open to having the text live somewhere else or otherwise improve the structure of this page (separate the charts into separate components / whatever).  But I didn't feel like making hard decisions like that right now.

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -104,4 +121,79 @@ const ChartsHolder = (props: {
   );
 };
 
+function caseGrowthStatusText(projection: Projection) {
+  const rt = projection.rt!;
+  if (rt === null) {
+    return 'No case load data is available.';
+  }
+  const level = determineZone(CASE_GROWTH_RATE, rt);
```
mikelehen, line 129: @igorkofman This is similar to what you'd want to do for the homepage recoloring.  We get the indicator out of the projection (`projection.rt` in this case) and then use `determineZone()` to see if it's `Level.Low`, `Level.Medium`, or `Level.High`.  You'd need to do that for each of the key indicators (see below 2 functions) and then compare to find the highest one, and then feed that to https://github.com/covid-projections/covid-projections/blob/2e0f17bda1fa7f412c584cbe76e843ec0ac45bea/src/enums/zones.ts#L23 to get the color.

mikelehen: Oh, and you'll need to either merge this PR first or branch off of it, since I added `currentTestPositiveRate` and `currentIcuUtilization`, which you'll need.

PR #608, opened 2020-04-27
Update to team bios and add missing header

schlosser:
Per JKT's feedback

### DIFF
--- a/src/enums/team.js
+++ b/src/enums/team.js
@@ -70,6 +70,11 @@ export default {
       link: 'https://www.linkedin.com/in/anna-blech-121277138',
     },
     {
+      name: 'Anya Lehr',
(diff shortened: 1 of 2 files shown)


src/enums/team.js
```diff
@@ -69,6 +69,11 @@ export default {
       title: 'Icahn School of Medicine at Mount Sinai',
       link: 'https://www.linkedin.com/in/anna-blech-121277138',
     },
+    {
+      name: 'Anya Lehr',
+      title: 'Ex-Vice Media Director of Operations',
```
mikelehen, line 74: nit: We seem to mix "Ex-" and "Former".  "Former" sounds a little better to me, but I don't have a strong preference.  Anyway, approving as-is, but maybe something to clean up the next time we're making edits.

PR #612, opened 2020-04-27
Colorize based on min status

igorkofman:


src/components/Map/Map.js
```diff
@@ -40,23 +40,23 @@ function Map({ hideLegend = false, setMobileMenuOpen, setMapOption }) {
         <Legend>
           <LegendItem
             key={'legend-3'}
-            title={'Cases are increasing'}
+            title={'No criteria met for reopening'}
             color={COLOR_MAP.RED.BASE}
             description={
               'Hospitals at elevated risk to be overloaded in the next 3 weeks. Act now to flatten the curve.'
```
mikelehen, line 46: This is outdated, but maybe it's no longer used, so it doesn't matter?

src/components/Map/Map.js
```diff
@@ -40,23 +40,23 @@ function Map({ hideLegend = false, setMobileMenuOpen, setMapOption }) {
...
+            title={'Some criteria met for reopening'}
             color={COLOR_MAP.ORANGE.BASE}
             description={
               'Hospitals at moderate risk to be overloaded in the next 3-6 weeks.'
             }
           />
           <LegendItem
             key={'legend-1'}
-            title={'Cases are decreasing'}
+            title={'All criteria met for reopening'}
```
mikelehen, line 59: FYI- I'm nervous about using language that's too strong here.  But I assume we'll review / fine-tune.

src/models/Projections.ts
```diff
@@ -7,6 +7,8 @@ import {
 } from '../enums/interventions';
 import { STATES } from '../enums';
 import { RegionSummaryWithTimeseriesMap } from 'api';
+import { CASE_GROWTH_RATE, POSITIVE_TESTS, HOSPITAL_USAGE, determineZone } from 'enums/zones';
+import { worstStatusColor } from 'enums/status';
```
mikelehen, line 11: Run prettier?

src/enums/status.ts
```diff
@@ -40,3 +40,17 @@ export const INFECTION_RATE_STATUSES: Statuses = {
...
+    return Level.HIGH;
+  } else if (statusList.indexOf(Level.MEDIUM) > -1) {
+    return Level.MEDIUM;
+  } else {
+    return Level.LOW;
+  }
+}
+
+export function worstStatusColor(statusList: Level[]): string {
+  return INFECTION_RATE_STATUSES[worstStatus(statusList)].color;
```
mikelehen, line 55: nit: Maybe the colors should be moved to a separate enum rather than tied to the specific indicator?

mikelehen: (just flagging for future cleanup)

Issue #609, opened 2020-04-27
Charts don't always have all 3 sections; section labels are not evenly padded


igorkofman:
## summary
https://staging.covidactnow.org/us/ca
https://www.dropbox.com/s/ig4hhreaqo77fb6/Screenshot%202020-04-27%2011.49.02.png?dl=0
https://www.dropbox.com/s/8ql38wue4fjai84/Screenshot%202020-04-27%2011.49.15.png?dl=0

## expected result
Charts always have 3 sections (and are always the same static height ideally).
Padding around labels is even 

## actual result
Charts don't always have all 3 sections; 
section labels are not evenly padded

## environment
*(device, browser, mobile / web)*

## screenshot
https://www.dropbox.com/s/ig4hhreaqo77fb6/Screenshot%202020-04-27%2011.49.02.png?dl=0
https://www.dropbox.com/s/8ql38wue4fjai84/Screenshot%202020-04-27%2011.49.15.png?dl=0

*(you can copy/paste images directly in)*

pnavarrc:
### 1. Charts don't always have all 3 sections

Currently, the maximum value shown in the chart is computed from the data, so if the max value for Rt is in the "medium" zone, the red region is not going to show.

https://github.com/covid-projections/covid-projections/blob/ca6589d9305144d19081062a02f93e61f6c53227/src/components/Charts/zoneUtils.js#L41-L51

We can fix this by forcing the red area to be the maximum between the data and the point where the orange and red regions would have the same height

```js
// Something like this should work
const [minYAxis, maxYValues] = const [minYAxis, maxYAxis] = roundAxisLimits(0, getMaxY(data));
const  mediumRegionHeight = CASE_GROWTH_RATE.MEDIUM.upperLimit - CASE_GROWTH_RATE.LOW.upperLimit;
const maxYAxis = Math.max(maxYValues, CASE_GROWTH_RATE.MEDIUM.upperLimit + mediumRegionHeight) 
```

## 2. Section labels are not evenly padded 

The section labels are vertically centered in their respective regions. The boundaries between the regions are defined in https://github.com/covid-projections/covid-projections/blob/develop/src/enums/zones.ts - I just put values to mimic the mocks, I don't think that the values are correct. The values might not be evenly spaced either.

@maxwhenderson do we know how we will determine what is low, medium, and high for each one of the charts (Rt, positive test rate, hospital usage rate)?

[...]

PR #613, opened 2020-04-27
Phase 3 Charts: Ensure high region is at least as large as the medium region.

mikelehen:
Fixes part 1 (Charts don't always have all 3 sections) of https://github.com/covid-projections/covid-projections/issues/609.

If you'd rather not merge this since you'll probably be mucking with this code later today, that's 100% fine. But it seems to work, so sending it your way for consideration.


pnavarrc:
Tested with mock data, looks good!

image

PR #614, opened 2020-04-27
Update utilization calculation

igorkofman:


src/models/Projection.ts
```diff
@@ -241,8 +241,15 @@ export class Projection {
     timeseries: Timeseries,
     lastUpdated: Date,
   ): Array<number | null> {
```
mikelehen, line 243: This code hurts my brain, but I think I convinced myself it makes sense. 😁 

Can you throw a comment above it, since this will probably be a mystery to anybody except you and maybe Eric?

```
    // The API gives us the beds in use *by covid*, and the total capacity *for
    // covid*, using an assumption that ICUs are usually 75% full with non-covid
    // patients. We've decided to show full ICU utilization (not just covid), so
    // we need to undo that assumption.
    // TODO(igor): Update this on the API side so we can undo this logic.
```

igorkofman: done

PR #616, opened 2020-04-28
Fix the y-axis to ensure that the 3 zones are visible

pnavarrc:
Testing a potential solution for #609 - Using fixed axis limits for zone charts.

- The y-axis limits are now defined in [`src/enums/zones.ts`](https://github.com/covid-projections/covid-projections/pull/616/files#diff-07c5caee9b2a8b10c99d1fd2f4d7d18a) to make them easier to modify in a centralized place
- I also adjusted the position of the current value label so it's visible even if the value is too close to the lower limit.
- Some code cleanup in `zoneUtils.js`

### Examples for Growth Rate
image

### Examples for Positive Tests and Hospital Usage
image

We still need to fix the tooltips when the series goes outside the chart, I can try to fix that this evening on a separate PR. 

cc @igorkofman @mikelehen


mikelehen:
I suspect this isn't exactly what we want, especially if we stop forcing the y-axis to start at `0` (the medium region will no longer end up in the middle necessarily).

Per Josh's feedback it sounds like maybe we want to move towards having the axes be a bit more static instead of dynamically determined based on the data.  Like maybe we need to be able to specify either specific limits for the yAxis or at least specific min/maxes for them.

pnavarrc:
How about we specify the lower and upper limits for each one of the zones here https://github.com/covid-projections/covid-projections/blob/develop/src/enums/zones.ts so the configuration of that value is centralized? If the boundaries of the charts are static, that would be the right place to put the values.

I can update this PR to do that, can @mikelehen confirm if that would work?

mikelehen:
@pnavarrc Sounds good, let's give it a try.

mikelehen:
Thanks for putting this together!  I didn't realize it would require quite so much code change. 😬 

I kinda' think we should put a pin in this, since offhand I am guessing this needs more iteration still, but I'm not sure exactly where we're headed.  Just browsing https://covid-projections-git-pablo-fix-zone-chart-extreme-values.covidactnow.now.sh/ I notice that:
1. The test positive label for WA (https://covid-projections-git-pablo-fix-zone-chart-extreme-values.covidactnow.now.sh/us/wa) is falling out of view:
image
2. We end up with a lot of empty space in the charts, which I'm not really a fan of.

Are there any charts on http://staging.covidactnow.org/ that we think are really problematic right now?  Else, maybe we just leave things as-is and revisit?  I think working on the tooltips or the new chart for future projections would probably be a better use of your time. (Thanks / sorry!)

src/enums/zones.ts
```diff
@@ -43,93 +44,106 @@ export const COLOR_ZONE = {
...
-// Case Growth Rate
 export const CASE_GROWTH_RATE: Zones = {
   [Level.LOW]: {
+    lowerLimit: 0.85,
     upperLimit: 1.05,
     name: 'Low',
     color: COLOR_MAP.GREEN.BASE,
   },
   [Level.MEDIUM]: {
+    lowerLimit: 1.05,
```
mikelehen, line 80: Not a huge fan of having to duplicate these values between the different zones.  I feel like we probably want to move the graph bounds out of the zone definitions. Or have like a wrapper interface.

```
interface ChartZones {
  lowerLimit: number;
  upperLimit: number;
  zones: Zone[];
}
```

pnavarrc: yeah, I thought about that as well - I don't have much time today so feel free to update the PR, if this fix can wait I can polish a bit more this evening or tomorrow

mikelehen:
FYI- I merged `develop` in so we get the smoothed out data, and I tweaked the zone limits (start hospital usage at 60%, cap test positive at 100%) so that we get a better idea of what this looks like with real data.

src/enums/zones.ts
pnavarrc: maybe we should have "chart" objects with title, subtitle, zones, upper and lower limits for the axes instead of just defining the zones and the chart data separately?

pnavarrc:
We are leaving the charts as they are for now - closing!

PR #617, opened 2020-04-28
Fixes missing current values for zone charts

pnavarrc:
Fixes an issue that was causing the current value for zone charts to be hidden.

The value for the current point annotation was taken as the last item of the passed `data` array. Since the future values have data, but it's undefined, the values for `x` and `y` were `undefined` too. This PR gets the last value such that both `x` and `y` are finite numbers.

Before
image

After
image
(I took a screenshot scaling down, it might look a bit funky, but the values are shown now).


mikelehen:
Thanks!

PR #625, opened 2020-04-29
Update the Future Projections Chart

pnavarrc:
Update the Future Projections Chart

Designs in [Figma](https://www.figma.com/file/UUCr3ZpwaPV4o1KuGdSCNl/Covid?node-id=2719%3A4830)

Screenshot: with and without inferred data
image

### Notes

I'm still trying to have this annotation outside the chart, but maybe we should have a plan B for it (Josh?)

image

I'm using the `projected` series to show the "previous estimate" section, let me know if we need to update it.

### TODO
- [x] Get colors from enums/variables
- [x] Fix the color of available beds and the available beds annotation
- [ ] ~~Add back the y-axis, adding only a tick for the available number of beds (with the correct label)~~ that would take too long, we are adding the label on top of the available beds line
- [x] Arrange the legends
- [x] Remove available beds from the legends
- [x] Adjust colors to match the mocks
- [x] Add a marker at the end of the "previous estimates" line
- [x] Fix the color of the markers so they match the series

### DIFF
--- a/src/components/Charts/ModelChart.js
+++ b/src/components/Charts/ModelChart.js
@@ -1,13 +1,11 @@
 import React, { useMemo } from 'react';
-import { dateFormat } from 'highcharts';
+import _ from 'lodash';
+import { dateFormat, numberFormat } from 'highcharts';
 import moment from 'moment';
 import { INTERVENTIONS } from 'enums/interventions';
 import LightTooltip from 'components/LightTooltip/LightTooltip';
 import ClaimStateBlock from 'components/ClaimStateBlock/ClaimStateBlock';
 import Chart from './Chart';
-import { isEmpty } from 'lodash';
-import { COLOR_MAP } from 'enums/interventions';
-import ReactDOMServer from 'react-dom/server';
 import { useModelLastUpdatedDate } from 'utils/model';
 import {
   ChartContainer,
@@ -20,21 +18,13 @@ import {
   CondensedLegendItemStyled,
 } from './ModelChart.style';
 
-function dateIsPastHalfway(dateToCompare, dateArray, itemKey) {
-  if (dateArray.length === 0) return true;
-  const midpoint = Math.floor(dateArray.length / 2);
-  // Enforce date objects in case these are date strings
-  return (
-    new Date(dateToCompare) >
-    new Date(itemKey ? dateArray[midpoint][itemKey] : dateArray[midpoint])
-  );
-}
-
 const formatIntervention = (intervention, optCase) =>
   `3 months of ${intervention.toLowerCase()}${optCase || ''}`;
 
 const condensedFormatIntervention = (intervention, optCase) =>
   `${intervention}${optCase || ''}`;
+
+const isFuture = d => d.x > new Date().valueOf();
 
 const ModelChart = ({
   height,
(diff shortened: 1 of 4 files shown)


src/components/Charts/ModelChart.js
```diff
@@ -19,23 +17,16 @@ import {
...
-function dateIsPastHalfway(dateToCompare, dateArray, itemKey) {
-  if (dateArray.length === 0) return true;
-  const midpoint = Math.floor(dateArray.length / 2);
-  // Enforce date objects in case these are date strings
-  return (
-    new Date(dateToCompare) >
-    new Date(itemKey ? dateArray[midpoint][itemKey] : dateArray[midpoint])
-  );
-}
+import { COLORS } from 'enums';
```
mikelehen, line 20: unused

src/components/Charts/ModelChart.js
```diff
@@ -282,21 +233,56 @@ const ModelChart = ({
           },
         },
       },
-      series: projection.isInferred
-        ? [noAction, projected, shelterInPlace, availableBeds]
-        : [noAction, shelterInPlace, availableBeds],
```
mikelehen, line 287: So I think maybe this needs to be:
```
      series: [noAction, projection.isInferred ? projected : shelterInPlace, availableBeds, previousEstimates]
```

Basically https://covid-projections-git-pablo-update-projections-chart.covidactnow.now.sh/us/wa/county/kittitas_county

should look a bit more like:
https://staging.covidactnow.org/us/wa/county/kittitas_county

Because we don't have inference results for many counties, so we have to show the static shelter-in-place (aka stay-at-home) scenario.  See https://blog.covidactnow.org/inference-projections-for-states/ for more details.

src/components/Charts/ModelChart.style.js
```diff
@@ -120,14 +118,63 @@ export const Wrapper = styled.div`
...
   .limited-action {
-    fill: ${props =>
-      props.projections.getChartSeriesColorMap().limitedActionSeries};
-    stroke: ${props =>
-      props.isInferred
-        ? props.projections.getChartSeriesColorMap().limitedActionSeries
-        : 'white'};
-    fill-opacity: 1;
+    .highcharts-graph {
+      stroke: ${COLORS.TRENDS_CONTINUE};
```
mikelehen, line 122: Hrm.  Should this be called "LIMITED_ACTION" or "RESTRICTIONS_LIFTED" instead of "TRENDS_CONTINUE"?

src/components/Charts/ModelChart.js
pnavarrc: Updated, see screenshot above

src/components/Charts/ModelChart.style.js
pnavarrc: updated (`LIMITED_ACTION`)

src/components/Charts/ModelChart.js
```diff
@@ -282,23 +233,67 @@ const ModelChart = ({
...
+    projection.isInferred,
     availableBeds,
-    projections,
     condensed,
-    dateOverwhelmedIsPastHalfway,
+    previousEstimates,
+    shelterInPlace,
   ]);
 
+  console.log({ options });
```
mikelehen, line 295: Stray console.log()? 😄

pnavarrc: ooops! 😅

PR #626, opened 2020-04-29
UX polish, add designers to about page

schlosser:


public/index.html
```diff
@@ -53,6 +53,8 @@
     <meta data-react-helmet="true" name="twitter:title" content="This model shows COVID projections for your state" />
     <meta data-react-helmet="true" name="twitter:description" content="The only thing that matters right now is the speed of your response" />
 
+    <link href="https://fonts.googleapis.com/css2?family=Source+Code+Pro:wght@700&display=swap" rel="stylesheet">
```
goldblatt, line 56: do we need to do this? i thought this would have already existed since we use it   font-family: source-code-pro, Menlo, Monaco, Consolas, 'Courier New',?

PR #635, opened 2020-04-29
Default to null if there is 0 for hospital Use data

goldblatt:
If there is a zero for the icu data, return null values (otherwise it should probably default to the average Capacity rate?)


goldblatt:
eg Nevada in staging

PR #637, opened 2020-04-29
Fix the y-axis limit for zone charts

pnavarrc:
Fixes the upper limit for zone charts

The upper limit for the y-axis was taken only from the data. When the max y value was below the upper limit of the medium region, the mid point between the max y value and the upper limit of the medium region, making it show below the upper limit of the medium region:

**Example: Before the change**
```
maxY = 0.4
mediumUpperLimit = 0.8
yHighLabel = (maxY + mediumUpperLimit) / 2 = 0.6
```

This changes the logic to take either the max of the data, or `1.5 * mediumUpperLimit`, to ensure that the high region is visible, also aligning the high region label

```
maxY = 0.4
maxYAxis = max(maxY, 1.5 * mediumUpperLimit) = 1.6
mediumUpperLimit = 0.8
yHighLabel = (maxYAxis + mediumUpperLimit) / 2 = 1.2
``` 

**Examples**
image

### DIFF
--- a/src/components/Charts/utils.ts
+++ b/src/components/Charts/utils.ts
@@ -70,8 +70,15 @@ export const getTickPositions = (
 
 export const getYAxisLimits = (minY: number, maxY: number, zones: Zone[]) => {
   const tickPositions = getTickPositions(minY, maxY, zones);
-  return [_.min(tickPositions), _.max(tickPositions)];
+  const minTickPosition = _.min(tickPositions) || minY;
+  const maxTickPosition = _.max(tickPositions) || maxY;
+  return roundAxisLimits(minTickPosition, maxTickPosition);
 };
+
+export const roundAxisLimits = (axisMin: number, axisMax: number) => [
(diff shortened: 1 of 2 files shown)


src/components/Charts/utils.ts
```diff
@@ -70,9 +70,16 @@ export const getTickPositions = (
 
 export const getYAxisLimits = (minY: number, maxY: number, zones: Zone[]) => {
   const tickPositions = getTickPositions(minY, maxY, zones);
-  return [_.min(tickPositions), _.max(tickPositions)];
+  const minTickPosition = _.min(tickPositions) || minY;
+  const maxTickPosition = _.max(tickPositions) || maxY;
```
pnavarrc, line 74: TS Lint complains that `_.min(tickPositions)` could be `undefined`

pnavarrc:
- Updated for Rt too
- Updated the PR to round the axis limits as well

PR #638, opened 2020-04-29
Add Outcomes back.

mikelehen:
Add outcomes table back and fix / cleanup issues.

The bulk of changes are in the last commit (and it may be useful to review it separately since I renamed files in the original commit).

The Outcomes table had a bunch of dead code from before we had inference, and there were a bunch of bugs (e.g. https://covidactnow.org/us/sd/county/pennington_county shows "Limited Action" twice, https://covidactnow.org/us/nd/county/mc_lean_county shows "Social Distancing" when really we want it to show "Stay at home" to mach the chart), etc.

It also had broken `timeHorizon` logic that I ripped out, and we had corresponding broken `Projection.cumulative...After(days: numbers)` functions that I replaced.

I think everything works, and is cleaner / simpler. But I apologize for the extra code changes.


mikelehen:
There's a phantom scrollbar on desktop I want to fix...

src/components/Charts/ModelChart.js
```diff
@@ -310,6 +311,17 @@ const ModelChart = ({
     <ChartContainer>
       <Wrapper projections={projections} isInferred={projection.isInferred}>
         <Chart options={options} />
+        <Outcomes
+          title="Predicted Outcomes"
+          projections={[projections.baseline, projections.primary]}
+          colors={[
```
goldblatt, line 317: huh the colors + the projections + the intervention? seems like we could just pass in projections object

mikelehen: So I ditched intervention, but kept projections / colors, since the logic for which ones to show should match the logic for the charts above, so it seems better to keep it in one place, though _really_ we should share the logic, but unfortunately that's not straightforward to do with the way things are set up right this moment.

src/components/Charts/ModelChart.js
```diff
@@ -307,20 +308,28 @@ const ModelChart = ({
       </ChartContainer>
     );
   }
+
+  // TODO(michael): I think this logic matches what we show in the chart, but we should
+  // unify this logic.
+  let outcomesProjections = [projections.baseline],
+    outcomesColors = [projections.getSeriesColorForLimitedAction()];
```
goldblatt, line 315: ooooo can we make these two different vars?

goldblatt: also why are they arrays?

goldblatt: oh i see nvm

src/components/Charts/ModelChart.js
```diff
@@ -307,20 +308,28 @@ const ModelChart = ({
...
-            projections.getSeriesColorForLimitedAction(),
-            projections.getSeriesColorForPrimary(),
-          ]}
-          asterisk={['', '*', '*', '**']}
-          timeHorizon={120}
-          currentIntervention={projections.stateIntervention}
+          title={`Predicted Outcomes by ${formatDate(
+            projections.primary.finalDate,
+          )} (90 days from now)`}
+          projections={outcomesProjections}
```
goldblatt, line 331: i just wish that these could come the projection instead of projections?

PR #645, opened 2020-04-30
changes the font of the axis to match other text

pnavarrc:
## Changes
- makes the font-family of the axes consistent with the rest of the text on the charts
- Adds a background for the current value annotation, so if it overlaps with the grid lines it doesn't look as bad

## Notes
- Highcharts renders axes as `<span>` (HTML) or `<text>` (SVG) elements, so I had to style both. It appears that chooses which one randomly 🤷‍♂️


pnavarrc:
sorry! Highcharts add inline styles, it's the only way to win over specificity 😅

PR #661, opened 2020-04-30
redo share styling

schlosser:
Our sharing and "claim state" modules were weirdly aligned on some large screen sizes. Fixed that and removed box shadow to align with our styling approach.

### DIFF
--- a/src/screens/HomePage/HomePage.js
+++ b/src/screens/HomePage/HomePage.js
@@ -26,7 +26,7 @@ export default function HomePage() {
                 padding: '0 1rem',
               }}
             >
-              <ShareBlock />
+              <ShareBlock centered={true} />
             </div>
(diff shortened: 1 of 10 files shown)


src/screens/HomePage/HomePage.js
```diff
@@ -26,7 +26,7 @@ export default function HomePage() {
                 padding: '0 1rem',
               }}
             >
-              <ShareBlock />
+              <ShareBlock centered={true} />
```
goldblatt, line 29: i think we always want it centered?

goldblatt: i feel like it hsould be centered on the location page

schlosser: updated to a wider layout

PR #670, opened 2020-05-01
Fix county name in ClaimStateBlock.

### LINKED ISSUE #658: County names not appearing correctly in "Claim" section
## summary
County names not appear in "Claim" section

## expected result
Do you work for San Bernardino County?

## actual result
Do you work for san_bernardino_county?

## environment
*(device, browser, mobile / web)*

## screenshot
Real-time models and metrics to understand where we stand against COVID  - Covid Act Now

mikelehen:
* Also refactored out a getFormattedCountyName() helper function.

Fixes #658

### DIFF
--- a/src/components/ClaimStateBlock/ClaimStateBlock.js
+++ b/src/components/ClaimStateBlock/ClaimStateBlock.js
@@ -9,12 +9,14 @@ import {
   ClaimStateButtonWrapper,
 } from './ClaimStateBlock.style';
 import { STATES } from 'enums';
+import { getFormattedCountyName } from 'utils';
 
 const EMAIL = 'gov@covidactnow.org';
 
-const ClaimStateBlock = ({ stateId, countyName }) => {
-  const stateName = stateId && STATES[stateId];
-  const headerName = countyName ? `${countyName}` : `the state of ${stateName}`;
+const ClaimStateBlock = ({ stateId, countyId }) => {
+  const headerName = countyId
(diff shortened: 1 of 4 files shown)


src/components/ClaimStateBlock/ClaimStateBlock.js
```diff
@@ -9,12 +9,14 @@ import {
...
 
 const EMAIL = 'gov@covidactnow.org';
 
-const ClaimStateBlock = ({ stateId, countyName }) => {
-  const stateName = stateId && STATES[stateId];
-  const headerName = countyName ? `${countyName}` : `the state of ${stateName}`;
+const ClaimStateBlock = ({ stateId, countyId }) => {
+  const headerName = countyId
+    ? getFormattedCountyName(stateId, countyId)
+    : `the state of ${STATES[stateId]}`;
```
virsaviya, line 19: Since `stateName = stateId && STATES[stateId];` is removed, is there a check to make sure that `STATES[stateId]` is truthy?  Perhaps add a default string.

mikelehen: So, yeah...  If we somehow get here with a bad `stateId` then it'll say "the state of undefined", but the old code would have too...

I could rework it to say "this state" in the case of a bad stateId, but it'd require extra code, and I'd actually rather it say "the state of undefined" since that'll get reported as a bug faster. 😁

So I'm leaving it as-is.  If you disagree, let me know.  I'm happy to continue to iterate.

PR #674, opened 2020-05-01
Fix issue affecting 506 counties that don't have actuals.ICUBeds data.

mikelehen:
Projection.totalICUCapacity and ProjectiontypicallyFreeICUCapacity are now nullable.

Without this, 506 county pages fail to load.

### DIFF
--- a/src/components/LocationPage/ChartsHolder.tsx
+++ b/src/components/LocationPage/ChartsHolder.tsx
@@ -262,7 +262,11 @@ function positiveTestsStatusText(projection: Projection) {
 function hospitalOccupancyStatusText(projection: Projection) {
   const icuUtilization = projection.currentIcuUtilization;
   const currentlyInICU = projection.currentICUPatients;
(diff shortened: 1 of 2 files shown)


src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -262,7 +262,11 @@ function positiveTestsStatusText(projection: Projection) {
 function hospitalOccupancyStatusText(projection: Projection) {
   const icuUtilization = projection.currentIcuUtilization;
   const currentlyInICU = projection.currentICUPatients;
-  if (icuUtilization === null || currentlyInICU == null) {
+  if (
+    icuUtilization === null ||
+    currentlyInICU == null ||
+    projection.typicallyFreeICUCapacity == null
```
mikelehen, line 268: Note that I added this null check to make TypeScript happy, but if this is null, the other two values are likely null too, so we would have returned the "no data available" text anyway.

PR #687, opened 2020-05-01
New about page design with photos and partners

schlosser:


src/components/StapledSidebar/StapledSidebar.style.tsx
```diff
@@ -50,12 +59,27 @@ export const StyledSidebarLink = styled.a`
...
+  padding-top: ${NAV_HEIGHT + INNER_PADDING - 1}px;
+  margin-top: -${NAV_HEIGHT - OUTER_PADDING}px;
+
+  &:before {
+    display: block;
+    width: 100%;
+    content: ' ';
+    transform: translateY(-${INNER_PADDING}px);
+    border-top: 1px solid rgba(0, 0, 0, 0.12);
+  }
```
mikelehen, line 78: Wowsers...  I'm not sure exactly what's going on here (my CSS is weak), but it seems pretty complicated for a section heading...  Consider adding a comment (optional)?

src/screens/HomePage/HomePage.js
```diff
@@ -21,9 +19,8 @@ export default function HomePage() {
       <main>
         <div className="App">
           <Content>
-            {!isMobile && <CriteriaExplanation />}
+            <CriteriaExplanation />
```
mikelehen, line 22: I assume this change is intentional, but it wasn't mentioned in the PR description so please confirm.

schlosser: Yep – we show above the map in all cases now. Before it was top on desktop only.

src/components/StapledSidebar/StapledSidebar.style.tsx
schlosser: Comment added! this positions the scroll-to links perfectly aligned with the headers.

PR #694, opened 2020-05-02
Brand new newsletter and sharing UI

schlosser:


src/components/ShareBlock/ShareBlock.tsx
```diff
@@ -68,90 +81,110 @@ const ShareBlock = ({
...
+            <Newsletter county={countyName} stateId={stateId} />
+          </NewsletterTextArea>
+        </ShareRowContentArea>
+      </ShareRow>
+      <ShareRow newsletter={false}>
+        <ShareRowContentArea>
+          <SocialTextAreaWrapper>
+            <SocialTextArea>
+              <ShareInstructionHeader>
+                Share data, save lives
```
goldblatt, line 107: this seems like a lot of div wrappers?

goldblatt:
hmm the embed didn't work in the preview from the home page?

goldblatt:
I think it's a lil confusing that the embed card we show to share and the actual embedded objects/shared data is different?

src/components/ShareBlock/ShareBlock.tsx
schlosser: Yeah, lots of flexing in alternating... but I don't think any are superfluous

schlosser:
Fixed this to remove embedding on the homepage, per Mike.

Michael is working on updating the actual embed images to match these "mockups" – will create a job to screenshot them.

PR #698, opened 2020-05-02
Make sure we don't provide hospital overload predictions if we don't have inference data.

mikelehen:
See https://covidactnow.org/us/ut/county/utah_county

Right now the chart and the text disagree since the text ends up using our "social distancing" static intervention to predict an overload date.

image

With my change, the text should match the chart better.
image

### DIFF
--- a/src/components/LocationPage/ChartsHolder.tsx
+++ b/src/components/LocationPage/ChartsHolder.tsx
@@ -169,8 +169,23 @@ function generateChartDescription(
   projection: Projection,
   noInterventionProjection: Projection,
 ) {
-  // TODO(sgoldblatt): figure out how to get people number data from projection
+  const noInterventionDate = noInterventionProjection.dateOverwhelmed;
+  const reopeningText = noInterventionDate
+    ? `happen in a slow and phased fashion. If all restrictions were completely lifted today, ` +
+      `hospitals are projected to become overloaded by ${formatDate(
+        noInterventionDate,
+      )}.`
+    : `happen in a slow and phased fashion.`;
+
+  if (!projection.isInferred) {
+    return (
+      `We don't have enough data to project whether hospitals are likely to become overloaded. ` +
+      `Reopening should always ${reopeningText}`
+    );
+  }
+
   if (projection.dateOverwhelmed) {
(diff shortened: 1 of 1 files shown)


src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -169,8 +169,23 @@ function generateChartDescription(
...
+    ? `happen in a slow and phased fashion. If all restrictions were completely lifted today, ` +
+      `hospitals are projected to become overloaded by ${formatDate(
+        noInterventionDate,
+      )}.`
+    : `happen in a slow and phased fashion.`;
+
+  if (!projection.isInferred) {
+    return (
+      `We don't have enough data to project whether hospitals are likely to become overloaded. ` +
+      `Reopening should always ${reopeningText}`
```
goldblatt, line 183: i guess it's maybe a lil confusing to me to say we don't have enough data to project whether hospitals are overloaded and then say hospitals are projected to become overloaded?

mikelehen: What do you suggest? 😁

Basically we don't have enough data to indicate how well in-place interventions are working, so we can't make a projection.  But we do know that if all restrictions are lifted, we'll return to a r-value of ~3 and things will go poorly. So we can project a hospital overload date for that.

mikelehen:
That said, this may be about to become irrelevant because I believe the new snapshot in https://github.com/covid-projections/covid-projections/pull/699 moves us to a world where we either have no projections or they're inference projections.  So we don't have to worry about these "only have projections for the static intervention scenarios" cases.

mikelehen:
Making this a draft for now.  If we move forward with #699, then this isn't necessary.

mikelehen:
I think this is officially no longer necessary as we've moved to a world where all projections are inference projections.  

https://covidactnow.org/us/ut/county/utah_county now says "Assuming current trends and interventions continue, Utah County, Utah hospitals are unlikely to become overloaded in the next 3 months." so the specific issue is addressed.

Now we can/should do a bunch of cleanup on the frontend to remove inference special-casing.

PR #702, opened 2020-05-03
Remove use of county_summaries/XX.summary.json files in the frontend.

mikelehen:
We used to need these files to render the county map without having to read all the projection files. But the calculated_county_interventions.json file took the place of that a while ago so it's been redundant for a while.

Additionally, the summary files are slightly wrong. E.g. they indicate Weber County, UT (fips 49057) should have data, but it does not. I'm not sure why the summary file is wrong, but we don't need it anymore, so who cares.

My next PR (to move the calculated interventions script onto the batch API will remove use of the summary file completely).

### DIFF
--- a/src/components/CountyMap/CountyMap.js
+++ b/src/components/CountyMap/CountyMap.js
@@ -12,24 +12,17 @@ import { FIPS_CODE_TO_CALCULATED_INTERVENTION_COLOR } from 'enums/interventions'
 import { COLOR_MAP } from 'enums/interventions';
 import { CountyMapWrapper, CountyMapLayerWrapper } from './CountyMap.style';
 
-const CountyMap = ({
-  selectedCounty,
-  setSelectedCounty,
-  fill,
-  stateSummary = {},
-}) => {
+const CountyMap = ({ selectedCounty, setSelectedCounty }) => {
   let { stateId } = useParams();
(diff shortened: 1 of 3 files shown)


src/components/CountyMap/CountyMap.js
```diff
@@ -12,24 +12,17 @@ import { FIPS_CODE_TO_CALCULATED_INTERVENTION_COLOR } from 'enums/interventions'
...
   const state = STATE_CENTERS[stateId];
   const counties = require(`./countyTopoJson/${stateId}.json`);
-  const countiesWithData =
-    stateSummary && stateSummary.counties_with_data
-      ? stateSummary.counties_with_data
-      : [];
   const [content, setContent] = useState('');
 
   const getFillColor = geoId => {
-    return FIPS_CODE_TO_CALCULATED_INTERVENTION_COLOR[geoId] || fill;
```
mikelehen, line 32: `fill` was actually never getting used because this function only gets called for the counties in the summary file and (because of how calculated-interventions.ts works) every county in the summary file is guaranteed an entry in the FIPS_CODE_TO_CALCULATED_INTERVENTION_COLOR map.

Also the passed in `fill` color was "wrong" (it was the state's intervention color), so I have removed the `fill` prop entirely.

PR #703, opened 2020-05-03
Move calculated_interventions.ts to use batch API endpoints.

mikelehen:
It takes ~20 seconds now and only downloads ~8 files instead of many thousands.

We could get it down to ~5 seconds and only 2 files with some refactoring.


mikelehen:
NOTE that this is based off of #702 because that needs to be merged first so I can delete fetchStateSummary() in here.

So we either need to merge this into #702 and then merge #702, or merge #702 and then change this to be based off develop before merging.

PR #712, opened 2020-05-05
about page updates

schlosser:
Small updates from JKT


schlosser:
[About page copy tweaks](https://trello.com/c/sy5kKblN/1-about-page-copy-tweaks)

PR #718, opened 2020-05-06
Fix county sharing and About page typo

schlosser:
I wasn't sure where to put the `County` type declaration.

### DIFF
--- a/src/components/LocationPage/ChartsHolder.tsx
+++ b/src/components/LocationPage/ChartsHolder.tsx
@@ -43,6 +43,7 @@ import {
   formatInteger,
 } from 'components/Charts/utils';
 
+// TODO(michael): figure out where this type declaration should live.
 type County = {
   county: string;
   county_url_name: string;
@@ -81,7 +82,7 @@ const ChartsHolder = (props: {
     <>
       {!projection ? (
         <NoCountyDetail
-          countyId={props.county.county_url_name}
+          countyId={props.county?.county_url_name}
           stateId={props.stateId}
         />
       ) : (
@@ -172,7 +173,7 @@ const ChartsHolder = (props: {
             </MainContentInner>
             <ClaimStateBlock
               stateId={props.stateId}
-              countyId={props.county && props.county.county_url_name}
+              countyId={props.county?.county_url_name}
             />
           </ChartContentWrapper>
           <ShareModelBlock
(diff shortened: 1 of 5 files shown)


src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -158,13 +172,13 @@ const ChartsHolder = (props: {
             </MainContentInner>
             <ClaimStateBlock
               stateId={props.stateId}
-              countyId={props.countyId}
+              countyId={props.county.county_url_name}
```
mikelehen, line 175: Looks like the state pages are broken. 😬  https://covid-projections-git-county-share.covidactnow.now.sh/us/wa

`props.county?.county_url_name` might fix it?

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -69,7 +80,10 @@ const ChartsHolder = (props: {
   return (
     <>
       {!projection ? (
-        <NoCountyDetail countyId={props.countyId} stateId={props.stateId} />
+        <NoCountyDetail
+          countyId={props.county.county_url_name}
```
mikelehen, line 84: This code probably only gets hit for counties, but to be safe, `props.county?.county_url_name` here too.

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -43,10 +43,21 @@ import {
...
+type County = {
+  county: string;
+  county_url_name: string;
+  county_fips_code: string;
+  state_fips_code: string;
+  state_code: string;
+  full_fips_code: string;
+  cities: string[];
+  population: string;
+};
```
mikelehen, line 55: Thank you for this! 😁  I'm not sure where it should go either.  I feel like we probably need a util/counties.ts file with this type and some helpers for dealing with the county dataset.  Maybe add a TODO, and I might try to grab it as tech debt later this week.

src/components/LocationPage/ChartsHolder.tsx
schlosser: Updated!

src/components/LocationPage/ChartsHolder.tsx
schlosser: Added a TODO!

src/components/LocationPage/ChartsHolder.tsx
schlosser: Done!

schlosser:
Added code to put the right name: `countyName, countyCode`

PR #721, opened 2020-05-06
Generate screenshots for social sharing

mikelehen:
* Added /internal/share/... pages that show the share image we want to capture.  E.g. https://covid-projections-git-mikelehen-share-screenshots.covidactnow.now.sh/internal/share-image/ for the home page and https://covid-projections-git-mikelehen-share-screenshots.covidactnow.now.sh/internal/share-image/states/wa for a state.
* Added scripts/share_images script that generates the share images by using `pageres` package to visit URLs and screenshot.
* A bit of refactoring:
  * Moved all internal pages under /internal/ for consistency / cleanup.
  * Refactored findCountyByFips() into common/locations.ts.


src/App.js
```diff
@@ -13,7 +12,9 @@ import Contact from 'screens/Contact/Contact';
 import Terms from 'screens/Terms/Terms';
 import Privacy from 'screens/Terms/Privacy';
 import Embed from 'screens/Embed/Embed';
-import CompareModels from 'screens/CompareModels/CompareModels';
+import AllStates from 'screens/internal/AllStates/AllStates';
+import CompareModels from 'screens/internal/CompareModels/CompareModels';
+import ShareImage from 'screens/internal/ShareImage/ShareImage';
```
mikelehen, line 17: I reorganized all the 'internal' pages to be under an `internal/` folder and changed the routing to match (see below) but left redirects in place so the old URLs work too.

package.json
```diff
@@ -111,6 +112,7 @@
     "lint-staged": "^10.1.3",
     "nyc": "^15.0.0",
     "prettier": "2.0.2",
+    "puppeteer": "^3.0.4",
```
mikelehen, line 115: This actually isn't used directly by our code (only indirectly via pageres), but I was worried about pageres using a pretty old version.

mikelehen:
@pnavarrc I _think_ you volunteered to review, so I'm sending your way but holler if you don't have time or don't feel like you're the right person.  I can ask somebody else.

scripts/generate_share_images/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  while (screenshots.length > 0) {
+    const pageres = new Pageres().dest(
+      OUTPUT_DIR,
+    );
+    const urls = [];
+    const batchSize = Math.min(PAGERES_BATCH_SIZE, screenshots.length);
+    for (let i = 0; i < batchSize; i++) {
+      const s = screenshots.pop()!;
+      urls.push(s.url);
+      pageres.src(BASE_URL + s.url, [OUTPUT_SIZE], {
```
pnavarrc, line 62: consider using `url-join` now that it was installed for the metatags PR

scripts/generate_share_images/index.ts
```diff
@@ -0,0 +1,75 @@
+// You can run via `yarn generate-share-images`
```
pnavarrc, line 1: Q: Do we need to start the web application before running the script? Consider adding instructions if that's the case

src/screens/internal/ShareImage/ShareImage.tsx
```diff
@@ -0,0 +1,82 @@
...
+interface ShareCardProps {
+  stateId?: string;
+  countyFipsId?: string;
+}
+
+const Header = () => {
+  return (
+    <>
+      <TitleWrapper>Real-time COVID metrics</TitleWrapper>
+      <LastUpdatedWrapper>Updated {formatDate(new Date())}</LastUpdatedWrapper>
```
pnavarrc, line 48: Maybe the date should come from the data instead?

scripts/generate_share_images/index.ts
mikelehen: Good suggestion, done!

scripts/generate_share_images/index.ts
mikelehen: Good call, done!

src/screens/internal/ShareImage/ShareImage.tsx
mikelehen: So that would be more accurate (it'd match when the data was generated), but I decided we want this date to be as fresh as possible, especially since Twitter / Facebook cache images for ~24hrs so people are going to be seeing ~day-old images already.  So I'm cheating a bit by just using the current date.

PR #731, opened 2020-05-08
Ignore the last 7 days of r(t) data when calculating official metric.

mikelehen:
Our r(t) indicator algorithm uses smooting such that new data points can
substantially shift the prior ~7 days of data. So we are not including it in
the calculation. @pnavarrc is additionally working on a change to the charts to
show this data as a dotted line.

### DIFF
--- a/src/models/Projection.ts
+++ b/src/models/Projection.ts
@@ -1,5 +1,12 @@
 import { RegionSummaryWithTimeseries, Timeseries } from 'api';
 
+/**
+ * We truncate (or in the case of charts, switch to a dashed line) the last
+ * seven days of r(t) data because it is susceptible to continued change as we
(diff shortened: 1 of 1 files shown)


src/models/Projection.ts
```diff
@@ -1,5 +1,12 @@
 import { RegionSummaryWithTimeseries, Timeseries } from 'api';
 
+/**
+ * We truncate (or in the case of charts, switch to a dashed line) the last
+ * seven days of r(t) data because it is susceptible to continued change as we
+ * get future data points.
+ */
+export const RT_TRUNCATION_DAYS = 7;
```
mikelehen, line 8: @pnavarrc I exported this constant so perhaps you can pull it in for your charts work?

PR #736, opened 2020-05-09
Align actuals timeseries with the projections timeseries.

mikelehen:


src/models/Projection.ts
```diff
@@ -166,12 +172,12 @@ export class Projection {
...
       const latestCurrentUssage = this.lastValue(
-        this.actualTimeseries.map(row => row.ICUBeds.currentUsageTotal),
+        this.actualTimeseries.map(row => row && row.ICUBeds.currentUsageTotal),
       );
       const latestUsageCovid = this.lastValue(
-        this.actualTimeseries.map(row => row.ICUBeds.currentUsageCovid),
+        this.actualTimeseries.map(row => row && row.ICUBeds.currentUsageCovid),
       );
-      return latestCurrentUssage - latestUsageCovid;
+      return latestCurrentUssage! - latestUsageCovid!;
```
goldblatt, line 180: lol

src/models/Projection.ts
```diff
@@ -229,12 +235,29 @@ export class Projection {
...
+  private getActualTimeseries(
+    actuals: ActualsTimeseries,
+  ): Array<CANActualsTimeseriesRow | null> {
+    if (actuals.length === 0) {
+      return actuals;
     }
+
+    const firstDate = moment(this.dates[0]);
+    const firstActualDate = moment(actuals[0].date);
+    const padFrontDays = firstActualDate.diff(firstDate, 'days');
```
goldblatt, line 247: is the data daily for sure now?

goldblatt: otherwise the dates might not line up is what i'm thinking jst seems tenuous

mikelehen:
I reworked this completely to basically walk the two timeseries in parallel and match trim / pad / fill in missing dates in the actuals.  I added a bunch of asserts and verified that the /all page doesn't crash, so I think it probably is working as intended. :-)

But I don't like this approach, so feel free to come up with something better.

PR #741, opened 2020-05-09
padding all the datasets

goldblatt:
fill both the timeseries and the actualTimeseries and the dates with every single date from timeseries day by day

### DIFF
--- a/src/models/Projection.ts
+++ b/src/models/Projection.ts
@@ -1,8 +1,14 @@
+import moment from 'moment';
+
 import {
   RegionSummaryWithTimeseries,
   Timeseries,
   ActualsTimeseries,
 } from 'api';
+import {
+  CANPredictionTimeseriesRow,
+  CANActualsTimeseriesRow,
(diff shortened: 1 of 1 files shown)


src/models/Projection.ts
```diff
@@ -251,6 +277,64 @@ export class Projection {
...
+      [date: string]: CANPredictionTimeseriesRow | CANActualsTimeseriesRow;
+    } = {};
+    ts.forEach((row: CANPredictionTimeseriesRow | CANActualsTimeseriesRow) => {
+      const ts_date = moment(row.date).toString();
+      dict[ts_date] = row;
+    });
+    return dict;
+  }
+
+  private fillDates(summaryWithTimeseries: RegionSummaryWithTimeseries) {
```
mikelehen, line 295: Okay, this method is pretty nice (fixing up _all_ the timeseries so they align).  Perhaps add a nice comment to complement it?  

I'd also be down to rename to something like alignTimeseriesByDates() or something, but up to you.

PR #745, opened 2020-05-10
Stop using /case_summary/ API endpoints in favor of API v1.

mikelehen:
https://trello.com/c/g44w2D1c/54-stop-all-use-of-v0-api

@ghop02 If you do me the favor of reviewing this, you can have the honor of deleting execute_summaries on the backend. 😛

### DIFF
--- a/src/screens/Embed/Embed.js
+++ b/src/screens/Embed/Embed.js
@@ -11,7 +11,7 @@ import '../../App.css'; /* optional for styling like the :hover pseudo-class */
 import LocationPageHeader from '../../components/LocationPage/LocationPageHeader';
 import ShareModelBlock from '../../components/ShareBlock/ShareModelBlock';
 
-import { useProjections, useStateSummaryData } from 'utils/model';
+import { useProjections } from 'utils/model';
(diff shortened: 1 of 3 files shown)


src/screens/Embed/Embed.js
```diff
@@ -58,28 +58,21 @@ export default function Embed() {
...
+  if (!primary) {
     return <span>'No data available for county.'</span>;
   }
 
-  const { cases, deaths } = summaryData;
+  const cases = primary.currentCumulativeCases;
+  const deaths = primary.currentCumulativeDeaths;
 
-  const totalPopulation = projections.baseline.totalPopulation;
+  const totalPopulation = primary.totalPopulation;
```
mikelehen, line 75: projections.baseline is the NO_INTERVENTION endpoint whereas primary will be OBSERVED_INTERVENTION, but it doesn't matter which one we use. I'm just using `primary` to be consistent.

mikelehen:
[Stop all use of V0 API](https://trello.com/c/g44w2D1c/54-stop-all-use-of-v0-api)

PR #748, opened 2020-05-11
Tweak site functionality to enable cleanup.

mikelehen:
This is a precursor to make my next PR possible which deletes ~750 lines of code (related to state interventions, social distancing, stay-at-home, etc.).  The changes are separated into 3 different commits if that's easier to review.

It contains 4 behavioral changes:

### 1. No longer show intervention in search dropdown.
**Before:**
image

**After:**
image

---
### 2. It removes "current intervention" from embeds:
**Before:**
image

**After:**
image

---
### 3. It fixes the legend on the embed projections tab:
**Before:**
image

**After:**
image

---
### 4. It removes the county blacklist for using inference projections.  
The 14 counties on the blacklist were the only places on the site where we were still showing a "stay at home" projection.  But I checked all of them, and the inference projection is virtually identical to the stay-at-home projection, so I don't think the blacklist is necessary any more.  If we want to keep it, I suggest we just treat these counties as no-data counties rather than showing stay-at-home.

Example: Hall County, NE
**Before:**
image

**After:**
image

### DIFF
--- a/src/components/Charts/ModelChart.js
+++ b/src/components/Charts/ModelChart.js
@@ -11,6 +11,7 @@ import {
   CondensedLegendItemStyled,
 } from './ModelChart.style';
 import Outcomes from '../Outcomes/Outcomes';
+import { COLORS } from 'enums';
 import { formatDate } from 'utils';
 
 const formatIntervention = (intervention, optCase) =>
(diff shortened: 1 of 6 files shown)


src/components/Charts/ModelChart.js
```diff
@@ -49,35 +50,11 @@ const ModelChart = ({
     },
     visible: !forCompareModels,
     condensedLegend: {
-      bgColor: projections.getChartSeriesColorMap().limitedActionSeries,
+      bgColor: COLORS.LIMITED_ACTION,
```
mikelehen, line 53: These condensedLegend.bgColor values are used in the embed legend and weren't using the same constants that the charts now use (as of phase 3), so they could be wrong in some cases.

src/components/Charts/ModelChart.js
```diff
@@ -49,35 +50,11 @@ const ModelChart = ({
...
-          ? condensedFormatIntervention(
-              INTERVENTIONS.SHELTER_IN_PLACE,
-              ' (lax)',
-            )
-          : condensedFormatIntervention(INTERVENTIONS.SOCIAL_DISTANCING),
-
-      bgColor: projections.getChartSeriesColorMap().socialDistancingSeries,
-    },
-  };
-
```
mikelehen, line 80: This was unused except it sometimes showed up in the embed legend (even though it shouldn't).

src/components/Charts/ModelChart.js
```diff
@@ -288,8 +265,8 @@ const ModelChart = ({
         <Wrapper projections={projections} isInferred={projection.isInferred}>
           <Chart options={options} />
           <CondensedLegend>
-            {[noAction, socialDistancing, shelterInPlace, availableBeds]
-              .filter(intervention => intervention.visible !== false)
+            {[previousEstimates, noAction, projected]
+              .filter(series => series.visible !== false)
```
mikelehen, line 269: socialDistancing is never shown in the charts, but we were still including it in the legend sometimes (see screenshot in PR description).

src/components/MapSelectors/GlobalSelector.js
```diff
@@ -48,10 +46,7 @@ const StateItem = ({ dataset }) => {
       <div>
         <div>{dataset.state}</div>
         <StyledResultsMenuSubText>
-          <span>
-            {intervention && <span>{intervention} • </span>}{' '}
-            {ShortNumber(dataset.population)} residents
-          </span>
```
mikelehen, line 54: Removing the intervention from the search dropdown text (see PR description).

mikelehen:
[Remove all code dealing with: isInferred (everything is inferred now). currentIntervention (and the checked-in interventions.json file too). The  “social distancing” projection (we don’t show it on any graphs anymore).](https://trello.com/c/8LiFjhee/55-remove-all-code-dealing-with-isinferred-everything-is-inferred-now-currentintervention-and-the-checked-in-interventionsjson-file)

PR #749, opened 2020-05-11
Clean up code related to currentIntervention, social distancing, and stay-at-home.

mikelehen:
This PR should have no functionality changes (with the exception of a slight performance improvement from only downloading 2 intervention API endpoints) and is mostly just deleting code that is no longer necessary.

In particular:
* We no longer use the current state intervention anywhere, so I deleted STATE_TO_INTERVENTION and the interventions.json file (actually I moved it to covid-data-public [here](https://github.com/covid-projections/covid-data-public/commit/7571e26cc95958f6f202ae04668d9e5d84333d94)).
* I removed the filter-by-intervention functionality from CompareModels (the /compare page).
* We no longer use social-distancing or stay-at-home projections, so I deleted code / constants that deal with them.
* Projections always include an inference projection, so we no longer need `isInferred` and can delete any code paths that deal with not having inference projections.
* I simplified Projection.getDataset() to just return the data instead of a `{data, label}` tuple, since we never use the label.
* We didn't use the StateSelector, CountySelector, or Status components so I deleted them.
* ModelChart.style.js had a bunch of unused chart styles, so I deleted them.

It may be easier to review the commits individually.


src/components/Charts/ModelChart.js
```diff
@@ -14,37 +13,23 @@ import Outcomes from '../Outcomes/Outcomes';
...
-  const data = [
-    projections.baseline.getDataset('hospitalizations'),
-    projections.distancingPoorEnforcement.now.getDataset('hospitalizations'),
-    projections.primary.getDataset('hospitalizations'),
-    projections.distancing.now.getDataset('hospitalizations'),
-    projections.baseline.getDataset('beds', 'Available hospital beds'),
-  ];
+  const noActionDataset = projections.baseline.getDataset('hospitalizations');
+  const projectedDataset = projections.projected.getDataset('hospitalizations');
+  const bedsDataset = projections.primary.getDataset('beds');
```
mikelehen, line 26: The data[] array seemed like unnecessary indirection so I just created direct variables for the datasets we use.

src/components/Charts/ModelChart.style.js
```diff
@@ -134,24 +131,6 @@ export const Wrapper = styled.div`
     }
   }
 
-  /* Stay at home */
```
mikelehen, line 137: All these styles are unused as of phase 3...

mikelehen:
[Remove all code dealing with: isInferred (everything is inferred now). currentIntervention (and the checked-in interventions.json file too). The  “social distancing” projection (we don’t show it on any graphs anymore).](https://trello.com/c/8LiFjhee/55-remove-all-code-dealing-with-isinferred-everything-is-inferred-now-currentintervention-and-the-checked-in-interventionsjson-file)

igorkofman:
i am soooo excited

PR #753, opened 2020-05-12
Charts Remake - ChartRt

pnavarrc:
Implements the Rt Chart using the @vx library

The idea is to create charts by putting together the chart building blocks; `Group`, `LinePath`, `Area`, `AxisLeft`, and `AxisBottom`, among others. The structure of a chart is something like:

```jsx
<svg width={width} height={height}>
  <Group left={marginLeft} top={marginTop}>
    <LinePath />
    <Area />
    <TextAnnotation />
    <CircleMarker />
    <AxisLeft />
    <AxisRight />
  </Group>
</svg>
```

Most building blocks of the chart need similar properties, the `data` array and functions to transform a data item into the x and y coordinates, in pixel units. 

### Scales
To transform data units into pixel units we use a _scale_. The _domain_ is the range of values the data can take (for the x-dimension in this case), and the _range_ is the corresponding range in pixels. For example, if our data goes from `2020-01-01` to `2020-01-10`, and the chart has width `200` (after discounting the margins), we can define the scale as:

```js
const xScale = scaleTime({
  domain: [new Date('2020-01-01'), new Date('2020-01-10')],
  range: [0, 200]
});
```

The `xScale` is a function that takes a `Date` and returns the corresponding number of pixels:

```
xScale(new Date('2020-01-01'));  // 0 
xScale(new Date('2020-01-05'));  // 100 
xScale(new Date('2020-01-10'));  // 200 
```

The `xScale` and `yScale` are used to calculate the `(x, y)` position of the points on a line, for example, when creating a `LinePath` for the R(t) chart, we would use them as:

```js
<LinePath
  data={data}
  x={d => xScale(d.x)}
  y={d => yScale(d.y.rt)}
/>
```

### Styling

To style the chart building blocks, we use styled-components, which allows us to keep the building blocks simple, without needing to pass properties to change how they look. We can alter how the `LinePath` looks by wrapping with styled-components as

```jsx
<Style.LineDashed stroke={"green"}>
  <LinePath data={data} {/* ... */} />
</Style.LineDashed>
<Style.LineDotted stroke={"red"}>
  <LinePath data={data} {/* ... */} />
</Style.LineDotted>
```

## Learn More
- https://vx-demo.now.sh/
- [Medium · Getting Started with vx](https://medium.com/vx-code/getting-started-with-vx-1756bb661410)


src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -108,6 +108,14 @@ const ChartsHolder = (props: {
...
                     for two weeks before reopening.
                   </Disclaimer>
+                  <hr />
+                  <ChartRt
+                    projectionDataset={projection.getDataset('rtRange')}
+                  />
+                  <Disclaimer metricName="infection growth rate">
+                    Most experts recommend an infection rate of less than 1.0
+                    for two weeks before reopening.
+                  </Disclaimer>
```
pnavarrc, line 118: **TODO**: Remove the old rt chart

src/components/Charts/Charts.style.ts
```diff
@@ -0,0 +1,117 @@
...
+    color: '#666',
+  },
+  annotations: {
+    color: '#333',
+  },
+  series: {
+    lineWidth: '4px',
+    defaultColor: '#000',
+  },
+};
```
pnavarrc, line 24: **TODO**: Grab the colors from a central location

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,256 @@
...
+  marginRight?: number;
+}) => {
+  const chartWidth = width - marginLeft - marginRight;
+  const chartHeight = height - marginTop - marginBottom;
+
+  const data = projectionDataset.data.filter(hasData);
+
+  // TODO(@pnavarrc): This is so TS doesn't complain
+  const minDate = d3min(data, x) || new Date('2020-01-01');
+  const maxDate = CHART_END_DATE;
```
pnavarrc, line 70: TODO: Is there a better way to avoid TS complaining about `minDate` might be `undefined`? cc @goldblatt

src/components/Charts/BoxedAnnotation.tsx
```diff
@@ -0,0 +1,33 @@
...
+  y,
+  children,
+  xPadding = 4,
+  yPadding = 6,
+}: {
+  x: number;
+  y: number;
+  children: React.ReactNode;
+  xPadding?: number;
+  yPadding?: number;
```
pnavarrc, line 16: **TODO**: Maybe we should make this only one padding (and hard-code it)

src/components/Charts/ChartRt.tsx
pnavarrc: **TODO**: I'm not sure if we need to force `CHART_END_DATE` anymore, we could just calculate how much space we need and extend the timeframe internally. cc @mikelehen

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+  // Tooltip
+
+  const { tooltipData, tooltipOpen, showTooltip, hideTooltip } = useTooltip();
+
+  const onMouseOver = (
+    event: React.MouseEvent<SVGPathElement, MouseEvent>,
+    d: any,
+  ) => {
+    // @ts-ignore - typing bug
```
pnavarrc, line 111: **TODO**: I might have fixed this, confirm and remove the comment

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,256 @@
...
+    const coords = localPoint(event.target.ownerSVGElement, event);
+    showTooltip({
+      tooltipLeft: coords?.x || 0,
+      tooltipTop: coords?.y || 0,
+      tooltipData: d,
```
pnavarrc, line 116: **TODO**: TS is complaining about `x` and `y` potentially being `undefined`

src/components/Charts/Charts.style.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export const ChartContainer = styled.div`
+  /* TODO(@pnavarrc): This negative margin breaks the auto-size of the chart */
+  @media (min-width: 996px) {
+    margin-left: -3rem;
+  }
+`;
```
pnavarrc, line 31: **TODO**: This breaks the width of the chart a bit, review

src/components/Charts/BoxedAnnotation.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { useRef } from 'react';
+import { Group } from '@vx/group';
+import useSvgBox from './useSvgBox';
+
+export const BoxedAnnotation = ({
+  x,
```
goldblatt, line 6: lil confused what x and y are here

src/components/Charts/BoxedAnnotation.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const { top, left, height, width } = useSvgBox(textRef);
+  return (
+    <Group left={x} top={y}>
+      <rect
+        y={top - padding}
+        x={left - padding}
+        width={width + 2 * padding}
+        height={height + 2 * padding}
+      />
+      <text ref={textRef}>{children}</text>
```
goldblatt, line 26: if this is just text why use children?

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,265 @@
...
+          <RectClipPath
+            id={mainClipPathId}
+            width={chartWidth}
+            height={chartHeight}
+          />
+          <Group clipPath={`url(#${mainClipPathId})`}>
+            <Style.SeriesArea>
+              <Area
+                data={data}
+                x={xCoord}
```
pnavarrc, line 132: `x={d => xScale(x(d))}`

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,265 @@
...
+            </BoxedAnnotation>
+          </Style.TextAnnotation>
+          <Style.CircleMarker
+            cx={xScale(x(truncationDataPoint))}
+            cy={yScale(yRt(truncationDataPoint))}
+            r={6}
+          />
+          <Style.Axis>
+            <AxisBottom top={chartHeight} scale={xScale} numTicks={3} />
+          </Style.Axis>
```
pnavarrc, line 203: TODO: Fix the number of ticks

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,256 @@
...
+  const maxZones = zones.MEDIUM.upperLimit;
+  const maxTick = maxY < maxZones ? 1.5 * maxZones : maxY;
+  return [minY, zones.LOW.upperLimit, zones.MEDIUM.upperLimit, maxTick];
+};
+
+// Accessing the data
+const x = (d: any) => new Date(d.x);
+const yRt = (d: any) => d?.y?.rt;
+const yHigh = (d: any) => d?.y?.high;
+const yLow = (d: any) => d?.y?.low;
```
pnavarrc, line 39: **TODO**: Since we know the type of `projectionDataset`, we _should_ be able to get rid of all the `any` here, not sure what the correct type would be though.

goldblatt: yeah would be good preferrable to make theset type safe

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,265 @@
...
+import * as Style from './Charts.style';
+
+const computeTickPositions = (minY: number, maxY: number, zones: Zones) => {
+  const maxZones = zones.MEDIUM.upperLimit;
+  const maxTick = maxY < maxZones ? 1.5 * maxZones : maxY;
+  return [minY, zones.LOW.upperLimit, zones.MEDIUM.upperLimit, maxTick];
+};
+
+// Accessing the data
+const x = (d: any) => new Date(d.x);
```
goldblatt, line 36: I think maybe could call this like getX or getDate?

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,265 @@
...
+const yRt = (d: any) => d?.y?.rt;
+const yHigh = (d: any) => d?.y?.high;
+const yLow = (d: any) => d?.y?.low;
```
goldblatt, line 39: I'd probably also lbel these as for the area chart?

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,265 @@
...
+  marginLeft?: number;
+  marginRight?: number;
+}) => {
+  const chartWidth = width - marginLeft - marginRight;
+  const chartHeight = height - marginTop - marginBottom;
+
+  const data = projectionDataset.data.filter(hasData);
+
+  // TODO(@pnavarrc): This is so TS doesn't complain
+  const minDate = d3min(data, x) || new Date('2020-01-01');
```
goldblatt, line 70: i guess it's odd that this is function right?

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+  const xScale = scaleTime({
+    domain: [minDate, maxDate],
+    range: [0, chartWidth],
+  });
+
+  const yDataMin = 0;
+  const yDataMax = d3max(data, yRt);
+
+  const yScale = scaleLinear({
```
goldblatt, line 81: I'm kind of a fan of using get for functions like this so since xScale and yScale are functions i would maybe call this getYScale

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,265 @@
...
+    domain: [yDataMin, yDataMax],
+    range: [chartHeight, 0],
+  });
+
+  const xCoord = (d: any) => xScale(x(d));
+  const yCoord = (d: any) => yScale(yRt(d));
+
```
goldblatt, line 88: typesafety here as well

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,265 @@
...
+  const yTicks = computeTickPositions(yDataMin, yDataMax, CASE_GROWTH_RATE);
+  const yAxisTicks = _tail(yTicks);
+
+  const { x: lastValidDate } = last(data);
```
goldblatt, line 92: hmm ya so above x is a functoin and here it's value?

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,265 @@
...
+  const dateTruncation = getTruncationDate(lastValidDate, RT_TRUNCATION_DAYS);
+  const prevData = data.filter((d: any) => x(d) <= dateTruncation);
+  const restData = data.filter((d: any) => x(d) >= dateTruncation);
```
goldblatt, line 95: type safety here

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,265 @@
...
+  const truncationDataPoint = last(prevData);
+
+  const regions = getChartRegions(yDataMin, yDataMax, CASE_GROWTH_RATE);
```
goldblatt, line 98: what's the difference betwen a region and a zone and a level?

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,265 @@
...
+    CASE_GROWTH_RATE,
+  );
+
+  const { tooltipData, tooltipOpen, showTooltip, hideTooltip } = useTooltip();
+  const onMouseOver = (
+    event: React.MouseEvent<SVGPathElement, MouseEvent>,
+    d: any,
+  ) => {
+    // @ts-ignore - typing bug
+    const coords = localPoint(event.target.ownerSVGElement, event);
```
goldblatt, line 110: you could probably have a default value here instead of ? to check if x or y exists?

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,265 @@
...
+            width={chartWidth}
+            height={chartHeight}
+          />
+          <Group clipPath={`url(#${mainClipPathId})`}>
+            <Style.SeriesArea>
+              <Area
+                data={data}
+                x={xCoord}
+                y0={(d: any) => yScale(yLow(d))}
+                y1={(d: any) => yScale(yHigh(d))}
```
goldblatt, line 134: type safety here would be nice

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,265 @@
...
+                curve={curveNatural}
+              />
+            </Style.SeriesArea>
+            {regions.map((region, i) => {
+              const clipPathZoneId = randomizeId(`clip-region-${region.name}`);
```
goldblatt, line 139: I think a comment here about what this is doing would be helpful

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,265 @@
...
+    // @ts-ignore - typing bug
+    const coords = localPoint(event.target.ownerSVGElement, event);
+    showTooltip({
+      tooltipLeft: coords?.x || 0,
+      tooltipTop: coords?.y || 0,
+      tooltipData: d,
+    });
+  };
+
+  const mainClipPathId = randomizeId('chart-clip-path');
```
goldblatt, line 118: huh?

src/components/Charts/utils.ts
```diff
@@ -196,3 +198,37 @@ export const zoneAnnotations = (
...
+
+export const getTruncationDate = (date: Date, truncationDays: number) =>
+  moment(date).subtract(truncationDays, 'days').toDate();
+
+export const randomizeId = (name: string): string =>
+  `${name}-${Math.random().toFixed(9)}`;
+
+export const getChartRegions = (minY: number, maxY: number, zones: Zones) => [
+  {
+    valueFrom: minY,
```
goldblatt, line 210: could we just add this to the og files and not have this function?

src/components/Charts/BoxedAnnotation.tsx
pnavarrc: nice catch! yeah, that's better

src/components/Charts/ChartRt.tsx
pnavarrc: good point, I'm renaming the `x` function and improve the code here as well

```ts
const lastValidDate = getDate(last(data));
```

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,274 @@
...
+  const chartWidth = width - marginLeft - marginRight;
+  const chartHeight = height - marginTop - marginBottom;
+
+  const data: PointRt[] = projectionDataset.data.filter(hasData);
+
+  const minDate = d3min(data, getDate) || new Date('2020-01-01');
+  const maxDate = CHART_END_DATE;
+
+  const yDataMin = 0;
+  const yDataMax = d3max(data, getRt) || 1;
```
pnavarrc, line 81: **TODO**: How to avoid having to pass a default just so TS doesn't complain?

src/components/Charts/HoverOverlay.tsx
```diff
@@ -0,0 +1,44 @@
...
+  onMouseOut,
+}: {
+  data: any[];
+  x: (d: any) => number;
+  y: (d: any) => number;
+  width: number;
+  height: number;
+  onMouseOver: (
+    event: React.MouseEvent<SVGPathElement, MouseEvent>,
+    d: any,
```
pnavarrc, line 20: **TODO**: Use templates for the types here

src/components/Charts/ChartRt.tsx
pnavarrc: what do you mean? `d3min` calculates the minimum value of `getDate(d)` for all the data items `d` in the array `data`. It's the same as `d3min(data.map(getDate))`, maybe that's more clear?

I could also use something like:

```ts
const minDate = data.map(getDate).reduce(Math.min)
```

but I find it a bit convoluted. Thoughts?

src/components/Charts/ChartRt.tsx
pnavarrc: I created a component to abstract the clipping behavior and added an explanation of how it works, I hope it's more clear now.

src/components/Charts/ChartRt.tsx
pnavarrc: yeah, I created a new component to abstract this and put the explanation there 👍

src/components/Charts/utils.ts
pnavarrc: Not really, because `minY` and `maxY` depend on the data, they are calculated so all the regions are visible (even if the R(t) curve doesn't get to that region) and to crop the values for R(t) if they are too far from the medium and low zones, that way they are still legible in the chart. Makes sense?

Note: I still need to explain better region vs. zone, I will get to that soon!

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,261 @@
...
+import * as Style from './Charts.style';
+
+type PointRt = {
+  x: number;
+  y: {
+    rt: number;
+    low: number;
+    high: number;
+  };
+};
```
goldblatt, line 37: i'd imagine this is coming from the projection so we should use the data type from that?

pnavarrc: almost ... `Column` has `{ x: number; y: any; }`, so I'm taking `RtRange` from `Projection.ts` to redefine `PointRt`

```ts
type PointRt = Omit<Column, 'y'> & {
  y: RtRange;
};
```

Not sure what the best practice is, let me know if there is a better way!

src/components/Charts/Charts.style.ts
pnavarrc: Fixed!

src/components/Charts/ChartRt.tsx
pnavarrc: The regions have similar properties than zones, except that they have `valueFrom` and `validTo`, which are calculated from the data and the zone limits. We need those values to calculate the position of the High, Medium, and Low labels and to know which part of the line to give each color.

src/components/Charts/ChartRt.tsx
pnavarrc: Updated!

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,298 @@
...
+import {
+  formatDecimal,
+  getChartRegions,
+  getTruncationDate,
+  getZoneByValue,
+  last,
+} from './utils';
+import * as Style from './Charts.style';
+
+type PointRt = Omit<Column, 'y'> & {
```
goldblatt, line 29: hmm i'm not familiar with this syntax

src/components/Charts/utils.ts
goldblatt: hmm seems like we oculd use negative infinitey and infinitey and then clip the charts, but if that doesn't work this is chill

src/components/Charts/ChartRt.tsx
```diff
@@ -0,0 +1,298 @@
...
+    showTooltip({
+      tooltipLeft: coords.x,
+      tooltipTop: coords.y,
+      tooltipData: d,
+    });
+  };
+
+  const getTooltipBody = (d: PointRt): string => {
+    const Rt = formatDecimal(getRt(d));
+    return getDate(d) < truncationDate ? `Rt ${Rt}` : `Rt ${Rt} (preliminar)`;
```
goldblatt, line 147: should this be preliminary?

src/components/Charts/utils.ts
pnavarrc: can't be done that way, the clipping area has to be bounded (although that's weird, the SVG surface is supposed to go to infinity in all directions)

src/components/Charts/ChartRt.tsx
pnavarrc: https://www.typescriptlang.org/docs/handbook/utility-types.html#omittk - it allows omit one property from the type - in this case, `y` is `any` in `Column`, which is not compatible with `RtRange`

pnavarrc:
[Rewriting the Charts - Chart R(t)](https://trello.com/c/rv4oIPZN/95-rewriting-the-charts-chart-rt)

pnavarrc:
[Rewriting the Charts - Chart R(t)](https://trello.com/c/rv4oIPZN/95-rewriting-the-charts-chart-rt)

pnavarrc:
[Rewriting the Charts - Chart R(t)](https://trello.com/c/rv4oIPZN/95-rewriting-the-charts-chart-rt)

PR #758, opened 2020-05-14
About page updates from Anna

schlosser:
From https://trello.com/b/pPyaxWaK/copy

### DIFF
--- a/src/screens/About/About.tsx
+++ b/src/screens/About/About.tsx
@@ -239,9 +239,9 @@ const About = ({ children }: { children: React.ReactNode }) => {
             1.0. This means that, on average, each person with COVID will
             transmit to 0.9 other people, which means the disease is shrinking.
             An R<sub>0</sub> of 1.0 means COVID is in perfect equilibrium,
-            neither growing nor shrinking. And an R<sub>0</sub> greater than 1.0
-            means that COVID is growing exponentially — and the greater the
-            value above 1.0, the faster the doubling speed.
+            neither growing nor shrinking. An R<sub>0</sub> greater than 1.0
+            means that COVID is growing exponentially. The greater the value
+            above 1.0, the faster the doubling speed.
           </Typography>
 
           <Typography variant="h6" component="h6">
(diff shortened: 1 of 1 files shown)


src/screens/About/About.tsx
```diff
@@ -239,9 +239,9 @@ const About = ({ children }: { children: React.ReactNode }) => {
             1.0. This means that, on average, each person with COVID will
             transmit to 0.9 other people, which means the disease is shrinking.
             An R<sub>0</sub> of 1.0 means COVID is in perfect equilibrium,
-            neither growing nor shrinking. And an R<sub>0</sub> greater than 1.0
-            means that COVID is growing exponentially — and the greater the
-            value above 1.0, the faster the doubling speed.
+            neither growing nor shrinking. An R<sub>0</sub> greater than 1.0
+            means that COVID is growing exponentially. The greater the value
+            above 1.0, the faster the doubling speed.
```
mikelehen, line 244: Sorry to flag text you didn't change, but it feels like this whole paragraph should be saying R-effective instead of R<sub>0</sub>?

My armchair epidemiology understanding is that R<sub>0</sub> is the "initial conditions" value of R.  You can't "drive it down".

But I'm sure the public doesn't understand, so maybe we don't care.

src/screens/About/About.tsx
```diff
@@ -334,27 +337,44 @@ const About = ({ children }: { children: React.ReactNode }) => {
...
 
+          <Typography variant="h6" component="h6">
+            What does the dotted line on the “infection growth” graph mean?
+          </Typography>
+
+          <Typography variant="body1" component="p">
+            Because of how we weight our data (Gaussian smoothing), and because
+            of potential reporting delays and errors in the incoming case data,
+            we need 7 preceding days of data before we can calculate a final R
+            <sub>t</sub> value. Therefore, we notate preliminary R<sub>t</sub>{' '}
```
mikelehen, line 348: Flag (feel free to ignore), I don't think we use R<sub>t</sub> anywhere else. We use R-effective above and in our blog post: https://blog.covidactnow.org/modeling-metrics-critical-to-reopen-safely/

PR #766, opened 2020-05-16
Adds press grid to homepage

schlosser:


mikelehen:
PressLogoGrid doesn't seem to be laying out great when I put Chrome in mobile mode.  :-/

image

schlosser:
Ahh, while typical zoom levels don't result in this happening, below 340px or with zoom on this will break. Fix incoming!

schlosser:
https://github.com/covid-projections/covid-projections/pull/769

Issue #633, opened 2020-04-29
Add County Lines to US Map on homepage


yvonnehung:
## summary
Show county lines on map on homepage to indicate we have county level data.

## expected result
Show county lines in white 0.5px white (or if can't do 0.5px, then 1px, #ffffff at 50% opacity).  State lines in 2px, white.

## actual result
Currently no county lines, just state lines.  

## environment
*(device, browser, mobile / web)*
All

## link
Mocks in Figma:
https://www.figma.com/file/UUCr3ZpwaPV4o1KuGdSCNl/Covid?node-id=3334%3A0

## screenshot
*(you can copy/paste images directly in)*
![image](https://user-images.githubusercontent.com/2531984/80634496-6f9e2880-8a0f-11ea-8ab3-0e005deebfaf.png)

PR #772, opened 2020-05-17
USA Map - Add county lines

pnavarrc:
Close https://github.com/covid-projections/covid-projections/issues/633 - Add county lines to the USA map

image

I just added the county geographies on top of the state map and style so they are transparent and ignore pointer events. I also changed the style wrappers a bit to make them work better with SVG.


pnavarrc:
Obsolete, closing!

PR #777, opened 2020-05-18
Generate index.html files with proper share meta tags.

mikelehen:
Successor to #721.  Adds a build step to generate index.html files with proper meta tags for sharing that point to our generated screenshots.


src/components/AppMetaTags/AppMetaTags.js
```diff
@@ -13,32 +13,18 @@ export default function AppMetaTags({
   canonicalUrl,
   pageTitle,
   pageDescription,
-  // BUG: shareTitle and shareDescription don't work! See note below.
-  shareTitle,
-  shareDescription,
```
mikelehen, line 18: We were using react-helmet (a React component that rewrites your head tags for title, meta tags, etc.) for these sharing tag stuff, but it was ineffective since it only shows up for crawlers that run your JavaScript, which Facebook and Twitter do not.  So I've ripped it out of here and we do it purely in the generate-index-pages script.

scripts/generate_index_pages.ts
```diff
@@ -0,0 +1,161 @@
...
+  // replace them.
+  return {
+    'og:image:url': imageUrl,
+    'twitter:image': imageUrl,
+  };
+}
+
+function locationPageTags(fullImageUrl: string, canonicalUrl: string, locationName: string): MetaTags {
+  const title = `Is ${locationName} ready to reopen?`;
+  const description = 'Real-time modeling and metrics to understand where we stand against COVID.';
```
mikelehen, line 35: Note that these are provisional.  I'll work with @schlosser, etc. on final copy.

mikelehen:
@pnavarrc Again, let me know if you're not up for these reviews. 😁

scripts/generate_index_pages.ts
```diff
@@ -0,0 +1,161 @@
...
+    const imageUrl = builder.fullImageUrl(`states/${stateCode.toLowerCase()}.png`);
+    const page = path.join(relativeUrl, 'index.html');
+    await builder.writeTemplatedPage(page, locationPageTags(imageUrl, canonicalUrl, stateName));
+  }
+
+  for (const fips of COUNTIES) {
+    const county = findCountyByFips(fips);
+    if (!county) {
+      console.warn('Skipping missing county: ', fips);
+      continue;
```
pnavarrc, line 68: Just an FYI, it seems that there is only one missing

```
Building index.html pages...
Skipping missing county:  11001
```

mikelehen: Oh, yeah...  That's Washington, DC which we treat as a state, not a county.   For the other PR, I explicitly blacklisted it, so I went ahead and did the same thing here.

scripts/generate_index_pages.ts
```diff
@@ -0,0 +1,161 @@
...
+    await builder.writeTemplatedPage(page, locationPageTags(imageUrl, canonicalUrl, locationName));
+  }
+}
+
+class IndexPageBuilder {
+  // TODO(michael): I know parsing HTML via regex is super fraught (could be
+  // fragile) but I'm also hesitant to parse / emit modified HTML. This feels
+  // safer, at least for now.
+  private readonly META_TAG_REGEX =
+      new RegExp(/<meta\s+(?<propKey>(?:property|name))="(?<name>[^"]*?)"\s+content="(?<value>[^"]*?)"/, 'g');
```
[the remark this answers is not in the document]
mikelehen: Thanks! Yeah, that's probably a really good idea. I think it'll mean we'll potentially see the template placeholders (`{{canonicalUrl}}` etc.) when we're debugging locally with `yarn start`, but that shouldn't matter. I doubt anybody is looking at the meta tags anyway.

Since the existing code is working, I may merge as-is. But I'll take a look at swapping in handlebars soon (probably tomorrow). 😁

PR #781, opened 2020-05-19
Add Contact Tracing Metric

goldblatt:
- updates to use a rolling weekly average for contact tracing
- adds the current copy
- moves the status update text for each of the charts into their own chart file


src/common/metrics/contact_tracing.ts
```diff
@@ -0,0 +1,112 @@
...
+      currentContactTracers,
+    )} contact tracers.` +
+    ` With an average of ${formatInteger(
+      currentWeeklyAverage,
+    )} new daily cases, ` +
+    `we estimate ${location} needs ${formatInteger(
+      currentWeeklyAverage * TRACERS_NEEDED_PER_CASE,
+    )} ` +
+    `contact tracing staff to trace all new cases in 48 hours, before too ` +
+    `many other people are infected.`;
```
adpirz, line 91: Nit: For readability, may be better to hoist the formatting calls out of here and set them to vars before the overview

src/common/metrics/future_projection.ts
```diff
@@ -0,0 +1,31 @@
...
+          noInterventionDate,
+        )}.`
+      : `However, any reopening should happen in a slow and phased fashion.`;
+
+    return (
+      `Assuming current trends and interventions continue, ${projection.locationName} hospitals are unlikely to become overloaded in the next 3 months. ` +
+      `${restrictionsLiftedText}`
+    );
+  }
+}
```
adpirz, line 31: Nit: for readability, I think this makes more sense:

```js
const { dateOverwhelmed } = projection;

if (dateOverwhelmed && dateOverwhelmed < new Date()} {
 // path 1
 return;
}

if (dateOverwhelmed) {
 // path 2
 return;
}

// path 3
return;

```

src/common/models/Projection.ts
```diff
@@ -161,6 +173,13 @@ export class Projection {
       summaryWithTimeseries.actuals.cumulativeDeaths;
     this.currentCumulativeCases =
       summaryWithTimeseries.actuals.cumulativeConfirmedCases;
+    this.currentContactTracerMetric = this.contractTracers
+      .filter(x => x !== null)
```
adpirz, line 177: Just confirming you won't hit `undefined` here, in which case you could do `!=` instead

src/common/utils/chart.tsx
```diff
@@ -0,0 +1,18 @@
...
+  lowText: string,
+  mediumText: string,
+  highText: string,
+) {
+  return level === Level.LOW
+    ? lowText
+    : level === Level.MEDIUM
+    ? mediumText
+    : highText;
+}
```
adpirz, line 18: Non-blocking: I'd recommend a higher order function here. I imagine something like this:
```typescript
function textLevelerGenerator(): (level: string) => string {};

const textLeveler = textLevelerGenerator({ lowText: 'low', mediumText: 'medium', highText: 'high' });

// At call site
const leveledText = textLeveler(level);

```

I think at least making explicit what the first 3 args are by using an object will be clearer when you're calling the function.

PR #790, opened 2020-05-20
Chart Remake - Positive Test Rate

pnavarrc:
Closes https://trello.com/c/2IEGjPVL/150-rewriting-charts-positive-tests

This chart re-implements the Positive Test Rate chart and extracts common code from `ChartRt` into components so it can be reused. The `ChartZone` component can be also used for ICU hospitalization rate and contact tracing, but I prefer to change those in separate PRs to make it easier to test.

image

Can @goldblatt review?

### DIFF
--- a/src/components/Charts/ChartContainer.tsx
+++ b/src/components/Charts/ChartContainer.tsx
@@ -1,0 +1,77 @@
+import React from 'react';
+import { localPoint } from '@vx/event';
+import { Group } from '@vx/group';
+import { useTooltip } from '@vx/tooltip';
+import HoverOverlay from './HoverOverlay';
+import * as Style from './Charts.style';
+
+const ChartContainer = <T extends unknown>({
+  width,
+  height,
+  marginTop = 5,
+  marginBottom = 40,
+  marginLeft = 40,
+  marginRight = 5,
+  x,
+  y,
+  data,
+  renderMarker,
+  renderTooltip,
+  children,
+}: {
+  width: number;
+  height: number;
+  marginTop?: number;
+  marginBottom?: number;
+  marginLeft?: number;
+  marginRight?: number;
+  x: (d: T) => number;
+  y: (d: T) => number;
+  data: T[];
+  renderMarker: (d: T) => React.ReactNode;
+  renderTooltip: (d: T) => React.ReactNode;
+  children: React.ReactNode;
+}) => {
+  const chartWidth = width - marginLeft - marginRight;
+  const chartHeight = height - marginTop - marginBottom;
+
+  const tooltip = useTooltip<T>();
+  const { tooltipData, tooltipOpen, showTooltip, hideTooltip } = tooltip;
+
+  const onMouseOver = (
+    event: React.MouseEvent<SVGPathElement, MouseEvent>,
+    d: T,
+  ) => {
+    // @ts-ignore - typing bug
+    const coords = localPoint(event.target.ownerSVGElement, event);
+    if (!coords) return;
+    showTooltip({
+      tooltipLeft: coords.x,
+      tooltipTop: coords.y,
+      tooltipData: d,
+    });
+  };
+
+  return (
+    <Style.PositionRelative>
+      <svg width={width} height={height}>
+        <Group left={marginLeft} top={marginTop}>
+          {children}
+          {tooltipOpen && tooltipData && renderMarker(tooltipData)}
+          <HoverOverlay
+            width={chartWidth}
+            height={chartHeight}
+            data={data}
+            x={x}
+            y={y}
+            onMouseOver={onMouseOver}
+            onMouseOut={hideTooltip}
+          />
+        </Group>
+      </svg>
+      {tooltipOpen && tooltipData && renderTooltip(tooltipData)}
+    </Style.PositionRelative>
+  );
+};
+
+export default ChartContainer;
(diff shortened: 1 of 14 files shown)


pnavarrc:
[Rewriting Charts - Positive Tests](https://trello.com/c/2IEGjPVL/150-rewriting-charts-positive-tests)

src/components/Charts/ZoneLinePath.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import RectClipGroup from './RectClipGroup';
+import { curveNatural } from '@vx/curve';
+import { CurveFactory } from 'd3-shape';
+import { LinePath } from '@vx/shape';
+import { Region } from './utils';
+
+const ZoneLinePath = <T extends unknown>({
```
goldblatt, line 8: hmmm

src/components/Charts/ChartContainer.tsx
```diff
@@ -0,0 +1,77 @@
...
+  height: number;
+  marginTop?: number;
+  marginBottom?: number;
+  marginLeft?: number;
+  marginRight?: number;
+  x: (d: T) => number;
+  y: (d: T) => number;
+  data: T[];
+  renderMarker: any;
+  renderTooltip: any;
```
goldblatt, line 32: ? could we add types here

src/components/Charts/ChartContainer.tsx
```diff
@@ -0,0 +1,77 @@
...
+  children: React.ReactNode;
+}) => {
+  const chartWidth = width - marginLeft - marginRight;
+  const chartHeight = height - marginTop - marginBottom;
+
+  const tooltip = useTooltip<T>();
```
goldblatt, line 38: what's the benefit of the <T> here? guess i'm a lil confused

src/components/Charts/ChartZones.tsx
```diff
@@ -0,0 +1,211 @@
...
+                yScale={yScale}
+                curve={curveLinear}
+              />
+            </Style.SeriesLine>
+            <ZoneAnnotation
+              color={region.color}
+              name={region.name}
+              isActive={lastPointZone.name === region.name}
+              x={xScale(maxDate) - 10}
+              y={yScale(0.5 * (region.valueFrom + region.valueTo))}
```
goldblatt, line 145: why .5 here?

src/components/Charts/ChartZones.tsx
```diff
@@ -0,0 +1,211 @@
...
+            />
+          </Group>
+        ))}
+      </RectClipGroup>
+      <Style.LineGrid>
+        <GridRows width={chartWidth} scale={yScale} tickValues={yTicks} />
+      </Style.LineGrid>
+      <Style.TextAnnotation>
+        <BoxedAnnotation
+          x={getXCoord(lastPoint) + 30}
```
goldblatt, line 155: also why plus 30?

src/components/Charts/utils.ts
```diff
@@ -250,3 +261,25 @@ export const getZoneByValue = (value: number, zones: Zones) => {
...
+    maxTick,
+  ];
+};
+
+export const getAxisLimits = (minY: number, maxY: number, zones: Zones) => {
+  const tickPositions = computeTickPositions(minY, maxY, zones);
+  const minTickPosition = _.min(tickPositions) || minY;
+  const maxTickPosition = _.max(tickPositions) || maxY;
+  return roundAxisLimits(minTickPosition, maxTickPosition);
+};
```
goldblatt, line 285: did we not need these for the rt graph?

src/components/Charts/ZoneLinePath.tsx
pnavarrc: no good? I want to be able to pass different types of points to it, I thought this was the best practice 😂

src/components/Charts/ChartZones.tsx
pnavarrc: It's the midpoint between `valueFrom` and `valueTo`, maybe `(valueFrom + valueTo) / 2` would be more clear?

src/components/Charts/ChartZones.tsx
pnavarrc: so it's not directly over the point (the last point in the line wouldn't be visible otherwise)

src/components/Charts/utils.ts
pnavarrc: `computeTickPositions` was in the chart file, I moved to `utils` so I can reuse it for other charts.

src/components/Charts/ChartContainer.tsx
pnavarrc: `useTooltip` is a hook with an internal state, it stores the currently hovered data point. Since this logic is used for more than one type of chart, the data of the hovered point might have a different type. For the R(t) chart, the shape is `PointRt = { x: date, y: { low: number, high: number, rt: number }}`, for the other charts is `Point = { x: date, y: number }`

If we don't use the generic, TS is going to complain because we later try to access the type-specific properties when rendering the tooltip

src/components/Charts/ChartContainer.tsx
pnavarrc: yeah, adding the types!

src/components/Charts/ZoneLinePath.tsx
pnavarrc: This is used for `ChartRt` whose points have a different shape than the points for the Positive test rate chart

src/components/Charts/utils.ts
pnavarrc: (also `getAxisLimits`)

PR #792, opened 2020-05-20
Add github actions workflow (and shell script to kick it off) for updating to a new data snapshot.

mikelehen:


.github/workflows/update-snapshot.yml
```diff
@@ -0,0 +1,91 @@
...
+
+      # Generate Share Images and Upload to S3
+      - run: yarn serve -l 3000 &
+      - run: npx wait-on -t 120000 http://localhost:3000/
+      - run: yarn generate-share-images
+      - uses: jakejarvis/s3-sync-action@master
+        with:
+          args: '--acl public-read --follow-symlinks --delete'
+        env:
+          #TODO(michael): Make default secrets work.
```
mikelehen, line 46: Working with Simon to get this figured out.  Right now we have two different AWS keys for our different jobs, which isn't ideal.

.github/workflows/update-snapshot.yml
```diff
@@ -0,0 +1,91 @@
...
+          body: |
+            Update to data snapshot ${{env.SNAPSHOT_ID}}.
+
+            * Regenerate calculated interventions files.
+            * Point at updated share images.
+
+      - name: Slack notification
+        if: job.status != 'success'
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
```
mikelehen, line 75: FYI- I just used the existing webhook secret that was configured for the repo.  If we want to send this somewhere other than #eng or we want the name to be different than "Publishing New API", I'll need some help setting it up.  But I think it's fine? 🤷

.github/workflows/update-snapshot.yml
```diff
@@ -0,0 +1,91 @@
...
+          STATUS: ${{job.status}}
+        uses: Ilshidur/action-slack@fb92a78
+        with:
+          args: |
+            PR to update the website to snapshot ${{env.SNAPSHOT_ID}} (with updated map colors and share images) is available.
+            PR: https://github.com/covid-projections/covid-projections/pull/${{env.PULL_REQUEST_NUMBER}}
+            Preview: https://covid-projections-git-bump-snapshot-${{env.RUN_ID}}.covidactnow.now.sh/
+            Compare: https://covid-projections-git-bump-snapshot-${{env.RUN_ID}}.covidactnow.now.sh/compare/
```
mikelehen, line 91: This is where we could add extra report output from our other automated QA...

yarn.lock
```diff
@@ -3237,6 +3237,11 @@
   resolved "https://registry.yarnpkg.com/@xtuc/long/-/long-4.2.2.tgz#d291c6a4e97989b5c61d9acf396ae4fe133a718d"
   integrity sha512-NuHqBY1PB/D8xU6s/thBgOAiAP7HOYDQ32+BFZILJ8ivkUkAHQnWfn6WhL79Owj1qmUnoN/YPhktdIoucipkAQ==
 
+"@zeit/schemas@2.6.0":
```
mikelehen, line 3240: All these diffs are from adding `serve` package which we need to make `yarn serve` work... 🤷

mikelehen:
@goldblatt Can you review this since it sort of ties into the snapshot QA automation stuff? 😁 

@chasulin If you want to take a look as well, to start getting some familiarity with how our share image generation stuff is hooked up, that would be good.  I expect some of this will be foreign, but we can chat through the details.

.github/workflows/update-snapshot.yml
goldblatt: yeah i was thinking we should change the name of that webhook anyway to be something more general?

.github/workflows/update-snapshot.yml
```diff
@@ -0,0 +1,91 @@
+name: Update to new data snapshot (regenerate colors and sharing images).
+
+on:
```
goldblatt, line 3: so this is still manually called right now right?

.github/workflows/update-snapshot.yml
```diff
@@ -0,0 +1,91 @@
...
+        with:
+          repository: covid-projections/covid-projections
+          ref: '${{ env.BRANCH }}'
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+
+      # Update snapshot URL and regenerate map colors.
+      - run: yarn install
+      - run: 'echo "{ \"data_url\": \"https://data.covidactnow.org/snapshot/${{env.SNAPSHOT_ID}}/\" }" | python -mjson.tool > src/assets/data/data_url.json'
```
goldblatt, line 35: do we not need to install mjson pacakge?

.github/workflows/update-snapshot.yml
```diff
@@ -0,0 +1,91 @@
...
+            Update to data snapshot ${{env.SNAPSHOT_ID}}.
+
+            * Regenerate calculated interventions files.
+            * Point at updated share images.
+          title: Update to data snapshot ${{env.SNAPSHOT_ID}}
+          body: |
+            Update to data snapshot ${{env.SNAPSHOT_ID}}.
+
+            * Regenerate calculated interventions files.
+            * Point at updated share images.
```
goldblatt, line 70: does the merge the pr into develop?

.github/workflows/update-snapshot.yml
mikelehen: Yeah.  My thinking is that we can add a scheduled workflow that just triggers it every morning.

And if we need to kick one off manually (e.g. the morning snapshot was bad), then we can do that too.

.github/workflows/update-snapshot.yml
mikelehen: I think it's builtin or something?  I don't know how python works TBH. 😁

.github/workflows/update-snapshot.yml
mikelehen: It should generate a PR (against develop by default), but won't merge it.  A human has to do that.

.github/workflows/update-snapshot.yml
mikelehen: Yes, thanks!

PR #796, opened 2020-05-21
phase 4 copy from JKT, Max, and James

schlosser:


src/common/metrics/hospitalizations.ts
```diff
@@ -23,30 +23,30 @@ export const HOSPITAL_USAGE_LEVEL_INFO_MAP: LevelInfoMap = {
...
     level: Level.UNKNOWN,
     upperLimit: 0,
     name: UNKNOWN,
     color: COLOR_MAP.GRAY.BASE,
-    detail: SHORT_DESCRIPTION_UNKNOWN,
+    detail: () => SHORT_DESCRIPTION_UNKNOWN,
   },
 };
 
 export const HOSPITALIZATIONS_DISCLAIMER =
```
igorkofman, line 51: this should just contain the link..

schlosser: Yeah I wasn't sure how to do that, without something like `dangerouslySetInnerHTML={createMarkup()} `

PR #801, opened 2020-05-21
Chart Share Buttons

chasulin:
- Adds ShareButtons component (social links, copy link, save) to each individual chart that has data
- Adds routes into App.js that are specific to share links (…/chart/:chartId). These links will scroll to whichever chart is identified by chartId
- Adds makeChartShareQuote function to generate share quotes specific to each location and chart

Example share link: https://covid-projections-j6wdzl5xm.now.sh/us/pa/chart/2


src/components/LocationPage/ShareButtons.js
```diff
@@ -0,0 +1,79 @@
...
+
+  useOutsideClickHandler(shareButtonsRef, setShowShareIcons);
+
+  const theme = useTheme();
+  const isMobile = useMediaQuery(theme.breakpoints.down('xs'));
+
+  return (
+    <Fragment>
+      {isMobile && (
+        <MobileButtonsWrapper isFirst={isFirst} ref={shareButtonsRef}>
```
mikelehen, line 54: I think `isFirst` is unused by `MobileButtonsWrapper`?

src/components/LocationPage/ShareButtons.style.js
```diff
@@ -0,0 +1,109 @@
...
+`;
+
+export const DesktopButtonsWrapper = styled.div`
+  display: none;
+
+  @media (min-width: 600px) {
+    ${WrapperStyles};
+    top: 4.875rem;
+    margin-top: ${({ isFirst, chartAbove }) =>
+      !isFirst && chartAbove && '-24px'};
```
mikelehen, line 20: I think you could maybe get rid of `isFirst` and just rely on `chartAbove`?  (i.e. if chartAbove is true, then isFirst will always be false, so you don't need that check and can drop the `isFirst` property?)

src/components/LocationPage/ShareButtons.js
```diff
@@ -0,0 +1,79 @@
...
+  );
+};
+
+const ShareButtons = props => {
+  const { isFirst, chartAbove } = props;
+  const [showShareIcons, setShowShareIcons] = useState(false);
+
+  const shareButtonsRef = useRef(null);
+
+  useOutsideClickHandler(shareButtonsRef, setShowShareIcons);
```
mikelehen, line 46: If we can switch to https://material-ui.com/components/click-away-listener/ I suspect we can get rid of `shareButtonsRef` and move `showShareIcons` into the `InnerContent` component?

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -103,11 +113,15 @@ const ChartsHolder = (props: {
                   </Disclaimer>
                 </>
               )}
+              {!isMobile && getChartSummarys(projection)[1] && (
+                <ShareButtons chartAbove={getChartSummarys(projection)[0]} />
```
mikelehen, line 117: Hrm...  This `chartAbove` thing strikes me as unfortunate (ideally the component wouldn't need to know this much about its surroundings).  Can you explain why this is necessary, exactly?  Would it be possible / make sense to handle whatever the "has chart above" situation is here (in `ChartsHolder`) so the component doesn't need to deal with it?

chasulin: chartAbove was for styling purposes (different top margin if there was a chart above). I refactored and wrapped the buttons with the header which removed the need for chartAbove!

src/components/LocationPage/ShareButtons.js
chasulin: implemented clickAwayListener and got rid of shareButtonsRef! I kept showShareIcons because it is used by clickAwayListener

src/components/LocationPage/ShareButtons.style.js
chasulin: Rearranged in ChartsHolder and removed the need for isFirst and chartAbove

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -88,13 +95,17 @@ const ChartsHolder = (props: {
...
             <SummaryStats stats={getChartSummarys(projection)} />
             <MainContentInner>
-              <ChartHeader>
-                {getMetricName(Metric.CASE_GROWTH_RATE)}
-              </ChartHeader>
+              <ChartHeaderWrapper>
+                <ChartHeader>
+                  {getMetricName(Metric.CASE_GROWTH_RATE)}
+                </ChartHeader>
+                {!isMobile && rtRangeData && <ShareButtons isFirst />}
```
mikelehen, line 102: I think `isFirst` is unused and can be removed now?

src/components/LocationPage/ShareButtons.js
```diff
@@ -0,0 +1,74 @@
...
+  const { isFirst } = props;
+  const [showShareIcons, setShowShareIcons] = useState(false);
+
+  const theme = useTheme();
+  const isMobile = useMediaQuery(theme.breakpoints.down('xs'));
+
+  return (
+    <Fragment>
+      {isMobile && (
+        <ClickAwayListener onClickAway={() => setShowShareIcons(false)}>
```
mikelehen, line 49: Rather than duplicate the `ClickAwayListener` for both `isMobile` and `!isMobile` could you move it outside (i.e. right below `<Fragment>` ?

Alternatively would it work to put this inside `InnerContent`?  And then you could move `showShareIcons` state 100% inside of `InnerContent`.  Not sure if that'll cause some issue though.

src/components/LocationPage/ChartsHolder.tsx
mikelehen: Nice!

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -118,13 +155,29 @@ const ChartsHolder = (props: {
...
             <MainContentInner>
-              <ChartHeader ref={rtRangeRef}>
-                {getMetricName(Metric.CASE_GROWTH_RATE)}
-              </ChartHeader>
+              <ChartHeaderWrapper>
+                <ChartHeader ref={rtRangeRef}>
+                  {getMetricName(Metric.CASE_GROWTH_RATE)}
+                </ChartHeader>
+                {!isMobile && rtRangeData && (
+                  <ShareButtons
```
mikelehen, line 163: I'm not wild about all the repetition going on (10 of these ShareButtons instantiations).  I think we could clean it up a bit by creating a `metricToShareButtons` map up top or something, but we should probably rework this whole page to be more componentized rather than repeat ourselves for each metric.  So we can punt for now.

src/components/LocationPage/ShareButtons.js
```diff
@@ -0,0 +1,135 @@
...
+    countyId,
+    chartIdentifier,
+  } = props;
+
+  const [showShareIcons, setShowShareIcons] = useState(false);
+
+  const hideSocialButtons = () => {
+    const timeoutId = setTimeout(() => {
+      setShowShareIcons(false);
+    }, 1500);
```
mikelehen, line 31: Why the delay?  Can you add a comment?

src/components/LocationPage/ShareButtons.style.js
```diff
@@ -102,3 +102,5 @@ export const CopyLinkButton = styled.div`
     font-size: 0.875rem;
   }
 `;
+
+export const ClickAwayWrapper = styled.div``;
```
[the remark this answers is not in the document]
chasulin: ClickAwayListener only accepts one child- it wasn't working with just a Fragment so I added a wrapper. I made the div a named styled component just to keep it descriptive

src/common/utils/makeChartShareQuote.js
```diff
@@ -0,0 +1,44 @@
...
+  chartType,
+  projections,
+) {
+  const stateName = STATES[stateId];
+  const countyName = county && county.county;
+  const displayName = countyName ? `${countyName}, ${stateName}` : stateName;
+
+  if (chartType === Metric.CASE_GROWTH_RATE) {
+    return `${displayName}’s infection growth rate is ${formatDecimal(
+      stats[0],
```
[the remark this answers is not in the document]
chasulin: Adding it to TODOs

src/common/utils/makeChartShareQuote.js
```diff
@@ -0,0 +1,44 @@
+import { STATES } from 'common';
+import { formatDecimal, formatPercent } from 'components/Charts/utils';
+import { Metric } from 'common/metric';
+import { formatDate } from 'common/utils';
+
+export default function makeChartShareQuote(
```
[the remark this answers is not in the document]
chasulin: Ah yes ok

src/components/LocationPage/ShareButtons.js
```diff
@@ -0,0 +1,135 @@
...
+  const {
+    stateId,
+    county,
+    stats,
+    chartType,
+    projections,
+    isMobile,
+    chartId,
+    countyId,
+    chartIdentifier,
```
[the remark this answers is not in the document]
chasulin: Ooo yes. I can combine 'chartType' and 'chartIdentifier'. 

'chartIdentifier' and 'chartId' do resolve to the same thing. The different uses are:
- 'chartId' is the actual param name & is used to dictate the scrollTo when you land on a share link.
-'chartIdentifier' is used to create the shareUrl (when there usually won't be a chartId in the URL).

src/components/LocationPage/ShareButtons.js
chasulin: Yvonne requested that the 'copy link' button changes to say 'copied' when it is clicked. I delayed the dismissal of the social buttons so the user could see that change briefly.
I'll add a comment!

PR #802, opened 2020-05-21
Ignore days with negative cases when calculating cases/day for contact tracing.

mikelehen:
Removes days with negative values when calculating avg cases/day.  This makes WA closer to reality (139 cases/day instead of 97), but still is questionable, since in addition to a negative value (now removed) there are several very low values pulling the average down:
```
[267, 225, 130, 30, 5, 178]
```

**Before:**
image

**After:**
image

I spot-checked ~5 other states and they were completely unaffected.

### DIFF
--- a/src/common/models/Projection.ts
+++ b/src/common/models/Projection.ts
@@ -1,3 +1,4 @@
+import _ from 'lodash';
 import moment from 'moment';
 
 import {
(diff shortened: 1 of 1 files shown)


src/common/models/Projection.ts
```diff
@@ -372,24 +382,23 @@ export class Projection {
   }
 
   /**
-   * G
+   * Gets the cases/day on day i, by averaging the trailing week of cases/day
+   * data.
```
goldblatt, line 386: can we add a reason why we are doing this? here

mikelehen: Oh, yeah.  I just added the comment because it was missing, but then I forgot to document my change. Hah.  Done.

PR #805, opened 2020-05-22
Fix broken share previews

adpirz:
675rem is maybe a bit too large


mikelehen:
Susan rolled this into her PRs.

PR #808, opened 2020-05-22
Refactor ShareImage into two components.

mikelehen:
This is a precursor minor refactor before actually generating the sharing charts.

I'm splitting ShareImage into two components (ShareImage which is now just a wrapper and router, and ShareCardImage which contains the preexisting homepage / location share image stuff).  

It will probably be easier to review the 2 commits separately. The first one is just a rename to help github show the diffs better in the second commit.

* Moved existing `ShareImage` component to `ShareCardImage` and it's used for the default homepage / location sharing (it should really be split into two components but I didn't want to make things more complicated).
* Created _new_ ShareImage component (i.e. I reused the name) that now does the routing for all of the share images we need to create, including the new chart share images.
* Added a dummy placeholder ChartImage component for the new chart images


src/App.js
```diff
@@ -85,17 +85,7 @@ export default function App() {
...
-              exact
-              path="/internal/share-image/states/:stateId"
-              component={ShareImage}
-            />
-            <Route
-              exact
-              path="/internal/share-image/counties/:countyFipsId"
-              component={ShareImage}
-            />
+            <Route path="/internal/share-image/" component={ShareImage} />
```
mikelehen, line 88: All the routing moved inside the new `ShareImage` component.

src/screens/internal/ShareImage/ShareCardImage.style.ts
```diff
@@ -1,27 +1,5 @@
...
-`;
-
-export const ScreenshotWrapper = styled.div`
-  width: 1200px;
-  height: 630px;
-  background-color: black;
-  color: white;
-  overflow: hidden;
-`;
-
```
mikelehen: Moved these into `ShareImage` component, since they'll be used by all of the share images we generate.

src/screens/internal/ShareImage/ShareImage.style.ts
```diff
@@ -0,0 +1,23 @@
+import styled from 'styled-components';
```
mikelehen: This was all just moved from ShareCardImage.style.ts

PR #816, opened 2020-05-22
Chart Remake - ICU Utilization

pnavarrc:
Please review this after #790

NOTE: #790 needs to be merged first, and then we need to change the base branch to `develop` in this PR

### DIFF
--- a/src/components/Charts/ChartICUHeadroom.tsx
+++ b/src/components/Charts/ChartICUHeadroom.tsx
@@ -1,0 +1,33 @@
+import React from 'react';
+import { HOSPITAL_USAGE_LEVEL_INFO_MAP } from 'common/metrics/hospitalizations';
+import { Column } from 'common/models/Projection';
+import ChartZones from './ChartZones';
+import { formatPercent } from './utils';
+
+const CAP_Y = 1;
+
+const getPointText = (valueY: number) => formatPercent(valueY, 0);
+
+const getTooltipText = (valueY: number) =>
+  valueY > CAP_Y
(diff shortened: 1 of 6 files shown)


src/components/Charts/ChartZones.tsx
```diff
@@ -184,12 +186,14 @@ const ChartZoneAutosize = ({
   zones,
   capY,
   getTooltipText,
+  getPointText = (valueY: number) => formatPercent(valueY, 1),
```
goldblatt, line 189: why do you set the default here?

goldblatt: maybe we can add a comment i think this sets the floats to show

src/components/Charts/ChartICUHeadroom.tsx
```diff
@@ -0,0 +1,32 @@
...
+const getTooltipText = (valueY: number) =>
+  valueY > CAP_Y
+    ? `ICU headroom used > ${getPointText(CAP_Y)}`
+    : `ICU headroom used ${getPointText(valueY)}`;
+
+const ChartICUHeadroom = ({
+  columnData,
+  height = 400,
+}: {
+  columnData: any[];
```
goldblatt, line 19: can we give this a type

src/components/Charts/ChartZones.tsx
pnavarrc: is not really that useful, removing it as default and moving it to its corresponding chart

pnavarrc:
Thanks again @goldblatt - updated with your suggestions, merging!

PR #817, opened 2020-05-22
Chart Remake - Contact Tracing

pnavarrc:
Review after #816 - rebase once #816 is merged

### DIFF
--- a/src/components/Charts/ChartContactTracing.tsx
+++ b/src/components/Charts/ChartContactTracing.tsx
@@ -1,0 +1,31 @@
+import React from 'react';
+import ChartZones from './ChartZones';
+import { CONTACT_TRACING_LEVEL_INFO_MAP } from 'common/metrics/contact_tracing';
+import { Column } from 'common/models/Projection';
+import { formatPercent } from './utils';
+
+const CAP_Y = 1;
+
+const getPointText = (valueY: number) => formatPercent(valueY, 0);
+
+const getTooltipText = (valueY: number) =>
+  `Contacts traced ${getPointText(valueY)}`;
+
+const ChartContactTracing = ({
+  columnData,
+  height = 400,
+}: {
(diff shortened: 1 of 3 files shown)


src/components/Charts/ChartContactTracing.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import ChartZones from './ChartZones';
+import { CONTACT_TRACING_LEVEL_INFO_MAP } from 'common/metrics/contact_tracing';
+import { formatPercent } from './utils';
+
+const CAP_Y = 1;
+
+const getPointText = (valueY: number) => formatPercent(valueY, 0);
+
+const getTooltipText = (valueY: number) =>
```
goldblatt, line 10: do we need to do the capping stuff we did in icu headroom?

src/components/Charts/ChartContactTracing.tsx
```diff
@@ -0,0 +1,30 @@
...
+  `Contacts traced ${getPointText(valueY)}`;
+
+const ChartContactTracing = ({
+  columnData,
+  height = 400,
+}: {
+  columnData: any[];
```
goldblatt, line 17: also a type here would be nice

src/components/Charts/ChartContactTracing.tsx
pnavarrc: I'm assuming that we can't go over 100% contact tracing. For ICU headroom, there are some cases that go over 100% (I checked on the `/all` page), although I don't know how that works in real life 🤷‍♂️

src/components/Charts/ChartContactTracing.tsx
pnavarrc: Fixed!

pnavarrc:
Fixed the missing type - merging!

PR #822, opened 2020-05-24
Charts Remake - Future Hospitalizations

pnavarrc:
This PR re-implements the Future Hospitalizations chart

image

**Other Changes**
- beb01cfb629f0ecde29d442c965b64921af3e90b Added padding to the clipping region, to avoid cutting half of the line when they hit the top of the chart.
- 04d8c55fbf1fdc9eaae48d2521abf989eec5ad52 Update the Tooltip component so it can receive children and minor styling fixes. This is needed because the tooltip for the Future Hospitalizations chart is formatted (some text is in bold).

## Notes
- The Highcharts version allows us to toggle the series on and off by clicking on the legend. I vaguely remember that this behavior wasn't desired, because it allowed the user to hide all of them, so I didn't implement that. We can implement it if that's something we want to have (or a variation of it)
- I haven't removed the old chart code yet - I want to make sure that this is stable in production before removing it.

### DIFF
--- a/src/components/Charts/ChartFutureHospitalization.tsx
+++ b/src/components/Charts/ChartFutureHospitalization.tsx
@@ -1,0 +1,239 @@
+import React, { ReactNode } from 'react';
+import moment from 'moment';
+import { last, isDate } from 'lodash';
+import { min as d3min, max as d3max } from 'd3-array';
+import { AxisBottom, AxisLeft } from '@vx/axis';
+import { LinePath } from '@vx/shape';
+import { ParentSize } from '@vx/responsive';
+import { scaleLinear, scaleTime } from '@vx/scale';
+import { Projections } from 'common/models/Projections';
+import { COLORS } from 'common';
+import BoxedAnnotation from './BoxedAnnotation';
+import ChartContainer from './ChartContainer';
+import RectClipGroup from './RectClipGroup';
+import Tooltip from './Tooltip';
+import { LegendMarker, LegendLine } from './Legend';
+import * as Style from './Charts.style';
+import { formatDate, formatInteger } from './utils';
+import palette from 'assets/theme/palette';
+
+type Point = {
+  x: number;
+  y: number;
+};
+
+type PointTooltip = Point & {
+  color: string;
+  name: string;
+};
+
+const getDate = (p: Point) => new Date(p.x);
+const getY = (p: Point) => p.y;
+const hasData = (d: any) => isDate(getDate(d)) && Number.isFinite(getY(d));
+const getLastPoint = (points: Point[]): Point => last(points) || { x: 0, y: 0 };
+const isFuture = (p: Point) => getDate(p) > new Date();
+
+const getTooltipPoint = (
+  points: Point[],
+  color: string,
+  name: string,
+): PointTooltip[] => points.map(p => ({ ...p, color, name }));
+
+const getTooltipBody = (p: PointTooltip): ReactNode => (
+  <span>
+    <b>{formatInteger(getY(p))}</b>{' '}
+    {p.name === 'beds' ? 'beds available on' : 'hospitalizations expected by'}{' '}
+    <b>{formatDate(getDate(p), 'MMMM D')}</b>
+  </span>
+);
+
+const ChartFutureHospitalization = ({
+  projections,
+  width,
+  height,
+  marginTop = 5,
+  marginBottom = 40,
+  marginLeft = 48,
+  marginRight = 5,
+}: {
+  projections: Projections;
+  width: number;
+  height: number;
+  marginTop?: number;
+  marginBottom?: number;
+  marginLeft?: number;
+  marginRight?: number;
+}) => {
+  const chartWidth = width - marginLeft - marginRight;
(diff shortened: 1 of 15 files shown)


src/components/Charts/ChartFutureHospitalization.tsx
```diff
@@ -0,0 +1,239 @@
...
+  const allData: PointTooltip[] = [
+    ...getTooltipPoint(dataProjectedFuture, COLORS.PROJECTED, 'projected'),
+    ...getTooltipPoint(dataNoActionFuture, COLORS.LIMITED_ACTION, 'no-action'),
+    ...getTooltipPoint(dataProjectedPast, palette.black, 'past'),
+    ...getTooltipPoint(dataBeds, palette.chart.axis, 'beds'),
+  ];
+
+  const dateTwoWeeks = moment().add(2, 'weeks').toDate();
+  const minDate = d3min(allData, getDate) || new Date('2020-01-01');
+  const maxDate = d3max(allData, getDate) || dateTwoWeeks;
```
mikelehen, line 94: To be clear, these min/max dates should never get used, right?  If there's no data, then the chart will be empty and other stuff will probably break (e.g. firstPointBeds and lastPastPoint).

So I'd be a little tempted to do:
```
  const minDate = d3min(allData, getDate);
  const maxDate = d3max(allData, getDate);
  assert(minDate && maxDate, 'Data must not be empty');
```

(make sure to import assert from 'common/utils')

src/components/Charts/Charts.style.ts
```diff
@@ -117,19 +125,57 @@ export const RegionAnnotation = styled(TextAnnotation)<{ isActive: boolean }>`
...
+      margin-right: 0;
+    }
+  }
+`;
+
+export const LegendItem = styled.div`
+  flex: 0 1 auto;
+  align-items: center;
+  > * {
+    margin-right: 5px;
```
mikelehen: FYI- I've pushed a change to tweak the margins on the legend items.

src/components/Charts/ChartRt.tsx
```diff
@@ -99,18 +100,16 @@ const ChartRt = ({
...
-      />
-    );
-  };
+  const renderTooltip = (d: PointRt) => (
+    <Tooltip
+      left={marginLeft + getXCoord(d)}
+      top={marginTop + getYCoord(d)}
+      title={formatDate(getDate(d))}
+    >
+      {`Rt ${formatDecimal(getRt(d), 2)}`}{' '}
```
mikelehen: FYI- I pushed a change to make this "Infection rate"

src/components/Charts/ChartFutureHospitalization.tsx
```diff
@@ -0,0 +1,239 @@
...
+const getY = (p: Point) => p.y;
+const hasData = (d: any) => isDate(getDate(d)) && Number.isFinite(getY(d));
+const getLastPoint = (points: Point[]): Point => last(points) || { x: 0, y: 0 };
+const isFuture = (p: Point) => getDate(p) > new Date();
+
+const getTooltipPoint = (
+  points: Point[],
+  color: string,
+  name: string,
+): PointTooltip[] => points.map(p => ({ ...p, color, name }));
```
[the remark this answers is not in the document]
pnavarrc: Good suggestions, that's better - I added a string so we could differentiate the 3 cases, but the copy on the tooltip is actually the same for all the curves, will use the boolean instead and rename.

src/components/Charts/ChartFutureHospitalization.tsx
pnavarrc: I added the defaults to TS won't complain - if I don't, it says that `minDate` and `maxDate` could be `undefined`. Would the `assert` get rid of the warnings? I'm not sure what the best practices are

mikelehen: Yeah, if you use the assert() from `common/utils` it'll get rid of the warning:

```
export function assert(condition: any, msg?: string): asserts condition {
  if (!condition) {
    fail(msg);
  }
}
```

The `asserts condition` part basically tells TypeScript that the condition will be true if the function returns and so TypeScript will stop complaining.

src/components/Charts/ChartFutureHospitalization.tsx
```diff
@@ -0,0 +1,239 @@
...
+};
+
+type PointTooltip = Point & {
+  color: string;
+  name: string;
+};
+
+const getDate = (p: Point) => new Date(p.x);
+const getY = (p: Point) => p.y;
+const hasData = (d: any) => isDate(getDate(d)) && Number.isFinite(getY(d));
```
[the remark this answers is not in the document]
pnavarrc: Not yet, because the points in the dataset could have `null` or `undefined` values - the idea of this function is to filter out data that is not a `Point`

https://github.com/covid-projections/covid-projections/pull/822/files#diff-def6ce2c3c90cde92f1ea7ff7fc1e832R70-R73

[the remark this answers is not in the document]
mikelehen: Oh, I see!  Makes sense.

PR #825, opened 2020-05-24
Adds pages to generate the charts we need for share / export images.

mikelehen:
Implements the styling for the chart images for sharing and exporting. See:

Share image: https://covid-projections-git-mikelehen-share-charts.covidactnow.now.sh/internal/share-image/counties/53033/chart/0
Export image: https://covid-projections-git-mikelehen-share-charts.covidactnow.now.sh/internal/share-image/counties/53033/chart/0/export

PR #826, opened 2020-05-24
Make charts 'theme-able' and add a 'dark' theme.

mikelehen:
This makes the charts minimally "theme-able" and adds a dark set of colors for them.

To see how this gets used in the chart sharing, see [here](https://github.com/covid-projections/covid-projections/pull/825/files#diff-f5680937bf6b070548b0898693f6fc04R69)

And preview: https://covid-projections-git-mikelehen-share-charts.covidactnow.now.sh/internal/share-image/states/wa/chart/0

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,7 +1,11 @@
 import React from 'react';
 import CssBaseline from '@material-ui/core/CssBaseline';
-import { ThemeProvider, StylesProvider } from '@material-ui/core/styles';
+import {
+  ThemeProvider as MuiThemeProvider,
+  StylesProvider,
+} from '@material-ui/core/styles';
 import { BrowserRouter, Route, Redirect, Switch } from 'react-router-dom';
+import { ThemeProvider as ScThemeProvider } from 'styled-components';
 
 import LocationPage from 'screens/LocationPage/LocationPage';
 import HomePage from 'screens/HomePage/HomePage';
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -21,89 +25,95 @@ import theme from 'assets/theme';
...
   return (
-    <ThemeProvider theme={theme}>
-      <StylesProvider injectFirst>
-        <CssBaseline />
-        <BrowserRouter>
-          <AppBar />
-          <Switch>
-            <Route exact path="/" component={HomePage} />
+    <MuiThemeProvider theme={theme}>
+      <ScThemeProvider theme={theme}>
```
mikelehen, line 29: This seems to be the recommended way to use a theme with both material-ui and styled-components. 😬 
https://github.com/mui-org/material-ui/issues/10098#issuecomment-574162533

(note: the rest of this file is unchanged; turn on "Hide whitespace changes" to make the diff much better)

src/components/Charts/Charts.style.ts
```diff
@@ -18,10 +11,14 @@ const charts = {
...
 const tooltip = {
   width: '200px',
-  boxShadow: `3px 3px 5px ${palette.chart.tooltip.shadow}`,
   fontSizeTitle: '11px',
 };
 
+/** Gets the chart palette based on the current theme. */
+function palette(props: any) {
+  return props.theme.palette.chart;
+}
```
mikelehen, line 20: This is the core change.  Instead of accessing `palette` directly, we always access it via `props.theme.palette` so that it picks up the current theme.  The rest of the changes are just changing to use this helper function.

See [here](https://github.com/covid-projections/covid-projections/pull/825/files#diff-f5680937bf6b070548b0898693f6fc04R69) for how I override the theme for the share charts.

src/assets/theme/palette.tsx
```diff
@@ -4,6 +4,36 @@ const white = '#FFFFFF';
...
+  tooltip: {
+    background: colors.grey[900],
+    text: white,
+    shadow: colors.grey[500],
+  },
+  annotation: black,
+
+  // TODO(michael): Not loving this complexity. Perhaps we could just always use regionColor?
+  regionAnnotationStroke: (isActive: boolean, regionColor: string) =>
+    isActive ? regionColor : lightGray,
```
mikelehen, line 22: The UX mocks for the dark-mode chart always outline (stroke) the region labels with the region color, but the current light-mode charts only use the region color if the region is active, so I had to make this into a function so I could have different logic for the dark charts. 😦

src/components/Charts/Charts.style.ts
```diff
@@ -105,20 +103,23 @@ export const TextAnnotation = styled.g`
...
 `;
 
 export const RegionAnnotation = styled(TextAnnotation)<{ isActive: boolean }>`
   rect {
-    fill: ${props => (props.isActive ? props.color : palette.white)};
-    stroke: ${props => (props.isActive ? props.color : palette.lightGray)};
+    fill: ${props =>
+      props.isActive ? props.color : palette(props).background};
+    stroke: ${props =>
+      palette(props).regionAnnotationStroke(props.isActive, props.color)};
```
pnavarrc, line 117: Is there a way to know which theme is active? If not, I think that the function is good enough 

```js
stroke: ${props => (props.isActive || theme.isDarkMode) ? props.color : lightGrey}
```

mikelehen: Done, thanks!

PR #829, opened 2020-05-25
Create MetricChart helper component and use it in a few places.

mikelehen:
* Used for /compare and /all and will be used in my share charts pages.
* Not used in ChartsHolder yet because it'd require additional refactoring to make it worthwhile.

### DIFF
--- a/src/components/Charts/MetricChart.tsx
+++ b/src/components/Charts/MetricChart.tsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import { Projections } from 'common/models/Projections';
+import {
+  ChartRt,
+  ChartPositiveTestRate,
+  ChartICUHeadroom,
+  ChartContactTracing,
+  ChartFutureHospitalization,
+} from 'components/Charts';
+import { Metric } from 'common/metric';
+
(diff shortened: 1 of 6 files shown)


src/components/Charts/MetricChart.tsx
```diff
@@ -0,0 +1,60 @@
...
+  ChartContactTracing,
+  ChartFutureHospitalization,
+} from 'components/Charts';
+import { Metric } from 'common/metric';
+
+// TODO(michael): Rename to `Chart` once we get rid of existing (highcharts) Chart component.
+// TODO(michael): Update ChartsHolder to use this component instead of the individual chart components.
+export default function MetricChart({
+  metric,
+  height,
```
pnavarrc, line 16: Consider giving `height` a default value if you make it optional.

mikelehen: I'm a bit worried about specifying a default here since the charts each specify their own default too, and so it could get confusing if they got out-of-sync.  Maybe we should have a DEFAULT_CHART_HEIGHT constant somewhere that everything uses?  Anyway, I'll leave it as-is for now but am definitely open to cleaning this up somehow.

PR #833, opened 2020-05-26
Better UX for Preliminary Rt Data 

pnavarrc:
Close [Better UX for preliminary Rt data + add Confidence Intervals](https://trello.com/c/fQX9Jxxj/115-better-ux-for-preliminary-rt-data-add-confidence-intervals)

I tried to replicate the styles from [Figma - Preliminary Data for Rt](https://www.figma.com/file/xGLbdgTR348eQmmIpvBrvF/Preliminary-data-for-Rt?node-id=0%3A1), but the width of the tooltip wasn't big enough in some cases.

I also updated the tooltips for the other charts (see screenshots below) trying to keep the styles consistent with the new and existing design. The body of the future hospitalizations chart is the one that departs most from the others.

Note: Used the [triangle trick](https://css-tricks.com/snippets/css/css-triangle/) for the tooltip arrow. I only implemented the arrow down since that's the only one we need for now.

## Minor Changes
- Removed the `formatDate` function from `components/Charts/utils.ts` to use `formatDate` from `common/utils` instead. Added a default format to keep backwards compatibility.

### Growth Rate

**Confirmed**
image

**Preliminary**
image

### Positive Tests
image

### ICU Headroom Used
image

### Contacts Traced
image

### Future Hospitalizations
image

### DIFF
--- a/src/components/Charts/ChartFutureHospitalization.tsx
+++ b/src/components/Charts/ChartFutureHospitalization.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, ReactNode } from 'react';
+import React, { useContext } from 'react';
 import { ThemeContext } from 'styled-components';
 import { last, isDate } from 'lodash';
 import { extent as d3extent } from 'd3-array';
@@ -6,15 +6,15 @@ import { AxisBottom, AxisLeft } from '@vx/axis';
 import { LinePath } from '@vx/shape';
 import { ParentSize } from '@vx/responsive';
 import { scaleLinear, scaleTime } from '@vx/scale';
-import { Projections } from 'common/models/Projections';
 import { COLORS } from 'common';
-import { assert } from 'common/utils';
+import { Projections } from 'common/models/Projections';
+import { assert, formatDate, formatInteger } from 'common/utils';
 import BoxedAnnotation from './BoxedAnnotation';
 import ChartContainer from './ChartContainer';
 import RectClipGroup from './RectClipGroup';
 import Tooltip from './Tooltip';
+import * as TooltipStyle from './Tooltip.style';
 import { LegendMarker, LegendLine } from './Legend';
-import { formatDate, formatInteger } from './utils';
 import * as Style from './Charts.style';
 
 type Point = {
@@ -38,14 +38,6 @@ const getProjectionsPoints = (
   color: string,
   isBeds: boolean,
 ): PointProjections[] => points.map(p => ({ ...p, color, isBeds }));
-
-const getTooltipBody = (p: PointProjections): ReactNode => (
-  <span>
-    <b>{formatInteger(getY(p))}</b>{' '}
-    {p.isBeds ? 'beds available on' : 'hospitalizations expected by'}{' '}
-    <b>{formatDate(getDate(p), 'MMMM D')}</b>
-  </span>
-);
 
 const ChartFutureHospitalization = ({
   projections,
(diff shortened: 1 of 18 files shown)


pnavarrc:
[Better UX for preliminary Rt data + add Confidence Intervals](https://trello.com/c/fQX9Jxxj/115-better-ux-for-preliminary-rt-data-add-confidence-intervals)

src/components/Charts/ChartRt.tsx
```diff
@@ -100,17 +101,24 @@ const ChartRt = ({
...
+    const rtHigh = formatDecimal(getYAreaHigh(d), 2);
+    return (
+      <Tooltip
+        left={marginLeft + getXCoord(d)}
+        top={marginTop + getYCoord(d)}
+        title={formatDate(getDate(d), 'MMM D, YYYY')}
+        subtext={isConfirmed ? undefined : 'Data might change'}
+      >
+        <TooltipStyle.Body>
+          {`Rt ${isConfirmed ? '' : '~'}${formatDecimal(getRt(d), 2)}`}
```
mikelehen, line 116: `Rt` should be `Infection rate` still, probably.

src/components/Charts/ChartFutureHospitalization.tsx
```diff
@@ -117,7 +118,7 @@ const ChartFutureHospitalization = ({
 
   const renderTooltip = (p: PointProjections) => (
     <Tooltip left={marginLeft + getXCoord(p)} top={marginTop + getYCoord(p)}>
-      {getTooltipBody(p)}
+      <TooltipStyle.BodyMuted>{getTooltipBody(p)}</TooltipStyle.BodyMuted>
```
mikelehen, line 121: It seems weird to me that we're using `BodyMuted` instead of `Body` for all of the non-rt charts.  Is that because they get too big, otherwise?  Can we not just make the tooltip bigger or something?

src/components/Charts/ChartFutureHospitalization.tsx
```diff
@@ -8,13 +8,14 @@ import { ParentSize } from '@vx/responsive';
...
-import { assert } from 'common/utils';
+import { assert, formatDate } from 'common/utils';
 import BoxedAnnotation from './BoxedAnnotation';
 import ChartContainer from './ChartContainer';
 import RectClipGroup from './RectClipGroup';
 import Tooltip from './Tooltip';
+import * as TooltipStyle from './Tooltip.style';
 import { LegendMarker, LegendLine } from './Legend';
-import { formatDate, formatInteger } from './utils';
+import { formatInteger } from './utils';
```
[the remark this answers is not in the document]
pnavarrc: I moved the formatting functions for now, I will move the rest after we remove the Highcharts code to keep the diff small.

mikelehen:
Oh sorry!  I got mixed up and thought we were referring to the same file
two different ways.  Carry on!

src/components/Charts/ChartFutureHospitalization.tsx
pnavarrc: Updated - I think we need to make the tooltip for hospitalizations more consistent with the others so we don't need to tweak it here.

pnavarrc:
> Oh sorry!  I got mixed up and thought we were referring to the same file
two different ways.  Carry on!

Those functions were being used outside the context of charts, I wanted to move them anyways :)

PR #839, opened 2020-05-27
Add getMetricValue() and ALL_METRICS helpers.

mikelehen:
This makes some of the chart sharing code cleaner.

### DIFF
--- a/src/common/models/Projection.ts
+++ b/src/common/models/Projection.ts
@@ -199,9 +199,7 @@ export class Projection {
       summaryWithTimeseries.actuals.cumulativeDeaths;
     this.currentCumulativeCases =
       summaryWithTimeseries.actuals.cumulativeConfirmedCases;
-    this.currentContactTracerMetric = this.contractTracers
-      .filter(x => x !== null)
-      .slice(-1)[0];
+    this.currentContactTracerMetric = this.lastValue(this.contractTracers);
   }
 
   get currentContactTracers() {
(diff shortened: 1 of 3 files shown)


src/common/models/Projection.ts
```diff
@@ -199,9 +199,7 @@ export class Projection {
       summaryWithTimeseries.actuals.cumulativeDeaths;
     this.currentCumulativeCases =
       summaryWithTimeseries.actuals.cumulativeConfirmedCases;
-    this.currentContactTracerMetric = this.contractTracers
-      .filter(x => x !== null)
-      .slice(-1)[0];
```
mikelehen, line 204: This ended up setting currentContactTracingMetric to `undefined` instead of `null` so I changed it to use the existing `lastValue` helper.

goldblatt: I can't remember why i used this initially I think there was some like undefined vs null situation

src/common/models/Projections.ts
```diff
@@ -62,41 +63,41 @@ export class Projections {
...
   } {
-    const projection = this.primary;
-    if (!projection)
-      return {
-        rt_level: Level.UNKNOWN,
-        hospitalizations_level: Level.UNKNOWN,
-        test_rate_level: Level.UNKNOWN,
-        contact_tracing_level: Level.UNKNOWN,
-      };
-
```
goldblatt, line 79: hmm don't we still need this case?

mikelehen: GetMetricValue() returns `null` if `projections.primary` is null, and `getLevel()` returns UNKNOWN if you pass a `null` metric value, so we don't need to special case this `!projection` case anymore.

goldblatt: cool cool

PR #850, opened 2020-05-28
Implements the fourth color

pnavarrc:
This PR adds a fourth zone between the Medium and High zones.

Copy updates based on [CAN Phase 3 Webapp Text](https://docs.google.com/document/d/1Jsj4zOksCJcgjQLUo3rwVAobEUkGLFukD1K71IqVNpo/edit#), last updated on Jun 3 at 5:30 pm PDT.

**Preview**: https://covid-projections-git-pablo-add-new-zone.covidactnow.now.sh/

## TODO

- [x] UX Changes (from [Slack](https://covidactnow.slack.com/archives/C011CKR2B2B/p1591042549003000))
  - [x] Remove grey boxes around labels
  - [x] Changed the axes fonts from bold to medium.  They were looking too strong
  - [x] Update colors for the levels, see [Figma mocks](https://www.figma.com/file/IsocRlNjHrRBdbDMe5aKDN/Charts?node-id=6%3A0) for the colors.
  - [x] Move the map legend from the bottom to the top of the map
- [x] Final copy review from [CAN Phase 3 Webapp Text](https://docs.google.com/document/d/1Jsj4zOksCJcgjQLUo3rwVAobEUkGLFukD1K71IqVNpo/edit#) - Jun 3, 5:30pm
- [x] Fix the link for "Experts recommend that at least 90%..."
- [x] Change the number of required tracers per case to **5**
- [x] The legend on the map needs to be adjusted on mobile

## Notes
We might want to rename the levels to be more copy-agnostic, for example, `Level.HIGH` now corresponds to "Critical" and `LEVEL.MEDIUM_HIGH` corresponds to "High". This is particularly confusing for contact tracing, where Level.LOW corresponds to "Critical" and `Level.MEDIUM` corresponds to "Low".

In the copy document, the levels are categorized as Green, Yellow, Orange, and Red, maybe we should use that for the levels?


pnavarrc:
[Implement 4th color for ChartZone](https://trello.com/c/0AwaS8Yk/172-implement-4th-color-for-chartzone)

src/common/colors.js
```diff
@@ -41,6 +41,7 @@ export const COLOR_MAP = {
 export const LEVEL_COLOR = {
   [Level.LOW]: COLOR_MAP.GREEN.BASE,
   [Level.MEDIUM]: COLOR_MAP.ORANGE.BASE,
+  [Level.MEDIUM_HIGH]: '#ff6d00', // Material UI orange A700
```
pnavarrc, line 44: **TODO**: Add to `COLOR_MAP`

src/common/metrics/case_growth.ts
```diff
@@ -36,6 +39,13 @@ export const CASE_GROWTH_RATE_LEVEL_INFO_MAP: LevelInfoMap = {
     color: COLOR_MAP.ORANGE.BASE,
     detail: () => SHORT_DESCRIPTION_MEDIUM,
   },
+  [Level.MEDIUM_HIGH]: {
+    level: Level.MEDIUM_HIGH,
+    upperLimit: LIMIT_MEDIUM_HIGH,
+    name: MEDIUM_HIGH_NAME,
+    color: '#ff6d00',
```
mikelehen, line 46: Huh.  I wonder if there's a good reason we don't use LEVEL_COLOR constants for these...  🤔  Seems like something we should maybe clean up at some point (not in this PR).

src/components/Charts/ChartZones.tsx
```diff
@@ -94,7 +94,7 @@ const ChartZones = ({
 
   const lastPoint = last(data);
   const lastPointY = getY(lastPoint);
-  const lastPointZone = getZoneByValue(lastPointY, zones);
+  const lastPointRegion = getZoneByValue(lastPointY, zones);
```
mikelehen, line 97: Hrm... I'm not 100% sure why you renamed this.  It seems like the actual type is `LevelInfo` which I think is the type we use for zones, but TBH I'm a little fuzzy on levels vs zones vs regions. :) Perhaps another thing to consider cleaning up at some point (not in this PR).

pnavarrc: I need to rename it back - I tried implementing `getRegionByValue` since regions already have the upper and lower limit, but then decided against it because regions don't have an explicit sense of order. For example, in `getRegionByValue`, I would have to do

```js
if (value <= region[0].valueFrom) {
  return region[0];
}
```

which implicitly assumes that `region[0]` corresponds to "Low". The code for zones is more explicit in that regard

```js
if (value <= zones[Level.LOW].upperLimit) {
  return zones[Level.LOW];
}
```

so I changed it back to use `zones`. I'm going to rename it right away, it's a pretty small change.

src/common/metrics/case_growth.ts
pnavarrc: I'm going to add that color to `COLOR_MAP` for now as `ORANGE_DARK` so the color is not hardcoded here, but yes, I think that ideally, we should use `LEVEL_COLOR`. I can fix that on a separate PR but want to check with @goldblatt first

src/common/models/Projections.ts
```diff
@@ -119,6 +119,11 @@ export class Projections {
       reverseList.some(level => level === Level.LOW)
     ) {
       return Level.HIGH;
+    } else if (
+      levelList.some(level => level === Level.MEDIUM_HIGH) ||
+      reverseList.some(level => level === Level.MEDIUM_HIGH)
+    ) {
+      return Level.MEDIUM_HIGH;
```
pnavarrc, line 126: @mikelehen @maxwhenderson we need to double-check if this is what we want

src/components/Charts/utils.ts
```diff
@@ -238,32 +238,49 @@ export const getChartRegions = (
...
+
+  if (isBetween(zones[Level.LOW], zones[Level.MEDIUM], value)) {
+    return zones[Level.MEDIUM];
+  }
+
+  if (isBetween(zones[Level.MEDIUM], zones[Level.MEDIUM_HIGH], value)) {
+    return zones[Level.MEDIUM_HIGH];
+  }
+
+  return zones[Level.HIGH];
```
pnavarrc, line 269: I _think_ that this code almost duplicates https://github.com/covid-projections/covid-projections/pull/850/files?file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#diff-f81cf374a211ca4db3c5e39b93306773R65, I will look into that after we merge to see if we can centralize the logic (@mikelehen 's code is much cleaner)

src/common/colors.js
pnavarrc: Done

src/common/models/Projections.ts
pnavarrc: Updated!

src/common/metrics/case_growth.ts
pnavarrc: I think it's because the colors are reversed for contact tracing https://github.com/covid-projections/covid-projections/pull/850/files#diff-b88c1dc9a0f512264e2d89dc4032053a

yvonnehung:
Here are some thoughts on improving the charts how that we have 4 colors:
- new colors on [Figma](https://www.figma.com/file/iCoitMqZ6eVZZs1A9QXhIv/Wireframes-v2?node-id=229%3A48)
- Shorten label copy to "hi, med, lo" when possible, pending copy.
- Auto adjust y-axis range to take up the vertical space of the chart. Right now there are many charts with lots of white space. For example, in R(t) charts,  the lower bound can be  0.1 below the actual lowest point, OR 0.1 below the green area just to ensure we show the line for the goal.

yvonnehung:
Here are some thoughts on improving the charts how that we have 4 colors:

- new colors on Figma
- Shorten label copy to "hi, med, lo" when possible, pending copy.

src/common/metrics/location_summary.ts
```diff
@@ -3,6 +3,7 @@ import { Level, LevelInfoMap } from 'common/level';
 
 const LOW_NAME = 'Low';
 const MEDIUM_NAME = 'Moderate';
+const MEDIUM_HIGH_NAME = 'Moderate High';
```
mikelehen, line 6: Hrm...  It took me a while to figure out where these get used (in SocialLocationPreview)...  Can we add a comment mentioning that?

Also, have you confirmed this is the copy we want?  It seems a bit inconsistent with the rest of the copy (low/medium/high/critical)

src/common/models/Projections.ts
```diff
@@ -124,6 +124,11 @@ export class Projections {
       reverseList.some(level => level === Level.LOW)
     ) {
       return Level.HIGH;
+    } else if (
```
mikelehen, line 127: Hrm...  I had suggested a change to this code [here](https://github.com/covid-projections/covid-projections/commit/5729b6f44b3c7771bd5a4680974bb6874c82534d#r39517842) bit it may have gotten hidden due to force-pushing...

Basically I think we need to swap MEDIUM_HIGH and MEIDUM when using the reverseList.  As-is, for example, Kansas is getting ranked yellow when it should be orange (because contact tracing is "low" which is orange):

image

And on the flip side, Wyoming is orange even though it should be yellow based on the metrics:
image

FWIW- I think this is all super confusing and should probably be reworked somehow.

src/common/metrics/location_summary.ts
pnavarrc: Good catch! I also missed the MEDIUM_HIGH label on the `SocialLocationPreview` 🤦‍♂️  - I will add a note.

pnavarrc: I updated them to be consistent. I changed that based on the screenshot on the [copy document](https://docs.google.com/document/d/1Jsj4zOksCJcgjQLUo3rwVAobEUkGLFukD1K71IqVNpo/edit#) before the 4th color updates, but forgot about it after the updates

image

I flagged this to @maxwhenderson in the document to double-check what we want.

src/common/models/Projections.ts
pnavarrc: Updated! I probably lost the change with a bad rebase. I added it back and confirmed that it works well for the states above.

src/common/colors.js
```diff
@@ -14,21 +14,21 @@ export default {
...
 export const COLOR_MAP = {
   ORANGE: {
-    BASE: '#FFAD16',
+    BASE: '#FFC900',
     LIGHT: '#FFC020',
     DARK: '#933500',
   },
   ORANGE_DARK: {
-    BASE: '#FF6D00', // Material UI orange A700
+    BASE: '#FF9600', // Material UI orange A700
```
mikelehen, line 22: This comment probably isn't right anymore. 😁

src/common/colors.js
```diff
@@ -14,18 +14,21 @@ export default {
 
 export const COLOR_MAP = {
   ORANGE: {
-    BASE: '#FFAD16',
+    BASE: '#FFC900',
     LIGHT: '#FFC020',
     DARK: '#933500',
   },
+  ORANGE_DARK: {
+    BASE: '#FF9600',
```
goldblatt, line 22: why doesnt this need light/dark?

pnavarrc: The designs don't have light or dark versions of this color, do we need them?

mikelehen: I don't really know the history of BASE/LIGHT/DARK, but In general, I think we're trying to avoid having multiple shades of a color on the site (I did some unifications of the greens a while ago at Josh's request).

PR #869, opened 2020-05-31
Charts - Improve x-axis ticks

pnavarrc:
This PR improves the generation of tick values for the charts. The charts were using the default ticks, which were not optimal for the timespan of our charts. Instead, I'm forcing ticks every 2 weeks. I also moved the axes to their own module so the logic is centralized and to simplify the code on the charts.

**Updated!**: I updated the logic so it will put ticks every 14 days on desktop and 28 days for mobile, and changed the date format so it's always like **May 31**.

## Desktop

image

### Mobile

image

Can @mikelehen or @chasulin review?


pnavarrc:
oops, I forgot to put back handling for mobile, will update this evening, thanks @mikelehen !

pnavarrc:
Updated @mikelehen !

src/components/Charts/Axis.tsx
```diff
@@ -1,10 +1,18 @@
 import React from 'react';
-import { timeWeek } from 'd3-time';
+import { timeDay } from 'd3-time';
```
pnavarrc, line 2: Changed it to use days to make sure that the first tick matches the first day in the range, when using `timeWeek` it was rounding up to the first Sunday after the first date.

pnavarrc:
thanks for the review @mikelehen! I changed it to month divisions for mobile as it used to be, merging!

mikelehen:
Awesome, thank you!  😁

PR #872, opened 2020-06-02
Embed 2.0

chasulin:
Implements new embed design. Clicking a metric redirects to the location page + scrolls to the selected metric's chart. Clicking the logo redirects to the homepage. 

State example: https://covid-projections-2odcj5nj6.now.sh/embed/us/ks
County example: https://covid-projections-2odcj5nj6.now.sh/embed/us/ny/county/suffolk_county
Preview modal viewable via "embed on your website" link

Still to do:
- [x] Add click functionality for each metric

Screen Shot 2020-06-05 at 4 01 40 PM


src/components/ShareBlock/EmbedPreview.style.js
```diff
@@ -35,7 +35,7 @@ export const CenteredModal = styled(Modal)`
 
 export const EmbedPreviewScrollContainer = styled(Box)`
   max-height: 95vh;
-  overflow-y: hidden;
+  overflow-y: scroll;
```
adpirz, line 38: Would check with @mikelehen ; was able to show unwanted scroll bars in his view

src/common/utils/hooks.tsx
```diff
@@ -1,8 +1,8 @@
 import { useLocation, useParams } from 'react-router-dom';
 
 // TODO: Mv to enums?
-const EMBED_WIDTH = 350;
-const EMBED_HEIGHT = 700;
+const EMBED_WIDTH = 375;
+const EMBED_HEIGHT = 393;
```
adpirz, line 6: Would check the js snippet to ensure that we're using the same dimensions when we dynamically create

src/components/ShareBlock/EmbedPreview.style.js
mikelehen: Looks okay to me.

FWIW- To duplicate my config, you can either plug a mouse into your macbook, use Windows, or [set this option in OSX](https://osxdaily.com/2011/08/03/show-scroll-bars-mac-os-x-lion/).

src/common/utils/hooks.tsx
mikelehen: Yeah, a quick test suggests something is off.

image

src/components/ShareBlock/EmbedPreview.style.js
chasulin: Ah ok. I had changed it to scroll because when the embed preview + the snippets are stacked in a column and the browser is short, the snippets aren't accessible. But I guess if this wasn't of concern before, I can keep it as it was if that's preferred

mikelehen: I think as long as it's not causing problems, your change is fine.  Though I'm not sure what "when the browser is short" means, exactly.  As long as it's embedded as an iframe with a fixed width / height, can we assume that won't be an issue?

For context, the issue we had previously was that for some reason, the content would always be _slightly_ too big for the container (by a pixel or something) and so you'd end up with a dummy scrollbar all the time.

chasulin: Got it. Yeah, this is the wrapper for the entire preview modal that opens when you first click "embed on your website" in the CAN footer (the entire white modal below), not for the iframe itself

src/common/utils/hooks.tsx
chasulin: Ooh yes! Fixed in embed.js

mikelehen:
Are we intentionally adding all the vertical whitespace between the metrics in the share card?

BEFORE:
image

AFTER:
image

This means we don't see all the metrics on the share image:
https://covid-projections-git-casulin-new-embed.covidactnow.now.sh/internal/share-image/states/wa
image

chasulin:
Fixed! Good catch, thank you! That was an unwanted by-product of changing styles with prop 'condensed'

mikelehen:
👍

PR #877, opened 2020-06-04
Fix several issues with conflating null and '0', leading to our metrics disappearing when the metric goes to 0.

mikelehen:
In particular, this fixes:
1. The Montana ICU chart truncating at May 18, due to the value going to 0%.
2. The Hawaii ICU chart disappearing entirely because it was all 0's.

**Before:**
image

**After:**
image

**Before:**
image

**After:**
image

### DIFF
--- a/src/common/models/Projection.ts
+++ b/src/common/models/Projection.ts
@@ -550,7 +550,7 @@ export class Projection {
           row &&
           this.totalICUCapacity &&
           this.totalICUCapacity > 0 &&
-          row.ICUBedsInUse > 0
+          row.ICUBedsInUse != null
         ) {
           const predictedNonCovidPatientsAtDate =
             this.totalICUCapacity! *
--- a/src/common/models/Projection.ts
+++ b/src/common/models/Projection.ts
@@ -550,7 +550,7 @@ export class Projection {
           row &&
           this.totalICUCapacity &&
           this.totalICUCapacity > 0 &&
-          row.ICUBedsInUse != null
+          row.ICUBedsInUse !== null
         ) {
           const predictedNonCovidPatientsAtDate =
             this.totalICUCapacity! *
--- a/src/components/Charts/MetricChart.tsx
+++ b/src/components/Charts/MetricChart.tsx
@@ -21,7 +21,7 @@ export default function MetricChart({
   height?: number;
 }) {
   const projection = projections.primary;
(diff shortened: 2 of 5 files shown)


src/components/Charts/MetricChart.tsx
```diff
@@ -21,37 +21,35 @@ export default function MetricChart({
   height?: number;
 }) {
   const projection = projections.primary;
-  if (!projection) {
+  if (projection == null || projections.getMetricValue(metric) == null) {
```
goldblatt, line 24: this will catch undefined too do you mean ===

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -322,29 +322,29 @@ export function getChartData(
...
   const rtRangeData =
-    projection &&
-    projection.rt &&
-    projection.getDataset('rtRange').map(d => ({
-      x: d.x,
-      y: d.y?.rt,
-      low: d.y?.low,
-      hi: d.y?.high,
-    }));
+    projection?.rt == null
```
goldblatt, line 325: same with the == catching undefine dhere

src/common/models/Projection.ts
```diff
@@ -550,7 +550,7 @@ export class Projection {
           row &&
           this.totalICUCapacity &&
           this.totalICUCapacity > 0 &&
-          row.ICUBedsInUse > 0
+          row.ICUBedsInUse != null
```
goldblatt, line 553: this will catch undefined as well (here and below).

src/components/LocationPage/ChartsHolder.tsx
mikelehen: So here it is definitely intentional / necessary because `projection?.rt` will return `undefined` if `projection` is `null` (It's just how the optional chaining operator works. I don't love it).

src/common/models/Projection.ts
mikelehen: Eh. It was sorta' intentional because I just don't trust our code not to accidentally have `undefined`s in it right now (since it's half JavaScript, etc.) but it _seems_ like `!== null` is sufficient.  Changed.

src/components/Charts/MetricChart.tsx
mikelehen: Again, half intentional but I went ahead and made it more strict.  Seems to still work.

PR #878, opened 2020-06-04
Ignore negative deltas in cumulatives. Fixes test positive rate for several states.

mikelehen:
A number of states seemingly made large corrections to their test reporting in the last week or two resulting in large _negative_ deltas in the day/day numbers, which was wagging the positive test rate metric around...

This PR fixes our deltasFromCumulatives() code to always ignore negative deltas (replace them with a null) and then consequentially our calcTestPositiveRate() needed to deal with nulls properly.

A bunch of states were affected modestly, but a few had big changes, e.g.:

**DC Before**
image

**DC After**
image

---
**GA Before**
image

**GA After**
image

---
**MO Before**
image

**MO After**
image

### DIFF
--- a/src/common/models/Projection.ts
+++ b/src/common/models/Projection.ts
@@ -481,13 +481,16 @@ export class Projection {
     );
 
     return dailyPositives.map((dailyPositive, idx) => {
-      const positive = dailyPositive || 0;
-      const negative = dailyNegatives[idx] || 0;
-      const total = positive + negative;
+      const positive = dailyPositive;
+      const negative = dailyNegatives[idx];
       // If there are no negatives (but there are positives), then this is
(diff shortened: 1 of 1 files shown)


src/common/models/Projection.ts
```diff
@@ -481,13 +481,16 @@ export class Projection {
     );
 
     return dailyPositives.map((dailyPositive, idx) => {
-      const positive = dailyPositive || 0;
-      const negative = dailyNegatives[idx] || 0;
-      const total = positive + negative;
+      const positive = dailyPositive;
+      const negative = dailyNegatives[idx];
```
mikelehen, line 485: Note: Previously the values were never null, so we didn't actually hit the `|| 0` case.

PR #885, opened 2020-06-05
Charts - Fix missing label on hospitalizations chart

pnavarrc:
Close [193](https://trello.com/c/ALh0HTSt/193-make-sure-capacity-horizontal-line-is-labeled-in-projection-charts) - Make sure that the capacity horizontal line is labeled in projections chart

The vertical scale maps the min and max values on the data to the min and max values of the inner chart dimensions in pixels. If the number of beds is the maximum data value, the label ends up outside the chart because it's above the line.

This PR moves the label outside the clipping path boundaries to avoid cropping the label and adjusts the chart margins a bit so it fits. I also adjusted where the minimum value goes (`chartHeight - MAX_LINE_WIDTH`) to avoid hiding the lines if they are too close to the x-axis.

**Before** https://covidactnow.org/us/ak?s=41740
image

**After** https://covid-projections-git-pablo-fix-hospitalization-bed-count.covidactnow.now.sh/us/ak?s=41740
image

Can @chasulin review?

### DIFF
--- a/src/components/Charts/ChartFutureHospitalization.tsx
+++ b/src/components/Charts/ChartFutureHospitalization.tsx
@@ -27,6 +27,8 @@ type PointProjections = Point & {
   isBeds: boolean;
 };
 
+const MAX_LINE_WIDTH = 3;
+
 const getDate = (p: Point) => new Date(p.x);
 const getY = (p: Point) => p.y;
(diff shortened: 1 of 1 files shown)


pnavarrc:
[Make sure capacity horizontal line is labeled in projection charts](https://trello.com/c/ALh0HTSt/193-make-sure-capacity-horizontal-line-is-labeled-in-projection-charts)

PR #888, opened 2020-06-06
allow users to sign to email list for alerts or to generic newsletter

goldblatt:
some trickery to get the form to work right, but allow users to sign up to different states/counties

### DIFF
--- a/src/common/locations.ts
+++ b/src/common/locations.ts
@@ -1,7 +1,29 @@
 /** Helpers for dealing with the State / Counties dataset. */
 import US_STATE_DATASET from 'components/MapSelectors/datasets/us_states_dataset_01_02_2020.json';
+import { each, uniqBy } from 'lodash';
 
 // TODO(michael): Add types, move more common code here.
+
+export function getLocationNames() {
+  const locations = US_STATE_DATASET.state_dataset.map(state => {
+    return {
+      ...state,
+      full_fips_code: state.state_fips_code,
+    };
+  });
+
+  each(US_STATE_DATASET.state_county_map_dataset, (value, key) => {
+    locations.push(
+      ...value.county_dataset.map(county => {
+        return { ...county, state: county.state_code };
+      }),
+    );
+  });
+
+  // Currently check for the unique fips so there aren't duplicates, but maybe
+  // match the way that the homepage gets unique values
+  return uniqBy(locations, 'full_fips_code');
+}
 
 export function findCountyByFips(fips: string) {
   // NYC HACK.
--- a/src/common/locations.ts
+++ b/src/common/locations.ts
@@ -14,9 +14,11 @@ export function getLocationNames() {
 
   each(US_STATE_DATASET.state_county_map_dataset, (value, key) => {
     locations.push(
-      ...value.county_dataset.map(county => {
-        return { ...county, state: county.state_code };
-      }),
+      ...value.county_dataset
+        .filter(county => !county.county.includes('/'))
+        .map(county => {
+          return { ...county, state: county.state_code };
+        }),
     );
   });
 
--- a/src/common/locations.ts
+++ b/src/common/locations.ts
@@ -1,6 +1,6 @@
 /** Helpers for dealing with the State / Counties dataset. */
 import US_STATE_DATASET from 'components/MapSelectors/datasets/us_states_dataset_01_02_2020.json';
-import { each, uniqBy } from 'lodash';
+import { each } from 'lodash';
 
 // TODO(michael): Add types, move more common code here.
 
@@ -22,9 +22,7 @@ export function getLocationNames() {
     );
   });
 
-  // Currently check for the unique fips so there aren't duplicates, but maybe
-  // match the way that the homepage gets unique values
-  return uniqBy(locations, 'full_fips_code');
+  return locations;
 }
 
 export function findCountyByFips(fips: string) {
--- a/src/components/Newsletter/Newsletter.js
+++ b/src/components/Newsletter/Newsletter.js
@@ -1,12 +1,20 @@
 import React from 'react';
-import { StyledNewsletter } from './Newsletter.style';
+import { StyledNewsletter, InputHolder } from './Newsletter.style';
+import Chip from '@material-ui/core/Chip';
+import Autocomplete from '@material-ui/lab/Autocomplete';
+import TextField from '@material-ui/core/TextField';
+import { each } from 'lodash';
+import US_STATE_DATASET from 'components/MapSelectors/datasets/us_states_dataset_01_02_2020';
 
 class Newsletter extends React.Component {
   constructor() {
     super();
     this.form = null;
     this.emailInput = null;
+    this.alertsSelectionArray = [];
+    this.state = { alertSignUps: '' };
     this.submitForm = this.submitForm.bind(this);
+    this.handleSelectChange = this.handleSelectChange.bind(this);
   }
 
   submitForm(e) {
@@ -26,7 +34,6 @@ class Newsletter extends React.Component {
       .then(response => {
         response.text().then(text => {
           this.form.action = text;
-          console.log(new FormData(this.form));
           this.form.submit();
         });
       })
@@ -35,8 +42,32 @@ class Newsletter extends React.Component {
       });
   }
 
+  handleSelectChange = selectedOption => {
+    this.alertsSelectionArray = selectedOption;
+    this.setState({
+      alertSignUps: this.alertsSelectionArray
+        .map(item => item.full_fips_code)
+        .join(','),
+    });
+  };
+
   render() {
     const { stateId, county } = this.props;
+
+    const stateDataset = US_STATE_DATASET.state_dataset.map(state => {
+      return {
+        ...state,
+        full_fips_code: state.state_fips_code + '000',
+      };
+    });
+    const countyDataset = [];
+
+    each(US_STATE_DATASET.state_county_map_dataset, (value, key) => {
+      countyDataset.push(...value.county_dataset);
+    });
+
+    const autocomplete_options = stateDataset.concat(countyDataset);
+
     return (
       <StyledNewsletter>
         <form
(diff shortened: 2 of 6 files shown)


src/components/Newsletter/Newsletter.js
```diff
@@ -1,12 +1,20 @@
...
+import TextField from '@material-ui/core/TextField';
+import { each } from 'lodash';
+import US_STATE_DATASET from 'components/MapSelectors/datasets/us_states_dataset_01_02_2020';
 
 class Newsletter extends React.Component {
   constructor() {
     super();
     this.form = null;
     this.emailInput = null;
+    this.alertsSelectionArray = [];
```
mikelehen, line 14: Can we initialize this with the current state / county page that you're on?

src/components/Newsletter/Newsletter.js
```diff
@@ -35,50 +42,131 @@ class Newsletter extends React.Component {
...
+    });
+  };
+
   render() {
     const { stateId, county } = this.props;
+
+    const stateDataset = US_STATE_DATASET.state_dataset.map(state => {
+      return {
+        ...state,
+        full_fips_code: state.state_fips_code + '000',
```
mikelehen, line 60: FWIW, adding '000' seems nonstandard.  The FIPS code for Washington is just `53`, not `53000` and this is what we use e.g. in our [ingested data sets](https://github.com/covid-projections/covid-data-public/blob/master/data/test-and-trace/state_data.csv).  It maybe doesn't matter _too_ much as long as we know this is what we're using and are consistent everywhere, but it could be a surprise / cause bugs if somebody new comes to this code and assumes we're using normal state FIPS codes.

Since it would simplify our code to _not_ add '000', it seems strictly better to use the raw state FIPS code.

src/components/Newsletter/Newsletter.js
```diff
@@ -35,50 +42,131 @@ class Newsletter extends React.Component {
...
-            id="fieldjlkulhk"
-            maxLength="200"
-            name="cm-f-jlkulhk"
           />
+          <InputHolder>
+            <input
+              type="checkbox"
+              value="wurhhh"
+              id="wurhhh"
+              name="cm-ol-wurhhh"
```
mikelehen, line 103: Should we default this to checked so people will opt-in to our daily newsletter by default?  Or no?

src/components/Newsletter/Newsletter.js
```diff
@@ -35,50 +42,131 @@ class Newsletter extends React.Component {
...
             readOnly
             aria-label="county"
+            id="fieldjrdtwd"
+            maxlength="200"
+            name="cm-f-jrdtwd"
             value={county || ''}
-            id="fieldjlkulhk"
-            maxLength="200"
-            name="cm-f-jlkulhk"
           />
```
mikelehen, line 97: Do we still want / need these state / county custom fields anymore?  It seems like the new alert_list_csv field essentially supersedes them.

src/components/Newsletter/Newsletter.js
```diff
@@ -35,50 +42,131 @@ class Newsletter extends React.Component {
...
-            name="cm-wurhhh-wurhhh"
-            required=""
-            type="email"
+            hidden
+            aria-label="alert_list_csv"
+            id="fieldjrdtwy"
+            maxlength="200"
+            name="cm-f-jrdtwy"
+            onChange={value => {
+              console.log(value);
```
mikelehen, line 118: maybe remove before submitting?

src/components/Newsletter/Newsletter.js
```diff
@@ -35,50 +42,131 @@ class Newsletter extends React.Component {
...
+            }}
+            value={this.state.alertSignUps}
+          />
+          <Autocomplete
```
mikelehen, line 122: Nit: I'd like to be able to just type "Washi<enter>" to add Washington, but I have to actually click on it or press down arrow and then enter or whatever...  Maybe if we set `autoHighlight` and/or `autoSelect` it'll work?

mikelehen: Maybe kinda' Related: Right now, pressing <enter> actually seems to submit the form which maybe we don't want?

src/components/Newsletter/Newsletter.js
```diff
@@ -35,50 +42,131 @@ class Newsletter extends React.Component {
...
+        full_fips_code: state.state_fips_code + '000',
+      };
+    });
+    const countyDataset = [];
+
+    each(US_STATE_DATASET.state_county_map_dataset, (value, key) => {
+      countyDataset.push(...value.county_dataset);
+    });
+
+    const autocomplete_options = stateDataset.concat(countyDataset);
```
mikelehen, line 69: As I mentioned in chat, the overall site search gives me much better results if I try to search for "King County".  It filters out irrelevant stuff and sorts by population (or something) so my result is at the top.  I think we should try to do the same thing here.

image

image

src/components/Newsletter/Newsletter.js
```diff
@@ -35,50 +42,131 @@ class Newsletter extends React.Component {
...
+                : option.state
+            }
+            renderTags={(tagValue, getTagProps) =>
+              tagValue.map((option, index) => (
+                <Chip
+                  label={
+                    option.county
+                      ? `${option.county}, ${option.state_code}`
+                      : option.state
+                  }
```
mikelehen, line 144: I'm curious how getOptionLabel / renderTags are used. They seem kinda' redundant...  But I assume there's a good reason we need both.

src/components/Newsletter/Newsletter.js
```diff
@@ -35,50 +42,131 @@ class Newsletter extends React.Component {
...
+                  {...getTagProps({ index })}
+                />
+              ))
+            }
+            renderInput={params => (
+              <TextField {...params} placeholder="Enter alert locations" />
```
mikelehen, line 150: FWIW- It's a bit weird that "Enter alert locations" continues to show up even after you have selected 1 or more locations, but I guess it's fine.

image

src/components/Newsletter/Newsletter.js
```diff
@@ -35,50 +42,131 @@ class Newsletter extends React.Component {
...
+
     return (
       <StyledNewsletter>
         <form
           ref={f => (this.form = f)}
           className="js-cm-form"
           id="subForm"
           method="post"
-          data-id="2BE4EF332AA2E32596E38B640E90561930C9C3A433D015D9D4BD88E99175E51395EF5EBFFD527179E032AC15455BB1208D87A6CE87843E524B0EA520CBFF446E"
+          data-id="2BE4EF332AA2E32596E38B640E905619E90CD5DAC48A878CDEBFFE3B420D8CD24E4AEABAB52A4CE3526219C7A966AE965B84F99C823C89EFF1F01B28DE4F975E"
```
mikelehen, line 78: FWIW- Since you had to puzzle out how these IDs and hidden form element things work, it might be worth adding some comments somewhere in this file to help the next unfortunate soul who has to mess with this stuff. 😛

src/components/Newsletter/Newsletter.js
goldblatt: i think it's probably useful to have the signup page idk it might break the form to remove them so i'm gonna leave

src/components/Newsletter/Newsletter.js
goldblatt: yes default opted in

src/components/Newsletter/Newsletter.js
goldblatt: ah one is the label and one is the input itself i think

src/components/Newsletter/Newsletter.js
goldblatt: default is opted out my b

src/components/Newsletter/Newsletter.js
goldblatt: done added some comments above

src/components/Newsletter/Newsletter.js
goldblatt: done

src/components/Newsletter/Newsletter.js
goldblatt: hmmmm i'm not seeing that >.<

src/components/Newsletter/Newsletter.js
goldblatt: cool fixed

src/components/Newsletter/Newsletter.js
goldblatt: I switched this to use a unique list (so solves some of the issue-- I guess the thing is that to use this in autocomplete + chip + other stuff it's just a lil easier to leave it like this now)

src/common/locations.ts
```diff
@@ -1,8 +1,30 @@
...
+  each(US_STATE_DATASET.state_county_map_dataset, (value, key) => {
+    locations.push(
+      ...value.county_dataset.map(county => {
+        return { ...county, state: county.state_code };
+      }),
+    );
+  });
+
+  // Currently check for the unique fips so there aren't duplicates, but maybe
+  // match the way that the homepage gets unique values
```
mikelehen, line 24: Isn't this going to be kinda' arbitrary?

E.g. now "Hancock County, IN" isn't in the list because instead there's a weird "Hancock County / Marion County, IN" entry (which I guess is using the Hancock County FIPS)...

I think we either need to:
1) Actually use the same logic as the homepage, or
2) Just delete these "Foo County / Bar County" entries from the dataset so we don't need special handling.  I think these existed in order to allow you to search by city (for cities that span multiple counties), but we don't really support that anymore.

src/components/Newsletter/Newsletter.js
```diff
@@ -51,25 +52,35 @@ class Newsletter extends React.Component {
...
+    const { stateId, county } = this.props;
+    this.defaultValues =  this.autocompleteOptions.filter(location => {
+      const matching_state =
+        location.state_code === stateId && location.full_fips_code.length === 2;
+      let matching_county = false;
+      if (county) {
+        matching_county =
+          location.county === county &&
+          location.state_code === stateId &&
+          location.full_fips_code.length === 5;
```
mikelehen, line 69: is this length check necessary?

src/components/Newsletter/Newsletter.js
```diff
@@ -51,25 +52,35 @@ class Newsletter extends React.Component {
...
     });
 
-    const autocomplete_options = stateDataset.concat(countyDataset);
-
     return (
       <StyledNewsletter>
+        {/* This form comes from the signup form builder
+        (https://covidactnow.createsend.com/subscribers/signupformbuilder/58a40b3258be0d56) within the subscribers page
+        From there choose the option where you add code to your website without css.
+        To update the grab the data-id, classNames, ids and names for each of the inputs in order to subscribe users.
```
mikelehen, line 79: "To update the grab the ..." ?

src/components/Newsletter/Newsletter.js
```diff
@@ -114,21 +125,20 @@ class Newsletter extends React.Component {
             id="fieldjrdtwy"
             maxlength="200"
             name="cm-f-jrdtwy"
-            onChange={value => {
-              console.log(value);
-            }}
+            onChange={value => {}}
```
mikelehen, line 129: presumably we can just remove the onChange handler now?

src/components/Newsletter/Newsletter.js
mikelehen: Blah, sorry. I didn't escape my &lt; and &gt; signs so my message didn't make sense.  😞 

What I meant is that I should be able to type "Wash&lt;enter&gt;" to add "Washington" but instead I get sent to an error page that says "Invalid email address" because it submits the form.

src/components/Newsletter/Newsletter.js
goldblatt: mhmm otherwise this would match the state if county is null which i guess would be fine but is a lil less clear

src/components/Newsletter/Newsletter.js
goldblatt: mm no react complains if we take it out because it's a readonly input

src/common/locations.ts
goldblatt: okay yes removed the slash in code to prevent this

src/common/locations.ts
```diff
@@ -1,8 +1,32 @@
...
+        .filter(county => !county.county.includes('/'))
+        .map(county => {
+          return { ...county, state: county.state_code };
+        }),
+    );
+  });
+
+  // Currently check for the unique fips so there aren't duplicates, but maybe
+  // match the way that the homepage gets unique values
+  return uniqBy(locations, 'full_fips_code');
```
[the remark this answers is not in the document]
goldblatt: removed (although it did have 4 fewer locations so some overlap in fips codes...)

PR #898, opened 2020-06-08
Chart footer copy update

chasulin:
- Makes disclaimer copy change apply only to contact tracing chart

### DIFF
--- a/src/components/LocationPage/ChartsHolder.tsx
+++ b/src/components/LocationPage/ChartsHolder.tsx
@@ -12,7 +12,6 @@ import {
 import NoCountyDetail from './NoCountyDetail';
 import { Projections } from 'common/models/Projections';
 import { Projection, TRACERS_NEEDED_PER_CASE } from 'common/models/Projection';
-import SummaryStats from 'components/SummaryStats/SummaryStats';
 import Disclaimer from 'components/Disclaimer/Disclaimer';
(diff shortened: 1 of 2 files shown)


src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -208,7 +207,7 @@ const ChartsHolder = (props: {
               {testPositiveData && (
                 <>
                   <ChartPositiveTestRate columnData={testPositiveData} />
-                  <Disclaimer metricName="positive test rate">
+                  <Disclaimer metricName={Metric.POSITIVE_TESTS}>
                     {POSITIVE_RATE_DISCLAIMER}
```
mikelehen, line 211: Hrm...  This seems fine, but FWIW- It seems a bit weird that we are passing in _part_ of the disclaimer text and then the Disclaimer component optionally adds some more disclaimer text based on the metricName...

It feels like ideally we should either pass in _all_ of the text or have the Disclaimer component generate _all_ of the text (based on metricName).

Like this it's kinda' fragmented / confusing how the text is getting generated.

chasulin: hmm yeah this is true. added in a TODO

mikelehen: Perfect! 👍

PR #904, opened 2020-06-09
Chart Save Button

chasulin:


src/components/LocationPage/ShareButtons.js
```diff
@@ -41,17 +42,58 @@ const InnerContent = props => {
     : imageBaseUrl +
       `states/${stateId.toLowerCase()}/chart/${chartIdentifier}/export.png`;
 
+  const downloadDate = moment().format('YYYY-MM-DD');
+
+  function makeDownloadFilename(chartIdentifier) {
+    const chartDownloadType = {
+      0: 'infection_growth',
```
mikelehen, line 49: Can you use the actual metric values (e.g. `Metric.INFECTION_GROWTH`) instead of raw numbers?  If somebody were to e.g. reorder the enum, then this would break.

src/components/LocationPage/ShareButtons.js
```diff
@@ -41,17 +42,58 @@ const InnerContent = props => {
...
+    a.click();
+    a.remove();
+  }
+
+  function downloadChartOnClick(url) {
+    const filename =
+      makeDownloadFilename(chartIdentifier) || `CovidActNow_${downloadDate}`;
+    fetch(url, {
+      headers: new Headers({
+        Origin: window.location.origin,
```
mikelehen, line 76: Hrm.  Just to confirm, is this header and the mode: 'cors' really necessary?  I was hoping it would "just work"...  If it doesn't, then maybe I didn't fully set up CORS right for content.covidactnow.org.

src/components/LocationPage/ShareButtons.js
```diff
@@ -41,17 +42,58 @@ const InnerContent = props => {
...
+      4: 'future_projections',
+    };
+    const chartType = chartDownloadType[chartIdentifier];
+    const location = countyId
+      ? `${countyId}_${stateId.toLowerCase()}`
+      : `${stateId.toLowerCase()}`;
+    return `${location}_${chartType}_${downloadDate}`;
+  }
+
+  function downloadChart(blob, filename) {
```
mikelehen, line 62: Since this is a bit hacky, I'd maybe call it out in a comment.

// HACK: Add a link to the blob and immediately click it as described in _link-to-stack-overflow_

src/components/LocationPage/ShareButtons.js
chasulin: ah yes true

PR #916, opened 2020-06-11
Script to generate alert summary files by comparing snapshots.

mikelehen:
Example output: https://gist.github.com/mikelehen/69bc69c727fa90c58dcc3ec07bae31a6


scripts/alert_emails/generate_daily_alerts.ts
```diff
@@ -0,0 +1,120 @@
...
+  const {oldSnap, newSnap} = await parseArgs();
+
+  console.log('Fetching projections from snapshots...');
+  const projectionsSet = ProjectionsSet.fromProjections(
+    await fetchAllProjections(snapshotUrl(oldSnap)),
+    await fetchAllProjections(snapshotUrl(newSnap))
+  );
+
+  const alerts: Alert[] = [ ];
+  for(const pair of projectionsSet.pairs) {
```
goldblatt, line 53: is projectionSet new?

scripts/alert_emails/generate_daily_alerts.ts
```diff
@@ -0,0 +1,120 @@
...
+        oldLevel,
+        newLevel,
+        changedMetrics
+      });
+    }
+  }
+
+  await fs.ensureDir(outputFolder);
+  const file = path.join(outputFolder, `${oldSnap}-${newSnap}.json`);
+  await fs.writeFile(file, JSON.stringify(alerts));
```
goldblatt, line 78: why did you do Alert[] vs {[fips: string]: Alert}. easier only the 'finding people to email side' to have it be in an object-- do we expect there for any reason to be duplicates

scripts/alert_emails/generate_daily_alerts.ts
```diff
@@ -0,0 +1,120 @@
...
+      alerts.push({
+        fips: pair.fips,
+        oldLevel,
+        newLevel,
+        changedMetrics
+      });
+    }
+  }
+
+  await fs.ensureDir(outputFolder);
```
goldblatt, line 76: hmm do we want any of these not to be overwritten? i think we should keep track of some of these so we can confirm the emails are sending as we expect

scripts/alert_emails/generate_daily_alerts.ts
```diff
@@ -0,0 +1,120 @@
...
+  const file = path.join(outputFolder, `${oldSnap}-${newSnap}.json`);
```
goldblatt, line 77: hmm will need some of the snapshot determination in a common file so my user list generation stuff can use it

src/common/models/ProjectionsPair.ts
```diff
@@ -32,6 +32,10 @@ export class ProjectionsPair {
     return rootMeanSquareDiff(leftDataset, rightDataset);
   }
 
+  get fips(): string {
+    return this.left.fips;
```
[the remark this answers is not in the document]
mikelehen: I honestly don't know what to do with code review feedback that just says "huh?".  Sorry. 😢

scripts/alert_emails/generate_daily_alerts.ts
mikelehen: 1. It seems intuitive to me to just generate a list of alerts to send.
2. There will not be duplicates, but I don't think we need to enforce that by using a different data structure.
3. Why is an object easier?  Array seems simpler since you can just loop over the values?

scripts/alert_emails/generate_daily_alerts.ts
mikelehen: Yeah, the github workflow should probably pass in the snapshots to both your script and my script explicitly.

scripts/alert_emails/generate_daily_alerts.ts
mikelehen: Yeah, I was thinking we should have the github workflow commit the files to git so we have a historical record of the diffs we generated and sent emails based on.

scripts/alert_emails/generate_daily_alerts.ts
mikelehen: Yeeeaaah...  

So I added that Monday when adding county support to the /compare tool for the modeling team.  Apologies that:
1) I didn't realize this would coincide so nicely with the email alerts stuff.  If I had, I would have coordinated with you on it.
2) I added it without code review because /compare is kinda' my baby and nobody else cares about the code.  If you _want_ to review, I'm happy to make changes.  In particular, the naming could probably be better.

scripts/alert_emails/generate_daily_alerts.ts
```diff
@@ -0,0 +1,120 @@
...
+
+const outputFolder = path.join(__dirname, 'alerts');
+
+// Disable counties; mostly useful for faster testing / debugging.
+const STATES_ONLY = false;
+
+interface Alert {
+  fips: string;
+  oldLevel: Level;
+  newLevel: Level;
```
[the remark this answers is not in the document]
mikelehen: Yeah, and it cleverly avoids using "threat", "alarm", or "overall" to avoid creating additional inconsistencies with our terminology. 😛

scripts/alert_emails/generate_daily_alerts.ts
mikelehen: Switched this over to a map so it plays better with your script. 😁

PR #919, opened 2020-06-11
Given a json of places to send alerts, look through the users who hav…

goldblatt:
Generate a list of users to email


scripts/alerts/create_lists_to_email.ts
```diff
@@ -0,0 +1,96 @@
+//@ts-ignore
```
mikelehen, line 1: Can you append to this comment what we're suppressing exactly / why?

scripts/alerts/create_lists_to_email.ts
```diff
@@ -0,0 +1,96 @@
...
+    while (pageToGet <= numberPages) {
+        const result = await getSubscribersForList(pageToGet, api, ALERT_SIGNUPS_LIST_ID)
+        result.Results.filter(subscriber => subscriber.State === 'Active').forEach((subscriber: Subscriber) => {
+            subscriber.CustomFields
+                .filter(field => field.Key === ALERT_LIST_CUSTOM_FIELD)
+                .map(field => field.Value.split(","))
+                .forEach((alertLocations: string[]) => {
+                    alertLocations.forEach((fips: string) => {
+                        // if the location the user signed up for has an alert
+                        if (locationsWithAlerts[fips]) {
```
mikelehen, line 77: Oh, heh.  I see why you wanted it to be a map now.  You're iterating all subscriptions every time and then checking if they have an alert or not.

I was imagining we'd iterate over the alerts and then find the subscribers for each one.

I guess either way will get the job done. I can change mine to output a map.

scripts/alerts/create_lists_to_email.ts
```diff
@@ -0,0 +1,96 @@
...
+    // Sample fips, will eventually be json loaded from a file
+    const locationsWithAlerts: { [fips: string]: { [key: string]: any } } = {
+        "06": { "overall": true },
+        "13127": { "overall": true }
+    }
+
+    // fips to object with a list of users and the previous meta data
+    const usersToEmail: { [fips: string]: Array<string> } = {};
+
+    // TODO: use oauth here and add secrets into github.
```
mikelehen, line 61: Out of curiosity, why is oauth better than using an api key?  (I don't have much context, but an API key _seems_ simpler).

scripts/alerts/create_lists_to_email.ts
```diff
@@ -0,0 +1,96 @@
...
+            subscriber.CustomFields
+                .filter(field => field.Key === ALERT_LIST_CUSTOM_FIELD)
+                .map(field => field.Value.split(","))
+                .forEach((alertLocations: string[]) => {
+                    alertLocations.forEach((fips: string) => {
+                        // if the location the user signed up for has an alert
+                        if (locationsWithAlerts[fips]) {
+                            const userListForFips = usersToEmail[fips] || [];
+                            userListForFips.push(subscriber.EmailAddress)
+                            usersToEmail[fips] = userListForFips;
```
mikelehen, line 80: Presumably we'll want to pass along the locationsWithAlerts[fips] contents too so the emailer script can use it to fill in the email template / whatever?

scripts/alerts/create_lists_to_email.ts
```diff
@@ -0,0 +1,96 @@
...
+                    });
+            });
+        });
+        numberPages = result.NumberOfPages;
+        pageToGet += 1
+    }
+
+    const outputFolder = path.join(__dirname, 'alerts');
+    await fs.ensureDir(outputFolder);
+    const file = path.join(outputFolder, `users-to-email.json`);
```
mikelehen, line 91: Like you said, we'll probably want to have some shared code for plumbing the before/after snapshot numbers through both of our scripts and we should probably include them in the json filename here too so we have an "audit log" type of thing for understanding what data was used to generate the alerts, etc.

scripts/alerts/create_lists_to_email.ts
goldblatt: oh hmm just general best practice to use oauth if it exsits

scripts/alerts/create_lists_to_email.ts
goldblatt: yeah figured since they are both files we could just use them both instead of making one that has both, but maybe one object is simpler

scripts/alerts/create_lists_to_email.ts
goldblatt: cool cool will leave this the way it is then!

scripts/alerts/create_lists_to_email.ts
goldblatt: yeah it's like i don't exactly understand where these files would live in the github action file system?

.gitignore
```diff
@@ -23,6 +23,10 @@ token.json
 .env.test.local
 .env.production.local
 
+# email alert files
+/scripts/alerts/users-to-email.json
+/scripts/alerts/example_alert.json
```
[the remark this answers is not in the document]
goldblatt: oh this is exceeedingly useful

scripts/alerts/create_lists_to_email.ts
```diff
@@ -0,0 +1,117 @@
...
+        const fipsToAlertFilename = args[0];
+        const outputFilename = args[1];
+        return {fipsToAlertFilename, outputFilename};
+    } else {
+        exitWithUsage();
+    }
+  }
+
+function exitWithUsage(): never {
+    console.log('Usage: yarn create-lists-to-email [fipsToAlertFilename] [outputfileName, optional]');
```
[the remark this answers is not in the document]
goldblatt: ohh yes thank you!

scripts/alerts/create_lists_to_email.ts
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Generates a file of fips and corresponding users to email with alerts
+ * for that fips.
+ *
+ * Takes in a file with {[fips: string]: {...data}}
+ *
+ * You can run via `yarn create-lists-to-email`
+ *
+ * Generates a file alerts/users-to-email.json containing an object of fips and
+ * users to send an email to. See the Alert interface below for schema.
```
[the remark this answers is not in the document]
goldblatt: lol a classic

PR #923, opened 2020-06-11
adding a scroll to alert sign up

goldblatt:
if you link to /alert from the homepage it will take you to the alert signup section

### DIFF
--- a/src/screens/HomePage/HomePage.js
+++ b/src/screens/HomePage/HomePage.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useRef, useEffect } from 'react';
 import HomePageHeader from 'components/Header/HomePageHeader';
 import Map from 'components/Map/Map';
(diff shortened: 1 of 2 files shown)


src/screens/HomePage/HomePage.js
```diff
@@ -14,7 +15,26 @@ import {
   PartnerHeader,
 } from './HomePage.style';
 
+// TODO: 180 is rough accounting for the navbar and searchbar;
+// could make these constants so we don't have to manually update
```
mikelehen, line 19: Seems like we don't really use this 180 default (we explicitly pass in -80), so maybe remove the default and then you don't need this TODO? 😁 

Else, change the default to -80 if that makes more sense... not sure what the -80 represents.

jamestamplin:
@mikelehen - Correct, we'd like a place to link to sign up (emails, tweets, etc)

PR #932, opened 2020-06-12
Adjust decomp factor from 30% to 25%

maxwhenderson:
Adjust decomp based on new analysis, pending a more dynamic decomp factor

### DIFF
--- a/src/common/models/Projection.ts
+++ b/src/common/models/Projection.ts
@@ -29,9 +29,17 @@ export const TRACERS_NEEDED_PER_CASE = 5;
  * get from the API to account for hospitals being able to decrease their utilization
  * (by cancelling elective surgeries, using surge capacity, etc.).
  */
-const ICU_DECOMP_FACTOR = 0.3;
+const ICU_DECOMP_FACTOR = 0.25;
 const ICU_DECOMP_FACTOR_STATE_OVERRIDES: { [key: string]: number } = {
+  Alabama: 20,
   Arizona: 0,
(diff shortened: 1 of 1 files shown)


src/common/models/Projection.ts
```diff
@@ -29,9 +29,17 @@ export const TRACERS_NEEDED_PER_CASE = 5;
  * get from the API to account for hospitals being able to decrease their utilization
  * (by cancelling elective surgeries, using surge capacity, etc.).
  */
-const ICU_DECOMP_FACTOR = 0.3;
+const ICU_DECOMP_FACTOR = 0.25;
 const ICU_DECOMP_FACTOR_STATE_OVERRIDES: { [key: string]: number } = {
+  Alabama: 20,
```
mikelehen, line 34: Needs to be a decimal percent (i.e. `0.2` not `20`)

src/common/models/Projection.ts
```diff
@@ -29,9 +29,17 @@ export const TRACERS_NEEDED_PER_CASE = 5;
...
   Arizona: 0,
+  Delaware: 30,
+  "District of Columbia": 10,
+  Georgia: 20,
+  Mississippi: 20,
+  Nevada: 30,
+  "Rhode Island": 15,
+
```
mikelehen, line 42: Can you run `yarn prettier-fix` to fix the formatting complaints above?

And while you're at it, run `yarn install` which _should_ install the git hook that automatically formats code on commit.

PR #937, opened 2020-06-13
Fix future projections y axis to 0.

mikelehen:


src/components/Charts/ChartFutureHospitalization.tsx
```diff
@@ -96,8 +96,10 @@ const ChartFutureHospitalization = ({
     range: [0, chartWidth],
   });
 
-  const [minY, maxY] = d3extent(allData, getY);
-  assert(minY !== undefined && maxY !== undefined, 'Data must not be empty');
+  const [yAxisMin, yAxisMax] = d3extent(allData, getY);
```
pnavarrc, line 99: nit: if we don't need `yAxisMin` anymore you can use `d3max` instead

```js
const yAxisMin = 0;
const yAxisMax = d3max(allData, getY);
```

BrettBoval:
Thank you!

PR #941, opened 2020-06-14
improve the alignment of map legend items on mobile

pnavarrc:
Close [221](https://trello.com/c/jJE61NeP/221-ui-fix-on-homepage-make-homepage-legend-2x2-on-mobile) - Make the map legend 2x2 on mobile

I set `flex-direction: column` instead of `flex-direction: row` to make sure that the orange item is below red (instead than on the first row).

image

Can @mikelehen review? Thanks!

### DIFF
--- a/src/components/Map/Legend.style.js
+++ b/src/components/Map/Legend.style.js
@@ -5,13 +5,18 @@ import { Typography } from '@material-ui/core';
 export const LegendContainer = styled.div`
   display: flex;
   flex-direction: column;
+  flex-wrap: wrap;
+  height: 3rem;
+  width: 100%;
+
   @media (min-width: 960px) {
+    height: unset;
+    width: unset;
     flex-direction: ${props => (props.condensed ? 'column' : 'row')};
   }
 `;
 
 export const LegendItemHeader = styled.div`
-  width: 100%;
   display: flex;
   flex-direction: row;
   align-items: center;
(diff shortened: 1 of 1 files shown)


pnavarrc:
[UI fix on homepage: Make homepage legend 2x2 on mobile.](https://trello.com/c/jJE61NeP/221-ui-fix-on-homepage-make-homepage-legend-2x2-on-mobile)

mikelehen:
Nice!  Though now, between 600-960px, it looks kinda' funky:

image

Fixable?

pnavarrc:
So, I played around with it, but the fix is pretty ugly https://codepen.io/pnavarrc/pen/XWXKaed

In desktop, we have a container with `flex-direction: column` to have the title and the legend items in the second div, no problem there.

image

For mobile, in order to have orange under red, I changed the direction of the legend items to be `column` and set a maximum height and set `flex-wrap` to `wrap`, so the items distribute from top to bottom and left to right (red, orange, and then yellow and green on the second column)

I _think_ that the width of the container of the items becomes the width of only one column (I highlighted it here), which makes the legend to be off-center by one column to the right

image

In the pen (https://codepen.io/pnavarrc/pen/XWXKaed), I tried another solution, keeping the container of the legend items to have `flex-direction: row` and setting the order of the items on mobile to make the second item (orange) to be the 3rd, and the third item (yellow) to be the 2nd, so they will have the right order.

With this solution, alignment is better but the code is awful (it's kind of hard to explain why we are setting 2nd → 3rd and 3rd → 2nd on mobile, the solution feels weird. @mikelehen any ideas? I probably spent more time than I should on this and couldn't make it much better. :/

pnavarrc:
Haha no problem!

PR #946, opened 2020-06-15
Email Alerts: Create a script to generate html snippets for threat level changes.

mikelehen:
This is a proof-of-concept.  I just threw in the old/new risk thermometers as a test.  We'd need to update this to use the actual React components for the email.

It generates html files like [this](https://gist.github.com/mikelehen/f3c4a3827562cd279e1e5d80a344aef6).

Which render to something like:
image

cc/ @chasulin


mikelehen:
This didn't end up being useful.

PR #950, opened 2020-06-16
Truncate future projections to 30 days

igorkofman:


src/common/models/Projection.ts
```diff
@@ -403,10 +413,13 @@ export class Projection {
       currDate = currDate.clone().add(1, 'days');
     }
 
+    // only keep futureDaysToInclude days ahead of today
+    const now = new Date();
+    const days = dates.findIndex(date => date > now) + futureDaysToInclude;
```
mikelehen, line 418: I was going to point out that we should do `_.findIndex(dates, date => date > now)` if we care about IE support, but it looks like the site is already broken on IE, so I guess we don't care. 😁

src/common/models/Projection.ts
```diff
@@ -403,10 +413,13 @@ export class Projection {
...
     return {
-      timeseries,
-      actualsTimeseries,
-      dates,
+      timeseries: timeseries.slice(0, days),
+      actualTimeseries: actualsTimeseries.slice(0, days),
+      dates: dates.slice(0, days),
```
mikelehen, line 422: FWIW- Instead of doing this truncation logic here at the end I think you could just change latestDate at the top to:

```
const latestDate = moment().add(futureDaysToInclude, 'days');
```

src/common/models/Projection.ts
```diff
@@ -18,6 +18,11 @@ import {
  */
 export const RT_TRUNCATION_DAYS = 7;
 
+/**
+ * We truncate our projections to 4 weeks out.
```
mikelehen, line 22: nit: "30 days out"

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -307,7 +307,7 @@ const ChartsHolder = (props: {
                   <Outcomes
                     title={`Predicted outcomes by ${formatUtcDate(
                       props.projections.projected.finalDate,
-                    )} (90 days from now)`}
+                    )} (30 days from now)`}
```
mikelehen, line 310: nit: since you exported the PROJECTIONS_TRUNCATION_DAYS constant, you could use that here.

src/components/Outcomes/Outcomes.js
```diff
@@ -46,7 +46,7 @@ const OutcomesRow = ({ projection, label, color }) => {
       <div>
         {projection.dateOverwhelmed
           ? formatUtcDate(projection.dateOverwhelmed)
-          : 'Not in the next 90 days'}
+          : 'Not in the next 30 days'}
```
mikelehen, line 49: nit: since you exported the PROJECTIONS_TRUNCATION_DAYS constant, you could use that here.

PR #963, opened 2020-06-19
Generate summaries.json file to use for map colors, email alerting, and comparison tables.

mikelehen:
NOTE: All the code changes are in https://github.com/covid-projections/covid-projections/commit/254604d703e07d66674ccef23a65b5f0bc7f61de and can be reviewed separately from the commit that added the actual summary.json files.

* Refactors update-calculated-interventions script into update-location-summaries script.
  * Generated summaries.json file contains summary info (current threat level, and the value/threat level of each metric) for all states/counties.
* Reworked map coloring to use the summaries.json file now.
* We save historical copies of the summaries.json file in the scripts/alerts dir for alert diffing.


src/common/location_summaries.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export interface MetricSummary {
+  value: number | null;
+  level: Level;
+}
+
+export interface LocationSummary {
+  level: Level;
+  metrics: {
+    [metric in Metric]?: MetricSummary;
```
goldblatt, line 14: ohhhh fascinating typescript, does this make it so that not all Metrics are required?

mikelehen: Yeah!  It took a lot of Googling to find, but it made my day.

PR #967, opened 2020-06-20
Implement LocationMap component

pnavarrc:
Implements the map component to move the county map above the charts, to make space for the prevalence metric. See designs in [Figma](https://www.figma.com/file/Y0XivjSNr3kolTDWBy6kIz/Design-Sprint-v2?node-id=476%3A31787)

image

### Notes
- The style change from the designs will be implemented later, the style of this component should ideally match what we have in the site today.
- @mikelehen do you know where we got the scale factor for the states? This one is a bit off, so it looks like the state has a lot of padding around.

### DIFF
--- a/src/components/LocationPage/ChartsHolder.tsx
+++ b/src/components/LocationPage/ChartsHolder.tsx
@@ -18,6 +18,7 @@ import ShareModelBlock from 'components/ShareBlock/ShareModelBlock';
 import LocationPageHeader from 'components/LocationPage/LocationPageHeader';
 import Outcomes from 'components/Outcomes/Outcomes';
 import ShareButtons from 'components/LocationPage/ShareButtons';
+import LocationMap from 'components/LocationMap';
 import useMediaQuery from '@material-ui/core/useMediaQuery';
 import { useTheme } from '@material-ui/core/styles';
 import {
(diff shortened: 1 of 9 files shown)


pnavarrc:
[Implement LocationMap component](https://trello.com/c/8EG6rGmH/255-implement-locationmap-component)

src/components/Map/Legend.js
```diff
@@ -22,14 +21,9 @@ export function Legend(props) {
...
-export function LegendItem(props) {
-  const { title, color, condensed } = props;
-  return (
-    <LegendItemContainer condensed={condensed}>
-      <LegendItemHeader>
-        <ColorBox color={color} />
-        {title}
-      </LegendItemHeader>
-    </LegendItemContainer>
-  );
```
pnavarrc, line 34: The `LegendItemContainer` wrapper has margins, which we don't want on the `LocationMap` component, so I exported the `LegendItem` without the margin and using the wrapper to leave the legend as it was in other places.

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -68,6 +69,8 @@ const ChartsHolder = (props: {
   county: County;
   chartId: string;
   countyId: string;
+  selectedCounty: County;
+  setSelectedCounty: (input: string) => {};
```
pnavarrc, line 73: We need to pass these properties through the `ChartHolder` component so they are available for the `LocationMap` component, maybe we should move things outside the `ChartHolder` component?

pnavarrc:
We are not going with this design, closing!

PR #973, opened 2020-06-23
Set up Firestore and start recording alert subscriptions there.

mikelehen:
* Adds plumbing so we can distinguish dev / staging / prod environments in the website. (src/common/utils/environment.ts, package.json, and .github/workflows changes)
* Adds website code to initialize firebase for dev / staging / prod (src/common/firebase.ts).
* Checks in initial firebase config files (Firestore rules / indexes) for our projects (firebase/)
* Adds code to the newsletter signup to record alert subscriptions to Firestore (Newsletter.js).


src/components/Newsletter/Newsletter.js
```diff
@@ -46,6 +52,25 @@ class Newsletter extends React.Component {
...
+    // Merge the locations with any existing ones since that's _probably_ what the user wants.
+    await db
+      .collection('alerts-subscriptions')
+      .doc(email)
+      .set(
+        {
+          locations: firebase.firestore.FieldValue.arrayUnion(...locations),
+          subscribedAt: firebase.firestore.FieldValue.serverTimestamp(),
+        },
+        { merge: true },
```
mikelehen, line 70: @chasulin FYI- This is the only actual bit of real Firebase / Firestore code.

It just creates a document for the user, containing the subscribed locations and a timestamp.

We'll need to create some sort of "unsubscribe" page as well that deletes users from Firestore.  If you _want_ to work on that, you could.  See my overall PR comment.

firebase/firestore.rules
```diff
@@ -0,0 +1,8 @@
+rules_version = '2';
+service cloud.firestore {
+  match /databases/{database}/documents {
+    match /alerts-subscriptions/{email} {
+      allow write;
+    }
+  }
```
mikelehen, line 7: FWIW- These rules allow anybody to write to /alerts-subscriptions/{email}, but they can't read, which is suitable for our use case where we just want to let people write (and delete) their subscriptions.

PR #983, opened 2020-06-26
Update the MiniMap component

pnavarrc:
Updates the MiniMap component to accommodate the prevalence metric. See the designs in [Figma](https://www.figma.com/file/kQGQ5qsDh7ir4MhjMM20Tl/5th-metric?node-id=48%3A2025)

image

I updated the layout to match the designs and added a sized container for the state maps, we need this to adjust the map projection parameters later (center, rotation, scale) so they use the entire available area for the map. I will implement that in a follow-up PR.

Can @goldblatt and @chasulin review? Thanks!

### DIFF
--- a/src/components/MiniMap/MiniMap.tsx
+++ b/src/components/MiniMap/MiniMap.tsx
@@ -1,21 +1,12 @@
 import React, { FunctionComponent } from 'react';
-
-import {
-  MapMenuMobileWrapper,
-  MapMenuItem,
-  MapContentWrapper,
-  MapContentInner,
-  MapWrapper,
-  CountyMapAltWrapper,
-} from 'components/MiniMap/MiniMap.style';
-
+import { useHistory } from 'react-router-dom';
+import { find as _find } from 'lodash';
 import Map from 'components/Map/Map';
 import CountyMap from 'components/CountyMap/CountyMap';
-import _ from 'lodash';
 import { MAP_FILTERS } from 'screens/LocationPage/Enums/MapFilterEnums';
-import { useHistory } from 'react-router-dom';
 import US_STATE_DATASET from '../MapSelectors/datasets/us_states_dataset_01_02_2020.json';
 import { Projections } from 'common/models/Projections';
+import * as Styles from './MiniMap.style';
 
 interface MiniMapProperties {
   projections: Projections;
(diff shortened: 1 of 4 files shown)


pnavarrc:
[Prevalence · Update MiniMap Styling](https://trello.com/c/8EG6rGmH/255-prevalence-update-minimap-styling)

src/components/MiniMap/MiniMap.tsx
```diff
@@ -28,71 +19,81 @@ interface MiniMapProperties {
   setMapOption: (input: string) => {};
 }
 
+const getCounty = (stateId: string, fullFips: string) =>
+  _find(
+    // @ts-ignore: US_STATE_DATASET is .js, but this is valid
+    US_STATE_DATASET.state_county_map_dataset[stateId].county_dataset,
+    ['full_fips_code', fullFips],
+  );
```
pnavarrc, line 27: We are doing this on [LocationPage.js#L30](https://github.com/covid-projections/covid-projections/blob/4b0d75de7a5ccb53e2709464bbb1514a95908474/src/screens/LocationPage/LocationPage.js#L30) too, should we update `setSelectedCounty` so it takes the FIPS instead of the county object?

src/components/MiniMap/MiniMap.tsx
```diff
@@ -28,71 +19,81 @@ interface MiniMapProperties {
...
+
 /**
  * TODO(sgoldblatt): Don't pass in the setState vars here and use a provider
  * to set the selected MobileMenu and map options as well
  */
-const MiniMap: FunctionComponent<MiniMapProperties> = (
-  props: MiniMapProperties,
-) => {
+const MiniMap: FunctionComponent<MiniMapProperties> = ({
```
pnavarrc, line 33: @goldblatt Do we have a React context to do that or not yet? Maybe I could create one on the `LocationPage`

src/components/MiniMap/MiniMap.tsx
```diff
@@ -28,71 +19,81 @@ interface MiniMapProperties {
...
+  projections,
```
pnavarrc, line 34: I kept the existing API, but it looks like the projections are only used to get the state name, I can probably get that from a JSON file.

src/components/MiniMap/MiniMap.tsx
chasulin: Yeah it looks like its only used to get the fips so that would work (or could add a TODO and get this merged in for the time being)

PR #1002, opened 2020-06-30
Prevalence metric calculations

chasulin:
- Adds case density metrics (currentCaseDensityByCases, currentCaseDensityByDeaths) and calculations (calcCaseDensityByCases, calcCaseDensityByDeaths) into Projection.ts

### DIFF
--- a/src/common/metrics/contact_tracing.ts
+++ b/src/common/metrics/contact_tracing.ts
@@ -67,11 +67,11 @@ export const CONTACT_TRACING_LEVEL_INFO_MAP: LevelInfoMap = {
 
 export function contactTracingStatusText(projection: Projection) {
   const currentContactTracers = projection.currentContactTracers;
+  const weeklyAverageResult = projection.getWeeklyAverageForDay(
+    projection.dailyPositiveTests,
+  );
   const currentWeeklyAverage =
-    projection.getWeeklyAverageForDay(projection.dailyPositiveTests) &&
-    Math.round(
-      projection.getWeeklyAverageForDay(projection.dailyPositiveTests) || 0,
-    );
+    weeklyAverageResult && Math.round(weeklyAverageResult);
   const currentContactTracingMetric = projection.currentContactTracerMetric;
   if (
     currentContactTracingMetric === null ||
--- a/src/common/metrics/contact_tracing.ts
+++ b/src/common/metrics/contact_tracing.ts
@@ -68,8 +68,10 @@ export const CONTACT_TRACING_LEVEL_INFO_MAP: LevelInfoMap = {
 export function contactTracingStatusText(projection: Projection) {
   const currentContactTracers = projection.currentContactTracers;
   const currentWeeklyAverage =
-    projection.getWeeklyAverageCaseForDay() &&
-    Math.round(projection.getWeeklyAverageCaseForDay() || 0);
+    projection.getWeeklyAverageForDay(projection.dailyPositiveTests) &&
+    Math.round(
+      projection.getWeeklyAverageForDay(projection.dailyPositiveTests) || 0,
+    );
   const currentContactTracingMetric = projection.currentContactTracerMetric;
   if (
     currentContactTracingMetric === null ||
--- a/src/common/models/Projection.ts
+++ b/src/common/models/Projection.ts
@@ -80,7 +80,9 @@ export type DatasetId =
   | 'rtRange'
   | 'icuUtilization'
   | 'testPositiveRate'
-  | 'contractTracers';
+  | 'contractTracers'
+  | 'caseDensityByCases'
+  | 'caseDensityByDeaths';
 
 export interface RtRange {
   /** The actual Rt value. */
@@ -110,6 +112,9 @@ export class Projection {
   readonly currentCumulativeCases: number | null;
   readonly currentContactTracerMetric: number | null;
   readonly stateName: string;
+  readonly dailyPositiveTests: Array<number | null>;
+  readonly currentCaseDensityByCases: number | null;
+  readonly currentCaseDensityByDeaths: number | null;
 
   private readonly intervention: string;
   private readonly dates: Date[];
@@ -124,7 +129,6 @@ export class Projection {
   private readonly cumulativeInfected: Array<number | null>;
   private readonly cumulativePositiveTests: Array<number | null>;
   private readonly cumulativeNegativeTests: Array<number | null>;
-  private readonly dailyPositiveTests: Array<number | null>;
   private readonly dailyNegativeTests: Array<number | null>;
   private readonly rtRange: Array<RtRange | null>;
   // ICU Utilization series as values between 0-1 (or > 1 if over capacity).
@@ -132,8 +136,8 @@ export class Projection {
   // Test Positive series as values between 0-1.
   private readonly testPositiveRate: Array<number | null>;
   private readonly contractTracers: Array<number | null>;
-  private readonly caseDensityCases: Array<number | null>;
-  private readonly caseDensityDeaths: Array<number | null>;
+  private readonly caseDensityByCases: Array<number | null>;
+  private readonly caseDensityByDeaths: Array<number | null>;
   private readonly dailyDeaths: Array<number | null>;
 
   constructor(
@@ -183,7 +187,11 @@ export class Projection {
     this.dailyNegativeTests = this.deltasFromCumulatives(
       this.cumulativeNegativeTests,
     );
-    this.dailyDeaths = this.deltasFromCumulatives(this.cumulativeDeaths);
+
+    const cumulativeActualDeaths = actualTimeseries.map(
+      row => row?.cumulativeDeaths || null,
+    );
+    this.dailyDeaths = this.deltasFromCumulatives(cumulativeActualDeaths);
 
     this.rtRange = this.calcRtRange(timeseries);
     this.testPositiveRate = this.calcTestPositiveRate();
@@ -214,8 +222,11 @@ export class Projection {
 
     this.contractTracers = this.calcContactTracers(this.actualTimeseries);
 
-    this.caseDensityCases = this.calcCaseDensityCases(this.actualTimeseries);
-    this.caseDensityDeaths = this.calcCaseDensityDeaths(this.actualTimeseries);
+    this.caseDensityByCases = this.calcCaseDensityByCases();
+    this.caseDensityByDeaths = this.calcCaseDensityByDeaths();
+
+    this.currentCaseDensityByCases = this.lastValue(this.caseDensityByCases);
+    this.currentCaseDensityByDeaths = this.lastValue(this.caseDensityByDeaths);
 
     this.fixZeros(this.hospitalizations);
     this.fixZeros(this.cumulativeDeaths);
@@ -461,8 +472,9 @@ export class Projection {
    * to reporting weirdness), we skip that data point and may average fewer
    * than 7 days worth of data.
    */
-  getWeeklyAverageCaseForDay(i?: number) {
-    const lastIndex = this.indexOfLastValue(this.dailyPositiveTests);
+
+  getWeeklyAverageForDay(data: Array<number | null>, i?: number) {
+    const lastIndex = this.indexOfLastValue(data);
     if (i === undefined) {
       if (!lastIndex) return null;
       i = lastIndex;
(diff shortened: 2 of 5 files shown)


src/common/models/Projection.ts
```diff
@@ -180,6 +183,7 @@ export class Projection {
     this.dailyNegativeTests = this.deltasFromCumulatives(
       this.cumulativeNegativeTests,
     );
+    this.dailyDeaths = this.deltasFromCumulatives(this.cumulativeDeaths);
```
mikelehen, line 186: this.cumulativeDeaths is defined above as:
```
    this.cumulativeDeaths = timeseries.map(row => row && row.cumulativeDeaths);
```

Confusingly, our API has `timeseries` and `actualTimeseries`.  `timeseries` is what comes out of our SEIR model and they are _estimates_.  `actualTimeseries` is the raw actual data we have from CovidTracking or Corona Data Scraper or our other data sources.

So in this case, we want to use the actual deaths, which is probably something like:
```
    const cumulativeActualDeaths = actualTimeseries.map(row => row?.cumulativeDeaths || null);
    this.dailyDeaths = this.deltasFromCumulatives(cumulativeActualDeaths);
```

src/common/models/Projection.ts
```diff
@@ -470,6 +477,22 @@ export class Projection {
...
+    } else if (lastIndex === null || i > lastIndex) {
+      return null;
+    }
+    // Get last week of sane data (ignore negative values)
+    const lastWeekOfDeaths = _.filter(
+      this.dailyDeaths.slice(Math.max(0, i - 6), i + 1),
+      p => p !== null && p >= 0,
+    );
+    return _.mean(lastWeekOfDeaths);
+  }
```
mikelehen, line 494: I think you could combine getWeeklyAverageCaseForDay() and getWeeklyAverageDeathForDay() into a single `getWeeklyAverageForDay(data: Array<number|null>, i?: number)` helper.

And then the calling code would do `getWeeklyAverageForDay(this.dailyPositiveTests, ...)` or `getWeeklyAverageForDay(this.dailyDeaths, ...)` as appropriate.

FWIW- We _also_ have a `smoothWithRollingAverage()` helper that generates a new array with the averaged values.  I think we that would be more efficient and cleaner, but that's a bit of extra churn / effort (need to make sure it handles negative outliers in the same way), so that's probably best for a later project.

src/common/models/Projection.ts
```diff
@@ -132,6 +132,9 @@ export class Projection {
   // Test Positive series as values between 0-1.
   private readonly testPositiveRate: Array<number | null>;
   private readonly contractTracers: Array<number | null>;
+  private readonly caseDensityCases: Array<number | null>;
+  private readonly caseDensityDeaths: Array<number | null>;
```
mikelehen, line 136: nit: Maybe call these caseDenistyByCases and caseDensityByDeaths since they're the same metric, just calculated by two different mechanisms.

Also, because we want these to be usable via the getDataset() method which takes a string and expects a field with that name to exist on the class, we need to add these names to `DatasetId` above.

src/common/models/Projection.ts
```diff
@@ -491,6 +514,44 @@ export class Projection {
     );
   }
 
+  private calcCaseDensityCases(
+    actualTimeseries: Array<CANActualsTimeseriesRow | null>,
+  ): Array<number | null> {
+    return actualTimeseries.map(
+      (row: CANActualsTimeseriesRow | null, i: number) => {
```
mikelehen, line 521: Since you're not actually using anything from actualTimeseries, it's a bit weird to map over it.

Perhaps the simplest would just be:

```
for(let i = 0; i < this.dates.length; i++) {
...
}
```

Since dates represents all the dates in our timeseries.

src/common/models/Projection.ts
```diff
@@ -491,6 +514,44 @@ export class Projection {
...
+        const weeklyAverage = this.getWeeklyAverageCaseForDay(i);
+        const totalPopulation = this.totalPopulation;
+        if (weeklyAverage) {
+          console.log(
+            'calc cases',
+            weeklyAverage / (totalPopulation * 0.00001),
+          );
+          return weeklyAverage / (totalPopulation * 0.00001);
```
mikelehen, line 529: nit: I'd slightly prefer `(totalPopulation / 100000)` to be consistent with the PRD, etc.

src/common/models/Projection.ts
```diff
@@ -491,6 +514,44 @@ export class Projection {
     );
   }
 
+  private calcCaseDensityCases(
```
mikelehen, line 517: nit: Change to calcCaseDensityByCases() if you change the field name to caseDensityByCases.

src/common/models/Projection.ts
```diff
@@ -491,6 +514,44 @@ export class Projection {
...
+      },
+    );
+  }
+
+  private calcCaseDensityDeaths(
+    actualTimeseries: Array<CANActualsTimeseriesRow | null>,
+  ): Array<number | null> {
+    return actualTimeseries.map(
+      (row: CANActualsTimeseriesRow | null, i: number) => {
+        const weeklyAverage = this.getWeeklyAverageDeathForDay(i);
```
mikelehen, line 541: Looks like we are missing the .01 factor here.  So we need something like:
```
const estimatedCaseFatalityRatio = .01; // 1% of cases expected to turn into deaths.
const estimatedCases = weeklyAverage / estimatedCaseFatalityRatio;
```
And then use estimatedCases in the calculation below.

src/common/models/Projection.ts
```diff
@@ -210,6 +214,9 @@ export class Projection {
 
     this.contractTracers = this.calcContactTracers(this.actualTimeseries);
 
+    this.caseDensityCases = this.calcCaseDensityCases(this.actualTimeseries);
+    this.caseDensityDeaths = this.calcCaseDensityDeaths(this.actualTimeseries);
```
mikelehen, line 218: We'll probably also need to add something like:

```
    this.currentCaseDenistyByCases = this.lastValue(this.caseDensityByCases);
    this.currentCaseDenistyByDeaths = this.lastValue(this.caseDensityByDeaths);
```

So that we can get the current values for use in the actual metric value and perhaps in the status text above the chart, etc.

src/common/models/Projection.ts
chasulin: Keeping them separate for now since getWeeklyAverageCaseForDay() is used in contact_tracing.ts. Adding a TODO to combine them

src/common/models/Projection.ts
```diff
@@ -181,6 +188,11 @@ export class Projection {
       this.cumulativeNegativeTests,
     );
 
+    const cumulativeActualDeaths = actualTimeseries.map(
+      row => row?.cumulativeDeaths || null,
+    );
```
pnavarrc, line 193: Maybe we should smooth the `cumulativeActualDeaths` with `smoothCumulatives`? If the series has a hole, maybe we should use the value of the previous day (we know that the series should monotonically increase day to day)

**Example**

```
Date        CumulativeDeaths
----------------------------
2020-06-01  10
2020-06-02  12
2020-06-03  null // we know it must be at least 12
2020-06-04  16
// ...
```

src/common/models/Projection.ts
```diff
@@ -464,7 +483,7 @@ export class Projection {
     }
     // Get last week of sane data (ignore negative values)
     const lastWeekOfPositives = _.filter(
-      this.dailyPositiveTests.slice(Math.max(0, i - 6), i + 1),
+      data.slice(Math.max(0, i - 6), i + 1),
```
pnavarrc, line 486: Maybe set `Math.max(i + 1, data.length)` just to make sure that the end on `slice` is also valid.

src/common/models/Projection.ts
```diff
@@ -491,6 +514,41 @@ export class Projection {
...
+    const caseDensityByCases = [];
+    const totalPopulation = this.totalPopulation;
+    for (let i = 0; i < this.dates.length; i++) {
+      const weeklyAverage = this.getWeeklyAverageForDay(
+        this.dailyPositiveTests,
+        i,
+      );
+      if (weeklyAverage === null) {
+        caseDensityByCases.push(null);
+      } else {
```
pnavarrc, line 527: nit: Consider adding a check to ensure that `totalPopulation` is greater than `0`. Also, maybe you can simplify the expression

```js
const value = (weeklyAverage && totalPopulation > 0)
  ? weeklyAverage / (totalPopulation / 100000)
  : null;
caseDensityByCases.push(value);
```

(not sure if that's much better TBH)

src/common/models/Projection.ts
```diff
@@ -491,6 +514,41 @@ export class Projection {
...
+
+  private calcCaseDensityByDeaths(): Array<number | null> {
+    const caseDensityByDeaths = [];
+    const totalPopulation = this.totalPopulation;
+    for (let i = 0; i < this.dates.length; i++) {
+      const weeklyAverage = this.getWeeklyAverageForDay(this.dailyDeaths, i);
+      if (weeklyAverage === null) {
+        caseDensityByDeaths.push(null);
+      } else {
+        const estimatedCaseFatalityRatio = 0.01; // 1% of cases expected to turn into deaths.
```
pnavarrc, line 543: nit: Consider creating a constant at the top of the file for this ratio so it's easier to find if we have to change it (we might be using this number for other calculations, I'm going to check). I'm not sure, but I _think_ they use "mortality" no?

src/common/models/Projection.ts
```diff
@@ -454,8 +472,9 @@ export class Projection {
    * to reporting weirdness), we skip that data point and may average fewer
    * than 7 days worth of data.
    */
-  getWeeklyAverageCaseForDay(i?: number) {
-    const lastIndex = this.indexOfLastValue(this.dailyPositiveTests);
+
+  getWeeklyAverageForDay(data: Array<number | null>, i?: number) {
+    const lastIndex = this.indexOfLastValue(data);
```
pnavarrc, line 477: Isn't this the same as `smoothWithRollingAverage`? I think that the only difference is that `smoothWithRollingAverage` sets `null` for the first 6 days, whereas this one calculates the rolling average with a smaller window.

src/common/models/Projection.ts
mikelehen: Oh, yes.  Using `smoothCumulatives()` is a good idea.  It would fill that gap with 14 (split the difference between the day before and the day after).

src/common/models/Projection.ts
mikelehen: Yeah...   It's the same concept but implemented differently.  I think we should consolidate and use smoothWithRollingAverage(), but I don't feel strongly it's worth reworking things in this PR or not. 😁 

TBH this whole class has turned into a bit of a monster.  I think it'd be worth a pretty major overhaul, perhaps factoring out all of the timeseries-related code into a Timeseries class or something.

NOTE: One thing I was worried about is that this code has handling for negative values but smoothWithRollingAverage() does not...  But I _think_ we actually ended up solving negative values in `smoothCumulatives()`, so it probably doesn't matter.

src/common/models/Projection.ts
mikelehen: Hrm.  Do you mean `Math.min(i+1, data.length)`?  But I think it's probably invalid for `i` to be `>= data.length`.  So I don't think that should be necessary.  If we wanted to be cautious, we could `assert (i < data.length, ...)` above somewhere.  But it doesn't seem super valuable, especially if we're going to get rid of this code soon and use smoothWithRollingAverage() 😁

src/common/models/Projection.ts
mikelehen: just a nit on Pablo's suggestion: `weeklyAverage` could be `0` so I suspect we do need to do `(weeklyAverage !== null && ...` not just `(weeklyAverage && ...` 😁

src/common/models/Projection.ts
chasulin: punting for now

src/common/metrics/contact_tracing.ts
```diff
@@ -68,8 +68,10 @@ export const CONTACT_TRACING_LEVEL_INFO_MAP: LevelInfoMap = {
 export function contactTracingStatusText(projection: Projection) {
   const currentContactTracers = projection.currentContactTracers;
   const currentWeeklyAverage =
-    projection.getWeeklyAverageCaseForDay() &&
-    Math.round(projection.getWeeklyAverageCaseForDay() || 0);
+    projection.getWeeklyAverageForDay(projection.dailyPositiveTests) &&
+    Math.round(
+      projection.getWeeklyAverageForDay(projection.dailyPositiveTests) || 0,
+    );
```
mikelehen, line 74: FWIW (feel free to leave as-is since this is existing code)- this could be cleaned up to something like:
```
  const weeklyAverageResult = projection.getWeeklyAverageForDay(projection.dailyPositiveTests);
  const currentWeeklyAverage = weeklyAverageResult && Math.round(weeklyAverageResult);
```

It reduces duplication and avoids `|| 0` since if you use an intermediate variable, TypeScript knows that inside the Math.round(), `weeklyAverageResult` is guaranteed to be a number (not null).

src/common/models/Projection.ts
```diff
@@ -491,6 +516,42 @@ export class Projection {
     );
   }
 
+  private calcCaseDensityByCases(): Array<number | null> {
+    const caseDensityByCases = [];
+    const totalPopulation = this.totalPopulation;
+    const totalPopulationAboveZero = totalPopulation > 0;
```
mikelehen, line 522: super nit: I think this calculation (`totalPopulation > 0`) is so trivial that storing it to a variable probably decreases readability a tiny bit with no particular benefit.  The code would be simpler if you omitted the variable and changed the below check to just:

```
if (totalPopulation === 0 || weeklyAverage === null) { ...
```

But this is very stylistic, so feel free to leave as-is

src/common/models/Projection.ts
chasulin: punting this for now as well

PR #1011, opened 2020-07-02
Use actuals where available in ICU Headroom metric.

mikelehen:
This uses `ICUBeds.currentUsageCovid` and/or `ICUBeds.currentUsageTotal` when available in the API to calculate the ICU Headroom metric.  Since either one or both may be available, there are 4 possible combinations, and this code handles all of them and subtly includes the result in the copy to make it possible to decode the calculation.

---
If neither are available, you get the existing site behavior where both covid and non-covid patients are estimates:

image

---
If both are available, then we omit "we estimate" in both places:

image

---
If only `ICUBeds.currentUsageCovid` is available, then non-covid patients is an estimate and we include the typical utilization + decomp percentage that we're using to estimate it, e.g.:

image

---
If only `ICUBeds.currentUsageTotal` is available, then both covid and non-covid patients are estimates but we omit the percentage for non-covid patients, since we can avoid using the typical utilization+decomp percentage:

image


src/common/metrics/hospitalizations.ts
```diff
@@ -67,44 +68,40 @@ export const HOSPITALIZATIONS_DISCLAIMER =
   ', a pandemic think tank, recommends that hospitals maintain enough ICU capacity to double the number of COVID patients hospitalized.';
 
 export function hospitalOccupancyStatusText(projection: Projection) {
-  const currentIcuUtilization = projection.currentIcuUtilization;
-  const currentCovidICUPatients = projection.currentCovidICUPatients;
-  const totalICUCapacity = projection.totalICUCapacity;
-  const nonCovidPatients = Math.floor(projection.nonCovidPatients);
+  const icu = projection.icuHeadroomInfo;
```
mikelehen, line 71: This code didn't actually change much, but I created icuHeadroomInfo as a container for all of the data we were pulling out of the projection, including the new metadata about actuals that we use to tweak the copy below.

src/common/metrics/hospitalizations.ts
```diff
@@ -67,44 +68,40 @@ export const HOSPITALIZATIONS_DISCLAIMER =
...
   );
-  const remainingICUBeds = formatInteger(totalICUCapacity - nonCovidPatients);
-  const covidICUPatients = formatInteger(currentCovidICUPatients);
-  const covidICUBeds =
-    currentCovidICUPatients / (totalICUCapacity - nonCovidPatients);
+  const remainingICUBeds = formatInteger(icu.totalBeds - icu.nonCovidPatients);
+  const covidICUPatients = formatInteger(icu.covidPatients);
+  const icuHeadroom =
+    icu.metricValue > 1 ? '>100%' : formatPercent(icu.metricValue);
+
```
mikelehen, line 92: Below is the estimate copy that I tweaked.

src/common/models/ICUHeadroom.ts
```diff
@@ -0,0 +1,230 @@
+import { nonNull } from 'common/utils';
```
mikelehen, line 1: I refactored all of the ICU Headroom code into a new file since it's pretty sizable at this point.

I also basically rewrote it from scratch. Sorry.

src/common/models/ICUHeadroom.ts
```diff
@@ -0,0 +1,230 @@
...
+
+/**
+ * We subtract this "decomp" factor from the typical ICU Utilization rates we
+ * get from the API to account for hospitals being able to decrease their utilization
+ * (by cancelling elective surgeries, using surge capacity, etc.).
+ */
+const ICU_DECOMP_FACTOR = 0.21;
+const ICU_DECOMP_FACTOR_STATE_OVERRIDES: { [key: string]: number } = {
+  Alabama: 0.15,
+  Arizona: 0.4,
```
mikelehen, line 31: NOTE: I changed Arizona decomp from 0% to 40% to make it match up with the actuals from their state dashboard.  I think their capacity has changed or they've displaced non-covid patients or something.

With 0%, they were pegged at 100% since the beginning of their actuals (Apr 13).

src/common/models/Projection.ts
```diff
@@ -29,25 +31,6 @@ export const PROJECTIONS_TRUNCATION_DAYS = 30;
  */
 export const TRACERS_NEEDED_PER_CASE = 5;
```
mikelehen, line 33: As mentioned above, I moved all of the ICU Headroom code out of this file.

src/common/models/Projection.ts
```diff
@@ -198,37 +181,25 @@ export class Projection {
...
+      this.icuHeadroomInfo?.metricSeries || this.dates.map(date => null);
 
     this.contractTracers = this.calcContactTracers(this.actualTimeseries);
 
     this.caseDensityByCases = this.calcCaseDensityByCases();
     this.caseDensityByDeaths = this.calcCaseDensityByDeaths();
 
-    this.currentCaseDensityByCases = this.lastValue(this.caseDensityByCases);
-    this.currentCaseDensityByDeaths = this.lastValue(this.caseDensityByDeaths);
+    this.currentCaseDensityByCases = lastValue(this.caseDensityByCases);
```
mikelehen, line 201: I moved a few helpers into a utils.ts file for sharing with ICUHeadroom.ts.

src/common/models/Projections.ts
```diff
@@ -113,6 +113,14 @@ export class Projections {
     }
   }
 
+  getMetricValues(): { [metric in Metric]: number | null } {
```
mikelehen, line 116: This code was duplicated in several places (and would have required updating to use `projection.icuHeadroomInfo.metricValue` now), so I consolidated it here.

src/common/models/utils.ts
```diff
@@ -0,0 +1,54 @@
+import _ from 'lodash';
+
+// TODO(michael): We should perhaps create a Timeseries class that wraps an
+// Array<Date> and Array<number | null> and hang these utility functions off of
+// that.
```
mikelehen, line 5: FWIW- I think this would actually make a lot of things cleaner.  We have a lot of utility functions for smoothing / manipulating timeseries data at this point...

src/common/models/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export function lastValue<T>(data: Array<T | null>): T | null {
+  const i = indexOfLastValue(data);
+  return i === null ? null : data[i]!;
+}
+
+/**
+ * Replaces all values after the specified date with nulls.
+ */
+export function omitDataAfterDate<T>(
```
mikelehen, line 24: Note: The diff is hidden since I moved this function, but I changed it to omitDataAfterDate() instead of omitDataAtOrAfterDate() since I noticed the same thing you did that it was eating the latest data point.

src/common/models/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+): Array<T | null> {
+  return dates.map((date, idx) => {
+    const value = data[idx];
+    return date <= cutoffDate ? value : null;
+  });
+}
+
+/**
+ * Returns true if the series has at least 7 non-null values in the last 14
+ * days and has ever had a non-zero value.
```
mikelehen, line 37: This is an arbitrary criteria I came up with for when actuals are good enough to show.  Seems okay for now.

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -86,21 +86,12 @@ const ChartsHolder = (props: {
...
-    };
-  };
-
   const shareButtonProps = {
     chartId: props.chartId,
     stateId: props.stateId,
     countyId: props.countyId,
     county: props.county,
-    stats: projection ? getChartSummarys(projection) : {},
+    stats: projection ? props.projections.getMetricValues() : {},
```
mikelehen, line 94: Switched over to the helper I added...

src/screens/Embed/Embed.js
```diff
@@ -84,6 +68,7 @@ export default function Embed() {
     return <span>'No data available for county.'</span>;
   }
 
+  const stats = projections.getMetricValues();
```
mikelehen, line 71: switched over to the helper I added...

src/screens/internal/ShareImage/ShareCardImage.tsx
```diff
@@ -67,13 +66,7 @@ const LocationShareCard = ({ stateId, countyFipsId }: ShareCardProps) => {
...
     return null;
   }
-  const projection = projections.primary;
-  const stats = {
-    [Metric.CASE_GROWTH_RATE]: projection.rt,
-    [Metric.HOSPITAL_USAGE]: projection.currentIcuUtilization,
-    [Metric.POSITIVE_TESTS]: projection.currentTestPositiveRate,
-    [Metric.CONTACT_TRACING]: projection.currentContactTracerMetric,
-  };
+  const stats = projections.getMetricValues();
```
mikelehen, line 69: switched over to the helper I added...

mikelehen:
FWIW- You can sanity-check the states using the /all page.  Mostly things seem sane, especially after tweaking Arizona and DC's decomp.  I did notice that several states have pretty high (critical) peaks in the past (Maryland, Massachusetts, Rhode Island) but I don't know how correct or not that is.

src/common/models/ICUHeadroom.ts
```diff
@@ -0,0 +1,230 @@
...
+  timeseries: Array<CANPredictionTimeseriesRow | null>,
+  actuals: _Actuals,
+): { series: Array<number | null>; method: NonCovidPatientsMethod } {
+  const actualTotalPatients = actualTimeseries.map(
+    row => row?.ICUBeds.currentUsageTotal || null,
+  );
+  const actualCovidPatients = actualTimeseries.map(
+    row => row?.ICUBeds.currentUsageCovid || null,
+  );
+  if (hasRecentData(actualTotalPatients, dates)) {
```
igorkofman, line 183: In theory, the recent actuals might be non-overlapping I guess? That seems pretty unlikely to happen in practice so not worth worrying about.

igorkofman:
This is great - thanks!

src/common/models/ICUHeadroom.ts
mikelehen: Oops.  I must have forgotten to push the commit where I tweaked it. It's 15% now.

src/common/models/ICUHeadroom.ts
mikelehen: Yeah, that's true.  Something to keep an eye on, I guess.

PR #1015, opened 2020-07-03
Add rate limiting handling to email alerts.

mikelehen:
The good news is it works.

The bad news is I had to use patch-package to patch the createsend-node package since it doesn't expose the response headers. 🤮

### DIFF
--- a/patches/createsend-node+0.8.2.patch
+++ b/patches/createsend-node+0.8.2.patch
@@ -1,0 +1,39 @@
+diff --git a/node_modules/createsend-node/lib/interface/utils.js b/node_modules/createsend-node/lib/interface/utils.js
+index acc6561..b39f481 100644
+--- a/node_modules/createsend-node/lib/interface/utils.js
++++ b/node_modules/createsend-node/lib/interface/utils.js
+@@ -44,22 +44,20 @@ module.exports = {
+                 if (error) { return callback(error); }
+
+                 var code = response.statusCode;
+-                if (code == 200 || code == 201 || code == 202) {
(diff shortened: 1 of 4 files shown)


patches/createsend-node+0.8.2.patch
```diff
@@ -0,0 +1,39 @@
...
+-                            callback(null, JSON.parse(body));
+-                        } catch (e) {
+-                            callback(new Error('Unable to parse JSON.  Error: ' + e + '\nBody:\n' + body));
+-                        }
+-                    } else {
+-                        callback(null);
++                if (body) {
++                    try {
++                        body = JSON.parse(body);
++                        body.headers = response.headers;
```
mikelehen, line 21: This is the crux of my patch.  I parse the body and then add the headers to it.  In case it's easier to just look at the resulting code, it's below.  I consolidated the body parsing to make it cleaner.
```
        request(
            options,
            function (error, response, body) {
                if (error) { return callback(error); }

                var code = response.statusCode;
                if (body) {
                    try {
                        body = JSON.parse(body);
                        body.headers = response.headers;
                    } catch (e) {
                        callback(new Error('Unable to parse JSON.  Error: ' + e + '\nBody:\n' + body));
                        return;
                    }
                }

                if (code == 200 || code == 201 || code == 202) {
                    callback(null, body);
                } else {
                    callback(body);
                }
            }
        );
```

mikelehen:
Yeah, that's a good point...  We're probably not saving that much by using createsend-node at this point.

PR #1019, opened 2020-07-04
Case Density Chart - death-based, with uncertainty intervals

pnavarrc:
Testing one of the ideas from the meeting on July 2nd. This PR is just a draft, it will need more work to merge it to production.

The line chart is the case density based on deaths, calculated according to [PRD Case Density](https://docs.google.com/document/d/1qOqlK-Apw-8gn9ZDHY_KwgH1f960Qvpbsg9GLm_6XCA/edit). Since the case density is already calculated based on a 1% mortality rate, I scaled the series to calculate the lower and upper bounds of the case density estimate, using 1.5% and 0.5% mortality rate.

[**Preview**](https://covid-projections-git-pablo-case-density-v2.covidactnow.vercel.app/)

The case density was adjusted as follows:

```
case density (1%) = (7 day moving avg of deaths / 1%) / (population / 100k)

case density (m%) = (7 day moving avg of deaths / m%) / (population / 100k)
      = (m% / 1%) * (7 day moving avg of deaths / m%) / (population / 100k)
      = (m% / 1%) * case density (1%)
```

### Washington
image

### Texas
image

### New York
image


src/components/Charts/ChartCaseDensity.tsx
```diff
@@ -0,0 +1,205 @@
...
+/**
+ * Since the case density is already calculated with 1%, we need to scale
+ * accordingly to calculate the case density with a different mortality
+ * rate m%.
+ *
+ * case density (1%) = (7 day moving avg of deaths / 1%) / (population / 100k)
+ *
+ * case density (m%) = (7 day moving avg of deaths / m%) / (population / 100k)
+ *       = (m% / 1%) * (7 day moving avg of deaths / m%) / (population / 100k)
+ *       = (m% / 1%) * case density (1%)
```
mikelehen, line 46: These last two lines don't seem right and don't match the code below (the code does `(1% / m%) * case density (1%)`, right?)

I guess it should maybe be:

> = (1% / m%) * (7 day moving avg of deaths / 1%)) / (population / 100k)
> = (1% / m%) * case density (1%)

But honestly I'm not sure if including the derivation here is all that helpful.  And it'd be better to just move this logic into Projection.ts if we can so we don't need to convert back/forth.

pnavarrc: Oops, right, the coding is what I meant — it makes sense to move it to Projections.ts 👍

mikelehen:
Tests were failing because code in develop is relying on getMetricValue() working for CASE_DENSITY (even though it's not really used).  I merged in develop and fixed it up.

PR #1027, opened 2020-07-07
Updating schema

ghop02:
Updates the schemas with the latest types. 

I made changes with the goal of least disturbance, but would love additional guidance on how to better deal with types (a lot of union types before are now just one type).  I think we could in some cases or not export the types.  I'll flag a couple examples further down


src/api/index.ts
```diff
@@ -40,15 +34,13 @@ const ApiInterventions: { [intervention: string]: string } = {
 type InterventionKey = keyof typeof INTERVENTIONS;
 
 /** Represents timeseries for any kind of region. */
-export type Timeseries = CountyTimeseries | StateTimeseries;
+export type Timeseries = AreaTimeseries;
```
ghop02, line 37: Here's an example of awkward typing, should we just `import { Timeseries } from 'schema/CovidActNowAreaTimeseries;'` elsewhere?

ghop02:
Closing in favor of https://github.com/covid-projections/covid-projections/pull/1064

PR #1028, opened 2020-07-08
Sgoldblatt/email validation

goldblatt:
Validate the emails a bit before we write them to firestore

### DIFF
--- a/src/components/Newsletter/Newsletter.js
+++ b/src/components/Newsletter/Newsletter.js
@@ -1,11 +1,13 @@
 import firebase from 'firebase';
 import React from 'react';
-import { StyledNewsletter, InputHolder } from './Newsletter.style';
+import { StyledNewsletter, InputHolder, InputError } from './Newsletter.style';
 import Chip from '@material-ui/core/Chip';
 import Autocomplete from '@material-ui/lab/Autocomplete';
 import TextField from '@material-ui/core/TextField';
 import { getLocationNames } from 'common/locations';
 import { getFirebase } from 'common/firebase';
+
+const EMAIL_REGEX = /^([a-zA-Z0-9_\.\-])+\@(([a-zA-Z0-9\-])+\.)+([a-zA-Z0-9]{2,4})+$/;
 
 class Newsletter extends React.Component {
   constructor() {
@@ -15,17 +17,15 @@ class Newsletter extends React.Component {
     this.alertsSelectionArray = [];
     this.defaultValues = [];
     this.autocompleteOptions = getLocationNames();
-    this.state = { checked: true, email: '' };
+    this.state = { checked: true, email: '', errorMessage: '' };
     this.submitForm = this.submitForm.bind(this);
     this.handleSelectChange = this.handleSelectChange.bind(this);
   }
 
   async submitForm(e) {
     e.preventDefault();
-    // can't submit the form without the email entered
-    if (this.state.email) {
-      // TODO(michael): We can stop sending alert locations to Campaign Monitor
-      // once we've migrated everybody over to Firestore.
+    // can't submit the form without the email entered and email that is valid
+    if (this.state.email && EMAIL_REGEX.test(this.state.email)) {
       await this.subscribeToAlerts();
 
       window.gtag('event', 'subscribe', {
@@ -49,6 +49,8 @@ class Newsletter extends React.Component {
         .then(data => {
           console.log(data);
         });
+    } else {
+      this.setState({ errorMessage: 'Must supply a valid email address' });
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/components/Newsletter/Newsletter.js
```diff
@@ -62,7 +64,7 @@ class Newsletter extends React.Component {
       // Merge the locations with any existing ones since that's _probably_ what the user wants.
       await db
         .collection('alerts-subscriptions')
-        .doc(email)
+        .doc(email.toLocaleLowerCase())
```
goldblatt, line 67: i think we should lowercase all the emails in our db

src/components/Newsletter/Newsletter.js
```diff
@@ -1,12 +1,14 @@
...
 import React from 'react';
-import { StyledNewsletter, InputHolder } from './Newsletter.style';
+import { StyledNewsletter, InputHolder, InputError } from './Newsletter.style';
 import Chip from '@material-ui/core/Chip';
 import Autocomplete from '@material-ui/lab/Autocomplete';
 import TextField from '@material-ui/core/TextField';
 import { getLocationNames } from 'common/locations';
 import { getFirebase } from 'common/firebase';
 
+const EMAIL_REGEX = /^([a-zA-Z0-9_\.\-])+\@(([a-zA-Z0-9\-])+\.)+([a-zA-Z0-9]{2,4})+$/;
```
mikelehen, line 10: Did you make this yourself or copy it from somewhere?  If the latter, can you site the source?

I think email regexps are kinda' fragile... Like this will probably break with unicode, which is kinda' maybe supported in emails?  Not sure.  It's tempting to just use a suuuuper simple regex like https://stackoverflow.com/a/742455 which will catch the most common typos.  🤷 

FWIW- I also don't understand the `+` at the end...  It seems like `([a-zA-Z0-9]{2,4})` says you must have 2-4 letters or digits after the final dot, but then the `+` says you can have more than one of those sets of 2-4 letters or digits...  So it'd be equivalent and simpler to just have `[a-zA-Z0-9]+` (i.e. an arbitrary number of letters/digits) which makes sense to me (TLDs can definitely be longer than 4 chars these days (e.g. `.design`))

src/components/Newsletter/Newsletter.js
```diff
@@ -1,12 +1,14 @@
...
+
```
mikelehen, line 11: prettier (I guess?) wants you to unescape a bunch of things it seems.

src/components/Newsletter/Newsletter.js
mikelehen: 👍

src/components/Newsletter/Newsletter.js
```diff
@@ -75,11 +77,17 @@ class Newsletter extends React.Component {
...
     this.alertsSelectionArray = selectedOption;
-    this.setState({
-      alertSignUps: this.alertsSelectionArray
-        .map(item => item.full_fips_code)
-        .join(','),
-    });
+  };
+
+  handleSetEmail = email => {
+    // regex check the email
```
mikelehen, line 83: This comment is confusing to me.

src/components/Newsletter/Newsletter.js
```diff
@@ -202,13 +201,17 @@ class Newsletter extends React.Component {
               maxlength="200"
               name="cm-yddtsd-yddtsd"
               required=""
+              pattern="^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$"
```
mikelehen, line 204: Hrmmm...  Why is this a different regex?  And is it actually used?

goldblatt: ah ya i was hopping this would just...work but it didn't seem to do anything

src/components/Newsletter/Newsletter.js
goldblatt: ahh yeah i did copy it from somewhere (if only my personal regexs were so good and sorry ya i had another one that was just {2, }. Happy for the symbol one (maybe `^[^@\s]+@[^@\s\.]+\.[^@\.\s]+$` since it's a lil more specific

goldblatt: /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/ ended up using this one from https://emailregex.com/

src/components/Newsletter/Newsletter.js
goldblatt: coook okay fixed

src/components/Newsletter/Newsletter.js
goldblatt: okay once again https://ui.dev/validate-email-address-javascript/ using that now simpler and less confusing

src/components/Newsletter/Newsletter.js
goldblatt: removed it

PR #1044, opened 2020-07-09
Chris/actuals case density

ghop02:


mikelehen:
I opened a separate PR for this which has now been merged.

PR #1046, opened 2020-07-10
Prevalence - Update the critical threshold for contact tracing

pnavarrc:
Removes the `CRITICAL` level for the contact tracing chart and moves the upper limit of the orange level from 7% to 10%.

This is done by changing the upper limit of the `CRITICAL` threshold to `0` and updating the upper level of `HIGH` from 7% to 10%. I updated how regions (the equivalent of levels, but with upper and lower limits) to filter out regions such that their lower and upper limit are too close (< 1e3).

Can you review @mikelehen ?

### DIFF
--- a/src/common/metrics/contact_tracing.ts
+++ b/src/common/metrics/contact_tracing.ts
@@ -14,11 +14,6 @@ const SHORT_DESCRIPTION_MEDIUM_HIGH =
 const SHORT_DESCRIPTION_HIGH = 'Enough tracing to help contain COVID';
 const SHORT_DESCRIPTION_UNKNOWN = 'Insufficient data to assess';
 
-const LIMIT_LOW = 0.07;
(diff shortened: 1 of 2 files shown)


src/common/metrics/contact_tracing.ts
```diff
@@ -30,28 +25,28 @@ export const REVERSE_ORDER = true;
 export const CONTACT_TRACING_LEVEL_INFO_MAP: LevelInfoMap = {
   [Level.LOW]: {
     level: Level.LOW,
-    upperLimit: LIMIT_LOW,
+    upperLimit: 0,
```
mikelehen, line 28: Maybe add a comment that this is intentionally disabling the "critical" region.

mikelehen: added.

PR #1047, opened 2020-07-10
Handle state/county data without projections.

mikelehen:


mikelehen:
Thanks!  Merged both.

PR #1059, opened 2020-07-13
Enable location page for places with actuals but not projections.

mikelehen:


src/common/models/Projection.ts
```diff
@@ -188,7 +189,7 @@ export class Projection {
     );
 
     const cumulativeActualDeaths = this.smoothCumulatives(
-      actualTimeseries.map(row => row?.cumulativeDeaths || null),
+      actualTimeseries.map(row => row && row.cumulativeDeaths),
```
mikelehen, line 192: The old code was accidentally turning `0`s into `null`s. 😞

src/common/models/Projection.ts
```diff
@@ -342,22 +348,29 @@ export class Projection {
     summaryWithTimeseries: RegionSummaryWithTimeseries,
     futureDaysToInclude: number,
   ) {
```
mikelehen, line 350: We used to align the actuals timeseries to match the projections timeseries (filling in nulls for dates that didn't exist).  Now, we essentially do it in both directions.  Insert nulls into the projections timeseries to match the actuals timeseries.

src/common/models/Projection.ts
```diff
@@ -379,7 +392,9 @@ export class Projection {
 
     // only keep futureDaysToInclude days ahead of today
     const now = new Date();
-    const days = dates.findIndex(date => date > now) + futureDaysToInclude;
+    const todayIndex = dates.findIndex(date => date > now);
+    const days =
+      todayIndex >= 0 ? todayIndex + futureDaysToInclude : dates.length;
```
mikelehen, line 397: The old code didn't work right in the case that there are no future days (which is the norm if there are no projections)

src/common/models/Projections.ts
```diff
@@ -95,6 +95,17 @@ export class Projections {
...
+  hasMetric(metric: Metric): boolean {
+    if (metric === Metric.FUTURE_PROJECTIONS) {
+      return (
+        this.baseline?.hasHospitalProjections &&
+        this.projected.hasHospitalProjections
+      );
+    } else {
+      return this.getMetricValue(metric) !== null;
+    }
+  }
```
mikelehen, line 107: New helper to clean up some code now that future projections may not be present.

src/common/models/Projections.ts
```diff
@@ -95,6 +95,17 @@ export class Projections {
     };
   }
 
+  hasMetric(metric: Metric): boolean {
+    if (metric === Metric.FUTURE_PROJECTIONS) {
+      return (
+        this.baseline?.hasHospitalProjections &&
+        this.projected.hasHospitalProjections
```
ghop02, line 102: Should the metric simply return null for this case?  this logic almost feels like it could live on the metric: `metric.hasValue()`.  Note that I'm suggesting this without much context :) just a first impression

src/components/LocationPage/ChartBlock.tsx
```diff
@@ -51,6 +51,9 @@ function ChartBlock(props: {
   const futureProjectionsDisabled =
     FUTURE_PROJECTIONS_DISABLED_STATES.includes(props.stateId) &&
     !props.countyId;
+  const haveFutureProjections =
```
ghop02, line 54: should this be `hasFutureProjections`? hah what's the subject here ✏️ ?

src/common/models/Projection.ts
```diff
@@ -342,22 +348,29 @@ export class Projection {
...
-    const earliestDate = moment.min(
-      summaryWithTimeseries.timeseries.map(row => moment.utc(row.date)),
-    );
-    const latestDate = moment.max(
-      summaryWithTimeseries.timeseries.map(row => moment.utc(row.date)),
+    const timeseriesRaw = summaryWithTimeseries.timeseries;
+    const actualsTimeseriesRaw = summaryWithTimeseries.actualsTimeseries;
+    assert(
+      actualsTimeseriesRaw.length > 0,
+      `FIPS ${this.fips} missing actuals timeseries!`,
```
ghop02, line 355: My first impression is that I'm not sure i would expect this assertion to be true - are there not counties that don't have any data? (i.e. if they don't have any cases ever?)

src/common/models/Projections.ts
mikelehen: `Metric` is just an enum, so I'm not sure what you mean by `metric.hasValue()`.  As for returning null, `Metric.FUTURE_PROJECTIONS` is weird in that it has a metric chart, but no metric value (we don't choose a specific value and grade it)...  so it ends up special-cased in places like this.  It's not ideal, but I don't have a better solution offhand.

src/components/LocationPage/ChartBlock.tsx
mikelehen: Heh.  In my mind the subject is "we" ("do we have future projections? yes/no").

src/common/models/Projection.ts
mikelehen: I added this assert partially to test that hypothesis.  And since `yarn update-location-summaries` doesn't hit the assert, I think it means it is safe for now.  And since the code (currently) relies on the assert passing, we'll need to fix the code if it ever fails, so I think it's good to assert it to catch issues in the future.

To your specific question, my expectation is that either a county will have never had cases, so we should have an actuals timeseries with all `0`s or we will not know if it had cases, in which case we should continue to just not include it in our API (if we don't have actuals or projections, what's the point)?  This seems to be the current state of affairs.  But we can always revisit in the future.

PR #1060, opened 2020-07-13
Merge case density by cases and deaths.

mikelehen:


src/common/models/Projection.ts
```diff
@@ -88,9 +88,14 @@ export interface CaseDensityRange {
...
  * series and range).
  */
-const CASE_FATALITY_RATIO_LOWER = 0.005;
-const CASE_FATALITY_RATIO_UPPER = 0.015;
+// TODO: We were intending to calculate a low/high range for
+// caseDensityByDeath, based on a range of CFRs, but this doesn't work when
+// we merge with caseDensityByCases which has no range. So for now,
+// we are punting.
+// const CASE_FATALITY_RATIO_LOWER = 0.005;
+// const CASE_FATALITY_RATIO_UPPER = 0.015;
```
mikelehen, line 96: I left the CFR range stuff in here but commented out in case we want to bring it back, but I can also just delete it if you prefer.

pnavarrc: I think we can leave it for now, we will need to clean up more things if we confirm that the uncertainty intervals (we can reuse `ChartZones` instead of having a custom chart for case density, for example)

PR #1062, opened 2020-07-13
Refactor: Use smoothWithRollingAverage() instead of getWeeklyAverageForDay().

mikelehen:
This is a purely optional refactor that we could punt for now, but I started down this path while I was trying to track down an issue I was seeing.  If you're not a fan or are busy, we could hold off on this.

### DIFF
--- a/src/common/models/Projection.ts
+++ b/src/common/models/Projection.ts
@@ -108,7 +108,6 @@ export class Projection {
   readonly currentCumulativeCases: number | null;
   readonly currentContactTracerMetric: number | null;
   readonly stateName: string;
-  readonly dailyPositiveTests: Array<number | null>;
   readonly currentCaseDensityByCases: number | null;
   readonly currentCaseDensityByDeaths: number | null;
 
@@ -125,7 +124,8 @@ export class Projection {
   private readonly cumulativeInfected: Array<number | null>;
   private readonly cumulativePositiveTests: Array<number | null>;
   private readonly cumulativeNegativeTests: Array<number | null>;
-  private readonly dailyNegativeTests: Array<number | null>;
+  private readonly smoothedDailyNegativeTests: Array<number | null>;
+  private readonly smoothedDailyPositiveTests: Array<number | null>;
   private readonly rtRange: Array<RtRange | null>;
   // ICU Utilization series as values between 0-1 (or > 1 if over capacity).
   private readonly icuUtilization: Array<number | null>;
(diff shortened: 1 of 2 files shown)


src/common/models/Projection.ts
```diff
@@ -262,26 +273,6 @@ export class Projection {
     return this.dates[this.dates.length - 1];
   }
 
-  /**
-   * Returns the delta between the number of new cases in the past week and the
-   * number in the prior week.
-   */
-  get weeklyNewCasesDelta(): number {
```
mikelehen, line 269: This was just dead code that was never used and we have no plans to use...

src/common/models/Projection.ts
```diff
@@ -712,6 +654,12 @@ export class Projection {
     let sum = 0;
     let count = 0;
     for (let i = 0; i < data.length; i++) {
+      const oldValue = i < days ? null : data[i - days];
+      if (oldValue !== null) {
+        sum -= oldValue;
+        count--;
+      }
```
mikelehen, line 661: Embarrassingly, when I compared before/after this change, the shape of the curves changed a bit, and digging into it, it turns out that this code was accidentally computing an 8-day moving average instead of 7-day.  We were computing the average before removing values that had already fallen out of the 7-day window.  So I had to move this code from down below to here to do it in the right order. 😬

src/common/models/Projection.ts
```diff
@@ -394,88 +385,44 @@ export class Projection {
...
-          }
+    return actualTimeseries.map((row, i) => {
+      if (row && row.contactTracers) {
+        const contactTracers =
+          CONTACT_TRACER_STATE_OVERRIDES[this.stateName] || row.contactTracers;
+        const cases = this.smoothedDailyPositiveTests[i];
+        if (cases) {
+          return contactTracers / (cases * TRACERS_NEEDED_PER_CASE);
         }
+      }
```
pnavarrc, line 399: Should we return 100% if `cases = 0`? It's a bit of an edge case. We probably don't need to worry about it since we are using the smooth series now.

src/common/models/Projection.ts
pnavarrc: nice catch! hard to spot 🙌

src/common/models/Projection.ts
mikelehen: Oh, that's a good question.  Right now we only show contact tracing at the state-level, so this hasn't come up for current data (I don't think any state has reached 0 cases)...  

I'm going to leave it as-is for now...  In the case that we start showing contact tracing for counties (which may be more expected to have zero case counts), I'm not sure if it makes sense to omit the metric or show 100% when they haven't had any cases (yet).

PR #1065, opened 2020-07-13
Prevalence - minor dev fixes

pnavarrc:
Forgot to add keys to the zone annotations and removed a comma on the `index.html` template that was causing a warning as well.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1,21 +1,19 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="utf-8" />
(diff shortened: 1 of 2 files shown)


public/index.html
```diff
@@ -1,21 +1,19 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="utf-8" />
-    <!--link rel="icon" href="%PUBLIC_URL%/favicon.ico" /-->
-    <meta
-      name="viewport"
-      content="minimum-scale=1, initial-scale=1, width=device-width,"
```
pnavarrc, line 8: the leading comma makes the browser think that there is an empty key-value 🤷

PR #1066, opened 2020-07-13
Prevalence - Status text copy

pnavarrc:
This PR adds the copy from [CAN Phase 3 Webapp Text](https://docs.google.com/document/d/1Jsj4zOksCJcgjQLUo3rwVAobEUkGLFukD1K71IqVNpo/edit#) for the prevalence metric.

Note: The copy for case density is not complete in the document, adding what we have for now.

### DIFF
--- a/src/common/metrics/case_density.ts
+++ b/src/common/metrics/case_density.ts
@@ -1,7 +1,9 @@
 import { COLOR_MAP } from 'common/colors';
 import { Level, LevelInfoMap } from 'common/level';
 import { Projection, CASE_FATALITY_RATIO } from 'common/models/Projection';
-import { formatInteger, formatPercent } from 'common/utils';
+import { formatInteger, formatPercent, formatDecimal } from 'common/utils';
+import { levelText } from 'common/utils/chart';
+import { getLevel, Metric } from 'common/metric';
 
 export const METRIC_NAME = 'Case Density';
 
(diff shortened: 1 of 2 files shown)


src/common/models/Projection.ts
```diff
@@ -225,6 +226,10 @@ export class Projection {
     this.currentCaseDensity = lastValue(
       this.caseDensityRange.map(range => range && range.caseDensity),
     );
+    this.currentDailyDeaths =
+      this.currentCaseDensity === null
+        ? null
+        : CASE_FATALITY_RATIO * this.currentCaseDensity;
```
pnavarrc, line 232: @mikelehen do we have daily deaths that we can use? This is for the status text copy, where we show the number of daily deaths/100k and the number of daily cases/100k

> Over the last week, Mississippi has reported [10] new cases and [0.1] new deaths per day for every 100,000 residents.
>
> If these rates continue, we estimate less than 1-10% of Mississippi’s population will be infected in the next year.

For the cases, it's just the case density metric `currentCaseDensity`. I used `CASE_FATALITY_RATIO * currentCaseDensityByDeaths` first to calculate deaths, but in some cases (Benson County, ND) this is `0`, even though the case density is not zero (good catch @chasulin!).

image


See [CAN Phase 3 - Webapp Text](https://docs.google.com/document/d/1Jsj4zOksCJcgjQLUo3rwVAobEUkGLFukD1K71IqVNpo/edit#)

mikelehen: I think this is normal / expected, right?  Benson County has cases but no deaths right now?  In general we expect cases to increase first, since more people will start getting sick, and then eventually a percentage of them may die.

src/common/metrics/case_density.ts
```diff
@@ -52,5 +54,27 @@ rate). Note that this will not match reported cases in many states, as
...
+  const statusText1 = `Over the last week, ${locationName} has reported
+   ${dailyCases} new cases and ${dailyDeaths} new deaths per day for every
+    100,000 residents.`;
+
+  const statusText2 = levelText(
+    level,
+    `If these rates continue, we estimate less than 1% of ${locationName}’s population will be infected in the next year.`,
+    `If these rates continue, we estimate less than 1-10% of ${locationName}’s population will be infected in the next year.`,
+    `If these rates continue, we estimate less than 10-50% of ${locationName}’s population will be infected in the next year. Caution is warranted.`,
+    `If these rates continue, we estimate less than >50% of ${locationName}’s population will be infected in the next year. Aggressive action urgently needed.`,
```
mikelehen, line 76: FWIW- 'less than >50%' doesn't really make sense.  

But I think all of this copy will probably need tweaking.

pnavarrc: good catch - I'm updating it in the PR and adding a note in the copy doc.

PR #1077, opened 2020-07-14
Sharecard fix (location page)

chasulin:
- edits location page sharecard to make all 5 metrics visible

Before:

Screen Shot 2020-07-14 at 1 31 01 PM

After:

Screen Shot 2020-07-14 at 1 33 06 PM


chasulin:
going with a different fix

PR #1078, opened 2020-07-14
Sharecard fix (homepage)

chasulin:
- Edits sharecard so map+legend both fit in entirety

Before:

Screen Shot 2020-07-14 at 1 31 14 PM

After: 

Screen Shot 2020-07-14 at 1 31 57 PM


joshziman:
LGTM

PR #1080, opened 2020-07-14
charts - minor improvements for mobile and share images

pnavarrc:
### Changes
- Calculate the end date of the chart to ensure that the threat level labels don't overlap the last point of the time series instead of just adding 2 weeks and hope for the best.
- Changed the format of the x-axis tick labels from full month to 3-letter abbreviations (September → Sep).
- Moved the current point annotation from being on the right of the last point to be on top of the last point, to minimize the chance of overlapping with the zone labels.

#### Mobile

Before → After
image

#### Share Image

Before
image

After
image

### Notes
- There is some discussion around getting rid of the zone labels on mobile, but I'm going to submit a separate PR so we can get feedback on that change.
- The margins can be optimized as well, I will do that on a separate PR too.

### DIFF
--- a/src/components/Charts/Axis.tsx
+++ b/src/components/Charts/Axis.tsx
@@ -15,7 +15,7 @@ export const AxisBottom = (props: SharedAxisProps<Date>) => {
   const [dateStart, dateStop] = props.scale.domain();
   const [minPx, maxPx] = props.scale.range() as number[];
   const isSmallDevice = maxPx - minPx < 600;
-  const dateFormat = isSmallDevice ? 'MMMM' : 'MMM DD';
+  const dateFormat = isSmallDevice ? 'MMM' : 'MMM DD';
 
   const tickValues = isSmallDevice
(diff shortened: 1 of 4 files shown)


joshziman:
LGTM. 

A minor thing is that we're displaying the x axis label for Sept, which starts >30 days from now, and that seems unnecessary upon first thought.

pnavarrc:
> A minor thing is that we're displaying the x-axis label for Sept, which starts >30 days from now, and that seems unnecessary upon first thought.

yes, that's an artifact of the way I'm calculating spaces. Before, we were adding 2 weeks hoping that the equivalent in pixels would be enough to show the labels. It's good enough on desktop, but not on mobile. This PR changes the approach, it calculates what should be the end date so the labels don't overlap with the time series.

pnavarrc:
I don't want to increase the padding too much (that takes space from the data), I will keep more improvements coming!

PR #1086, opened 2020-07-15
alert emails - make create_lists_to_email.ts resumable

pnavarrc:
https://trello.com/c/ZldwpjOI/295-make-email-alert-sending-fully-resumable

The workflow to send email alerts creates a collection of emails to be sent for the locations that changed their threat level and then send the emails, updating the `sentAt` field on each email document with the timestamp of when the email was sent. 

If the workflow fails while sending the emails, we will need to re-run the workflow, however, this resets the `sentAt` field on the emails and some users would get send the alert email more than once. This PR updates the script that generates the list of emails to be sent so it doesn't overwrite documents that already exist.

The first two commits are linting fixes and removing unused dependencies. The commit with the actual changes to the script is https://github.com/covid-projections/covid-projections/pull/1086/commits/c11067d9883af8b6544d07db2841b297bdecb006

### Preparation
 - Download and unzip the [alerts.json.zip](https://github.com/covid-projections/covid-projections/files/4928598/alerts.json.zip) file in the `scripts/alert_emails` directory.
- Make sure that you have the `google-services-account.json` file in the `scripts/alert_emails` directory.

### Update collection
- Add an email on the [alert-subscriptions](https://console.firebase.google.com/u/1/project/covidactnow-dev/database/firestore/data~2Falerts-subscriptions) collection, add a `locations` array with the string `"11"` as element.
- Run `yarn create-lists-to-email alerts.json 497`
- The script should update the [snapshots/497/locations/11/emails](https://console.firebase.google.com/u/1/project/covidactnow-dev/database/firestore/data~2Fsnapshots~2F497~2Flocations~2F11~2Femails) collection, adding the email from before to the list.
- Check that the `sentAt` field of the email document is set to `null`
- Note: Since some documents subscribed to alerts for that location already exist, those won't be updated (check existing emails in the collection).

### Failure

Locally, comment-out on the [following lines](https://github.com/covid-projections/covid-projections/pull/1086/files#diff-d1cea5e1bd24348eaf5e4898221c4801R61-R64) and replace them with `await Promise.reject({ code: 5 })` to simulate a rejection from Firestore. 

```ts
try {
  // await db
  //   .collection('snapshots')
  //   .doc(`${snapshotId}/locations/${fips}/emails/${email}`)
  //   .create({ sentAt: null });
  await Promise.reject({ code: 5 });
} catch (updateError) {
  // ...
}
```

The script should fail and log the error code

```
$ yarn create-lists-to-email alerts.json 497                      
yarn run v1.22.4
$ ts-node-script -r tsconfig-paths/register scripts/alert_emails/create_lists_to_email.ts alerts.json 497
Done. Updated the emails for 1 locations.
Error updating emails for location 11 { code: 5 }
```


pnavarrc:
[Make email alert sending fully resumable.](https://trello.com/c/ZldwpjOI/295-make-email-alert-sending-fully-resumable)

scripts/alert_emails/create_lists_to_email.ts
```diff
@@ -1,69 +1,95 @@
...
+  alertsFileName: string,
+  snapshotId: string,
+) => {
+  const db = getFirestore();
+  const alertsByLocation = readAlertsFile(alertsFileName);
+  const fipsList = Object.keys(alertsByLocation);
 
-  const { fipsToAlertFilename, currentSnapshot } = await parseArgs();
-  const alertPath = path.join(outputFolder, fipsToAlertFilename);
+  fipsList.forEach(async (fips: string) => {
```
mikelehen, line 43: Is Array.forEach() aware of Promises?  If so, then this needs an `await`...  But I think this probably needs to be `for (const fips of fipsList) { ...` which is also just easier to read. 😁

scripts/alert_emails/create_lists_to_email.ts
```diff
@@ -1,69 +1,95 @@
...
- * You can run via `yarn create-lists-to-email fipsToAlertFilename currentSnapshot`
+ *   $ yarn create-lists-to-email alerts.json snapshotId
  */
-(async () => {
-  const outputFolder = path.join(__dirname);
-  await fs.ensureDir(outputFolder);
+const updateSnapshotEmails = async (
+  alertsFileName: string,
+  snapshotId: string,
+) => {
```
mikelehen, line 38: nit: Is there a reason to declare functions like this (`const foo = ...` instead of `async function foo(...)` ?

scripts/alert_emails/create_lists_to_email.ts
```diff
@@ -1,69 +1,95 @@
...
+      .collection('alerts-subscriptions')
+      .where('locations', 'array-contains', fips)
+      .get();
 
-  const rawdata = fs.readFileSync(alertPath, 'utf8');
-  const locationsWithAlerts: { [fips: string]: Alert } = JSON.parse(rawdata);
+    // Remove invalid and repeated emails for the location from the list
+    const emailList = _.uniq(
+      emailQuerySnapshot.docs
+        .map((emailDoc: { id: string }) => emailDoc.id)
```
mikelehen, line 52: Do you need `: { id: string }` ?  I'd expect TypeScript to know that emailDoc is a `firestore.DocumentSnapshot` object, which has an `id` property...

scripts/alert_emails/create_lists_to_email.ts
```diff
@@ -1,69 +1,95 @@
...
+        emailList.map(async (email: string) => {
+          try {
+            await db
+              .collection('snapshots')
+              .doc(`${snapshotId}/locations/${fips}/emails/${email}`)
+              .create({ sentAt: null });
+          } catch (updateError) {
+            // We don't want to overwrite existing emails for a location to avoid
+            // double-sending alerts, in case this is a re-run of the send-alert-emails
+            if (updateError.code !== ERROR_CODE_ALREADY_EXISTS) {
```
mikelehen, line 68: Is it worth doing a `console.error('Firestore write failed', updateError)` as well?  In particular if Promise.all() doesn't surface multiple errors?

scripts/alert_emails/create_lists_to_email.ts
```diff
@@ -1,69 +1,95 @@
...
-  const alertPath = path.join(outputFolder, fipsToAlertFilename);
+  fipsList.forEach(async (fips: string) => {
+    const emailQuerySnapshot = await db
+      .collection('alerts-subscriptions')
+      .where('locations', 'array-contains', fips)
+      .get();
 
-  const rawdata = fs.readFileSync(alertPath, 'utf8');
-  const locationsWithAlerts: { [fips: string]: Alert } = JSON.parse(rawdata);
+    // Remove invalid and repeated emails for the location from the list
```
mikelehen, line 49: Was there a particular reason you added this normalization?  FWIW:
1. The frontend (now) should be lowercasing emails (and preventing duplicates), I think.
2. The send emails script handles "invalid emails" errors from campaign monitor and deletes them (well, moves them to a different collection).

I guess this normalization seems like "general goodness" but it's also maybe not necessary?

scripts/alert_emails/create_lists_to_email.ts
```diff
@@ -1,69 +1,95 @@
-//@ts-ignore createsend has no types and throws an error
+import _ from 'lodash';
 import fs from 'fs-extra';
 import path from 'path';
-import { Alert } from './interfaces';
 import { getFirestore } from './firestore';
 
+const ERROR_CODE_ALREADY_EXISTS = 6;
```
mikelehen, line 6: I think this works:

```
import FirebaseFirestore from '@google-cloud/firestore';
...
FirebaseFirestore.GrpcStatus.ALREADY_EXISTS;
```

scripts/alert_emails/create_lists_to_email.ts
pnavarrc: Right, I didn't notice that the `send_email` script is constantly cleaning up the emails, plus they can't have duplicates (unless they have different capitalization I guess) - will remove thee normalization to keep it simple.

scripts/alert_emails/create_lists_to_email.ts
pnavarrc: Makes sense, the loop is cleaner, updating!

scripts/alert_emails/create_lists_to_email.ts
```diff
@@ -1,59 +1,94 @@
...
+      .where('locations', 'array-contains', fips)
+      .get();
+
+    // Remove invalid and repeated emails for the location from the list
+    const emailList = _.uniq(
+      emailQuerySnapshot.docs
+        .map(emailDoc => emailDoc.id)
+        .filter(isValidEmail)
+        .map(_.toLower),
+    );
```
pnavarrc, line 54: Since we are going to use the snapshots to do analytics on the email subscriptions, I'm going to keep the cleaning for now. Later, we can write a script to do a one-time cleanup of the `alert-subscriptions` collection to remove all the invalid emails.

PR #1087, opened 2020-07-15
change "deaths" to "additional deaths"

eleanoravril:
This is a PR to change the projection to show the additional deaths projected if restrictions are/aren't lifted, subtracting out all deaths that have already occurred to date.
Not ready to merge — currently, subtracting `currentCumulativeDeaths` from `projectedCumulativeDeaths` on line 250 of common/models/Projection.ts does not change the number at all — I think for some reason it's null, and I am not able to tell why, unfortunately.
Unrelated (I think), the numbers on the develop branch are higher than the numbers on the site, and I'm wondering why that is? I was seeing this looking at California — looking at the develop branch (with `yarn start`) shows 60k deaths if all restrictions are lifted and 39k otherwise, but looking at the official site we see 36k deaths if restrictions are lifted and 22k otherwise.


src/common/models/Projection.ts
```diff
@@ -240,6 +242,14 @@ export class Projection {
     return this.cumulativeDeaths[this.cumulativeDeaths.length - 1];
   }
 
+  get finalAdditionalDeaths() {
+    let projectedCumulativeDeaths: number =
+      this.cumulativeDeaths[this.cumulativeDeaths.length - 1] || 0;
+    let currentCumulativeDeaths: number =
+      this.cumulativeActualDeaths[this.cumulativeActualDeaths.length - 1] || 0;
```
mikelehen, line 249: So these timeseries arrays span from the beginning of our data (e.g. March) to 30 days into the future (e.g. August 15).

So `this.cumulativeActualDeaths` is going to have 30 days of nulls at the end of it. 😁 

We have a `lastValue()` helper function that you can use which will take the last value prior to the nulls, which I think will likely solve your issue.

eleanoravril: Ha, yup, that makes a lot of sense — whoops!! Thanks :)

eleanoravril:
@mikelehen — thanks for your help! It works now. Checked multiple states with and without the subtraction to confirm that it was working correctly, and in each case it was fine (confirmed number of deaths via NYT site so that I could double check that it was doing what it was supposed to.

mikelehen:
Great, thanks!  Yep, this looks good.  Let me sync with a few folks on slack to make sure this is ready to ship.

eleanoravril:
Changed header of predicted deaths to be more clear, changed rounding granularity so that we show "Under 500" or "500-1000" deaths

PR #1092, opened 2020-07-16
demo - prevalence by cases only 

pnavarrc:
- changed the prevalence threshold from HIGH to CRITICAL from 50 → 25
- prevalence is calculated by cases only 
- using the snapshot 712


pnavarrc:
Closing - we can still use the previews to compare, but we are not merging this.

PR #1096, opened 2020-07-17
Use cumulativeConfirmedCases instead of cumulativePositiveTests where applicable.

mikelehen:
* We were assuming that cumulativeConfirmedCases and cumulativePositiveTests were interchangeable, but they're not. E.g. Marion County, IN has good cumulativeConfirmedCases, but cumulativePositiveTests is garbage.  So we use cumulativeConfirmedCases now for the contact tracing and case density metrics.

Also added new case density metric to compare tool (mostly because it made it easier
for me to sanity-check all the graphs).

Before this change (with deaths disabled):
image

After this change (with deaths disabled):
image


mikelehen:
cc/ @ghop02 FYI

PR #1098, opened 2020-07-17
prevalence: calculate case density by cases only

pnavarrc:
Updates how the case density metric is calculated to use only cases. See the [discussion on Slack](https://covidactnow.slack.com/archives/C0177KERT89/p1594945641003000)

Note: I left the function just in case we need to put deaths back in the mix, I will remove the ranges and unused properties and functions once we confirm the calculation.

### DIFF
--- a/src/common/models/Projection.ts
+++ b/src/common/models/Projection.ts
@@ -450,27 +450,11 @@ export class Projection {
   }
 
   private calcCaseDensityRange(): Array<CaseDensityRange | null> {
-    return this.caseDensityByDeaths.map((caseDensityByDeaths, i) => {
-      // TODO: We were intending to calculate a low/high range for
-      // caseDensityByDeath, based on a range of CFRs, but this doesn't work when
-      // we merge with caseDensityByCases which has no range. So for now,
-      // we are punting.
-      // const caseDensityByDeathsLow = (CASE_FATALITY_RATIO / CASE_FATALITY_RATIO_UPPER) * caseDensityByDeaths;
-      // const caseDensityByDeathsHigh = (CASE_FATALITY_RATIO / CASE_FATALITY_RATIO_LOWER) * caseDensityByDeaths;
-
-      const caseDensityByCases = this.caseDensityByCases[i];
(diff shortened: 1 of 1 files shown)


src/common/models/Projection.ts
```diff
@@ -450,27 +450,11 @@ export class Projection {
   }
 
   private calcCaseDensityRange(): Array<CaseDensityRange | null> {
-    return this.caseDensityByDeaths.map((caseDensityByDeaths, i) => {
-      // TODO: We were intending to calculate a low/high range for
-      // caseDensityByDeath, based on a range of CFRs, but this doesn't work when
-      // we merge with caseDensityByCases which has no range. So for now,
-      // we are punting.
-      // const caseDensityByDeathsLow = (CASE_FATALITY_RATIO / CASE_FATALITY_RATIO_UPPER) * caseDensityByDeaths;
-      // const caseDensityByDeathsHigh = (CASE_FATALITY_RATIO / CASE_FATALITY_RATIO_LOWER) * caseDensityByDeaths;
```
mikelehen, line 459: Should we go ahead and remove the commented-out CASE_FATALITY_RATIO, etc. constants at the top of the file too?

src/common/models/Projection.ts
```diff
@@ -450,27 +450,11 @@ export class Projection {
...
+    return this.caseDensityByCases.map(caseDensity =>
+      caseDensity === null
         ? null
-        : {
-            caseDensity,
-            low: caseDensity,
-            high: caseDensity,
-          };
-    });
+        : { caseDensity, low: caseDensity, high: caseDensity },
```
mikelehen, line 456: Should we have a TODO to get rid of low/high?  (and caseDensityByDeaths)

mikelehen:
Oh.  I think you addressed my comments with your "Note:" in the PR description, which sounds fine.  I like to capture things like that in TODOs just to make 100% sure they don't get lost.

mikelehen:
Went ahead and merged since I'd like the preview site updated. 😁

PR #1139, opened 2020-07-22
Link Harvard key indicators framework.

mikelehen:
I assume this was an oversight and not intentional?

### DIFF
--- a/src/components/Disclaimer/Disclaimer.tsx
+++ b/src/components/Disclaimer/Disclaimer.tsx
@@ -23,7 +23,11 @@ const Disclaimer = ({ metricName }: { metricName: number }) => {
           <Fragment>
             <span>
               Our risk levels for daily new cases are based on the{' '}
-              <a href="https://ethics.harvard.edu/files/center-for-ethics/files/key_metrics_and_indicators_v4.pdf">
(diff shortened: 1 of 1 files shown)


src/components/Disclaimer/Disclaimer.tsx
```diff
@@ -22,9 +22,11 @@ const Disclaimer = ({ metricName }: { metricName: number }) => {
         {metricName === Metric.CASE_DENSITY && (
           <Fragment>
             <span>
-              Our risk levels for daily new cases are based on the “Key Metrics
-              for Covid Suppression” by Harvard Global Health Institute and
-              others.
+              Our risk levels for daily new cases are based on the{' '}
+              <a href="https://ethics.harvard.edu/files/center-for-ethics/files/key_metrics_and_indicators_v4.pdf">
+                “Key Metrics for Covid Suppression”
+              </a>{' '}
```
pnavarrc, line 28: Other external links open in a new tab `target="_blank" rel="noopener noreferrer"`, not sure if we want that in this case.

mikelehen: Oops! Good call.

PR #1144, opened 2020-07-22
Copy edits

chasulin:
Misc. copy tweaks per edits in the webapp doc

### DIFF
--- a/src/common/metrics/location_summary.ts
+++ b/src/common/metrics/location_summary.ts
@@ -9,7 +9,7 @@ const HIGH_NAME = 'Critical';
 const UNKNOWN = 'Unknown';
 
 const LEGEND_SUMMARY_LOW = 'On track to contain COVID';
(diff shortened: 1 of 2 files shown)


src/common/metrics/location_summary.ts
```diff
@@ -9,7 +9,7 @@ const HIGH_NAME = 'Critical';
 const UNKNOWN = 'Unknown';
 
 const LEGEND_SUMMARY_LOW = 'On track to contain COVID';
-const LEGEND_SUMMARY_MEDIUM = 'Controlled disease growth';
+const LEGEND_SUMMARY_MEDIUM = 'Slow disease growth';
```
pnavarrc, line 12: I don't see this in the copy doc, can you double-check?

chasulin: yep! it was suggested by nirav and +1'ed by debbie. ill confirm with debbie

chasulin: confirmed!

PR #1149, opened 2020-07-22
increase concurrency of email alert script

pnavarrc:


scripts/alert_emails/send_emails.ts
```diff
@@ -162,73 +168,95 @@ async function setLastSnapshotNumber(
...
+
+    try {
+      const response = await sendEmail(api, sendData, dryRun);
+      await onEmailSent(email, fips);
+      return response;
+    } catch (err) {
+      if (isInvalidEmail(err)) {
+        await onInvalidEmail(email);
+      } else {
+        errorCount += 1;
```
mikelehen, line 220: I guess this is all the old code did, but maybe we should `console.error(err)` here as well?

scripts/alert_emails/send_emails.ts
```diff
@@ -162,73 +168,95 @@ async function setLastSnapshotNumber(
...
+    const rateLimitRemaining = parseInt(pilotReq.headers['x-ratelimit-limit']);
+    const rateLimitResetSec =
+      parseInt(pilotReq.headers['x-ratelimit-reset']) + 15;
+
+    if (rateLimitRemaining < remainingEmails.length + 10) {
+      console.log(
+        `Rate limit remaining: ${rateLimitRemaining}. Sleeping ${rateLimitResetSec} seconds.`,
+      );
+      await delay(rateLimitResetSec * 1000);
+    }
```
mikelehen, line 242: I think we can drop the `firstEmail` special handling and just move the rate limit handling inside of sendAlertEmail() so that it works similar to before:

1. Try sending the email.
2. Check the response headers.
3. If we hit (or got close to) the rate limit, do the delay, and then just call sendAlertEmail() again to retry.

This means we might have more than one email from the batch hit the rate limit handling and have to sleep / retry.  But I think that's fine.  They'll all do it in parallel, so it doesn't actually add any extra time.  

The one messy bit is it's kinda' ugly if they all log a message about hitting the rate limit.  But since I think this will be rare we can either live with that (💯  okay by me) or else add some sort of flag so that only the first one that hits the rate limit logs the message.

scripts/alert_emails/send_emails.ts
```diff
@@ -162,73 +168,95 @@ async function setLastSnapshotNumber(
...
+
+    const rateLimitRemaining = parseInt(pilotReq.headers['x-ratelimit-limit']);
+    const rateLimitResetSec =
+      parseInt(pilotReq.headers['x-ratelimit-reset']) + 15;
+
+    if (rateLimitRemaining < remainingEmails.length + 10) {
+      console.log(
+        `Rate limit remaining: ${rateLimitRemaining}. Sleeping ${rateLimitResetSec} seconds.`,
+      );
+      await delay(rateLimitResetSec * 1000);
```
mikelehen, line 241: FYI- If we end up keeping this code as-is, I think you're missing the code to resend `firstEmail`

scripts/alert_emails/send_emails.ts
```diff
@@ -162,73 +168,95 @@ async function setLastSnapshotNumber(
...
+    }
+
+    await Promise.all(
+      remainingEmails.map(email => {
+        return sendAlertEmail(email, fips);
+      }),
+    );
+  }
+
+  for (const fips of locations) {
```
mikelehen, line 251: I think I mentioned this before but it would be more optimal if we precomputed a list of `[email, fips]` tuples and then we called `_.chunk()` on that so we get full parallelization, even across fips. Perhaps add a TODO for now.

scripts/alert_emails/send_emails.ts
```diff
@@ -138,99 +46,104 @@ async function setLastSnapshotNumber(
...
+      .collection(`snapshots/${currentSnapshot}/locations/${fips}/emails/`)
+      .doc(email)
+      .set({ sentAt: admin.firestore.FieldValue.serverTimestamp() });
+  }
+
+  async function sendAlertEmail(email: string, fips: string) {
+    const locationAlert = alertsByLocation[fips];
+    const sendData = generateAlertEmailData(email, locationAlert);
+    if (dryRun) {
+      await onEmailSent(email, fips);
```
mikelehen, line 110: I think this is going to mark the email as sent in Firestore, which we don't want to do for a dry run.

scripts/alert_emails/send_emails.ts
```diff
@@ -138,99 +46,104 @@ async function setLastSnapshotNumber(
...
+      return;
+    }
+
+    try {
+      const response = await createSend.sendClassicEmail(sendData);
+      await onEmailSent(email, fips);
+      return response;
```
mikelehen, line 117: Hrm...  what is this response exactly and why do we return it (I can't tell if it's used)?

In general it's good TypeScript practice to define the types of any return values on your functions.  I think there's a tslint rule we could enable that would enforce it. 😁

scripts/alert_emails/send_emails.ts
```diff
@@ -138,99 +46,104 @@ async function setLastSnapshotNumber(
...
+    } catch (err) {
+      if (isInvalidEmailError(err)) {
+        await onInvalidEmail(email);
+        return;
```
mikelehen, line 121: As an opinionated stylistic note, I vastly prefer if/else statements over early returns.  It's easy to overlook a "return" statement when reading the code and completely misunderstand the control-flow.  My one exception is for edge-case handling (e.g. error conditions) at the top of a function.  Then doing an early-return can make the rest of the function simpler to reason about.

But you're welcome to disregard my style preferences. 😁

scripts/alert_emails/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+    oldLevel,
+    locationName,
+    lastUpdated,
+    locationURL,
+  } = locationAlert;
+
+  const data: AlertTemplateData = {
+    change: 'new risk score',
+    location_name: locationName,
+    img_alt: `Image depicting that ${locationName} went from from level ${Level[oldLevel]} to ${Level[newLevel]}`,
```
mikelehen, line 41: This was probably preexisting but "from from" => "from"

scripts/alert_emails/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+  const {
+    newLevel,
+    oldLevel,
+    locationName,
+    lastUpdated,
+    locationURL,
+  } = locationAlert;
+
+  const data: AlertTemplateData = {
+    change: 'new risk score',
```
mikelehen, line 39: This is temporary for this one send, right?  Can you add a TODO or else just make sure to remember to put this back?

scripts/alert_emails/send_emails.ts
pnavarrc: I'm adding a type for the `sendClassicEmail` method (I was missing the body of the response), but here we don't really need a specific response, just to return so the promise resolves.

scripts/alert_emails/send_emails.ts
pnavarrc: makes sense, updating!

scripts/alert_emails/campaign-monitor.ts
```diff
@@ -55,18 +61,20 @@ class CampaignMonitor {
     });
   }
 
-  async sendClassicEmail(data: EmailSendData): Promise<AxiosResponse> {
+  async sendClassicEmail(
+    data: EmailSendData,
+  ): Promise<AxiosResponse<CampaignMonitorMessageSent[]>> {
```
mikelehen, line 66: nit: Returning an AxiosResponse leaks the implementation details of this function.  Would it make sense to just return `Promise<CampaignMonitorMessageSent>` and return `res.data`?

Feel free to punt.

PR #1151, opened 2020-07-22
improve the spacing of case incidence level annotations

pnavarrc:
From [Slack](https://covidactnow.slack.com/archives/C0177KERT89/p1595453168127300)

The "Low" level for case incidence goes from `0` to `1`, which is too narrow to show the label in the chart. We added some padding before to make room for it, but as a consequence, the chart ends up being a bit misleading, as the spacing between the x-axis and 1 is the same as the space between 1 and 10.

This fix hides the "Low" label and sets the min y-axis to 0 if the current level is not Low, and makes room and adds the "Low" label if the current Level for the location is "Low".

### Example: Level = Medium
image

### Example: Level = Low
image

### DIFF
--- a/src/components/Charts/ChartCaseDensity.tsx
+++ b/src/components/Charts/ChartCaseDensity.tsx
@@ -76,7 +76,7 @@ const ChartCaseDensity: FunctionComponent<{
   // Adjusts the min y-axis to make the Low label fit only if
   // the current level is Low
   const isLow = activeZone.level === Level.LOW;
-  const yAxisMin = isLow ? -9 : 0;
+  const yAxisMin = isLow ? -6 : 0;
   const yAxisMax = Math.min(yAxisLimits[1], capY);
 
   const yScale = scaleLinear({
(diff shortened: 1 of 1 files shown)


src/components/Charts/ChartCaseDensity.tsx
```diff
@@ -62,14 +62,22 @@ const ChartCaseDensity: FunctionComponent<{
...
   const xScale = getZonesTimeScale(minDate, currDate, 0, chartWidth);
 
   const yDataMax = d3max(data, getY) || 100;
   const yAxisLimits = getAxisLimits(0, yDataMax, zones);
-  const [yAxisMin, yAxisMax] = [-9, Math.min(yAxisLimits[1], capY)];
+
+  // Adjusts the min y-axis to make the Low label fit only if
+  // the current level is Low
+  const isLow = activeZone.level === Level.LOW;
+  const yAxisMin = isLow ? -9 : 0;
```
mikelehen, line 79: Do you think it's worth tweaking this `-9` down a bit so that "Low" still fits but there's a bit less extra padding?

Not a big deal at this point, so feel free to punt.

pnavarrc: `-6` seems to be better, updating!

PR #1159, opened 2020-07-24
fix handling of invalid emails

pnavarrc:
Fixes the handling of invalid emails in the `send_emails` script. It also fixes the name of the collection with invalid emails. 

This is how I tested:

### Testing
**Setup**
- Make sure that you have the `CREATE_SEND_TOKEN` environment variable set.
- Make sure that you have the `scripts/alert_emails/google-service-account.json` with the development credentials.
- Unzip the [alerts.json.zip](https://github.com/covid-projections/covid-projections/files/4973819/alerts.json.zip) on `scripts/alert_emails`
 
**Testing**
1. Run `yarn send-test-email invalid@email.com,` - This will try to send a test email to the address `invalid@email.com,` (note the `,` after the email, that passes the regex, but Campaign Monitor rejects it). It should fail and log the `Invalid email` error in the console.
2. Add the invalid email to Firebase and subscribe it to the location `11`
3. Run `yarn create-lists-to-email alerts.json 497` - this will update the `/snapshots/479/locations/11/emails` collection
4. Run `yarn send-emails alerts.json 479 true` - this will try to send an email to all the emails on `/snapshots/479/locations/11/emails`

### DIFF
--- a/scripts/alert_emails/campaign-monitor.ts
+++ b/scripts/alert_emails/campaign-monitor.ts
@@ -1,16 +1,6 @@
 import axios, { AxiosInstance, AxiosResponse } from 'axios';
 import delay from 'delay';
 
-const STATUS_RATE_LIMIT = 429;
-
-function isStatusOK(status: number) {
-  return 200 <= status && status < 300;
-}
-
-function isRateLimitExceeded(status: number) {
-  return status === STATUS_RATE_LIMIT;
-}
-
 export interface EmailSendData {
   Subject: string;
   From: string;
@@ -40,6 +30,30 @@ export interface CampaignMonitorMessageSent {
   MessageID: string;
 }
 
+class CampaignMonitorException {
+  Code: number;
+  Message: string;
+  constructor(code: number, message: string) {
+    this.Code = code;
+    this.Message = message;
+  }
+}
+
+function isStatusOK(status: number) {
+  return 200 <= status && status < 300;
+}
+
+function isRateLimitExceeded(status: number) {
+  return status === 429;
+}
+
+export function isInvalidEmailError(error: CampaignMonitorException) {
+  return (
+    error.Code === 1 &&
+    error.Message === 'A valid recipient address is required'
+  );
+}
+
 class CampaignMonitor {
   private api: AxiosInstance;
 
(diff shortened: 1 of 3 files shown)


scripts/alert_emails/send-test-email.ts
```diff
@@ -29,7 +29,11 @@ async function main(emailAddress: string) {
     await cm.sendClassicEmail(data);
     console.info(`Email sent to ${emailAddress}.`);
   } catch (err) {
-    console.error(err);
+    if (isInvalidEmailError(err)) {
+      console.log(`Invalid email: "${emailAddress}"`);
+    } else {
+      console.error(`Error sending email to "${emailAddress}"`, err);
```
mikelehen, line 35: Does this print something useful, or just like "[Object object]" ?

I'm worried that by using a class, we may lose visibility into the actual error details.  We may need to implement toString() or something.

 (I don't actually care about this test code, but I care about the similar error handling code in send_emails.ts)

Although it is verbose, it's nice that the existing log message included both our request and the error details.  Makes it straightforward to diagnose issues.

scripts/alert_emails/campaign-monitor.ts
```diff
@@ -40,6 +30,30 @@ export interface CampaignMonitorMessageSent {
...
+}
+
+function isRateLimitExceeded(status: number) {
+  return status === 429;
+}
+
+export function isInvalidEmailError(error: CampaignMonitorError) {
+  return (
+    error.Code === 1 &&
+    error.Message === 'A valid recipient address is required'
```
mikelehen, line 53: nit: This is existing code, but we should probably only check `Code` not `Message` since the latter probably isn't guaranteed to never change.

CM [documents](https://www.campaignmonitor.com/api/transactional/#send-classic-email) code 1 as "Invalid Email Address" so we don't need to check the message.

scripts/alert_emails/send-test-email.ts
pnavarrc: This is the output for a different error (I passed a bad token to check):

```
$ yarn send-test-email pnavarrc@gmail.com      
yarn run v1.22.4
$ ts-node-script -r tsconfig-paths/register scripts/alert_emails/send-test-email.ts pnavarrc@gmail.com
Error sending email to "pnavarrc@gmail.com" CampaignMonitorException {
  Code: 50,
  Message: 'Must supply a valid HTTP Basic Authorization header' }
```

I can return the original error (which includes the response) instead of the parsed response for unknown errors.

pnavarrc: Updated: It shows the entire error now, I tested with the test script after updating

PR #1161, opened 2020-07-25
alert email report

pnavarrc:
This PR implements a report for alert email subscriptions and overall engagement metrics for the alert emails. Basically, we want to track:

- Number of subscriptions by location (state and county)
- Overall engagement metrics for alert emails (open, click, and bounce rates) over time.

I'm using the development environment for Firebase - I will need to update the setup once I hook up the GitHub workflow. The plan to deploy is to write the workflow to be triggered manually first and then schedule it to run weekly or daily.

- [Alert Email Report - Dev](https://docs.google.com/spreadsheets/d/1M-GXVtFrv5NK1_Gceu8ynkK3OsHjD3GEQY4vaG0lDaA/edit#gid=0)
- [Alert Email Report - Prod](https://docs.google.com/spreadsheets/d/12JRyyAo6Thn4nCDf8nmdAU9CTVeJhjMRjMuK6vxfJbM/edit?usp=sharing)


pnavarrc:
[Analyze signup and CTR for alerting emails](https://trello.com/c/QsLxQWGq/307-analyze-signup-and-ctr-for-alerting-emails)

scripts/common/google-sheets.ts
```diff
@@ -0,0 +1,68 @@
...
+    await this.sheets.spreadsheets.values.append(request);
+  }
+}
+
+export default GoogleSheets;
+
+async function main() {
+  const keyFilePath = path.join(
+    __dirname,
+    '../alert_emails/google-service-account.json',
```
mikelehen, line 59: Huh.  Does this work?  This service account was meant for use with Firebase...  I have no idea if it is usable with google docs. 😁

package.json
```diff
@@ -72,7 +73,9 @@
     "generate-share-images": "ts-node-script -r tsconfig-paths/register scripts/generate_share_images/index.ts",
     "generate-daily-alerts": "ts-node-script -r tsconfig-paths/register scripts/alert_emails/generate_daily_alerts.ts",
     "send-emails": "ts-node-script -r tsconfig-paths/register scripts/alert_emails/send_emails.ts",
-    "send-test-email": "ts-node-script -r tsconfig-paths/register scripts/alert_emails/send-test-email.ts"
+    "send-test-email": "ts-node-script -r tsconfig-paths/register scripts/alert_emails/send-test-email.ts",
+    "report-alert-emails": "ts-node-script -r tsconfig-paths/register scripts/alert_emails_report/generate-report.ts",
+    "gs": "ts-node-script -r tsconfig-paths/register scripts/common/google-sheets.ts"
```
mikelehen, line 78: is this just temporary for testing?

scripts/alert_emails_report/generate-report.ts
```diff
@@ -0,0 +1,96 @@
...
+  __dirname,
+  '../alert_emails/google-service-account.json',
+);
+
+async function generateAlertEmailReport() {
+  const db = getFirestore();
+  const gsheets = new GoogleSheets(keyFile);
+  await gsheets.authenticate();
+
+  // TODO: Confirm, what is the best way to measure engagement stats?
```
mikelehen, line 29: Like I mentioned, maybe we want unique group IDs?

scripts/alert_emails_report/generate-report.ts
```diff
@@ -0,0 +1,96 @@
+import path from 'path';
+import _ from 'lodash';
+import moment from 'moment';
+import CampaignMonitor from '../alert_emails/campaign-monitor';
+import { getFirestore } from '../alert_emails/firestore';
+import { ALERT_EMAIL_GROUP, toISO8601 } from '../alert_emails/utils';
+import fipsByStateCode from '../what-the-fips/2018-census-fips-codes.json';
```
mikelehen, line 7: heh...  what-the-fips was a one-off hack I did a while ago.

Can we get this data from the https://github.com/covid-projections/covid-projections/blob/a35fd31c1d48e063b05af8a362c08ee637888054/src/components/MapSelectors/datasets/us_states_dataset_01_02_2020.json (perhaps adding a helper to https://github.com/covid-projections/covid-projections/blob/a35fd31c1d48e063b05af8a362c08ee637888054/src/common/locations.ts)?

scripts/alert_emails_report/generate-report.ts
```diff
@@ -0,0 +1,96 @@
...
+  const dateFrom = new Date('2020-07-24'); // initial send
+  const dateTo = moment(dateFrom).add(8, 'day').toDate();
+  const engagementStats = await fetchEngagementStats(dateFrom, dateTo);
+  await gsheets.appendRows(REPORT_SPREADSHEET_ID, 'data_engagement!A2:IE', [
+    [engagementStats],
+  ]);
+
+  const fipsStates = getAllStateFips();
+  const subscriptionStats = await Promise.all(
+    fipsStates.map(stateFips => fetchStateSubscriptionStats(db, stateFips)),
```
mikelehen, line 39: We would probably like to know about top counties as well...

Like I could imagine that the report has a tab with all states and their subscriber counts, and another tab with all counties (or top 100 if that's too many) with subscriber counts...

scripts/alert_emails_report/generate-report.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+async function fetchSubscriptionCountByFips(
+  db: FirebaseFirestore.Firestore,
+  fips: string,
+) {
+  const querySnapshot = await db
+    .collection('alerts-subscriptions')
+    .where('locations', 'array-contains', fips)
+    .get();
```
mikelehen, line 90: I think we end up running this query for every state/ county fips individually. Rather than do 3000 queries, it's probably more efficient to just do a single query for all subscriptions and then aggregate the per-fips counts in the client-side.

But if this seems plenty fast, it's probably fine too.

scripts/alert_emails_report/generate-report.ts
```diff
@@ -0,0 +1,96 @@
...
+    [engagementStats],
+  ]);
+
+  const fipsStates = getAllStateFips();
+  const subscriptionStats = await Promise.all(
+    fipsStates.map(stateFips => fetchStateSubscriptionStats(db, stateFips)),
+  );
+  await gsheets.appendRows(REPORT_SPREADSHEET_ID, 'data_locations!A2:F2', [
+    subscriptionStats,
+  ]);
```
mikelehen, line 43: Maybe log a link to the URL of the spreadsheet or something?

package.json
```diff
@@ -27,8 +27,9 @@
     "downshift": "^5.0.5",
     "esm": "^3.2.25",
     "firebase": "^7.15.4",
-    "googleapis": "^39.2.0",
+    "googleapis": "^57.0.0",
     "handlebars": "^4.7.6",
+    "install": "^0.13.0",
```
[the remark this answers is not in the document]
pnavarrc: oh, it's a leftover from a library that I tried for google sheets, will remove 👍

package.json
```diff
@@ -27,8 +27,9 @@
     "downshift": "^5.0.5",
     "esm": "^3.2.25",
     "firebase": "^7.15.4",
-    "googleapis": "^39.2.0",
```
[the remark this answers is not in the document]
pnavarrc: I looked it up - it's a dependency because our production system depends on this package to work [npm: Specifying dependencies and devDependencies in a package.json file](https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file).

The `build` script won't include this in the front-end bundle, so I think we are good. If you feel strongly I can move it to dev-dependency, but I don't think it does any harm.

package.json
pnavarrc: Yes, sorry, I forgot to remove it 😅 - it works though!

scripts/alert_emails_report/generate-report.ts
pnavarrc: 👌

scripts/common/google-sheets.ts
pnavarrc: It does - there are many options for authentication on the Google ecosystem, I had to authorize the same service account to be used for Google Sheets. 

I guess we could create a different one, but this is probably more convenient for development.

package.json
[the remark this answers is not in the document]
mikelehen: Don't feel strongly. 😁

scripts/common/google-sheets.ts
mikelehen: Ahh....  I see.

FYI- We have 3 different service accounts for firebase (dev, staging, and prod).  The email alerts workflow will use a different one depending on which environment you pass to the tools/send-alert-emails.sh script...  So we may need to authorize all 3 of them to be used with Google Sheets...

It feels like it might be slightly better to use a dedicated service account for Google Sheets, but I don't feel strongly.

pnavarrc: Ok, let's do that - I also want to create the "real" spreadsheet (we need to share that spreadsheet with the firebase emails for the corresponding environment). Maybe the spreadsheet ID should be an env variable as well?

scripts/alert_emails_report/generate-report.ts
pnavarrc: Campaign Monitor says that with ~100 groups it's OK, more than that might not work. Since we only send 1-2 alerts by week, we should have enough for a year or so. I can set a reminder to delete old groups and update the script so we can keep the old stats once we get closer to the limit.

scripts/alert_emails_report/generate-report.ts
pnavarrc: Updated!

scripts/alert_emails_report/generate-report.ts
pnavarrc: Changed it to use streams, the code is much cleaner too (IMHO)

scripts/alert_emails_report/generate-report.ts
pnavarrc: Added too - will update once we set up spreadsheets per environment.

scripts/alert_emails_report/README.md
```diff
@@ -0,0 +1,9 @@
+# Alert Email Report
+
+## Development Setup
+
+- Access to Firestore
+- google-service-account.json
+- CREATE_SEND_TOKEN
+- Access to the Spreadsheet
+- Share the spreadsheet with the email from the service account
```
pnavarrc, line 9: TODO: Improve details here, add links to the spreadsheets, etc.

scripts/alert_emails_report/generate-report.ts
```diff
@@ -0,0 +1,134 @@
...
+  const gsheets = new GoogleSheets(keyFile);
+  const range = 'data_engagement!A2:F';
+  const rows = await fetchEngagementStats(dateFrom, dateTo);
+  await gsheets.clearAndAppend(SPREADSHEET_ID, range, rows);
+}
+
+async function fetchEngagementStats(from: Date, to: Date): Promise<Cell[][]> {
+  const cm = new CampaignMonitor(process.env.CREATE_SEND_TOKEN);
+  const dateFrom = new Date('2020-07-23');
+  const dateTo = new Date();
```
mikelehen, line 106: Looks like you're passing in `from` and `to` but then ignoring them?

But also, would it make sense to drop the date range now that we are grouping emails by date?

scripts/alert_emails/campaign-monitor.ts
```diff
@@ -81,6 +94,26 @@ class CampaignMonitor {
 
     return response.data;
   }
+
+  async fetchTransactionalStats(
+    group: string,
+    from: Date,
+    to: Date,
```
mikelehen, line 101: Per my comment elsewhere I wonder if we don't need to filter by date anymore, in which case you can remove these or make them optional or something?

scripts/alert_emails/firestore.ts
```diff
@@ -2,6 +2,8 @@ import admin from 'firebase-admin';
 import fs from 'fs';
 import path from 'path';
 
+const ALERTS_SUBSCRIPTIONS = 'alerts-subscriptions';
```
mikelehen, line 5: Perhaps add _COLLECTION to the name?

scripts/alert_emails_report/generate-report.ts
```diff
@@ -0,0 +1,134 @@
...
+    await updateSubscriptionsByLocation();
+  } catch (err) {
+    console.error('Error updating subscription stats', err);
+  }
+
+  try {
+    await updateEngagementStats();
+  } catch (err) {
+    console.error('Error updating engagement stats', err);
+  }
```
mikelehen, line 43: This error handling feels a bit non-ideal.
1. main() is declared `async` which means it returns a `Promise`, but the calling code doesn't handle the Promise, which seems a bit misleading to me (not a huge deal though as long as we make sure that main()'s error handling is bulletproof).
2. I'm not really sure why you have 2 separate try/catch blocks.  Shouldn't the stack trace be enough to tell what failed?
3. Your catch block is _logging_ the error, but the script will still exit with error code 0 as if it was successful, which means if we run this script in a github workflow or whatever, the workflow will "succeed" even though there was an error, which means we probably won't notice it.

What I've done in other scripts is:

`main().catch(e => { console.error(e); process.exit(1); });`

or similar...  You don't have to do the same, but it would address all my feedback. 😁

scripts/alert_emails_report/generate-report.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  const subscriptions = await fetchAllAlertSubscriptions(db);
+  const groupedByFips = _.groupBy(subscriptions, fipsEmail => fipsEmail.fips);
+  const countByFips = _.map(groupedByFips, (items, fips) => {
+    return { fips, count: items.length };
+  });
+
+  const [stateCounts, countyCounts] = _.partition(countByFips, ({ fips }) =>
+    isStateFips(fips),
+  );
```
mikelehen, line 62: This code is very elegant! 👍

scripts/alert_emails_report/generate-report.ts
```diff
@@ -0,0 +1,134 @@
...
+  const countyStats = formatCountyStats(countyCounts);
+  const stateStats = formatStateStats(stateCounts);
+
+  const countyRange = 'subscriptions by county!A2:D';
```
mikelehen, line 66: Consider moving these to constants? 😁

scripts/alert_emails_report/generate-report.ts
```diff
@@ -0,0 +1,134 @@
...
+    getLocationNameForFips(fips),
+    count,
+  ]);
+  return _.sortBy(data, item => item[0]);
+}
+
+function formatCountyStats(stats: FipsCount[]): Cell[][] {
+  const data = stats.map(({ fips, count }) => {
+    const county = findCountyByFips(fips);
+    const fipsCode = `'${fips}`;
```
mikelehen, line 84: Hrm...  What is the `'` prefix for?  Maybe add a comment?

scripts/alert_emails_report/generate-report.ts
```diff
@@ -0,0 +1,134 @@
...
+    const countyName: string = county?.county
+      ? county?.county
+      : 'Unknown county';
```
mikelehen, line 87: nit: Can you do `county ? county.county : 'Unknown county'` or is there something funky with the typings preventing it?

You could also do `county?.county || 'Unknown county'` probably but I don't feel strongly.

scripts/alert_emails_report/generate-report.ts
```diff
@@ -0,0 +1,134 @@
...
+    const stateCode: string = county?.state_code ? county?.state_code : '-';
```
mikelehen, line 88: Same feedback here.

scripts/alert_emails_report/generate-report.ts
```diff
@@ -0,0 +1,134 @@
...
+    return [fipsCode, countyName, stateCode, count];
```
mikelehen, line 89: nit: It might be interesting to throw population in so we can easily spot counties that have more subscribers than expected based on population.

Same for states too, I guess.

scripts/alert_emails_report/generate-report.ts
```diff
@@ -0,0 +1,134 @@
...
+      );
+      return formatGroupStats(groupName, groupStats);
+    }),
+  );
+
+  return stats;
+}
+
+function formatGroupStats(groupName: string, stats: CampaignMonitorStats) {
+  return [groupName, stats.Sent, stats.Opened, stats.Clicked, stats.Bounces];
```
mikelehen, line 129: FWIW- I think converting opened and clicked to percentages might make sense, since they'll vary based on how many emails were sent.  So to compare, we need to "normalize" them.

I _think_ (but you might want to check) that in the CM UI they show:

Open rate = opened / sent
Click rate = clicked / opened

I would probably do the same here.

Alternatively (more flexible), maybe we can just set up the spreadsheet so that it calculates these automatically in the spreadsheet (as additional computed columns probably)?

src/common/locations.ts
```diff
@@ -1,6 +1,7 @@
 /** Helpers for dealing with the State / Counties dataset. */
 import US_STATE_DATASET from 'components/MapSelectors/datasets/us_states_dataset_01_02_2020.json';
 import { each, sortBy, takeRight } from 'lodash';
+import STATES from './us_states';
```
mikelehen, line 4: unused?

scripts/alert_emails/campaign-monitor.ts
pnavarrc: That makes sense, the default for `from` is 29 days ago, which should be good enough.

scripts/alert_emails_report/generate-report.ts
pnavarrc: 🎩

scripts/alert_emails_report/generate-report.ts
pnavarrc: Good point, adding a comment. It's to tell Google Sheets to interpret the value as text so it doesn't remove leading zeroes.

scripts/alert_emails_report/generate-report.ts
pnavarrc: much better 👍

scripts/alert_emails_report/generate-report.ts
pnavarrc: I think I'm going to set that up in Google Sheets - the formating is lost each time I write new data, so the plan is to have raw data sheets and then to use formulas to make the report nicer.

I checked, that's how CM calculates the rates.

PR #1177, opened 2020-07-29
Compare

chasulin:
- [x] lock background scroll behind modal

[Checklist of TODOs](https://www.dropbox.com/scl/fi/msz4wkf4sip97z5f1abcj/Compare-table-feedback.paper?dl=0&rlkey=7jiw0qvoncib6c4c1hwfygc37)


src/components/Compare/CompareMain.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const CompareMain = (props: {
+  stateId?: string;
+  stateName?: string;
+  county: any | null;
+  isModal?: Boolean;
+  locationsViewable?: number;
+  isLocationPage?: Boolean;
+  isHomepage?: Boolean;
+}) => {
```
pnavarrc, line 14: Consider making this a bit more generic instead of passing flags to customize the behavior between pages. For example, it seems that the only difference between the location and home pages is the list of locations to show. Maybe we could change it to something like:

```tsx
{/* Homepage */}
<CompareMain
  numLocations={5}
  locations={stateList}
  title='Compare states'
/>

{/* LocationPage */}
<CompareMain 
  currentLocation={okanoganCounty}
  numLocations={5}
  locations={countyList}
  title={isState ? 'Compare counties' : 'compare to other counties'}
/>
```

In the second case, we can pass a location item if we need it to highlight the current location.

src/components/Compare/CompareTable.tsx
```diff
@@ -0,0 +1,191 @@
...
+const CompareTable = (props: {
+  stateId?: string;
+  stateName?: string;
+  county?: any | null;
+  setShowModal: any;
+  isModal?: Boolean;
+  locationsViewable?: number;
+  isLocationPage?: Boolean;
+  isHomepage?: Boolean;
+}) => {
```
pnavarrc, line 44: I think it might be better to refactor this component to receive an array of locations with metrics and overall stats, that way we don't need to pass state, county, `isLocationPage`, etc. The parent component knows what is the current location (home page, state, or county page) so it has the context of which locations should be shown.

In this implementation, the parent passes some information `isLocationPage` and location information (`stateId?`, `stateName?`, `county?` and then the `CompareTable` component needs to calculate the items to show.

```tsx
<CompareTable
  locations={locationList}
  numLocations={5}
  title={`${locationName} State`}
/>
```

src/components/Compare/CompareTable.tsx
```diff
@@ -0,0 +1,191 @@
...
+    .map((stateInfo: Location) => {
+      if (stateInfo.state_code) {
+        return {
+          locationInfo: stateInfo,
+          metricsInfo: stateSummary(stateInfo.state_code),
+        };
+      } else {
+        fail('No state code');
+      }
+    });
```
pnavarrc, line 67: I think it would be better to pass this information as param for the component

src/components/Compare/CompareTable.tsx
```diff
@@ -0,0 +1,191 @@
...
+      if (
+        a.metricsInfo.metrics[sorter].value >
+        b.metricsInfo.metrics[sorter].value
+      ) {
+        return 1;
+      } else {
+        return -1;
+      }
+    }
+  });
```
pnavarrc, line 98: Consider using something simpler to sort, maybe use `lodash`'s `_.sortBy`. Example:

```tsx
const locations = [
  { name: 'AZ', rt: 1.3, contactTracing: 0.34 },
  { name: 'AL', rt: 1.8, contactTracing: 0.44 },
  { name: 'NY', rt: 1.6, contactTracing: 0.55 },
  { name: 'CA', rt: 1.4, contactTracing: 0.67 },
  // ...
]
const [sortingProp, setSortingProp] = useState('name');
const sorted = sortingProp === 'contactTracing' 
  ? _.sortBy(locations, 'contactTracing').reverse() 
  : _.sortBy(locations, sortingProp);
// render rows...
```

src/components/Compare/CompareTable.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  if (sortDescending) {
+    locationsArr.reverse();
+  }
+
+  const arrowColorSelected = props.isModal ? 'white' : 'black';
+  const arrowColorNotSelected = props.isModal
+    ? `${COLOR_MAP.GRAY_BODY_COPY}`
+    : `${COLOR_MAP.GRAY.BASE}`;
```
pnavarrc, line 107: Maybe this could be a theme? we can chat about that later

chasulin: added a TODO

src/components/Compare/CompareTable.tsx
chasulin: moved these functions to common/utils/compare

src/components/Compare/CompareTable.tsx
chasulin: Moved the location arrays out of CompareTable and into the top levels (Homepage + ChartsHolder), and removed stateId and isLocationPage props. Adding a TODO to refactor further and remove the need for the other page-specific props.

src/components/Compare/CompareMain.tsx
chasulin: made progress on this. adding TODO to refactor further and make the CompareMain/CompareTable fully generic

PR #1199, opened 2020-08-04
update the status text for case incidence

pnavarrc:
Update the copy of the status text for case incidence. The status text for case incidence (see [CAN Webapp Text](https://docs.google.com/document/d/1Jsj4zOksCJcgjQLUo3rwVAobEUkGLFukD1K71IqVNpo/edit#)) includes a link to an external source, so I updated the metric files to be `.tsx` and changed the functions that used to return the status text to return a `ReactElement` instead.

We have the same issue with the disclaimer text, some of them include links to sources, but we hacked it in the past by adding "special case" handling in the [`Disclaimer` component](https://github.com/covid-projections/covid-projections/blob/develop/src/components/Disclaimer/Disclaimer.tsx#L22).

## Proposal: Common interface + render functions

The indicators share a number of components that they need to implement to be complete. One idea to structure this is to define an `Indicator` interface with render functions for each one of these components, this way we have
- **Type safety**: We know that each instance of an `Indicator` has a render function implementation for each component
- **Separation of concerns**: The logic to render the components for each indicator is encapsulated in the indicator module/file.
- **Flexibility**: While most likely the components will share parts of their implementation, they can have completely different implementations.

## Example

The `Indicator` interface would hold render functions for each metric and potentially other common methods that might be useful when rendering all the metrics.

```ts
interface Indicator {
  formatValue: (currentValue: number) => string;
  renderChart: (projection: Projection) => ReactElement;
  renderStatus: (projection: Projection) => ReactElement;
  // renderDisclaimer, renderStats, etc...
}
```

On implementations of the Indicator, we need to implement each render method, likely using common components, but not necessarily:

```tsx
// metrics/contact_tracing.tsx
const contactTracing: Indicator = {
  renderChart(projection: Projection) {
    return (
      <ChartZones
        height={height}
        columnData={columnData}
        capY={CAP_Y}
        zones={CONTACT_TRACING_LEVEL_INFO_MAP}
        getTooltipBody={getTooltipBody}
        getPointText={getPointText}
      />
    );
  }
  renderStatus(projection: Projection) {
    // more code...
    return (
      <Fragment>
        {`Per best available data, ${location} has ${numTracers} contact...`}
      </Fragment>
    );
  }
}
```

Then, we could iterate through the indicators

```tsx
// LocationPage
const indicators: Indicator[] = [
  caseIncidence,  
  growthRate,
  positiveTestRate
  icuHeadroom,
  contactTracing,
];

// ...later
indicators.map((indicator:Indicator) => {
  <Fragment>  
    <h1>{indicator.chartTitle()}</h1>
    <ChartDescription>
      {indicator.renderStatusText(projection)}
    </ChartDescription>
    <ChartContainer>
      {indicator.renderChart(projection)}
    </ChartContainer>
    <Disclaimer>
      {indicator.renderDisclaimer()}
    </Disclaimer>
  </Fragment>
});
```

I'm not 100% sure that this is the best way to organize the indicators so I didn't implement it, but I did rewrite the code that returns the status text as a render function to keep the metrics uniform. Thoughts?


pnavarrc:
[Implement incidence copy updates](https://trello.com/c/4NEDiDY1/347-implement-incidence-copy-updates)

src/common/metrics/case_density.tsx
```diff
@@ -0,0 +1,101 @@
...
+    totalPopulation,
+  } = projection;
+  if (
+    currentCaseDensity === null ||
+    currentDailyDeaths === null ||
+    totalPopulation === null
+  ) {
+    return (
+      <React.Fragment>
+        {`Not enough case data is available to generate ${METRIC_NAME.toLowerCase()}.`}
```
mikelehen, line 65: Are you intentionally using JavaScript string interpolation instead of JSX expression syntax?

This could just be:

```
<React.Fragment>
Not enough case data is available to generate {METRIC_NAME.toLowerCase()}.
</React.Fragment>
```

right?

Same for all the other fragments you added...  If you're just trying to minimize churn for now, that's fine. It just looks a bit odd.

pnavarrc: 😅 I think I momentarily forgot that this syntax exists, updating!

src/common/metrics/case_density.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  const levelFactor = {
+    [Level.LOW]: 1,
+    [Level.MEDIUM]: 10,
+    [Level.HIGH]: 20,
+    [Level.CRITICAL]: 100,
+    [Level.UNKNOWN]: 0,
+  };
+
+  const casesPerYear = currentCaseDensity * 356 * levelFactor[level];
```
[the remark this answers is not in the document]
pnavarrc: Makes sense!

src/common/metrics/case_density.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const casesPerYearText = formatInteger(casesPerYear);
+  const estimatedCasesPerYear = casesPerYear * 5;
+  const estimatedCasesPerYearText = formatInteger(estimatedCasesPerYear);
+  const percentOfPopulation = Math.min(
+    1,
+    estimatedCasesPerYear / totalPopulation,
```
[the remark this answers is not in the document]
pnavarrc: Yes, this is wrong (but I think your proposal is also wrong). Can you review the calculations here?

```ts
const ESTIMATE_FACTOR = 5;
const dailyCasesPer100k = currentCaseDensity;
const yearlyCasesPer100k = dailyCasesPer100k * 365;
const yearlyEstimatedInfectionsPer100k = ESTIMATE_FACTOR * yearlyCasesPer100k;
const estimatedPercentOfPopulationInfectedYear =
    (100000 * yearlyEstimatedInfectionsPer100k) / totalPopulation;
```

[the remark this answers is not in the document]
pnavarrc: Actually, I think it's more simple to do:

```ts
  const ESTIMATE_FACTOR = 5;
  const casesPerDay = currentDailyAverageCases;
  const casesPerYear = 365 * casesPerDay;
  const estimatedInfectionsPerYear = ESTIMATE_FACTOR * casesPerYear;
  const estimatedPercentageInfectedPerYear = Math.min(
    1,
    estimatedInfectionsPerYear / totalPopulation,
  );
```

Plus, only the first phrase refers to "per 100k", the rest is about total cases and infections in a year.

> Over the last week, Washington has averaged 10.8 new confirmed cases per day for every 100,000 residents. Over the next year, this translates to 299,248 cases and an estimated 1,496,239 infections (19.6% of the population).

src/common/metric.tsx
```diff
@@ -95,17 +95,26 @@ export function getMetricDisclaimer(metric: Metric) {
...
+  [Metric.HOSPITAL_USAGE]: Hospitalizations.metricICUHeadroom,
+  [Metric.CONTACT_TRACING]: ContactTracing.metricContactTracing,
+  [Metric.FUTURE_PROJECTIONS]: FutureProjections.metricFutureProjection,
+  [Metric.CASE_DENSITY]: CaseDensity.metricCaseIncidence,
+};
+
+export function getMetricStatusText(metric: Metric, projections: Projections) {
+  if (!(metric in metricDefinitions)) {
+    fail('unknown metric');
+  }
```
pnavarrc, line 116: Do we even need this check? 🤔

src/common/metric.tsx
```diff
@@ -95,17 +95,20 @@ export function getMetricDisclaimer(metric: Metric) {
...
-      projection,
-    ),
-    [Metric.CONTACT_TRACING]: ContactTracing.contactTracingStatusText(
-      projection,
-    ),
-    [Metric.CASE_DENSITY]: CaseDensity.caseDensityStatusText(projection),
-  };
-  return METRIC_TO_STATUS_TEXT[metric];
+const metricDefinitions: { [metric in Metric]: MetricDefinition } = {
+  [Metric.CASE_GROWTH_RATE]: CaseGrowth.metricCaseGrowth,
```
pnavarrc, line 99: Eventually the metric definition will be the default export

src/components/LocationPage/ChartBlock.tsx
```diff
@@ -115,7 +114,7 @@ function ChartBlock(props: {
           ) : (
             <Fragment>
               <ChartDescription>
-                {generateChartDescription(projection, noInterventionProjection)}
+                {metricFutureProjection.renderStatus(props.projections)}
```
pnavarrc, line 117: We should be able to get rid of the special handling for future projections here, but I prefer to do that on a separate PR to keep the diff manageable.

mikelehen: 👍  to separate PRs! 😁

src/common/metric.tsx
pnavarrc: _actually_, Future Projections is not a core indicator, it doesn't share all the UI elements that the indicators have, so I might have to keep that distinction here.

src/common/metrics/case_density.tsx
```diff
@@ -0,0 +1,98 @@
+import React, { Fragment } from 'react';
+import { COLOR_MAP } from 'common/colors';
+import { Level, LevelInfoMap } from 'common/level';
+import { Projections } from 'common/models/Projections';
+import { formatDecimal, formatPercent, formatInteger } from 'common/utils';
+import ExternalLink from 'components/ExternalLink';
+import { MetricDefinition } from './interfaces';
+
+export const metricCaseIncidence: MetricDefinition = {
```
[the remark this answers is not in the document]
pnavarrc: I don't mind much one way or the other, let's go with your recommendation 👍

src/common/utils/index.ts
```diff
@@ -83,3 +83,20 @@ export const formatDecimal = (num: number, places = 2): string =>
  */
 export const formatPercent = (num: number, places = 0): string =>
   `${formatDecimal(100 * Math.min(1, num), places)}%`;
+
+/**
+ * Returns a formatted number, rounding to keep the specified number
+ * of significant digits. Example:
```
mikelehen, line 89: Nice!

src/common/utils/index.ts
```diff
@@ -83,3 +83,20 @@ export const formatDecimal = (num: number, places = 2): string =>
...
+ *   formatEstimate(123456, 3)     // 123,000
+ */
+export function formatEstimate(value: number, significantDigits = 2): string {
+  if (value <= 0) {
+    return formatInteger(value);
+  }
+  const numDigits = Math.floor(Math.log10(value));
+  const precision = Math.max(0, numDigits - significantDigits + 1);
+  const rounded = _.round(value, -1 * precision);
+  return formatInteger(rounded);
```
mikelehen, line 101: Did you consider something like https://stackoverflow.com/a/36369703 ?

pnavarrc: 😱  I didn't know about `toPrecision`, this is much better, updating!

PR #1231, opened 2020-08-10
Compare disclaimer copy

chasulin:
- Generates state-specific disclaimer copy

[Mocks](https://www.figma.com/file/CypfjFVRBhsC3vArhoIPq1/County-tables?node-id=318%3A18515)

 Todo:
- [x] Confirm copy

### DIFF
--- a/src/common/utils/compare.ts
+++ b/src/common/utils/compare.ts
@@ -1,6 +1,8 @@
 import { getLocationNames, Location } from 'common/locations';
 import { stateSummary, countySummary } from 'common/location_summaries';
 import { LocationSummary } from 'common/location_summaries';
+import { Metric } from 'common/metric';
+import { Level } from 'common/level';
 
 export interface SummaryForCompare {
   locationInfo: Location;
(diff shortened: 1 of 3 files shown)


src/common/utils/compare.ts
```diff
@@ -34,3 +36,29 @@ export function getCountiesArr(stateId: string) {
...
+      location.metricsInfo.metrics[Metric.CONTACT_TRACING].level !==
+      Level.UNKNOWN,
+  );
+  const positiveTestData = countiesArr.filter(
+    (location: any) =>
+      location.metricsInfo.metrics[Metric.POSITIVE_TESTS].level !==
+      Level.UNKNOWN,
+  );
+  if (contactTracingData.length && positiveTestData.length) {
+    return 'Most states only report contact tracing at the state level.';
```
pnavarrc, line 55: Do we need the disclaimer when we do have data for contact tracing and positive tests?

chasulin: my thought was: i think there's currently only a handful of counties with contact tracing data (maybe just 1 county?), so if there _is_ a table with contact tracing, it will be one county out of many, in which case there should probably still be a disclaimer re: why the column is nearly empty

chasulin: but yeah, can nix it all together if that feels like it makes more sense

pnavarrc: but if all the counties have both contact tracing and positive test data, then this condition would be `true`,  I'm not sure if we need the disclaimer in that case.

chasulin: updates: new mocks nix dynamic copy, so this is all going away..

src/components/Compare/Compare.style.tsx
```diff
@@ -223,13 +214,36 @@ export const ArrowContainer = styled.div<{
   }
 `;
 
-export const Footer = styled.div`
+export const Footer = styled.div<{ isCounty: any }>`
```
pnavarrc, line 217: do you want to make `isCounty: boolean`?

chasulin: its not a boolean-- isCounty={props.county}, which is either a location object or null

src/components/Compare/Compare.style.tsx
```diff
@@ -240,6 +254,29 @@ export const ViewAllLink = styled.div`
...
+  margin: 1rem auto 0;
+
+  span {
+    color: rgba(0, 0, 0, 0.7);
+  }
+
+  a {
+    color: ${COLOR_MAP.BLUE};
+    text-decoration: none;
+  }
```
pnavarrc, line 272: The blue color should work automatically (it's a global style), but you can keep it if it doesn't

chasulin: hm yeah it was being overwritten by another style. gonna keep

PR #1234, opened 2020-08-10
show current location in ranked position

pnavarrc:
This PR updates the compare table to make sure that the row for the current location (for counties) always shows in the inline version of the table in a position relative to its ranking. 

**Top**
The Karnes County is at the top given the current sorting of the table, so it shows at the top.
image

**Middle**
Cameron County is in the first 6 locations, so it shows in the corresponding location.
image

**Bottom**
Jim Wells County should be below the fold, so it shows at the end with its corresponding ranking.
image

### DIFF
--- a/src/components/Compare/LocationTable.style.tsx
+++ b/src/components/Compare/LocationTable.style.tsx
@@ -4,6 +4,7 @@ import { Cell, locationNameCellWidth, metricCellWidth } from './Compare.style';
 import { COLORS } from 'common';
 
 const minTableWidth = locationNameCellWidth + 5 * metricCellWidth;
+const pinnedBorderColor = '#CEBFAC';
 
 export const ModalContainer = styled.div`
   display: flex;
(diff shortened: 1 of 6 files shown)


src/components/Compare/LocationTable.style.tsx
```diff
@@ -30,10 +31,18 @@ export const Container = styled.div`
   overflow-x: auto;
 `;
 
-export const Head = styled.div`
+const PinnedTop = `
+  box-shadow: 0 4px 20px rgba(0, 0, 0, 0.12);
+  z-index: 0;
+  position: relative;
+  border-bottom: 1px solid ${pinnedBorderColor};
```
chasulin, line 38: i _think_ this is meant to be 2px

pnavarrc: Josh asked me to change it to 1px :)

joshziman:
A minor thing that we missed in the Jim Wells Co. case
https://covidactnow.org/us/tx/county/jim_wells_county?s=846164

If a county in a rank > #6, then the text underneath the table "Displaying 1-6 of 254" isn't accurate.

Instead, I suggest simplifying that always to "Displaying 6 of {total counties}".

PR #1241, opened 2020-08-11
Remove "New key indicator added" update notification.

mikelehen:
https://trello.com/c/MiXZaYyU/359-remove-the-new-cases-has-been-added-prompt-on-location-pages

### DIFF
--- a/src/components/LocationPage/LocationPageHeader.tsx
+++ b/src/components/LocationPage/LocationPageHeader.tsx
@@ -1,4 +1,4 @@
-import React, { Fragment, FunctionComponent } from 'react';
+import React, { Fragment } from 'react';
 import CheckIcon from '@material-ui/icons/Check';
 import {
   ColoredHeaderBanner,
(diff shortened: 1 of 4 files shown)


src/components/LocationPage/LocationPageHeader.tsx
```diff
@@ -198,13 +167,14 @@ const LocationPageHeader = (props: {
                 <Copy>{levelInfo.detail(locationName)}</Copy>
               </SectionColumn>
             </SectionHalf>
+            {/* TODO: Re-add this if we have new updates in the future.
```
mikelehen, line 170: I left this commented-out in case it's handy again in the future but I'm also happy to just rip it out and we can dig it out of git if/when we need it.

chasulin: hmm my vote is for delete and we can dig it out if/when needed

mikelehen:
Oh, yes!  Thank you for remembering!

PR #1257, opened 2020-08-13
metric definition: add metric names to metric definitions

pnavarrc:
Continuing to centralize the handling of metrics, this PR adds metric names for each metric to the interface. There are 3 versions of the names (not ideal) - but it makes sense since the daily cases metric has 3 versions on the site (compare table, chart title, and the header indicator).


pnavarrc:
Nice, thanks!

PR #1266, opened 2020-08-13
explore: use the real datasets

pnavarrc:
It looks like our rolling average function "translates" the data forward by a week, the effect is more noticeable with hospitalization data.

image

image

@mikelehen @chasulin can you double-check that I'm using the correct metrics in each case? I need to adjust the smoothed line to not go to zero before and after having data, but I will do that on a separate PR

### DIFF
--- a/src/common/models/Projection.ts
+++ b/src/common/models/Projection.ts
@@ -65,7 +65,15 @@ export type DatasetId =
   | 'contractTracers'
   | 'caseDensityByCases'
   | 'caseDensityByDeaths'
-  | 'caseDensityRange';
+  | 'caseDensityRange'
+  | 'smoothedDailyCases'
+  | 'smoothedDailyDeaths'
+  | 'rawDailyCases'
+  | 'rawDailyDeaths'
+  | 'rawHospitalizations'
+  | 'smoothedHospitalizations'
+  | 'rawICUHospitalizations'
+  | 'smoothedICUHospitalizations';
 
 export interface RtRange {
   /** The actual Rt value. */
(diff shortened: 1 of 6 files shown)


src/common/models/Projection.ts
```diff
@@ -201,21 +216,38 @@ export class Projection {
...
+      this.fillLeadingNullsWithZeros(
+        actualTimeseries.map(row => row && row.cumulativeDeaths),
+      ),
+    );
+    this.smoothedDailyDeaths = this.smoothWithRollingAverage(
+      this.rawDailyDeaths,
+    );
+
+    this.rawHospitalizations = actualTimeseries.map(
+      row => row && row?.hospitalBeds?.currentUsageCovid,
```
mikelehen, line 236: Annoyingly the `?.` operator defaults to `undefined` instead of null which doesn't match our type definition (`Array<number | null>`) and may break some of our assumptions in other parts of the code.  I _think_ you can do `row && row.hospitalBeds.currentUsageCovid` instead.

src/common/models/Projection.ts
```diff
@@ -201,21 +216,38 @@ export class Projection {
...
+    );
+    this.smoothedHospitalizations = this.smoothWithRollingAverage(
+      this.rawHospitalizations,
+    );
+    this.rawICUHospitalizations = actualTimeseries.map(
+      row => row && row?.ICUBeds?.currentUsageCovid,
```
mikelehen, line 242: same feedback here

pnavarrc:
you were right, it was a bug in the charting side - fixed, it looks really nice now!

PR #1268, opened 2020-08-13
explore: format x-axis labels and tick positions

pnavarrc:
Small fix to adjust tick positions to what we have in [the mocks](https://www.figma.com/file/nkoC6P02eXxkB95ukdU6Ms/Explore?node-id=260%3A24)

image

### DIFF
--- a/src/components/Explore/ExploreChart.tsx
+++ b/src/components/Explore/ExploreChart.tsx
@@ -1,4 +1,6 @@
 import React, { FunctionComponent } from 'react';
+import useMediaQuery from '@material-ui/core/useMediaQuery';
+import { useTheme } from '@material-ui/core/styles';
 import moment from 'moment';
 import { scaleTime, scaleLinear, scaleBand } from '@vx/scale';
 import { GridRows, GridColumns } from '@vx/grid';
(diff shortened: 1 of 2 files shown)


src/components/Explore/ExploreChart.tsx
```diff
@@ -79,17 +80,20 @@ const ExploreChart: FunctionComponent<{
...
       <Group key="axes" top={marginTop} left={marginLeft}>
         <Styles.GridLines>
-          <Grid
-            xScale={timeScale}
-            yScale={yScale}
-            width={innerWidth}
-            height={innerHeight}
-          />
+          <GridColumns<Date> scale={timeScale} height={innerHeight} />
+          <GridRows<number> scale={yScale} width={innerWidth} />
```
pnavarrc, line 84: `@vx/grid` has a bug that doesn't allow to pass specific values to the scale. It's picking the correct values somehow (I _think_ that the axis component mutates the scale ticks, and then the grid takes it from the scale). I'm going to dig into this.

pnavarrc: (later)

mikelehen:
Though I'm happy to defer to Josh, you or others if you think it's fine. 😁

pnavarrc:
The mobile version will have specific controls to manage that, however, it wouldn't look good in the "All" setting. I will update the format to make it mobile friendly and update the PR 👍 

< img src="https://user-images.githubusercontent.com/114084/90179545-72c0e880-dd62-11ea-8bc2-5bc075d81a20.png" width="375" />

pnavarrc:
Updated!

PR #1272, opened 2020-08-15
Remove legacy intervention-related code.

mikelehen:
NOTE: This is low-pri.  No rush. 

With the removal of future projections and the outcomes table, we no longer
need any intervention-related code.

We now only load the OBSERVED_INTERVENTION api endpoint, not NO_INTERVENTION.
This makes the location page, compare tool, update-location-summaries, etc. faster.

### DIFF
--- a/DEV_README.md
+++ b/DEV_README.md
@@ -70,5 +70,5 @@ export GITHUB_TOKEN=<YOUR PERSONAL GITHUB TOKEN>
 Alternatively, if you don't care about regenerating the sharing images (which takes a long time), you can update the snapshot and map colors with:
 
 * Update src/assets/data/data_url.json to point at the new snapshot.
-* Run `yarn update-calculated-interventions`
+* Run `yarn update-location-summaries`
 * Check in the result.
--- a/README.md
+++ b/README.md
@@ -1,18 +1,10 @@
 # Covid Act Now
 
-This is the code repository for https://covidactnow.org/ - a tool providing public leaders and health officials data to help answer the question: "how can the US reopen safely?"
-
-## How to use this tool
-This tool is built to enable political leaders to make decisions in their COVID-19 response informed by best available data and modeling.
-
-### Here are the questions we built this tool to answer:
-* What will the impact be in my region be and when can I expect it?
-* How long until my hospital system is under severe pressure?
-* What are my menu of interventions, and how will they address the spread of COVID-19?
+This is the code repository for https://covidactnow.org/
 
 ## Comments, questions, or want to get involved?
 If you have time to give us feedback, have questions, or otherwise want to get involved, please email us info@covidactnow.org
 
-## Already involved? 
-Check out the frontend 
+## Already involved?
+Check out the frontend
 [onboarding document](https://docs.google.com/document/d/1w5wlqynTOf8fFxWP_Cly3M8o8hrhkC6yQ43XpjzgOJY/edit#) for further information on how to get started
--- a/src/common/models/Projections.ts
+++ b/src/common/models/Projections.ts
@@ -3,7 +3,7 @@ import { STATES } from '..';
 import { Metric, getLevel, ALL_METRICS } from 'common/metric';
 import { Level } from 'common/level';
 import { LEVEL_COLOR } from 'common/colors';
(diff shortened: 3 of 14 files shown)


DEV_README.md
```diff
@@ -70,5 +70,5 @@ export GITHUB_TOKEN=<YOUR PERSONAL GITHUB TOKEN>
 Alternatively, if you don't care about regenerating the sharing images (which takes a long time), you can update the snapshot and map colors with:
 
 * Update src/assets/data/data_url.json to point at the new snapshot.
-* Run `yarn update-calculated-interventions`
+* Run `yarn update-location-summaries`
```
mikelehen, line 73: This was just out of date.

README.md
```diff
@@ -1,18 +1,10 @@
...
-This is the code repository for https://covidactnow.org/ - a tool providing public leaders and health officials data to help answer the question: "how can the US reopen safely?"
-
-## How to use this tool
-This tool is built to enable political leaders to make decisions in their COVID-19 response informed by best available data and modeling.
-
-### Here are the questions we built this tool to answer:
-* What will the impact be in my region be and when can I expect it?
-* How long until my hospital system is under severe pressure?
-* What are my menu of interventions, and how will they address the spread of COVID-19?
+This is the code repository for https://covidactnow.org/
```
mikelehen, line 3: This was out-of-date too, so I just gutted it.

src/common/models/Projections.ts
```diff
@@ -198,70 +193,8 @@ export class Projections {
     }
   }
 
-  // TODO(pablo): Remove this method once the prevalence metric is no longer
-  // an 'update'. This is only used to show the level before and after the
-  // introduction of the prevalence metric for locations where the overall
-  // level changed.
```
mikelehen, line 204: I forgot to remove this before. Oops!

src/common/utils/makeChartShareQuote.js
```diff
@@ -32,12 +32,7 @@ export default function makeChartShareQuote(
       0,
     )} of COVID cases, according to @CovidActNow. See the chart: `;
   } else if (chartIdentifier === 4) {
-    if (projections && projections.baseline.dateOverwhelmed) {
-      return `If all restrictions were completely lifted today, ${displayName}'s hospitals would overload on ${formatUtcDate(
-        projections.baseline.dateOverwhelmed,
-      )}, according to @CovidActNow. See the chart: `;
-    }
-    return `Assuming current trends and interventions continue, ${displayName}'s hospitals are unlikely to become overloaded in the next 3 months, according to @CovidActNow. See the chart: `;
+    fail('Future projections chart was deprecated.');
```
mikelehen, line 35: I could have cleaned this up when I moved the future projections chart but missed it.

src/common/utils/model.ts
```diff
@@ -23,10 +25,14 @@ export function fetchProjections(
...
-    const summaryWithTimeseriesMap = await new Api(
+    const summaryWithTimeseries = await new Api(
       snapshotUrl,
-    ).fetchSummaryWithTimeseriesMap(region);
-    return new Projections(summaryWithTimeseriesMap, stateId, countyInfo);
+    ).fetchSummaryWithTimeseries(region);
+    assert(
+      summaryWithTimeseries != null,
+      'Failed to fetch projections for ' + region,
+    );
```
mikelehen, line 34: Note: This assert is kinda' new.  We used to allow for missing API projections because we only produced them for locations where we had future projections, but when we added case incidence, the backend started producing API projections for all locations (even if we don't have any data / metrics).  So now we consider it an invariant that the API should have data for every state / county and we assert otherwise.

src/screens/Embed/Embed.js
```diff
@@ -61,11 +61,6 @@ function LocationEmbed() {
     return null;
   }
 
-  const primary = projections.primary;
-  if (!primary) {
-    return <span>'No data available for county.'</span>;
-  }
-
```
mikelehen, line 68: Per my comment on the assert in model.ts, it's no longer possible to hit this case.  The API exposes data for all states/counties, even if there's no actual non-null data in it.

mikelehen:
This is low-pri.  No rush.

PR #1279, opened 2020-08-17
explore: legend component

pnavarrc:
This PR adds the legend component for Explore. I added the legend by type with the series because they are related elements (the style of the series should match the style of the legend items), not sure if that's the best approach.

image


pnavarrc:
This is just for the legend component itself, the div that contains needs to have more controls (select linear/log scale, other locations, etc) - I will do a final UX/UI review once we have the core functionality, thanks @mikelehen !

PR #1300, opened 2020-08-19
Enhance Daily New Cases Tooltip

apeterson7:
This PR enhances the tooltip on the "Daily new cases per 100k population" chart.
The tooltip now includes 7 day average new cases for a given day. 

(Addresses https://trello.com/c/O9nRsq19/358-implement-new-new-cases-tool-tip.)
Screen Shot 2020-08-19 at 7 01 47 PM

### DIFF
--- a/src/common/models/Projection.ts
+++ b/src/common/models/Projection.ts
@@ -77,6 +77,7 @@ export interface RtRange {
 
 export interface CaseDensityRange {
   caseDensity: number;
+  newCases: number | null;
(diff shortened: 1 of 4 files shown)


apeterson7:
@mikelehen Just committed some changes to clean up the styling according to your feedback. 

Also, I'm wondering whether it might be good to let the user know that the number of new cases is based on 7 day average?   I don't see that noted near the chart but I might have missed it.

PR #1304, opened 2020-08-20
Compare: fips-to-msa list

chasulin:
Implements script to parse [county fips/MSA list from the DOL](https://www.dol.gov/owcp/regs/feeschedule/fee/Effective_May_16_2004_County_and_State_FIPS.htm)

- [x] remove generate_fips_to_msa.ts script + fips_to_msa.json + fips_to_msa.txt once combined script is good to go


scripts/county_adjacency_msa/index.ts
```diff
@@ -65,6 +85,10 @@ function parseAdjacencyFile(filePath: string, delimiter = TAB): CountyItem[] {
   }));
 }
 
+function readDelimitedFile(text: string, delimiter = TAB) {
```
pnavarrc, line 88: Consider changing it so it receives the file path (so it actually reads the file instead of parsing the text)

```ts
function readDelimitedFile(filePath:string, delimiter=TAB) {
  // ...
}
```

scripts/county_adjacency_msa/index.ts
```diff
@@ -54,9 +62,21 @@ interface CountyItem {
...
+  const msaFilePath = path.join(__dirname, 'fips_msa.txt');
+  const text = fs.readFileSync(msaFilePath, 'utf-8');
+  const countyGroups = readDelimitedFile(text);
+  const fipsToMsaMap = _.fromPairs(countyGroups.map(getFipsMsa));
+  return fipsToMsaMap;
+}
+
+function getFipsMsa(countyArr: any): any {
+  return [countyArr[0], countyArr[1]];
+}
```
pnavarrc, line 75: Consider adding the correct types here

```ts
function getFipsMsa(row: string[]):[string, string] {
  return [row[0], row[1]];
}
```

scripts/county_adjacency_msa/index.ts
```diff
@@ -84,6 +108,16 @@ function parseCountyData(data: CountyItem[]): AdjacencyMap {
...
   }
+
+  const fipsToMsa = parseFipsToMsa();
+  const fipsInMsaList = Object.keys(fipsToMsa);
+
+  fipsInMsaList.forEach(fips => {
+    if (adjacencyMap[fips]) {
+      adjacencyMap[fips].msa_code = fipsToMsa[fips];
+    }
+  });
```
pnavarrc, line 119: Note that if the `msa_code` is `0` it will still show up in the data, maybe you want to remove it? That way you won't need to handle `"0"` later in the front-end code.

```json
{
    "12107": {
      "adjacent_counties": [
        "12001",
        "12007",
        "12019",
        "12035",
        "12083",
        "12109",
        "12127"
      ],
      "msa_code": "0"
    },
}
```

scripts/county_adjacency_msa/index.ts
```diff
@@ -92,7 +126,7 @@ function main() {
   const data = parseAdjacencyFile(filePath);
   const adjacencyMap = parseCountyData(data);
 
-  const outputPath = path.join(__dirname, 'county_adjacency.json');
+  const outputPath = path.join(__dirname, 'county_adjacency_msa.json');
```
pnavarrc, line 129: Maybe it's a better idea to set the output path to be the final location of the file, that way we don't need to copy it over manually.

scripts/county_adjacency_msa/index.ts
```diff
@@ -84,15 +108,35 @@ function parseCountyData(data: CountyItem[]): AdjacencyMap {
...
   const adjacencyMap = parseCountyData(data);
-
-  const outputPath = path.join(__dirname, 'county_adjacency.json');
+  const outputPath = path.join(
+    __dirname,
+    '..',
+    '..',
+    'src',
+    'common',
+    'data',
```
pnavarrc, line 137: nit: you might want to put the segments together

```ts
const outputPath = path.join(
   __dirname,
  '../../src/common/data',
  'county_adjacency_msa.json'
);
```

PR #1318, opened 2020-08-22
Fix Escape Key Bug

apeterson7:
This PR adds a stateReducer function to the Downshift element that captures the escape key event and overrides the default behavior and closes the menu.  The default behavior is to pass null the onChange callback, which was causing the error.

### DIFF
--- a/src/components/MapSelectors/GlobalSelector.js
+++ b/src/components/MapSelectors/GlobalSelector.js
@@ -235,9 +235,11 @@ const GlobalSelector = ({ handleChange, extendRight }) => {
     }
   };
 
+  /*
+   * Downshift uses this state reducer function to close the menu when the escape key is pressed
+   * and executes default behavior for all other events.
+   */
   const escapeKeyStateReducer = (state, changes) => {
-    console.log(changes);
-
     switch (changes.type) {
       case Downshift.stateChangeTypes.keyDownEscape:
         return {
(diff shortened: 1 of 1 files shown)


src/components/MapSelectors/GlobalSelector.js
```diff
@@ -235,10 +235,24 @@ const GlobalSelector = ({ handleChange, extendRight }) => {
     }
   };
 
+  const escapeKeyStateReducer = (state, changes) => {
```
mikelehen, line 238: Can you add a comment explaining what's going on here?

src/components/MapSelectors/GlobalSelector.js
```diff
@@ -235,10 +235,24 @@ const GlobalSelector = ({ handleChange, extendRight }) => {
...
+    console.log(changes);
```
mikelehen, line 239: Perhaps remove this?  Or else add some surrounding text so somebody seeing it in the console knows what it is.

apeterson7: I meant to remove this, thanks!

src/components/MapSelectors/GlobalSelector.js
apeterson7: Will do!

mikelehen:
Thanks for this!

PR #1326, opened 2020-08-24
Show 'RISK INCREASED' / 'RISK DECREASED' as appropriate in email alerts instead of just `NEW RISK SCORE'.

mikelehen:
Before:
image

After:
image


scripts/alert_emails/template.html
```diff
@@ -775,7 +775,7 @@
                               normal; line-height: 12px;Margin-bottom: 4px;" align="center">
                                             <p style="Margin-top:4px;Margin-bottom:4px;font-family:Arial, Helvetica, sans-serif;
                                   text-align:center;font-size:14px;text-transform:uppercase;font-weight:bold;">
-                                                new risk score
+                                                {{change}}
```
mikelehen, line 778: We were already populating this template variable...  just not using it.

https://github.com/covid-projections/covid-projections/blob/d2e05dd7a9209f0e64a7bf9c483b463db3ad46cb/scripts/alert_emails/utils.ts#L46

PR #1328, opened 2020-08-25
Re enable harris county icu

ghop02:
Now that we're using the new icu model (and that these counties aren't in critical icu land anymore) figured we should remove the override.

would be nice to regenerate screenshots, but don't think its that urgent and can just be regenerated tomorrow


ghop02:
[Harris county is red due to disabled icu metric](https://trello.com/c/JrPTmYyh/403-harris-county-is-red-due-to-disabled-icu-metric)

src/common/models/ICUHeadroom.ts
```diff
@@ -14,10 +14,7 @@ import {
 // We sometimes need to override the ICU metric for locations due to bad data, etc.
 // TODO(https://trello.com/c/CPcYKmdo/): Review once we have better estimates for
```
mikelehen, line 15: Er, actually.  Can you remove this comment and close the trello ticket?

PR #1329, opened 2020-08-25
add links to state and county maps

pnavarrc:
This PR changes how we do navigation when clicking on locations on the map. In production, we pushed the new URL to the [`history`](https://reactrouter.com/web/api/history). 

This PR replaces navigation for the maps with the [`Link`](https://reactrouter.com/web/api/Link) component in order to allow opening links on a new tab, using the contextual menu to copy the link address, among others. If we go with this solution, we need to also replace links on:
- Compare table
- Top navigation bar
- Footer

but we can do that on a separate PR.

### DIFF
--- a/src/common/locations.ts
+++ b/src/common/locations.ts
@@ -1,6 +1,6 @@
 /** Helpers for dealing with the State / Counties dataset. */
 import US_STATE_DATASET from 'components/MapSelectors/datasets/us_states_dataset_01_02_2020.json';
-import { each, sortBy, takeRight } from 'lodash';
+import { each, sortBy, takeRight, find as _find } from 'lodash';
 import { assert } from './utils';
 import countyAdjacency from './data/county_adjacency_msa.json';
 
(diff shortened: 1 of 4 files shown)


ghop02:
Would this be easy to do for the compare table as well?

pnavarrc:
> Would this be easy to do for the compare table as well?

It should be easy enough, but I prefer to do it on a separate PR and check with Chelsi first :)

src/common/locations.ts
```diff
@@ -148,3 +148,11 @@ export function getAdjacentCounties(fips: string): string[] {
   assert(fips in ADJACENT_COUNTIES, `${fips} not found in adjacency list.`);
   return ADJACENT_COUNTIES[fips].adjacent_counties;
 }
+
+export function getCounty(stateId: string, fullFips: string) {
```
mikelehen, line 152: If we have the full fips, why don't we just use findCountyByFips()?

https://github.com/covid-projections/covid-projections/pull/1329/files#diff-4dab089aa9f4d6461076b7f21ec50a8fR69

PR #1335, opened 2020-08-25
feedback survey: add banner with tracking links

pnavarrc:
- Adds the feedback survey banner in the homepage
- It attaches a random ID to the Typeform URL, so we can track which responses belong to the same individual (we need this to link the first part of the feedback form to demographics, which are in the 3rd part of the survey)
- Adds a redirect so we can track users coming from social media with a unique ID as well. The link shared in social will be `https://covidactnow.org/feedback-survey`, upon clicking, the user will be redirected to the Typeform URL with `source=social&id={UNIQUE_ID}`

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "@types/d3-array": "^2.0.0",
     "@types/d3-shape": "^1.3.2",
     "@types/d3-time": "^1.0.10",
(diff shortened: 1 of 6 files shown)


pnavarrc:
[Feedback survey](https://trello.com/c/i0kL8Uuk/408-feedback-survey)

PR #1341, opened 2020-08-26
add redirect to track clicks to the covid response simulator

pnavarrc:
We want to track clicks to the COVID Response Simulator. Since Google Analytics already tracks page views for all our pages, one way to track clicks is to replace the URL of the CRS with an internal link to `/covid-response-simulator-redirect` and configure the router to redirect visits to  `/covid-response-simulator-redirect` to the actual CRS URL.

The `Redirect` component from `react-router-dom` only redirects to pages within the application, so I implemented a component to hide the implementation details of the external redirect.


pnavarrc:
[Track clicks to the CRS](https://trello.com/c/zqLyOGAk/411-track-clicks-to-the-crs)

PR #1344, opened 2020-08-26
explore: share component

pnavarrc:
### Notes
- Adapted the implementation of `LocationPage/ShareButtons` to make it a bit more generic, the new component doesn't include the logic that is specific for the metric charts.
- I used the `Button` component instead of a styled div to make it a bit more accessible.

### TODO
- Update the filename of the download image to include the location name
- Replace the existing component with the new component in the metric charts (maybe on a separate PR)
- Fix some styling issues
  - The share buttons should be aligned to the right
  - Check if it looks good on mobile
  - Spacing around the "Trends" title and subtitle


src/components/ShareButtons/ShareButtonGroup.tsx
```diff
@@ -0,0 +1,54 @@
...
+            <LinkedinShareButton url={url} quote={quote} />
+            <CopyLinkButton url={url} />
+          </SocialButtonsContainer>
+        </ClickAwayListener>
+      )}
+    </div>
+  );
+};
+
+export default ShareImageButtons;
```
pnavarrc, line 54: TODO: Rename the file to match the component name

mikelehen:
FYI- The tests are failing and the preview link doesn't work right now. 😁

pnavarrc:
> FYI- The tests are failing and the preview link doesn't work right now. 😁
 yeah, I think Cypress is not happy with `String.prototype.replaceAll` - trying with `replace` + regex now

pnavarrc:
updated!

src/common/locations.ts
```diff
@@ -110,6 +110,8 @@ export function getLocationNameForFips(fips: string): string {
 }
 
 export function getLocationUrlForFips(fips: string): string {
+  // TODO(pablo): Make sure that these urls are uniform, here, the state
+  // URL ends on / but not the county one
```
mikelehen, line 114: Hrm.  Why should the state end with / but not the county one?

pnavarrc: We probably missed that when implementing the function. I'm going to check where this function is used and update it so it's consistent once we ship Explore.

PR #1370, opened 2020-08-31
Website tweaks to enable dynamic screenshot generation.

mikelehen:
* Allow passing ?snapshot=xyz to most pages to choose what snapshot to use for rendering. This lets us generate screenshots for non-current snapshots as necessary.
* Refactor USACountyMap "summaries file" usage so that it can render maps for non-current snapshots.
* Add a <ScreenshotReady/> component that must be used to signal when the screenshot content is fully rendered and ready to be captured (i.e. all async requests are done, etc.).

For an example of what the first two bullets enable, see:
https://covid-projections-git-mikelehen-host-share-images-prework.covidactnow.vercel.app/?snapshot=983
vs
https://covid-projections-git-mikelehen-host-share-images-prework.covidactnow.vercel.app/?snapshot=974

### DIFF
--- a/src/common/utils/model.ts
+++ b/src/common/utils/model.ts
@@ -10,6 +10,7 @@ import { findCountyByFips } from 'common/locations';
 import moment from 'moment';
 import { RegionSummaryWithTimeseries } from 'api/schema/RegionSummaryWithTimeseries';
 import { assert } from '.';
+import { getSnapshotUrlOverride } from './snapshots';
 
 const cachedProjections: { [key: string]: Promise<Projections> } = {};
 export function fetchProjections(
@@ -17,6 +18,7 @@ export function fetchProjections(
   countyInfo: any = null,
   snapshotUrl: string | null = null,
 ) {
+  snapshotUrl = snapshotUrl || getSnapshotUrlOverride();
   let region: RegionDescriptor;
   if (countyInfo) {
     region = RegionDescriptor.forCounty(countyInfo.full_fips_code);
@@ -43,6 +45,7 @@ export function fetchProjections(
 /** Returns an array of `Projections` instances for all states. */
 const cachedStatesProjections: { [key: string]: Promise<Projections[]> } = {};
 export function fetchAllStateProjections(snapshotUrl: string | null = null) {
+  snapshotUrl = snapshotUrl || getSnapshotUrlOverride();
   async function fetch() {
     const all = await new Api(snapshotUrl).fetchAggregatedSummaryWithTimeseries(
       RegionAggregateDescriptor.STATES,
@@ -66,6 +69,7 @@ export function fetchAllStateProjections(snapshotUrl: string | null = null) {
 /** Returns an array of `Projections` instances for all counties. */
 const cachedCountiesProjections: { [key: string]: Promise<Projections[]> } = {};
 export function fetchAllCountyProjections(snapshotUrl: string | null = null) {
+  snapshotUrl = snapshotUrl || getSnapshotUrlOverride();
   async function fetch() {
     const all = await new Api(snapshotUrl).fetchAggregatedSummaryWithTimeseries(
       RegionAggregateDescriptor.COUNTIES,
--- a/src/common/utils/snapshots.ts
+++ b/src/common/utils/snapshots.ts
@@ -1,3 +1,4 @@
+import * as QueryString from 'query-string';
 import { assert } from 'common/utils';
 import DataUrlJson from 'assets/data/data_url.json';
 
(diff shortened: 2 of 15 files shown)


src/common/utils/snapshots.ts
```diff
@@ -25,3 +26,26 @@ export function snapshotUrl(snapshotNum: string | number) {
...
+  if (typeof window !== 'undefined') {
+    const search = window?.location?.search;
+    if (search !== undefined) {
+      const params = QueryString.parse(search);
+      if ('snapshot' in params) {
+        const snapshot = Number(params['snapshot']);
+        if (snapshot !== Number.NaN && snapshot !== 0) {
+          return snapshot;
+        }
+      }
```
pnavarrc, line 41: I think you can use `QueryString.parse(search, { parseNumbers: true })` to get the parameter already parsed.

src/common/utils/model.ts
```diff
@@ -10,13 +10,15 @@ import { findCountyByFips } from 'common/locations';
...
 import { assert } from '.';
+import { getSnapshotUrlOverride } from './snapshots';
 
 const cachedProjections: { [key: string]: Promise<Projections> } = {};
 export function fetchProjections(
   stateId: string,
   countyInfo: any = null,
   snapshotUrl: string | null = null,
 ) {
+  snapshotUrl = snapshotUrl || getSnapshotUrlOverride();
```
pnavarrc, line 21: I think it would be cleaner to have the caller of `fetchProjections` calling `getSnapshotUrlOverride` and passing the `snapshotUrl` parameter as usual instead? If this is much easier I think we can leave it though

src/components/Screenshot/index.ts
```diff
@@ -0,0 +1,10 @@
+export { default as ScreenshotReady } from './ScreenshotReady';
+
+/**
+ * This class name must be used to mark content that should be included in
+ * screenshotting purposes.
+ *
+ * It should be paired with a <ScreenshotReady /> element, which indicates when
+ * the content is fully-rendered and ready for the screenshot to be taken.
+ */
+export const SCREENSHOT_CLASS = 'screenshot';
```
pnavarrc, line 10: Maybe we can simplify this a bit by creating a component that renders both the wrapper and the sentinel at once:

```tsx
const ScreenshotWrapper: React.FC = ({ children }) => (
  <div className='screenshot'>
    {children}
    <div className='screenshot-ready' style={{display: 'none'}} />
  </div>
);

// later...

<ScreenshotWrapper>
  <SomeChart {...props} />
</ScreenshotWrapper>
```

src/screens/internal/ShareImage/ExploreChartImage.tsx
```diff
@@ -48,7 +49,7 @@ const ExploreChartImage = () => {
   }
 
   return (
-    <ScreenshotWrapper className={'screenshot'}>
+    <ScreenshotWrapper className={SCREENSHOT_CLASS}>
       <Wrapper>
```
pnavarrc, line 53: How come these don't have `<ScreenshotReady />`?

src/common/utils/snapshots.ts
mikelehen: Ah, cool.  Done!

src/common/utils/model.ts
mikelehen: I'm kinda' on the fence on this.  Right now most of our code isn't snapshot-aware.  In our location page, etc. we just call `useProjections(state, [county])` and don't think about snapshots.  The only code that is currently snapshot-aware is the compare tool, since it's comparing snapshots.

I think I'd like to keep it that way, meaning that this snapshot override just happens transparently without the calling code knowing about it.  

But I can see both points-of-view.  It's a little "magical" as-is which means somebody could accidentally break it or whatever.

src/components/Screenshot/index.ts
mikelehen: Yeah...  I originally tried to make this cleaner but I ran into two problems:

1. If we have a ScreenshotWrapper component and then you try to style it with styled-components (which _seems_ like a totally normal / sane thing to do, then the className gets lost (it gets minified or omitted or something) and so things break.
2. The reason we need the screenshot-ready div is because we can't rely on all of the children populating synchronously.  In particular, we don't know when the projections are done loading and the chart has rendered at its final width.  So it's a bit hacky, but right now you need to carefully place the `<ScreenshotReady>` component somewhere that's deeply nested enough that it'll only render once everything is actually ready.  So e.g. for the charts, we place it here: https://github.com/covid-projections/covid-projections/pull/1370/files/5882001a1fa82d49db87f3b3189741a38a904611#diff-1f39dac4db94a4170cf8add355312ee6R73

src/screens/internal/ShareImage/ExploreChartImage.tsx
mikelehen: Ack!  Good catch.  Added: https://github.com/covid-projections/covid-projections/pull/1370/files/5882001a1fa82d49db87f3b3189741a38a904611#diff-1f39dac4db94a4170cf8add355312ee6R73

src/common/utils/model.ts
pnavarrc: makes sense 👍

src/components/Screenshot/index.ts
pnavarrc: oh I see, that makes sense

PR #1371, opened 2020-08-31
Implement dynamic screenshot generation via Firebase Hosting + Functions

mikelehen:
Adds a hosting redirect to a "dynamicWebContent" Firebase function which uses
express + puppeteer + headless chrome to load the necessary page and generate /
serve a screenshot.

See the firebase/README.md file for how you can test this if you want.  Or you can see it in action on the dev firebase instance, e.g.: https://covidactnow-dev.web.app/share/983-123/states/ca/chart/0.png


firebase/firebase.json
```diff
@@ -7,6 +7,26 @@
     "predeploy": [
       "npm --prefix \"$RESOURCE_DIR\" run lint",
       "npm --prefix \"$RESOURCE_DIR\" run build"
+    ],
+    "source": "functions"
+  },
+  "hosting": {
+    "public": "public",
```
mikelehen, line 14: FYI- I set up "public" as a symlink to "../build" and now hosting serves it.  This isn't necessary for now, but it moves us towards being able to move the whole site to Firebase Hosting if we feel like it.  i.e. for a Firebase-hosted version of the site, see: https://covidactnow-dev.web.app/

firebase/firebase.json
```diff
@@ -7,6 +7,26 @@
...
+    "ignore": [
+      "firebase.json",
+      "**/.*",
+      "**/node_modules/**",
+      "us/**"
+    ],
+    "rewrites": [
+      {
+        "source": "/share/**",
+        "function": "dynamicWebContent"
```
mikelehen, line 24: This directs all /share/... requests to the "dynamicWebContent" firebase function.

In the future we could serve other requests as well (e.g. to do the meta tag replacement stuff.)

mikelehen:
FYI- Once this is merged, we can deploy it with:

```
firebase deploy -P prod
```

and then change the share_images_url to point at e.g. "https://covidactnow-prod.web.app/share/983-197/" instead of "https://content.covidactnow.org/share/983-197/"

firebase/public
```diff
@@ -0,0 +1 @@
+../build
```
mikelehen, line 1: This is a symlink from `firebase/public` to `../build` because Firebase won't let me point to a parent directory of firebase.json.  

It's a bit of a hack.  We could move things around to avoid the symlink, but it seems to work for now. 🤷

firebase/README.md
```diff
@@ -9,3 +16,17 @@ firebase -P staging deploy
...
+# Testing Hosting / Functions
+To test hosting / functions locally:
+
+```
+# Run this from the 'firebase/functions' directory to watch / recompile TypeScript files.
+`npm bin`/tsc -w
+
+# Run this to run the hosting / functions local servers.
+firebase serve
+```
```
pnavarrc, line 29: I tried running this, but it didn't work ... do I need to install Firebase globally? I tried running `yarn` in the top-level directory and in `firebase` with no luck 😢

firebase/functions/src/screenshots.ts
```diff
@@ -0,0 +1,59 @@
...
+  console.log('Waiting for "screenshot-ready" div.');
+  await tab.waitForSelector('.screenshot-ready', {
+    timeout: TIMEOUT
+  });
+
+  console.log('Capturing screenshot.');
+  const file = path.join(os.tmpdir(), 'temp.png')
+  await element.screenshot({
+    path: file
+  });
```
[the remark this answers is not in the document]
mikelehen: I think instances can be reused, but an instance will only handle one request at once, so there's no issue with reusing the filename.  And it simplifies things.  In particular, we have to be a bit mindful of disk usage, so if we generated random names, we'd probably have to ensure we deleted them as well.  So keeping it as-is, but I added a comment.

PR #1373, opened 2020-08-31
Fix: compare modal

chasulin:
- Fixes (removes) gap at the bottom of compare modal that occurs when metro menu is not displayed (mobile only)

Before:
Screen Shot 2020-08-31 at 1 41 34 PM

After:
Screen Shot 2020-08-31 at 1 43 21 PM

### DIFF
--- a/src/components/Compare/LocationTable.tsx
+++ b/src/components/Compare/LocationTable.tsx
@@ -189,7 +189,13 @@ const LocationTable: React.FunctionComponent<{
   const Container = isModal ? Styles.ModalContainer : Styles.Container;
 
   // Seemingly random numbers are the heights of each modal header
-  const modalHeaderOffset = isHomepage ? 159 : pinnedLocation ? 193 : 110;
+  const homepageOffset = viewAllCounties ? 159 : 73;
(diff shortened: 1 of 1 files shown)


src/components/Compare/LocationTable.tsx
```diff
@@ -189,7 +189,13 @@ const LocationTable: React.FunctionComponent<{
...
 
   // Seemingly random numbers are the heights of each modal header
-  const modalHeaderOffset = isHomepage ? 159 : pinnedLocation ? 193 : 110;
+  const homepageOffset = viewAllCounties ? 159 : 73;
+  const locationPageOffset = geoScope === GeoScopeFilter.NEARBY ? 107 : 198;
+  const modalHeaderOffset = isHomepage
+    ? homepageOffset
+    : pinnedLocation
+    ? locationPageOffset
+    : 110;
```
pnavarrc, line 198: I think there is a better way to manage the layout for the modal using flexbox (I made a quick demo here https://codesandbox.io/s/goofy-voice-8h4k6?file=/src/styles.css), but we can merge this for now if changing it would take too long.

chasulin: ooo thank you! im gonna merge for now and do the above separately

PR #1376, opened 2020-08-31
remove the feedback survey banner

pnavarrc:


src/App.js
```diff
@@ -119,17 +118,6 @@ export default function App() {
...
-              {/**
-               * This endpoint is to share the feedback survey link in social
-               * media. We redirec them to Typeform with URL parameters to
-               * track users through the survey, as well as their source.
-               */}
-              <Route
-                path="/feedback-survey"
-                component={() => (
-                  <ExternalRedirect url={getFeedbackSurveyUrl('social')} />
-                )}
```
mikelehen, line 131: Any idea how prominently we sent this link out and if it’s still getting traffic?  I wonder if we should leave the redirect for a while...

pnavarrc: It seems that we didn't get responses from social (vs. 1,500 from the website), but maybe you are right, since the link is out there, we might as well keep it for a little longer just in case.

PR #1381, opened 2020-09-01
explore: v1 feedback

pnavarrc:
Addressing feedback from [CAN Explore - Feedback](https://paper.dropbox.com/doc/CAN-Explore--A6wWR1iltDXmI1xe4ZTRyFhrAg-RIlOIJkMWHcX0AmXQA5PW)

- [ ] Add an empty state for when we don't have data
- [ ] Several copy updates to the tooltips
- [ ] Adjust the line to match the bar charts (the last data point didn't exactly match the last bar)


src/components/Explore/Explore.tsx
```diff
@@ -68,22 +73,37 @@ const Explore: React.FunctionComponent<{
...
               />
-            )
-          }
-        </ParentSize>
-      </Styles.ChartContainer>
+            )}
+          </ParentSize>
+        </Styles.ChartContainer>
+      ) : (
+        <EmptyChart height={400}>
```
mikelehen, line 92: FYI- I put the `<ScreenshotReady/>` marker inside `ExploreChart` so if somebody tried to share an empty chart, the screenshot will fail...  but probably nobody will do that, so it probably doesn't matter.  But it might be sane to move the marker up to this level.

src/components/Explore/ExploreChart.tsx
```diff
@@ -42,12 +45,18 @@ const ExploreTooltip: React.FC<{
...
       left={left(pointSmooth)}
       title={moment(date).format('MMM D, YYYY')}
     >
       <Styles.TooltipSubtitle>{seriesRaw.tooltipLabel}</Styles.TooltipSubtitle>
-      <Styles.TooltipMetric>{formatInteger(pointRaw.y)}</Styles.TooltipMetric>
+      <Styles.TooltipMetric>
+        {formatInteger(pointSmooth.y)}
+      </Styles.TooltipMetric>
+      <Styles.TooltipSubtitle>
+        {`${formatInteger(pointRaw.y)} cases`}
```
mikelehen, line 58: We seem to say "cases" even if it's deaths / hospitalizations / whatever...

image

src/components/Explore/utils.ts
```diff
@@ -77,13 +79,13 @@ export const exploreMetricData: {
...
         label: 'Cases',
-        tooltipLabel: 'Confirmed cases',
+        tooltipLabel: '7-day avg. cases',
         datasetId: 'rawDailyCases',
         type: ChartType.BAR,
       },
       {
         label: '7 Day Average',
-        tooltipLabel: 'Confirmed cases',
+        tooltipLabel: '7-day avg. cases',
```
mikelehen, line 88: We need to say "7-day avg. deaths" etc. for the other metrics, right?

src/components/Explore/ExploreChart.tsx
mikelehen: FWIW- Feel free to ignore, but if it were me I'd maybe make the tooltip say:

> HOSPITALIZATIONS 
> 46
> **7-DAY AVG. 55**

which means for cases it'd be:

> CASES
> 249
> **7-DAY AVG. 206**

Basically we only need to say "cases" once.  And it makes sense to put the raw number before the averaged number, since the average number is derived from the raw number.  But we can still emphasize the averaged number by making it bold still.

🤷

pnavarrc: oops, yeah, that's a bug

src/components/Explore/utils.ts
pnavarrc: yes, will fix soon!

src/components/Explore/Explore.tsx
pnavarrc: I'm adding it to the empty chart just in case

pnavarrc:
I renamed the branch and create a second PR with the same changes, GitHub couldn't figure out the correct diff for this one. I'm merging https://github.com/covid-projections/covid-projections/pull/1385 with the fixes instead.

PR #1386, opened 2020-09-01
explore: more fixes

pnavarrc:
- Remove leading zeroes in the series charts
- Add a disclaimer mentioning our data sources.


src/components/Explore/Explore.tsx
```diff
@@ -104,6 +113,19 @@ const Explore: React.FunctionComponent<{
           </p>
         </EmptyChart>
       )}
+      <DisclaimerWrapper>
+        <DisclaimerBody>
+          Last updated {lastUpdatedDateString}. Learn more about{' '}
+          <ExternalLink href="https://docs.google.com/document/d/1cd_cEpNiIl1TzUJBvw9sHLbrbUZ2qCxgN32IqVLa3Do/edit">
+            our methodology
```
mikelehen, line 120: FWIW this link feels a little pointless for "explore" since the only "methodology" we're applying is a 7-day average to the raw data.  So I'd be okay with dropping it.  But 🤷

src/components/Explore/utils.ts
```diff
@@ -151,10 +159,13 @@ export const EXPLORE_CHART_IDS = Object.values(exploreMetricData).map(
   metric => metric.chartId,
 );
 
-const hasValue = (point: Column) => isNumber(point.y);
+const missingValue = (point: Column) => !isNumber(point.y);
 
+/**
+ * Remove points without y values at the begining and end of the series
```
mikelehen, line 165: Perhaps add a TODO about the fact that we still should deal with missing y values in the middle of the series. 🤷

PR #1395, opened 2020-09-02
Mariana islands wip

ghop02:
Get the basics of Mariana Islands working. 

 * This does not include MP on the main map - still need to work out how to do that
 * I would like to get a better county map for MP - the islands are a bit round right now. 
 * Need to figure out how to handle counties that have no data. Right now, we don't produce api results if there is no timeseries data - this might not be the correct approach, but when you click on a county without data it throws an assertion error.


src/components/Map/USACountyMap.js
```diff
@@ -39,6 +56,25 @@ const USACountyMap = ({ stateClickHandler, setTooltipContent, condensed }) => {
...
                 const { name } = geo.properties;
                 const stateCode = getStateCode(name);
+
+                if (stateCode === 'MP') {
+                  return (
+                    <Link key={stateCode} to={`/us/${stateCode.toLowerCase()}`}>
+                      <MarianaIslands
+                        x={40}
+                        y={395}
+                        scale={0.8}
```
mikelehen, line 66: Would it make sense to just hardcode these in the MarianaIslands component?

src/components/Compare/CompareTable.tsx
```diff
@@ -86,7 +87,7 @@ const CompareTable = (props: {
   const getPopulation = (location: SummaryForCompare) =>
     location?.locationInfo?.population;
   const getMetricValue = (location: any) =>
-    location.metricsInfo.metrics[sorter].value;
+    location?.metricsInfo?.metrics[sorter]?.value;
```
chasulin, line 90: I don't think this change is needed? `location.metricsInfo.metrics[sorter].value` is always defined (either as a number or null)

src/components/Compare/CompareTable.tsx
```diff
@@ -155,6 +156,14 @@ const CompareTable = (props: {
   const isMobile = useMediaQuery(theme.breakpoints.down(700));
   const faqLinkCopy = isMobile ? 'FAQ' : 'Frequently Asked Questions';
 
+  // Disabling filters for Northern Mariana Islands because they don't have
+  // any data on metro vs non-metro islands.  There may be more elegant solutions
+  // that better handle any region without metro/non-metro regions.
+  const disableFilters = stateId === 'MP';
```
chasulin, line 162: So these 4 fips are all included in county_adjacency_msa.json--only 2 return a ‘nearby’ county, and a janky google search gives the impression that all 4 fall into the category of non-metro. Compare currently breaks+disappears when a filter selection returns nothing, so we can fix that (I prob should fix this regardless) and include the filters (though they’re not _super_ meaningful) if people are interested. Just an option !

src/components/MapSelectors/datasets/us_states_dataset_01_02_2020.json
```diff
@@ -48022,6 +48028,62 @@
...
+      ],
+      "county_dataset": [
+        {
+          "county": "Northern Islands Municipality",
+          "county_url_name": "northern_islands_municipality",
+          "county_fips_code": "085",
+          "state_fips_code": "69",
+          "state_code": "MP",
+          "full_fips_code": "69085",
+          "cities": [""],
```
mikelehen, line 48051: nit: Do we need a at least one entry, or could this just be `cities: []` ?

src/components/Map/USACountyMap.js
```diff
@@ -10,9 +10,26 @@ import { useSummaries } from 'common/location_summaries';
...
 
+const MarianaIslands = ({ x, y, fill, scale, onMouseEnter, onMouseLeave }) => {
+  return (
+    <g
+      transform={`translate(${x}, ${y}) scale(${scale})`}
+      onMouseEnter={onMouseEnter}
+      onMouseLeave={onMouseLeave}
+    >
+      <rect width="20" height="20" fill={fill} />
+      <text transform="translate(25, 15)">CNMI</text>
```
chasulin, line 28: Is this label just temporary?

mikelehen (approved):
Generally looks good.  Can you revert all the scripts/alert_emails/summaries/... files?  It probably doesn't matter assuming the only material change is adding MP, but in general I try not to modify these after we've shipped the snapshot public.

If you make major changes, let me know and I can take a second pass.

src/components/Map/USACountyMap.js
ghop02: The box with CNMI is the UX that we settled on for now.  It is possible that we may change it in the future, but no immediate plans

src/components/Map/USACountyMap.js
```diff
@@ -10,9 +10,26 @@ import { useSummaries } from 'common/location_summaries';
 import { ScreenshotReady } from 'components/Screenshot';
 
 function getStateCode(stateName) {
+  // hacky, name defined slightly differently in normal reversed states file.
+  if (stateName === 'Commonwealth of the Northern Mariana Islands') {
```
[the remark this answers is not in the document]
ghop02: Yeah this is a good question.  The main reason I went with Northern Mariana Islands on the location page + search was because it makes the minimap pretty cramped:
image

src/components/Map/USACountyMap.js
```diff
@@ -10,9 +10,26 @@ import { useSummaries } from 'common/location_summaries';
...
+    return 'MP';
+  }
   return REVERSED_STATES[stateName];
 }
 
+const MarianaIslands = ({ x, y, fill, scale, onMouseEnter, onMouseLeave }) => {
```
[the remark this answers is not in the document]
ghop02: Yup! that's exactly why.  There's a good chance it will change if we add in more territories - but not on the immediate roadmap

src/components/Compare/CompareTable.tsx
ghop02: Yeah - maybe worth a trello ticket for handling this case better? I initially was hoping to disable the menu item based on whether or not the filters returned any locations, but that required a bit more work than this if statement haha

src/components/MapSelectors/datasets/us_states_dataset_01_02_2020.json
ghop02: That's a good question - I was mostly just copying what we had from the other entries

src/components/Map/USACountyMap.js
```diff
@@ -39,6 +56,25 @@ const USACountyMap = ({ stateClickHandler, setTooltipContent, condensed }) => {
               geographies.map(geo => {
                 const { name } = geo.properties;
                 const stateCode = getStateCode(name);
+
+                if (stateCode === 'MP') {
```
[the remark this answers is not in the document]
ghop02: I was originally thinking that - when I asked pablo he recommended this route since the SVG work is a bit more custom (including the CNMI text).  I don't think there is an immediately easy way to do that.

src/components/Map/USACountyMap.js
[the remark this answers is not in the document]
ghop02: I think if people are okay with the truncated `Commonwealth of t...` on the minimap, I'd prefer to just go with `Commonwealth of the Northern Mariana Islands` for consistency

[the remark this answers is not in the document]
mikelehen: I'm not too worried about the minimap text, but I'm also not sure what the motivation is to prefer the longer name? Especially when we do have space constraints...

The wikipedia article is under "Norther Mariana Islands" (listing "Commonwealth ...") as the _official_ name..  https://en.wikipedia.org/wiki/Northern_Mariana_Islands. 🤷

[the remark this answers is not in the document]
ghop02: I was using Northern Mariana Islands everywhere, but the `counties-10m.json` file which defines the map and is the original list of states has it listed with Commonwealth.  I could modify that file to change the name to Northern Mariana Islands, but since we essentially downloaded it, i figured it'd be better to not modify it and shim in the state code in the javascript.  

The easier solution is probably just changing "Commonwealth of the Northern Mariana Islands" to "Northern Mariana Islands" in counties-10m.json.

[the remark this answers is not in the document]
mikelehen: Yeah, I'm fine with modifying those files.

src/components/Map/USACountyMap.js
ghop02: I hesitated because i was thinking of the MarianaIslands component as more standalone that could be placed anywhere, but don't think it adds much value so fine moving it in the component itself

src/components/Map/USACountyMap.js
ghop02: Added a comment above

PR #1412, opened 2020-09-02
Only use ICU bed capacity from the timeseries if it's from the last 7 days.

mikelehen:
https://trello.com/c/nvHzgI0r/420-icuutilization-total-beds-disconnect-between-actuals-and-actualstimeseries

Imperial County before (20 beds):
image

Imperial County after (28 beds):
image


ghop02 (approved):
LGTM, I'll make a note to update this change on the backend

PR #1413, opened 2020-09-03
explore v2: add lines and labels for multiple locations

pnavarrc:
This PR adds support to show metrics for multiple locations. The chart itself doesn't have many modifications because `ExploreChart` receives `series`, so we just pass series with data from different locations.

- Updated the `SeriesChart` components to it can take `params` to customize how the lines look.
- Added optional labels on `ExploreChart`
- Moved the body of `EmptyChart` to the component

Note that the tooltip for multiple locations is broken, that's coming up on a separate PR!


src/components/Explore/interfaces.ts
```diff
@@ -5,11 +5,16 @@ export enum ChartType {
   BAR,
 }
 
+export interface ChartParams {
```
mikelehen, line 8: nit: If you rename SeriesChart, maybe rename this to SeriesParams or something.

src/components/Explore/ExploreChart.tsx
```diff
@@ -176,7 +178,7 @@ const ExploreChart: React.FC<{
             <GridRows<number> scale={yScale} width={innerWidth} />
           </ChartStyle.LineGrid>
           <RectClipGroup width={innerWidth} height={innerHeight}>
-            {series.map(({ label, data, type }) => (
+            {series.map(({ label, data, type, params }) => (
               <SeriesChart
```
mikelehen, line 182: nit: It seems a bit weird that we're creating multiple "charts"...  I'd expect a single **chart** with multiple **series** or similar.  So maybe `SeriesChart` should be `ChartSeries` instead or something?

Anyway, feel free to punt for now.

src/components/Explore/utils.ts
```diff
@@ -325,3 +336,55 @@ export function getAutocompleteLocations(locationFips: string) {
         .filter(isCounty)
         .filter(location => belongsToState(location, stateFips));
 }
+
+const SERIES_COLORS = [
```
mikelehen, line 340: Heh.  Where did these colors come from?  Add a comment?

src/components/Explore/utils.ts
```diff
@@ -325,3 +336,55 @@ export function getAutocompleteLocations(locationFips: string) {
...
+    },
+    label: getLocationLabel(location),
+    tooltipLabel: '',
+  };
+}
+
+const getCountyForLocation = (location: Location) =>
+  isCounty(location) ? location : undefined;
+
+export function getChartSeries(metric: ExploreMetric, locations: Location[]) {
```
mikelehen, line 374: Does this return `Series[]`?  Can you make it explicit?

src/components/Explore/utils.ts
```diff
@@ -325,3 +336,55 @@ export function getAutocompleteLocations(locationFips: string) {
...
+  '#d62728',
+  '#9467bd',
+  '#8c564b',
+  '#e377c2',
+  '#7f7f7f',
+  '#bcbd22',
+  '#17becf',
+];
+
+function getSmoothedSeries(
```
mikelehen, line 353: nit: This should maybe be closer to getSeries() and maybe we should add comments to explain / differentiate the functions (and/or come up with clearer names)?

Feel free to punt for now.

mikelehen (approved):
Looks good! Left a few nits. but this is cool!

PR #1419, opened 2020-09-03
Generate sharing URLs for compare pages.

mikelehen:
1. Parse query string params and use them as default values for compare table options.
2. Use compare table options to generate a sharing URL and (for now) log it to the console.

### DIFF
--- a/src/components/Compare/CompareMain.tsx
+++ b/src/components/Compare/CompareMain.tsx
@@ -1,4 +1,6 @@
+import * as QueryString from 'query-string';
 import React, { useState, Fragment, useRef, useEffect } from 'react';
+import { useLocation } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


src/components/Compare/CompareMain.tsx
```diff
@@ -38,13 +42,61 @@ const CompareMain = (props: {
...
+    viewAllCounties,
+    geoScope,
+  };
+
+  const shareUrl = urls.getComparePageUrl(
+    props.stateId,
+    props.county,
+    shareParams,
+  );
+  console.log('Share URL: ', shareUrl);
```
mikelehen, line 99: This is a placeholder until we wire this into the share buttons, etc.

chasulin (approved):
looks good!

PR #1439, opened 2020-09-08
Using top level metrics

ghop02:
There are still a couple edge cases that I have to work around, but this is most of the way there for using the top level metrics instead of calculating them ourselves.

One more data change i'm making is only surfacing the metric if it is within the last 7 days.  There were cases where the timeseries would be empty if there was no data in the the past 7 days.  Now, the timeseries will be filled with the data we have, but the top level metric score will only appear from data in the previous 7 days.

As a total aside - I found the null vs undefined distinction a bit confusing.  I think i've seen shying away from `null` in a couple typescript style guides. was there a reason for the null usage?


mikelehen (commented):
On the topic of `null` vs `undefined`, I don't know which is most idiomatic.  On the last project I worked on, we preferred using `null` over `undefined` when you want explicit optional values since `undefined` can easily show up by mistake if you typo a variable or dictionary entry or whatever.  So generally we used `undefined` only in localized scopes, purely as a JavaScript language feature for determining existence, etc. but if it showed up in data structures / method arguments / etc., it's a bug.

Anyway, that's probably how I've been writing code.  But a lot of this code predates me and sometimes it's easier to be a bit loose / defensive than to necessarily clean everything up and guarantee consistency, so there may be a bit of both.

src/common/models/Projection.ts
```diff
@@ -252,35 +232,40 @@ export class Projection {
...
 
-    this.caseDensityByCases = this.calcCaseDensityByCases();
+    this.caseDensityByCases = metricsTimeseries.map(row =>
+      row ? row.caseDensity : null,
+    );
     this.caseDensityByDeaths = this.calcCaseDensityByDeaths();
     this.caseDensityRange = this.calcCaseDensityRange();
 
-    this.currentCaseDensityByCases = lastValue(this.caseDensityByCases);
+    this.currentCaseDensityByCases = metrics ? metrics!.caseDensity : null;
```
mikelehen, line 264: `!` shouldn't be necessary?

Also, FWIW we could clean up this code and get rid of caseDensityByCases/Deaths since I don't think we'll bring deaths back.

src/common/models/Projection.ts
```diff
@@ -252,35 +232,40 @@ export class Projection {
...
     this.currentCaseDensityByDeaths = lastValue(this.caseDensityByDeaths);
-    this.currentCaseDensity = lastValue(
-      this.caseDensityRange.map(range => range && range.caseDensity),
-    );
+    this.currentCaseDensity = metrics ? metrics!.caseDensity : null;
```
mikelehen, line 266: shouldn't need `!`

src/common/models/Projection.ts
```diff
@@ -299,13 +284,15 @@ export class Projection {
       summaryWithTimeseries.actuals.cumulativeDeaths;
     this.currentCumulativeCases =
       summaryWithTimeseries.actuals.cumulativeConfirmedCases;
-    this.currentContactTracerMetric = lastValue(this.contractTracers);
+    this.currentContactTracerMetric = metrics
+      ? metrics!.contactTracerCapacityRatio
```
mikelehen, line 288: shouldn't need `!`

src/common/models/Projection.ts
```diff
@@ -299,13 +284,15 @@ export class Projection {
...
+      : null;
   }
 
   get currentContactTracers() {
     return (
       CONTACT_TRACER_STATE_OVERRIDES[this.stateName] ||
-      lastValue(this.actualTimeseries.map(row => row && row.contactTracers)) ||
+      this.metrics?.contactTracerCapacityRatio ||
```
mikelehen, line 295: nit: This would coerce `0` to null, but since there are no states with 0 contact tracers, it's probably fine. 😁

src/common/models/Projection.ts
```diff
@@ -426,13 +415,19 @@ export class Projection {
...
+    const metricsTimeseriesRaw = summaryWithTimeseries.metricsTimeseries;
     assert(
       actualsTimeseriesRaw.length > 0,
       `FIPS ${this.fips} missing actuals timeseries!`,
     );
     let earliestDate, latestDate;
     // If we have projections, we use that time range; else we use the actuals.
-    if (timeseriesRaw.length > 0) {
+    // TODO(chris): Is there a reason that this was bound to the projections timeseries first?
+    // It cuts off some of the earlier dates
```
mikelehen, line 426: It's helpful to have a consistent start date so that we know that graphs for different locations will still render the same (so we can make sure it'll be legible at different resolutions, etc.)  Originally we did that by pinning everybody to the projections start date.

But this code is a mess and I think we could maybe just hardcode a start date of 3/1 at this point.

FWIW- This aligning code has always been kinda' yucky (I've been thinking we should just have a proper `Timeseries` class that wraps a timeseries and is date-aware)...  and we're probably really close to not needing it anymore, since we're no longer doing calculations between different timeseries that may not be date-aligned, since that's now in the backend...

src/common/models/Projection.ts
```diff
@@ -542,10 +517,10 @@ export class Projection {
...
   private calcRtRange(
-    timeseries: Array<PredictionTimeseriesRow | null>,
+    timeseries: Array<MetricsTimeseriesRow | null>,
   ): Array<RtRange | null> {
-    const rtSeriesRaw = timeseries.map(row => row && row.RtIndicator);
-    const rtCiSeriesRaw = timeseries.map(row => row && row.RtIndicatorCI90);
+    const rtSeriesRaw = timeseries.map(row => row && row.infectionRate);
+    const rtCiSeriesRaw = timeseries.map(row => row && row.infectionRateCI90);
 
     // This hides small gaps (less than 2 data points) in the rt series to make
```
mikelehen, line 525: FWIW- As part of this API v2 work, it would be nice to verify that none of this post-processing is needed anymore (we should fix it upstream if it is).  Though that doesn't need to be in this PR.

src/common/utils/index.ts
```diff
@@ -72,11 +72,16 @@ export function formatInteger(num: number): string {
  *   formatDecimal(1.2345)    // 1.23
  *   formatDecimal(1.2345, 3) // 1.234
  */
-export const formatDecimal = (num: number, places = 2): string =>
-  num.toLocaleString(undefined, {
+export const formatDecimal = (num: number, places = 2): string => {
+  if (num === null) {
+    return '--';
+  }
```
mikelehen, line 78: this is probably fine, but I'm mildly curious why it was necessary for this PR?

mikelehen (commented):
A few more comments...

In general, the code looks great and I'd be inclined to approve it, but I'd like to do a bit of actual validation on the data.  It might be nice to wire this into the compare tool (e.g. hack things up so we can load the same snapshot with both the old and new code and compare the resulting graphs). Thoughts on if this is doable / worthwhile?  Maybe check the old code back in as ProjectionOld.ts and then hack up enough plumbing so that the compare tool can opt into it?  Idunno, maybe a pain.

Also, on the `null` / `undefined` thing, I'm definitely flexible if you have an opinion on what direction this code should go and want to move it in that direction. 🤷

src/common/models/Projection.ts
ghop02: Aha - yes this is right, I think i was using `metrics?.caseDensity` before not realizing that the first part of the check unwraps the `metrics` variable

src/common/models/Projection.ts
ghop02: I think this will actually give a much more consistent start date since most of the metrics timeseries code goes back to early februrary.

ghop02: Also agree with the idea of a timeseries class of sorts! This was a good thought exercise to see where that might be useful.  Maybe part of it is breaking up the `Projection` class to be less powerful

mikelehen: Do we think that early February data is useful to show?

ghop02: We do not! but the code below will standardize it to starting at march 1st.  Maybe we could remove the earliest time all together and just find an appropriate latest time?

mikelehen: Yeah, I think that would be fine.  Like I said, this code has evolved into weirdness and doesn't make a ton of sense anymore.

src/common/models/Projection.ts
ghop02: done!

src/common/utils/index.ts
ghop02: There are a couple more null data points in some of the series that where causing some errors.  This was the best fix I came up with, but could put more effort into sussing out further up the chain. Also, I think other missing points might be an em-dash so i'll try to change that

src/common/models/Projection.ts
ghop02: I actually think this is wrong anyways, this should be the raw number of contact tracers, not the ratio

src/common/models/ICUHeadroom.ts
```diff
@@ -99,152 +85,23 @@ export function calcICUHeadroom(
...
-    actualTimeseries,
-    timeseries,
-    lastUpdated,
-  );
-
   // Use capacity from the timeseries if it's within the last 7 days, else use the
   // non-timeseries value.
   const finalTotalBeds =
     lastValue(actualTimeseries.map(r => r?.ICUBeds.capacity).slice(-7)) ||
     actuals.ICUBeds.totalCapacity;
```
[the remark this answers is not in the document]
ghop02: I ended up making a trello ticket for finishing that up: https://trello.com/c/CUcjDdtt/435-add-total-icu-beds-to-icu-headroom-metadata-instead-of-calculating-on-website

mikelehen (approved):

PR #1443, opened 2020-09-09
compare: share quotes and UI

chasulin:
- Adds share quote, save/share buttons, updated FAQ button, and some css cleanup

[Webapp doc with drafted share copy](https://docs.google.com/document/d/1Jsj4zOksCJcgjQLUo3rwVAobEUkGLFukD1K71IqVNpo/edit#heading=h.9ih5qfq369pk)

Todo:
- [ ] Finalize copy
- [x] UX/Josh signoff
- [ ] set `imageFilename`


src/common/utils/compare.ts
```diff
@@ -258,3 +273,78 @@ export function isCollegeCounty(location: Location) {
   const countyPopulation = location.population;
   return ftEnrollment && ftEnrollment / countyPopulation > threshold;
 }
+
+// For sharing:
+
+export function getShareQuote(
+  sorter: Metric,
+  countyTypeToView: MetroFilter,
+  sliderValue: GeoScopeFilter,
```
mikelehen, line 282: I think this type is wrong.  It's not a `GeoScopeFilter`, it's just a `number` (0, 50, or 99).

But it might be better if it _was_ the `GeoScopeFilter` because that's a much clearer type (doesn't involve magic numbers) and then you wouldn't need to have the exported `sliderValueMap` thing.

src/common/utils/compare.ts
```diff
@@ -258,3 +273,78 @@ export function isCollegeCounty(location: Location) {
...
+
+  const stateShareCopy = `Compare COVID metrics between ${getMetroPrefixCopy(
+    countyTypeToView,
+    true,
+  )} counties in ${stateName} with @CovidActNow.`;
+
+  const hasValidRank =
+    currentLocation &&
+    currentLocation.rank !== 0 &&
+    currentLocation.metricsInfo.metrics[sorter]?.value;
```
mikelehen, line 311: This breaks if you share counties with `0` for a metric value, I think?

e.g. https://covid-projections-git-casulin-compare-share-quotes.covidactnow.vercel.app/us/wa/county/lincoln_county

src/common/utils/compare.ts
```diff
@@ -258,3 +273,78 @@ export function isCollegeCounty(location: Location) {
...
+  sliderValue: GeoScopeFilter,
+  totalLocations: number,
+  sortDescending: boolean,
+  currentLocation?: RankedLocationSummary,
+  sortByPopulation?: boolean,
+  isHomepage?: boolean,
+  viewAllCounties?: boolean,
+  stateName?: string,
+): string {
+  const sliderValueShareCopy: any = {
```
mikelehen, line 291: This should be called geoScopeShareCopy probably?

src/components/Compare/Compare.style.tsx
```diff
@@ -243,77 +243,57 @@ export const TableHeadContainer = styled(TableHead)<{ isModal?: Boolean }>`
 export const Population = styled.span`
```
mikelehen, line 243: Looks like there's a slight placement issue with the slider on mobile...  (Nearby is falling off)

image

src/components/Compare/CompareMain.tsx
```diff
@@ -199,10 +200,15 @@ const CompareMain = (props: {
           locationsViewable={props.locationsViewable}
           isModal={false}
           viewMoreCopy={viewMoreCopy}
+          shareUrl={shareUrl}
```
mikelehen, line 203: Looks like you added this to sharedProps so we probably don't need this?

src/components/Compare/CompareMain.tsx
```diff
@@ -199,10 +200,15 @@ const CompareMain = (props: {
...
         />
       </DivForRef>
       <Modal open={showModal} onClose={handleCloseModal}>
-        <ModalCompare {...sharedProps} handleCloseModal={handleCloseModal} />
+        <ModalCompare
+          {...sharedProps}
+          handleCloseModal={handleCloseModal}
+          shareUrl={shareUrl}
```
mikelehen, line 210: Looks like you added this to sharedProps so we probably don't need this?

src/components/Compare/Filters.tsx
```diff
@@ -98,16 +99,10 @@ const Filters = (props: {
     }
   };
 
-  const valueMap: any = {
-    0: GeoScopeFilter.NEARBY,
-    50: GeoScopeFilter.STATE,
-    99: GeoScopeFilter.COUNTRY,
-  };
```
mikelehen, line 105: Per my other comment, if we can move the sliderValueMap back into this file and keep the magic 0, 50, 99 values contained to this file, I think that'll be cleaner.

Hopefully external code can use the geoScope value directly.

mikelehen (approved):
Left some comments, but overall this looks great.  Nice work covering all the edge cases. 😬

src/common/utils/compare.ts
chasulin: The name `sliderValue` is probably confusing things here-- the type is correct. `sliderValue` is the GeoScopeFilter that the numerical slider value (0, 50, 99) corresponds to.

As for the magic numbers themselves, the [material UI slider component](https://material-ui.com/api/slider/) calls for them unfortunately

src/components/Compare/Filters.tsx
chasulin: Ooo yes good call. Moving this map back into Filters.tsx

src/common/utils/compare.ts
chasulin: Oh just kidding actually. sliderValue is the magic #. hmmmm

src/common/utils/compare.ts
chasulin: Oo yep. Fixed

PR #1448, opened 2020-09-10
explore v2: misc TODOs

chasulin:
- Addresses todos in ['to fix' list](https://paper.dropbox.com/doc/CAN-Explore--A7XpRu5CTc2KZpwRT96hIb0fAg-RIlOIJkMWHcX0AmXQA5PW)

Still left todo:
- [x] Lock background scroll
- [x] QA
- [x] Edit checkbox color

### DIFF
--- a/src/components/Explore/LocationSelector.tsx
+++ b/src/components/Explore/LocationSelector.tsx
@@ -8,11 +8,28 @@ import { Location } from 'common/locations';
 import * as Styles from './LocationSelector.style';
 import AutocompleteLocations from 'components/AutocompleteLocations';
 
+import FormControlLabel from '@material-ui/core/FormControlLabel';
+import {
+  NormalizeCheckbox,
+  NormalizeDataContainer,
+  NormalizeSubLabel,
+} from './Explore.style';
+
 const LocationSelector: React.FC<{
   locations: Location[];
   selectedLocations: Location[];
   onChangeSelectedLocations: (newLocations: Location[]) => void;
-}> = ({ locations, selectedLocations, onChangeSelectedLocations }) => {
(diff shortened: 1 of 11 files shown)


src/components/Explore/LocationSelector.tsx
```diff
@@ -1,18 +1,32 @@
...
 
+import FormControlLabel from '@material-ui/core/FormControlLabel';
+import * as ExploreStyles from './Explore.style';
+
 const LocationSelector: React.FC<{
   locations: Location[];
   selectedLocations: Location[];
   onChangeSelectedLocations: (newLocations: Location[]) => void;
-}> = ({ locations, selectedLocations, onChangeSelectedLocations }) => {
+  hasMultipleLocations: boolean;
```
mikelehen, line 17: Rather than passing this into the component separately, couldn't we just do `const hasMultipleLocations = selectedLocations.length >= 1` or something?

src/components/Map/USACountyMap.js
```diff
@@ -9,7 +9,7 @@ import { REVERSED_STATES } from 'common';
 import { useSummaries } from 'common/location_summaries';
 import { ScreenshotReady } from 'components/Screenshot';
 
-function getStateCode(stateName) {
+export function getStateCode(stateName) {
```
mikelehen, line 12: nit: Would probably be better to move this into util/common/locations.ts if it's going to be a shared utility function.  But feel free to punt.

mikelehen (approved):
Couple nits, but looks good!

src/components/Explore/LocationSelector.tsx
chasulin: yep!

PR #1449, opened 2020-09-10
Use Firestore to store share parameters so we can create predictable URLs.

mikelehen:
This moves us away from using query params for compare sharing and instead uses Firestore to store your compare parameters when you share the table, so we can create predictable URLs.  This is necessary so we can pre-generate index.html pages that have meta tags that point to the corresponding share image URL.

One consequence of this change is we don't generate the share URL until the user clicks on the share / save buttons, so I had to rework ShareButtonGroup so you don't necessarily need to know the URLs up front but can instead pass in functions that generate a URL (asynchronously) and then return it.  The code's a bit more complicated and now there's a small delay when you click the share button. 😢  I may see if I can improve that by showing the buttons even before we have the right URL.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -42,7 +42,7 @@ export default function App() {
             <Switch>
               <Route exact path="/" component={HomePage} />
               <Route exact path="/alert_signup" component={HomePage} />
-              <Route exact path="/compare" component={HomePage} />
+              <Route exact path="/compare/:sharingId?" component={HomePage} />
 
               <Route
                 exact
@@ -68,7 +68,7 @@ export default function App() {
               />
               <Route
                 exact
-                path="/us/:stateId/compare"
+                path="/us/:stateId/compare/:sharingId?"
                 component={LocationPage}
               />
               <Route
@@ -83,7 +83,7 @@ export default function App() {
               />
               <Route
                 exact
-                path="/us/:stateId/county/:countyId/compare"
+                path="/us/:stateId/county/:countyId/compare/:sharingId?"
                 component={LocationPage}
               />
               {/* /state/ routes are deprecated but still supported. */}
@@ -165,9 +165,6 @@ export default function App() {
               <Route path="/internal/all" component={AllStates} />
 
               {/** Internal endpoint for comparing API snapshots. */}
-              <Route path="/compare">
-                <Redirect to="/internal/compare" />
-              </Route>
               <Route path="/internal/compare" component={CompareSnapshots} />
 
               {/** Internal endpoints we use to generate the content that we
--- a/src/App.js
+++ b/src/App.js
@@ -42,7 +42,11 @@ export default function App() {
             <Switch>
               <Route exact path="/" component={HomePage} />
               <Route exact path="/alert_signup" component={HomePage} />
-              <Route exact path="/compare/:sharingId?" component={HomePage} />
+              <Route
+                exact
+                path="/compare/:compareShareId?"
+                component={HomePage}
+              />
 
               <Route
                 exact
@@ -68,7 +72,7 @@ export default function App() {
               />
               <Route
                 exact
-                path="/us/:stateId/compare/:sharingId?"
+                path="/us/:stateId/compare/:compareShareId?"
                 component={LocationPage}
               />
               <Route
@@ -83,7 +87,7 @@ export default function App() {
               />
               <Route
                 exact
-                path="/us/:stateId/county/:countyId/compare/:sharingId?"
+                path="/us/:stateId/county/:countyId/compare/:compareShareId?"
                 component={LocationPage}
               />
               {/* /state/ routes are deprecated but still supported. */}
--- a/src/App.js
+++ b/src/App.js
@@ -29,6 +29,7 @@ import ScrollToTop from 'components/ScrollToTop';
 import theme from 'assets/theme';
 import { getFeedbackSurveyUrl } from 'components/Banner';
 import ExternalRedirect from 'components/ExternalRedirect';
+import HandleRedirectTo from 'components/HandleRedirectTo/HandleRedirectTo';
 
 export default function App() {
   return (
@@ -37,6 +38,7 @@ export default function App() {
         <StylesProvider injectFirst>
           <CssBaseline />
           <BrowserRouter>
+            <HandleRedirectTo />
             <ScrollToTop />
             <AppBar />
             <Switch>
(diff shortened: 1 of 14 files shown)


src/App.js
```diff
@@ -165,9 +165,6 @@ export default function App() {
               <Route path="/internal/all" component={AllStates} />
 
               {/** Internal endpoint for comparing API snapshots. */}
-              <Route path="/compare">
-                <Redirect to="/internal/compare" />
-              </Route>
```
mikelehen, line 170: This redirect was stale and no longer worked anyway.

src/components/Compare/CompareMain.tsx
```diff
@@ -20,7 +25,18 @@ import {
...
+const firestore = getFirebase().firestore();
+const paramsCollection = firestore.collection('compare-shared-params');
+const nextIdDocRef = paramsCollection.doc('__nextId');
+
+// For filters (0, 50, and 99 are numerical values required by MUI Slider component):
+const scopeValueMap = {
+  [GeoScopeFilter.NEARBY]: 0,
+  [GeoScopeFilter.STATE]: 50,
+  [GeoScopeFilter.COUNTRY]: 99,
+};
```
mikelehen, line 39: FYI- I moved this stuff to file-level constants because I use them in `useEffect()` and didn't want to have to list them as dependencies that should re-run the effect.

src/components/Compare/CompareMain.tsx
```diff
@@ -30,73 +46,38 @@ const CompareMain = (props: {
   isHomepage?: boolean;
   currentCounty?: any;
   stateId?: string;
-  autoScroll?: boolean;
```
mikelehen, line 33: FYI- The way autoScroll worked before was a little broken.  It would basically re-run on every render of the parent component, which wasn't desirable.  So I reworked it to use an effect, and just made the component automatically looks for /compare in the URL and scrolls itself into view if it's there (so `autoScroll` went away since the parent component doesn't need to pass it anymore).

src/components/Compare/CompareMain.tsx
```diff
@@ -30,73 +46,38 @@ const CompareMain = (props: {
...
+          left: 0,
+          top: tableRef.current.offsetTop - scrollOffset,
+          behavior: 'smooth',
+        });
+      }
+    }, 250);
+  }, [props.isHomepage, tableRef]);
 
   const location = useLocation();
-  const queryParams = QueryString.parse(location.search, {
```
mikelehen, line 46: I ripped out all the query param stuff in favor of the Firestore stuff.  😢  We can bring it back if we need / want it in the future.

src/components/ShareButtons/ShareButtonGroup.tsx
```diff
@@ -10,36 +10,60 @@ import CopyLinkButton from './CopyLinkButton';
 import { ShareButton, SocialButtonsContainer } from './ShareButtons.style';
 
 const ShareImageButtons: React.FC<{
-  imageUrl: string;
+  imageUrl: string | (() => Promise<string>);
   imageFilename: string;
-  url: string;
+  url: string | (() => Promise<string>);
```
mikelehen, line 15: I had to rework this so that url and imageUrl can be created lazily and asynchronously (we don't generate them until you actually click the share / save buttons).  So you can either pass a string or a function that runs and returns a `Promise<string>` now...

src/components/ShareButtons/ShareButtonGroup.tsx
```diff
@@ -10,36 +10,60 @@ import CopyLinkButton from './CopyLinkButton';
...
+  // Turn url / imageUrl into asynchronous getters if they aren't already.
+  const getUrl = typeof url === 'string' ? () => Promise.resolve(url) : url;
+  const getImageUrl =
+    typeof imageUrl === 'string' ? () => Promise.resolve(imageUrl) : imageUrl;
+
+  const [socialSharingProps, setSocialSharingProps] = useState<{
+    url: string;
+    quote: string;
+    socialIconSize: number;
+  } | null>(null);
```
mikelehen, line 28: Since we don't synchronously have the URL anymore I wrapped this up in a state variable that we set once we asynchronously generate the URL after the user clicks the share button.

chasulin (approved):

src/components/Compare/CompareMain.tsx
```diff
@@ -44,7 +46,6 @@ const CompareMain = (props: {
   isModal?: Boolean;
   locationsViewable: number;
   isHomepage?: boolean;
-  currentCounty?: any;
```
mikelehen, line 31: Note that `currentCounty` seemed to always be in-sync with `county` so I opted to populate it within the component (see line 78 below) instead of having the container component pass it in.  Let me know if I'm missing something.

src/App.js
```diff
@@ -37,6 +38,7 @@ export default function App() {
         <StylesProvider injectFirst>
           <CssBaseline />
           <BrowserRouter>
+            <HandleRedirectTo />
```
mikelehen, line 41: New component for handling ?redirectTo=... in the querystring and performing the desired redirect.

src/common/urls.ts
```diff
@@ -14,35 +15,31 @@ const SHARING_ID = sharingId();
...
- * - Uses window.origin to determine the base URL, so it should match localhost / staging / prod / etc.
+ * Generates the path for the homepage or a (state / county) location page.
  */
-function getPageBaseUrl(
-  stateId: string | undefined,
-  county: County | undefined,
-): string {
-  const origin = window.location.origin;
-  let shareURL = origin; // home page
+function getPagePath(stateId?: string, county?: County): string {
```
mikelehen, line 20: Did some refactoring and created this getPagePath() helper for generating the redirect path for shared compare URLs...

src/components/Compare/CompareMain.tsx
```diff
@@ -103,10 +113,10 @@ const CompareMain = (props: {
     getLocationPageViewMoreCopy(geoScope, countyTypeToView, props.stateName);
   const locationPageLocationsForCompare = getLocationPageLocations();
 
-  const locations = props.isHomepage
+  const locations = !stateId
```
mikelehen, line 116: I switched to using `!stateId` because to generate the share image I always leave isHomepage set to false since we don't know if compare was shared from the homepage or not.

src/components/Compare/CompareMain.tsx
chasulin: makes sense!

PR #1472, opened 2020-09-14
Compare modal bug fix

chasulin:
- Fixes bug that was causing some county-rows to appear 3x in modal (image # 1)
- Edits modal locations to exclude inline current-county-row if current county is ranked # 1 (causing consecutive repeating rows in image # 2) 

image

Screen Shot 2020-09-14 at 1 03 05 PM


pnavarrc (approved):
Looks good!

Issue #1471, opened 2020-09-14
Trends claims to start out "normalized" (per 100k) when there is only one county selected


urbanfly:
## summary
When first viewing the "Trends" (`Explore.tsx`), the sub-title is "CASES PER 100K POPULATION SINCE MARCH 1ST IN {county}".

## expected result
The chart is not actually "normalized" and the Y-axis appears to show >100/100K for an area that should be ~15/100K.

## actual result
The sub-title should not claim to be "per 100K" until a 2nd location is selected.

## environment
web; Chrome (but everywhere?)

## link
https://covidactnow.org/us/ga/county/cobb_county/?s=1033220

## screenshot
![image](https://user-images.githubusercontent.com/4717711/93085994-6af7ab00-f664-11ea-914a-9833babd935d.png)

urbanfly:
Appears to be [this line](https://github.com/covid-projections/covid-projections/blob/develop/src/components/Explore/Explore.tsx#L50) introduced with #1406 .

I think it should start out as false and be automatically set within `onChangeSelectedLocations` based on whether there are 1 or more locations selected.

urbanfly:
@mikelehen @pnavarrc @chasulin

urbanfly:
@pnavarrc should this issue be closed now or is the policy to close it only after it is merged to master and/or deployed to production site?

PR #1476, opened 2020-09-14
default to not-normalized

urbanfly:
When the component loads, there is only a single location and the data is not normalized.

When a 2nd location is selected, normalization is automatically enabled.

When the 2nd location is removed, normalization is disabled.

Fixes issue #1471

### DIFF
--- a/src/components/Explore/Explore.tsx
+++ b/src/components/Explore/Explore.tsx
@@ -64,14 +64,18 @@ const Explore: React.FunctionComponent<{
   ]);
 
   const onChangeSelectedLocations = (newLocations: Location[]) => {
-    const selectedLocations = uniq([currentLocation, ...newLocations]);
+    const changedLocations = uniq([currentLocation, ...newLocations]);
 
-    // if there is multiple locations, automatically normalize
-    // if there is only one location, no need to normalize
-    setNormalizeData(selectedLocations.length > 1);
+    if (selectedLocations.length > 1 && changedLocations.length === 1) {
+      // if switching from multiple to a single location, disable normalization
+      setNormalizeData(false);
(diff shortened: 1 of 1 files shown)


src/components/Explore/Explore.tsx
```diff
@@ -64,8 +64,14 @@ const Explore: React.FunctionComponent<{
   ]);
 
   const onChangeSelectedLocations = (newLocations: Location[]) => {
+    const selectedLocations = uniq([currentLocation, ...newLocations]);
```
pnavarrc, line 67: Consider renaming this variable so it has a different name than the state variable `selectedLocations`. I know that the local variable will take precedence, but a different name will make that more obvious.

pnavarrc (changes requested):
Hi @urbanfly, thanks for your contribution! It looks generally good, except that it overwrites the user preference if they have multiple locations and they don't want to normalize. For example, if I have Kansas and Washington enabled, and then uncheck normalization, and then add Texas, it will enable normalization back again, overwriting my preference. To summarize:
- The default value for a single location should be not normalized
- The default value for more than one location should be normalized
- The normalization setting should not change if the user adds more locations, unless when changing from one to two or from two to one.

Let me know if that makes sense and if you need help!

urbanfly:
Pinging @pnavarrc @chasulin since they are the authors of the related PRs #1406 and #1433 

Pinging @mikelehen since he is the approver of the related PR.

urbanfly:
@pnavarrc Thanks for the feedback. I understand and I've made the adjustments.

pnavarrc (approved):
Looks good, thanks @urbanfly !

PR #1484, opened 2020-09-15
Reduce metric-related code duplication in ChartsHolder, LocationPageHeader, and LocationHeaderStats

mikelehen:
There were a few places where we repeated very similar code 5 times, once for each metric, and I simplified it to just map over ALL_METRICS.

### DIFF
--- a/src/components/LocationPage/ChartBlock.tsx
+++ b/src/components/LocationPage/ChartBlock.tsx
@@ -26,7 +26,6 @@ function ChartBlock(props: {
   metric: Metric;
   projections: Projections;
   stateId: string;
-  countyId?: string;
 }) {
   const { projections, metric, isMobile } = props;
   const projection: Projection = projections.primary;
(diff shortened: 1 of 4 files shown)


src/components/LocationPage/ChartBlock.tsx
```diff
@@ -25,44 +25,38 @@ function ChartBlock(props: {
   shareButtonProps: { [key: string]: any };
   metric: Metric;
   projections: Projections;
-  data?: any;
```
mikelehen, line 28: I did some minor refactoring (elevated projections, metric, isMobile to local variables) but the only major change here is removing `data` and instead relying on `projections.hasMetric(metric)` to decide if there is data.

src/components/LocationPage/ChartBlock.tsx
```diff
@@ -25,44 +25,38 @@ function ChartBlock(props: {
...
   stateId: string;
-  countyId?: string;
```
mikelehen, line 30: unused

src/components/SummaryStats/LocationHeaderStats.tsx
```diff
@@ -61,7 +61,6 @@ const SummaryStat = ({
   onClick,
   beta,
   condensed,
-  flipSignalStatusOrder,
```
mikelehen, line 64: unused

chasulin (approved):
Sweet!

PR #1490, opened 2020-09-16
Donate page

chasulin:
- Adds /donate
- Adds donation banner to homepage

[Figma](https://www.figma.com/file/W0YTE0cw3MAxyef71yJJqs/Donations-and-other-banners?node-id=53%3A8)

### DIFF
--- a/src/screens/Donate/Donate.tsx
+++ b/src/screens/Donate/Donate.tsx
@@ -1,0 +1,27 @@
+import React from 'react';
+import styled from 'styled-components';
+
+const Wrapper = styled.div`
+  height: 1100px;
+`;
+const Donate = () => {
+  return (
+    <Wrapper>
+      <iframe
+        title="Donation embed"
+        src="https://givebutter.com/embed/c/hjgW7y"
+        width="100%"
+        height="100%"
+        style={{ maxWidth: '601px' }}
+        name="givebutter"
+        frameBorder="0"
+        scrolling="no"
+        seamless
+        data-allowPaymentRequest
+      ></iframe>
+      <script src="https://givebutter.com/js/widget.js"></script>
+    </Wrapper>
+  );
+};
+
+export default Donate;
--- a/src/screens/Donate/Donate.tsx
+++ b/src/screens/Donate/Donate.tsx
@@ -91,12 +91,14 @@ const Donate = () => {
             <SectionHeader>Our roadmap</SectionHeader>
             {roadmapListContent.map((section, i) => {
               return (
-                <ListSection key={section.header}>
+                <Fragment>
                   <ListHeader>{section.header}</ListHeader>
-                  {section.listItems.map((listItem, i) => {
-                    return <li key={`list-item-${i}`}>{listItem}</li>;
-                  })}
-                </ListSection>
+                  <ListSection key={section.header}>
+                    {section.listItems.map((listItem, i) => {
+                      return <li key={`list-item-${i}`}>{listItem}</li>;
+                    })}
+                  </ListSection>
+                </Fragment>
               );
             })}
           </ContentSection>
--- a/src/screens/Donate/Donate.tsx
+++ b/src/screens/Donate/Donate.tsx
@@ -54,6 +54,7 @@ const Donate = () => {
           style={{ maxWidth: '601px' }}
           name="givebutter"
           frameBorder="0"
+          scrolling="auto"
           seamless
           data-allowPaymentRequest
         ></iframe>
--- a/src/screens/Donate/Donate.tsx
+++ b/src/screens/Donate/Donate.tsx
@@ -1,6 +1,5 @@
 import React, { Fragment } from 'react';
 import { Link } from 'react-router-dom';
-import { Grid } from '@material-ui/core';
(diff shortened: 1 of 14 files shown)


src/screens/HomePage/HomePage.js
```diff
@@ -6,23 +6,21 @@ import EnsureSharingIdInUrl from 'components/EnsureSharingIdInUrl';
...
-  FeaturedHeader,
-  PartnerSection,
-  PartnerHeader,
   SearchBarThermometerWrapper,
   SectionWrapper,
+  BannerContainer,
 } from './HomePage.style';
 import { SelectorWrapper } from 'components/Header/HomePageHeader.style';
-
+import DonationBanner from 'components/Banner/DonationBanner';
```
pnavarrc, line 23: nit: Consider exporting `DonationBanner` in `components/Banner/index.ts` and import from `components/Banner` here as we do it with the `FeedbackBanner`.

The idea of the `index.ts` file in component folders is to differentiate public vs. private components in component folders. For example, in `components/Expore/index.ts` we export `ExploreChart`, but not `SingleLocationChart` or `MultipleLocationsChart`. In this case, it doesn't matter much, so feel free to ignore it!

src/screens/Donate/Donate.tsx
```diff
@@ -0,0 +1,111 @@
...
+          seamless
+          data-allowPaymentRequest
+        ></iframe>
+        <script src="https://givebutter.com/js/widget.js"></script>
+      </EmbedWrapper>
+      <ContentWrapper container justify="center">
+        <Grid item xs={12} sm={6}>
+          <ContentSection>
+            <Icon icon={loveLetter} />
+            <SectionHeader>How your donation helps</SectionHeader>
```
pnavarrc, line 67: Consider using `h1` and `h2` components for the headers instead of styled paragraphs. Assistive technologies such as screen readers rely on the structure of the page (some screen readers have keyboard shortcuts to navigate headers). See https://web.dev/headings-and-landmarks/

The `Typography` component can have a style (let's say `h4`) while using a different component (`h1`) which makes styling a bit easier. See [About.tsx#L54](https://github.com/covid-projections/covid-projections/blob/0074098e32a22d2e75ef53e054866947329b79c1/src/screens/About/About.tsx#L54) for an example of how to use semantic headers.

I think that the main "title" on the page is the embed title, so you probably want to use `h2` for "How your donation helps" and `h3` for "Improve our Data Quality".

Note: We don't use the correct tags in many places on our site, I would like to fix it eventually. We can skip this if we need to ship this quickly.

src/screens/Donate/Donate.tsx
```diff
@@ -0,0 +1,111 @@
...
+          <ContentSection>
+            <Icon icon={testTube} />
+            <SectionHeader>Our theory of change</SectionHeader>
+            <BodyCopy>
+              By providing clearer, more accessible, locally-relevant and
+              science-based guidance that the general public and decision-makers
+              can understand, rally behind and communicate, we can advance a
+              more unified and more effective response to COVID in the U.S.
+            </BodyCopy>
+            <a href="/about" target="_blank" rel="noopener noreferrer">
```
pnavarrc, line 84: Consider using the `ExternalLink` component

src/screens/Donate/Donate.tsx
```diff
@@ -0,0 +1,111 @@
...
+    ],
+  },
+];
+
+const Donate = () => {
+  return (
+    <Fragment>
+      <EmbedWrapper>
+        <iframe
+          title="Donation embed"
```
pnavarrc, line 50: Consider changing the title to something more meaningful `Donate to Covid Act Now on GiveButter`

src/screens/Donate/Donate.tsx
```diff
@@ -0,0 +1,111 @@
...
+            </a>
+          </ContentSection>
+        </Grid>
+        <Grid item xs={12} sm={6}>
+          <ContentSection>
+            <Icon icon={deliveryTruck} />
+            <SectionHeader>Our roadmap</SectionHeader>
+            {roadmapListContent.map((section, i) => {
+              return (
+                <ListSection>
```
pnavarrc, line 95: The items here need a unique key, also the `li` items below (see the React warnings in the console)

src/screens/Donate/Donate.tsx
```diff
@@ -0,0 +1,111 @@
...
+import {
+  EmbedWrapper,
+  ContentWrapper,
+  SectionHeader,
+  ContentSection,
+  ListHeader,
+  BodyCopy,
+  ListSection,
+} from 'screens/Donate/Donate.style';
+import PartnersSection from 'screens/HomePage/PartnersSection';
```
pnavarrc, line 16: nit: Consider moving `PartnersSection` to its own folder in `components` and import it from both `HomePage` and `Donate` screens.

pnavarrc (approved):
Looks good! Added a few optional notes

src/screens/Donate/Donate.tsx
chasulin: great note. going to be mindful of this going forward.

PR #1505, opened 2020-09-17
explore: optimize margin size

pnavarrc:
This PR optimizes the labels for the explore chart when there are multiple locations selected. There are 3 main cases:

- Desktop: We show the full name, hoping that it will fit (it will overflow if not)
- Mobile (states only): We show only the state code, so we need a narrow right margin
- Mobile (with counties): We show a cropped version of the county name and expand the margin so it will fit.

**Desktop**

**Mobile (states only)**

**Mobile (states + county)**

### DIFF
--- a/src/components/Explore/Explore.tsx
+++ b/src/components/Explore/Explore.tsx
@@ -35,6 +35,33 @@ import {
 } from './utils';
 import * as Styles from './Explore.style';
 
+const MARGIN_SINGLE_LOCATION = 20;
+const MARGIN_STATE_CODE = 60;
+const MARGIN_COUNTY = 120;
+
+function getMarginRight(
+  showLabels: boolean,
+  shortLabels: boolean,
+  seriesList: Series[],
+) {
+  const maxLabelLength =
+    max(seriesList.map(series => getLabelLength(series, shortLabels))) || 0;
+
+  // We only show the labels when multiple locations are selected. If only
+  // states are selected, we only need space for the state code, if at least
+  // one county is selected, we need more space.
+  return showLabels
+    ? maxLabelLength > 2
+      ? MARGIN_COUNTY
+      : MARGIN_STATE_CODE
+    : MARGIN_SINGLE_LOCATION;
+}
+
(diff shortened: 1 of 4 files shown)


src/components/Explore/Explore.tsx
```diff
@@ -104,6 +105,18 @@ const Explore: React.FunctionComponent<{
     setNormalizeData,
   };
 
+  const seriesLabels = chartSeries.map(series =>
+    getSeriesLabel(series, isMobileXs),
+  );
+
+  const maxLabelLength = max(seriesLabels.map(label => label.length)) || 0;
+
+  const adjustedMarginRight = hasMultipleLocations
```
mikelehen, line 114: This might be worth a comment explaining this logic.

src/components/Explore/utils.ts
```diff
@@ -358,9 +361,25 @@ export function weeksAgo(dateFrom: Date, dateTo: Date) {
...
-    : location.state;
+  const { county: countyName, state_code, state: stateName } = location;
+  const stateCode = state_code.toUpperCase();
+  return countyName
+    ? `${getAbbreviatedCounty(countyName)} ${stateCode}`
+    : stateName;
+}
+
+function truncateCountyName(countyName: string) {
+  return countyName.length < 8
```
mikelehen, line 372: nit: I think this should probably be `<= 8` since if it's 8 characters long, it doesn't make sense to truncate the last character and replace it with …

We could almost maybe even do `<=9` since '…' is wider than a normal letter. 🤷

mikelehen (approved):
LGTM with nits.  It pains me a bit that we have so waste so much space in the "Mobile (states + county)" case...  feels like we should probably move the labels either above or below the chart for better screen real-estate usage.

Also, FWIW- In the "Mobile (states + county)" case, I think it'd probably be fine to drop the state code for counties, in order to save a bit of space.  It'd be pretty rare that your comparing counties with the same name in different states.  Though not impossible.

But I think this is fine for now.

pnavarrc:
Good suggestions, I updated the code and will merge in a bit!

PR #1517, opened 2020-09-21
Rework compare sharing and implement explore sharing.

mikelehen:
FYI- This is split into two commits.  
1. A refactor / reimplementation of compare sharing to make it more general-purpose. The second one uses that infrastructure to implement explore sharing. It mostly has no functional changes though it does point at /share/xyz/ URLs instead of /compare/XYZ URLs which lets us unify the index.html pages we generate to just /share/xyz for compare, explore, and any future components that implement sharing.
2. Actual sharing implementation for the explore component (based on above refactor).

NOTE: Preexisting shared explore URLs should continue to work, but going forward instead of a URL like /us/wa/explore/cases it'll end up being /share/123?redierctTo=/us/wa/explore/123.

With this PR, the following should work:
1. You make an arbitrary explore chart and select copy link. You should get a URL something like: http://localhost:3000/share/15/?redirectTo=%2Fus%2Fwa%2Fexplore%2F15
2. Going to that URL should work (it'll redirect you to the right page and re-load the explore params).
3. If you run `firebase serve` from the `firebase` directory, you can go to http://localhost:5000/0-123/share/15.png or http://localhost:5000/0-123/share/15/export.png to get the correct share or download images.

The following won't work locally:
1. Actually sharing to Facebook / Twitter will insert a prod URL that doesn't work (since we haven't deployed these changes yet) but should look similar to the localhost URLs that do work.
2. The "Save" button won't work since it tries to load from prod, but if you look at the URL in the network tab it should be similar to the localhost URL that does work.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -39,7 +39,6 @@ export default function App() {
         <StylesProvider injectFirst>
           <CssBaseline />
           <BrowserRouter>
-            <HandleRedirectTo />
             <ScrollToTop />
             <AppBar />
             <Switch>
@@ -47,7 +46,7 @@ export default function App() {
               <Route exact path="/alert_signup" component={HomePage} />
               <Route
                 exact
-                path="/compare/:compareShareId?"
+                path="/compare/:sharedComponentId?"
                 component={HomePage}
               />
 
@@ -76,7 +75,7 @@ export default function App() {
               />
               <Route
                 exact
-                path="/us/:stateId/compare/:compareShareId?"
+                path="/us/:stateId/compare/:sharedComponentId?"
                 component={LocationPage}
               />
               <Route
@@ -91,7 +90,7 @@ export default function App() {
               />
               <Route
                 exact
-                path="/us/:stateId/county/:countyId/compare/:compareShareId?"
+                path="/us/:stateId/county/:countyId/compare/:sharedComponentId?"
                 component={LocationPage}
               />
               {/* /state/ routes are deprecated but still supported. */}
@@ -184,6 +183,7 @@ export default function App() {
               exports images. */}
               <Route path="/internal/export-image/" component={ExportImage} />
 
+              <HandleRedirectTo />
               <Route path="/*">
                 <Redirect to="/" />
               </Route>
--- a/src/App.js
+++ b/src/App.js
@@ -70,7 +70,7 @@ export default function App() {
               />
               <Route
                 exact
-                path="/us/:stateId/explore/:chartId"
+                path="/us/:stateId/explore/:sharedComponentId?"
                 component={LocationPage}
               />
               <Route
@@ -85,7 +85,7 @@ export default function App() {
               />
               <Route
                 exact
-                path="/us/:stateId/county/:countyId/explore/:chartId"
+                path="/us/:stateId/county/:countyId/explore/:sharedComponentId?"
                 component={LocationPage}
               />
               <Route
--- a/src/common/locations.ts
+++ b/src/common/locations.ts
@@ -124,17 +124,15 @@ export function getLocationNameForFips(fips: string): string {
   }
 }
 
-export function getLocationUrlForFips(fips: string): string {
-  // TODO(pablo): Make sure that these urls are uniform, here, the state
-  // URL ends on / but not the county one
+export function getRelativeUrlForFips(fips: string): string {
   if (isStateFips(fips)) {
     const state = findStateByFips(fips);
-    return `https://covidactnow.org/us/${state.state_code.toLowerCase()}/`;
+    return `/us/${state.state_code.toLowerCase()}/`;
   } else {
     const county = findCountyByFips(fips);
-    return `https://covidactnow.org/us/${county.state_code.toLowerCase()}/county/${
+    return `/us/${county.state_code.toLowerCase()}/county/${
       county.county_url_name
-    }`;
+    }/`;
   }
 }
 
--- a/src/common/sharing.ts
+++ b/src/common/sharing.ts
@@ -1,0 +1,141 @@
+import { getFirebase } from 'common/firebase';
+import { isEqual, pickBy } from 'lodash';
+import { useParams } from 'react-router-dom';
+import { useEffect, useState } from 'react';
+
+const firestore = getFirebase().firestore();
+const collection = firestore.collection('shared-component-params');
+const nextIdDocRef = collection.doc('__nextId');
+
+type Params = { [key: string]: any };
+
+type ComponentName = 'compare' | 'explore';
+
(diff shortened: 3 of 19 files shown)


src/App.js
```diff
@@ -39,15 +39,14 @@ export default function App() {
         <StylesProvider injectFirst>
           <CssBaseline />
           <BrowserRouter>
-            <HandleRedirectTo />
```
mikelehen, line 42: This is a component I created before to handle ?redirectTo=... query params for the sake of making https://covidactnow.org/share/xyz/?redirectTo=/us/wa/compare/xyz work (since we need to point at /share/xyz/ to get the right meta tags).

I had to move it lower in the page to work around an issue where react-router-dom sometimes ignored it. I think there was a conflict between whether the redirect or one of our routes took effect first or something. 🤷

src/App.js
```diff
@@ -39,15 +39,14 @@ export default function App() {
...
             <ScrollToTop />
             <AppBar />
             <Switch>
               <Route exact path="/" component={HomePage} />
               <Route exact path="/alert_signup" component={HomePage} />
               <Route
                 exact
-                path="/compare/:compareShareId?"
+                path="/compare/:sharedComponentId?"
```
mikelehen, line 49: `sharedComponentId` is now a "well-known" param used by e.g. the "useSharedComponentParams()" hook.

src/common/locations.ts
```diff
@@ -124,17 +124,15 @@ export function getLocationNameForFips(fips: string): string {
   }
 }
 
-export function getLocationUrlForFips(fips: string): string {
-  // TODO(pablo): Make sure that these urls are uniform, here, the state
-  // URL ends on / but not the county one
+export function getRelativeUrlForFips(fips: string): string {
```
mikelehen, line 127: In order to construct the /share/xyz/?redirectTo=/us/<state>/explore/xyz URLs I had to change this to be relative instead of the full URL.

I also fixed your TODO.

src/common/utils/model.ts
```diff
@@ -92,7 +92,7 @@ export function fetchAllCountyProjections(snapshotUrl: string | null = null) {
   return cachedCountiesProjections[key];
 }
 
-export function useProjections(location: string, county = null) {
+export function useProjections(location: string, county?: County) {
```
mikelehen, line 95: This wasn't typed properly before which caused some TypeScript errors from where I was calling it.

src/components/Compare/CompareMain.tsx
```diff
@@ -146,72 +145,41 @@ const CompareMain = (props: {
     countyTypeToView,
     viewAllCounties,
     geoScope,
+    stateId,
+    countyId: county?.full_fips_code,
```
mikelehen, line 149: I used to handle stateId / countyId specially below since they could be undefined, but instead I reworked the storeSharedComponentParams() function to deal with undefined better (Firestore doesn't support storing `undefined` explicitly).

src/components/Explore/Explore.tsx
```diff
@@ -63,16 +74,17 @@ function getLabelLength(series: Series, shortLabel: boolean) {
 }
 
 const Explore: React.FunctionComponent<{
-  projection: Projection;
-  chartId?: string;
-}> = ({ projection, chartId }) => {
-  const { locationName, fips } = projection;
+  fips: string;
```
mikelehen, line 77: The only thing we really needed from the projection was the fips, so I simplified this.

And we don't use chartId anymore, since we grab the sharedComponentId directly below.

src/components/Explore/Explore.tsx
```diff
@@ -106,6 +120,28 @@ const Explore: React.FunctionComponent<{
     setSelectedLocations(changedLocations);
   };
 
+  const exploreRef = useRef<HTMLDivElement>(null);
```
mikelehen, line 123: I reworked the way we make /explore/ URLs auto-scroll to the explore component so that it matches the way compare works.  The way we did it before in ChartsHolder couldn't be used anymore because it was based on chartId instead of the URL.

Also, this should automatically work if we end up adding explore to the homepage.

src/screens/internal/ShareImage/ChartShareImage.style.ts
```diff
@@ -20,6 +20,15 @@ export const Title = styled.div`
...
   text-transform: capitalize;
+  width: 400px;
+`;
+
+export const ExploreTitle = styled(Title)`
+  font-size: 18px;
+  line-height: 19px;
+  text-transform: capitalize;
+  height: 75px;
+  overflow: hidden;
```
mikelehen, line 31: I ended up reworking the styling of the title to make long / multi-line titles (with lots of locations) work better.

src/screens/internal/ShareImage/ExploreChartExportImage.tsx
```diff
@@ -1,85 +1,86 @@
...
+    return null;
   }
 
   const chartHeight = 415;
 
-  let url = `https://covidactnow.org/us/${stateId}`;
-  if (countyOption) {
-    url += `/county/${countyOption.county_url_name}`;
-  }
+  const url = `https://covidactnow.org/`;
```
mikelehen, line 54: Since there may be multiple locations I just link to the homepage.

pnavarrc (approved):
Looks good! I added a few notes - I think I finally get how this works 😅

src/common/sharing.ts
```diff
@@ -0,0 +1,141 @@
...
+import { useParams } from 'react-router-dom';
+import { useEffect, useState } from 'react';
+
+const firestore = getFirebase().firestore();
+const collection = firestore.collection('shared-component-params');
+const nextIdDocRef = collection.doc('__nextId');
+
+type Params = { [key: string]: any };
+
+type ComponentName = 'compare' | 'explore';
```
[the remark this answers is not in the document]
mikelehen: Good idea, done!

src/common/sharing.ts
```diff
@@ -0,0 +1,141 @@
...
+    storedComponentParams.push(cachedParams);
+  }
+  return cachedParams.id;
+}
+
+// A cache of component params that we have read from Firestore. Ensures that we
+// don't re-fetch the same params multiple times (causing extra delays).
+const fetchedComponentParams: {
+  [id: string]: Promise<Params | undefined>;
+} = {};
```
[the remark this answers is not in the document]
mikelehen: I was trying to make sure that e.g. when compare and explore are both on the page we don't end up needlessly fetching the parameters twice (since both components will use the useSharedComponentParams() hook).

The only way to do this is to cache the promise before it's resolved...  So it's a bit complicated, but I think it makes sense.

PR #1519, opened 2020-09-21
Voter registration banner

chasulin:


src/components/Banner/VoteBanner.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <Button
+        variant="contained"
+        color="primary"
+        disableRipple
+        disableFocusRipple
+      >
+        {cta}
+      </Button>
+    </ExternalLink>
+  );
```
pnavarrc, line 27: I think you are missing a key here 

image

pnavarrc (approved):
Looks great!

PR #1526, opened 2020-09-22
Specify image sizes so maybe Facebook will render a preview?

mikelehen:
Add og:image:{width,height} tags to hopefully help Facebook render previews.

I noticed the [docs](https://developers.facebook.com/docs/sharing/webmasters/images/) say:

> Use og:image:width and og:image:height Open Graph tags: Using these tags will specify the image dimensions to the crawler so that it can render the image immediately without having to asynchronously download and process it.


pnavarrc (approved):
Nice find!

PR #1535, opened 2020-09-23
Add Netlify CMS

chasulin:
- Adds [Netlify CMS](https://www.netlifycms.org/docs/intro/)
- Moves content of first two sections of About page to Netlify


public/admin/config.yml
```diff
@@ -0,0 +1,38 @@
+backend:
```
pnavarrc, line 1: Consider adding a comment at the start linking to the docs, and maybe explaining that the branch option is where the copy updates will go

src/common/utils/netlify.ts
```diff
@@ -0,0 +1,20 @@
...
+export enum PageType {
+  ABOUT,
+}
+
+const pageTypeToFileMap = {
+  [PageType.ABOUT]: aboutPageContent,
+};
+
+// Retrives page-specific json file:
+export function getContentFile(
```
pnavarrc, line 16: Consider renaming to `getPageContent` since it doesn't return a file

src/common/utils/netlify.ts
```diff
@@ -0,0 +1,20 @@
...
+  pageType: PageType,
+): { [fieldName: string]: any } {
+  return pageTypeToFileMap[pageType];
+}
```
pnavarrc, line 20: maybe we could define the interface for the about page here, so we can use it in the `About.tsx` component? Just an idea, maybe for later

src/screens/About/About.tsx
```diff
@@ -52,75 +58,41 @@ const About = ({ children }: { children: React.ReactNode }) => {
...
-            >
-              Harvard Global Health Institute
-            </a>
-            .
-          </Typography>
-          <PartnerLogoGrid />
+
+          {content.partnersContent.map((section: any, i: number) => {
+            return (
+              <Fragment>
```
pnavarrc, line 77: Remember to add a key when mapping over lists so react doesn't complain

src/screens/About/About.tsx
```diff
@@ -52,75 +58,41 @@ const About = ({ children }: { children: React.ReactNode }) => {
...
+                <BodyCopy source={section.copy} linkTarget="_blank" />
```
pnavarrc, line 78: do the links include `rel=noopener noreferrer`? If not we can try to figure this out later

src/screens/About/About.tsx
```diff
@@ -52,75 +58,41 @@ const About = ({ children }: { children: React.ReactNode }) => {
...
+                <Grid
+                  container
+                  spacing={1}
+                  alignItems="center"
+                  justify="center"
+                >
+                  {section.logos.map((logo: any, i: number) => (
+                    <LogoGridItem
```
pnavarrc, line 86: Remember to add a `key` so React doesn't complain

pnavarrc (approved):
Added some notes, looks good!

src/screens/About/About.tsx
chasulin: im going to punt this for now. im not seeing it as a prop in the react-markdown docs. gotta dig around a little more

src/common/utils/netlify.ts
chasulin: going to punt for the time being, but agreed that this would be useful

src/screens/About/About.tsx
```diff
@@ -75,7 +75,7 @@ const About = ({ children }: { children: React.ReactNode }) => {
           {content.partnersContent.map((section: any, i: number) => {
             return (
               <Fragment>
-                <BodyCopy source={section.copy} linkTarget="_blank" />
+                <BodyCopy key={i} source={section.copy} linkTarget="_blank" />
```
pnavarrc, line 78: you will need to put the key in the `Fragment` element, because that's the one repeating under `Wrapper`

src/screens/About/About.tsx
```diff
@@ -87,6 +87,7 @@ const About = ({ children }: { children: React.ReactNode }) => {
                       image={logo.image}
                       url={logo.url}
                       altText={logo.altText}
+                      key={i}
```
pnavarrc, line 90: Consider using a different letter here since both `i` are in scope here (the inner `i` wins, but just to make it explicit)

PR #1536, opened 2020-09-23
analytics: fix GA tracking of page views

pnavarrc:
Our current implementation of Google Analytics has a bug that causes page views to be reported multiple times for each visited page. This doesn't affect the reporting of other events, sessions, or unique users.

In static websites, the GA script automatically tracks page views when it initializes. This doesn't work well with Single Page Applications (SPA) like ours because the `index.html` page is loaded only once, and the JavaScript code updates the content of the page and the URL. For Single Page Applications, we need to record page views manually upon navigation. 

This was implemented for our application, however, it had 2 main problems:

- We were adding a new [listener](https://github.com/ReactTraining/history/blob/master/docs/api-reference.md#historylistenlistener-listener) for changes in `history` whenever `_AppBar` renders. If `_AppBar` were to re-render for some reason, we will add a new listener, leaving previous listeners active.
- The `location` parameter of the listener is an object that might change more often than when the `pathname` changes - it will change for sure when we update the URL to include `?s=snapshotId` (see https://developer.mozilla.org/en-US/docs/Web/API/Location)

The new implementation only triggers page view events when the `pathname` changes.

### Debugging and Testing

You can use the [Chrome GA Debugger Extension](https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna) to see the GA events being sent in the console or change the `debug` option to `true` in [src/components/Analytics/PageviewTracker.tsx#L11](https://github.com/covid-projections/covid-projections/pull/1536/files#diff-18d8e856bb3be0614177aed58c109f78R11).

#### Testing

- Make sure that you have enabled the "Preserve log" setting in the Development console to ensure that the logs persist through navigations.

image

- Click locations and links on the page, confirm that there is only one "page view" event per tracker for each page visited. 
- Compare with the logs shown using the extension above for `covidactnow.org`, it will show multiple page view events (among others).

#### Notes
- We have 2 tracking properties, the default, and the legacy. When debugging, you should see one page view event for each tracker. Trackers collect hits independently of each other and won't show up as duplicate page views.

```
Initializing Google Analytics.
Running command: ga("create", "UA-160622988-1", "auto")
Running command: ga("create", "G-HFCDC7K5G1", {name: "legacy"})
Running command: ga("send", {hitType: "pageview", page: "/us/tx/county/medina_county"})
Running command: ga("legacy.send", {hitType: "pageview", page: "/us/tx/county/medina_county"})
```

- We need to exclude `staging` and `localhost` from GA reporting, we can do this in the code or add a filter on the GA interface. Adding it in GA will remove existing events from the reports as well.


public/index.html
```diff
@@ -58,40 +58,28 @@
...
-        g.shutdown=function(){g("rec",!1)};g.restart=function(){g("rec",!0)};
-        g.log = function(a,b){g("log",[a,b])};
-        g.consent=function(a){g("consent",!arguments.length||a)};
-        g.identifyAccount=function(i,v){o='account';v=v||{};v.acctId=i;g(o,v)};
-        g.clearUserCookie=function(){};
-        g._w={};y='XMLHttpRequest';g._w[y]=m[y];y='fetch';g._w[y]=m[y];
-        if(m[y])m[y]=function(){return g._w[y].apply(this,arguments)};
-        g._v="1.2.0";
-    })(window,document,window['_fs_namespace'],'script','user');
+    (function (m, n, e, t, l, o, g, y) {
```
pnavarrc, line 67: I think we can remove the Fullstory script since the trial expired already

mikelehen: Seems reasonable.  I'd either just do it or throw it on the standup agenda for tomorrow.

mikelehen (approved):
This looks great!

And I confirm (via the GA debugger) that the events seem much more sane. :-)

pnavarrc:
[Fix GA page view tracking. We need to disable default tracking (which doesn't work with React) and properly tack navigations.](https://trello.com/c/dx8KAfwI/456-fix-ga-page-view-tracking-we-need-to-disable-default-tracking-which-doesnt-work-with-react-and-properly-tack-navigations)

PR #1555, opened 2020-09-25
analytics: record events for compare and explore

pnavarrc:


src/components/Compare/Filters.tsx
```diff
@@ -101,14 +108,23 @@ const Filters = (props: {
...
     }
   };
 
   const sliderHandleChange = (event: any, value: any) => {
     if (props.setGeoScope) {
       setSliderValue(value);
       props.setGeoScope(sliderNumberToFilterMap[value]);
+      trackCompareEvent(
+        EventAction.SELECT,
+        `GeoScope: ${trackingGeoScopeLabel[sliderNumberToFilterMap[value]]}`,
```
mikelehen, line 126: FWIW- I think you could do `GeoScopeFilter[sliderNumberToFilterMap[value]]` and it would return 'NEARBY' or whatever and you wouldn't need tracking GeoScopeLabel if we were okay just using the enum names.

Basically, TypeScript enums are a bidirectional map of the enum value<=>enum name

But this is also fine as-is if you just want to ship it.

mikelehen (approved):

pnavarrc:
[[Analytics] Add click events to CAN Compare and CAN Explore](https://trello.com/c/zk42sssV/444-analytics-add-click-events-to-can-compare-and-can-explore)

PR #1558, opened 2020-09-26
analytics: track sort and modal events for compare

pnavarrc:
- Add tracking for sorting and opening modals on Compare
- Restore the `titleCase: false` option (GA is case sensitive - I removed this option thinking that GA would consolidate the events)
- Remove the Full Story script


chasulin (approved):

pnavarrc:
[[Analytics] Add click events to CAN Compare and CAN Explore](https://trello.com/c/zk42sssV/444-analytics-add-click-events-to-can-compare-and-can-explore)

PR #1563, opened 2020-09-28
Removing projection uses

ghop02:
Removes some more unused projection code - will make it easier to move over to new API


mikelehen (approved):
Nice!

PR #1565, opened 2020-09-28
Disable metrics (and make it easier to do so).

mikelehen:
* Disable Test Positivity for Harris County, TX and Daily New Cases for Bexar County, TX.
* Make it more straightforward to disable metrics going forward.
* Unify the explanation text we use when we are missing a metric.


src/common/models/Projection.ts
```diff
@@ -14,6 +14,36 @@ import {
...
+  '25013',
+  '25015',
+  '25017',
+  '25019',
+  '25021',
+  '25023',
+  '25025',
+  '25027',
+  '48113',
+  '48215',
```
mikelehen, line 39: All of the above were preexisting disablements.  I just moved them up here to a top-level constant.

src/common/metrics/case_density.tsx
```diff
@@ -71,8 +71,9 @@ function renderStatus(projections: Projections): React.ReactElement {
   ) {
     return (
       <Fragment>
-        Not enough case data is available to generate{' '}
-        {CaseIncidenceMetric.extendedMetricName.toLowerCase()}.
+        Unable to generate{' '}
+        {CaseIncidenceMetric.extendedMetricName.toLowerCase()}. This could be
+        due to insufficient data.
```
ghop02, line 76: I bet this is out of scope for now, but it would be really great to link to an FAQ or explanation about why there could be insufficient data

ghop02 (approved):

src/common/metrics/case_density.tsx
mikelehen: Yeah...  I thought about making this more descriptive, but a FAQ or something would be great.

PR #1566, opened 2020-09-28
seo: update location page urls

pnavarrc:
https://trello.com/c/Q9qmbNP7/416-update-urls-for-states-and-counties

This PR updates the URL of the location page to include the name and code of the state in the URL to improve our search rankings.

- States: `https://covidactnow.org/us/montana_mt`
- Counties: `https://covidactnow.org/us/montana_mt/county/rosebud_county`


pnavarrc:
[Update URLs for states and counties](https://trello.com/c/Q9qmbNP7/416-update-urls-for-states-and-counties)

src/common/locations.ts
```diff
@@ -1,11 +1,20 @@
...
+  has,
+  partition,
+  isUndefined,
+  toLower,
+} from 'lodash';
 import { assert } from './utils';
 import countyAdjacencyMsa from './data/county_adjacency_msa.json';
 import collegesByFips from './data/colleges_by_fips.json';
 import { REVERSED_STATES } from 'common';
+import { isCounty } from '../components/AutocompleteLocations';
```
mikelehen, line 17: FWIW, I'd be a fan of moving `isCounty` (and probably `isState` and `belongsToState`) into this file.

src/common/locations.ts
```diff
@@ -100,6 +111,13 @@ export function findCountyByFips(fips: string) {
   return undefined;
 }
 
+export function isValidState(fips: string): boolean {
```
mikelehen, line 114: Feel like this should have "fips" in the name...  I was gonna' suggest `isStateFips`, but it looks like we already have that. 😁  So maybe either change `isStateFips` to include this functionality (making sure it's valid) or else this can be `isValidStateFips`.

src/common/locations.ts
```diff
@@ -186,3 +196,48 @@ export function getColleges(fips: string): CollegeData[] {
 export function getStateCode(stateName: string) {
   return REVERSED_STATES[stateName];
 }
+
+const ALL_LOCATIONS = getLocationNames();
```
mikelehen, line 200: FWIW (probably not worth worrying about in this PR)-
1. `getLocationNames` should probably be `getLocations` or similar?
2. I think we should do a pass at some point to try to unify all of our code to work with `State`, `County`, and `Location` data types and try to make sure all related logic is in this file and properly shared across parts of the site.  Right now it's pretty fragmented / ad-hoc.

src/common/locations.ts
```diff
@@ -186,3 +196,48 @@ export function getColleges(fips: string): CollegeData[] {
...
+const [COUNTIES, STATES] = partition(ALL_LOCATIONS, isCounty);
+
+export function getStateByUrlName(stateUrlName: string): Location | undefined {
```
mikelehen, line 203: Should this return `State | undefined` instead?  Or no?

src/common/locations.ts
```diff
@@ -186,3 +196,48 @@ export function getColleges(fips: string): CollegeData[] {
...
+  return STATES.find(
+    state =>
+      toLower(state.state_url_name) === toLower(stateUrlName) ||
+      toLower(state.state_code) === toLower(stateUrlName),
+  );
+}
+
+export function getCountyByUrlName(
+  countyUrlName: string,
+): Location | undefined {
```
mikelehen, line 213: Again, should this return `County | undefined`?

src/common/locations.ts
```diff
@@ -186,3 +196,48 @@ export function getColleges(fips: string): CollegeData[] {
...
+export function getStateByUrlName(stateUrlName: string): Location | undefined {
+  return STATES.find(
+    state =>
+      toLower(state.state_url_name) === toLower(stateUrlName) ||
+      toLower(state.state_code) === toLower(stateUrlName),
+  );
+}
+
+export function getCountyByUrlName(
+  countyUrlName: string,
```
mikelehen, line 212: I think we need to always use the stateUrlName and countyUrlName together, since county URL names aren't unique.  E.g. these both end up showing the same stats right now:

https://covid-projections-git-pablo-update-location-page-url.covidactnow.vercel.app/us/or/county/washington_county?s=1083238
https://covid-projections-git-pablo-update-location-page-url.covidactnow.vercel.app/us/mn/county/washington_county?s=1083238

src/common/locations.ts
```diff
@@ -186,3 +196,48 @@ export function getColleges(fips: string): CollegeData[] {
...
+  );
+}
+
+export function getLocationByUrlParams(
+  stateUrlName: string,
+  countyUrlName?: string,
+): Location | undefined {
+  const location = countyUrlName
+    ? getCountyByUrlName(countyUrlName)
+    : getStateByUrlName(stateUrlName);
```
mikelehen, line 225: Per comment above, I think we need to combine these, not use them separately since county URL names are not unique by themselves.

src/common/locations.ts
```diff
@@ -186,3 +196,48 @@ export function getColleges(fips: string): CollegeData[] {
...
+
+  assert(
+    isUndefined(location),
```
mikelehen, line 228: This assert seems backwards, which makes me think this code must not be used?

src/components/MapSelectors/datasets/us_states_dataset_01_02_2020.json
```diff
@@ -3,318 +3,371 @@
     {
       "state_code": "AK",
       "state": "Alaska",
+      "state_url_name": "alaska-ak",
```
mikelehen, line 6: is it weird that we use dashes in the state_url_name but underscores in county_url_name ?

I don't care much though if this is intentional.

src/screens/Embed/Embed.js
```diff
@@ -67,12 +69,9 @@ function LocationEmbed() {
...
     alarmLevel !== Level.UNKNOWN ? levelInfo.color : COLOR_MAP.GRAY.LIGHT;
 
+  const canonicalUrl = getCanonicalUrl(projections.fips);
   const embedOnClickBaseURL = projections
-    ? `https://covidactnow.org/us/${location.toLowerCase()}${
-        projections.isCounty
-          ? `/county/${projections.county.county_url_name}`
-          : ''
-      }`
+    ? `https://covidactnow.org/us/${canonicalUrl}`
```
mikelehen, line 74: This shouldn't have `/us/` first, right?

src/screens/LocationPage/LocationRouter.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { useParams, Redirect } from 'react-router-dom';
+import { getCountyByUrlName, getStateByUrlName } from 'common/locations';
+import LocationPage from './LocationPage';
+
+interface LocationPageUrlParams {
+  stateId: string;
+  countyId: string | undefined;
+}
+
+const LocationRouter: React.FC = () => {
```
mikelehen, line 11: It seems like maybe this isn't actually used?  Or am I confused?

mikelehen (changes requested):
Mostly looks good, but I think we're not handling non-unique `county_url_name`s properly right now.  I can take another look (and do a bit of testing) once that's fixed.

src/screens/LocationPage/LocationRouter.tsx
pnavarrc: It is, it's the default export in `LocationPage/index.ts` - before this PR, the AppBar was redirecting to the homepage when the state or county didn't exist (it took me a while to discover that), I moved that logic here to make it easier to find... maybe I should move this to `LocationPage.js` instead?

src/common/locations.ts
pnavarrc: moved!

src/screens/LocationPage/LocationRouter.tsx
mikelehen: oops! Yeah, I was confused.  This is probably fine, but I'd be fine moving it into LocationPage.js too. Up to you.

src/common/locations.ts
pnavarrc: agreed, we should try to unify that at some point

src/common/locations.ts
```diff
@@ -100,10 +102,19 @@ export function findCountyByFips(fips: string) {
...
+function _findStateByFips(fips: string): State | undefined {
+  return US_STATE_DATASET.state_dataset.find(
     state => state.state_fips_code === fips,
   );
+}
+
+export function isStateFips(fips: string): boolean {
+  const state = _findStateByFips(fips);
+  return !!state;
+}
```
pnavarrc, line 114: Sometimes we use `isStateFips` to differentiate state and county fips codes, so I wouldn't like to run the assertion in that case.

src/common/locations.ts
pnavarrc: good catch! It's working correctly now, I added the `stateCode` to the function

src/common/locations.ts
pnavarrc: removed the function

mikelehen (changes requested):
So I think this mostly looks good, but:
1. We still use the old URLs when you share a location or chart.  This is sort of good, because those URLs need to match the URLs of the index.html pages we generate [here](https://github.com/covid-projections/covid-projections/blob/27d8c5c3f92098021f42318607022bb8fd3322e6/scripts/generate_index_pages.ts#L154) which is _also_ using the old URLs.  But *that* means that if you directly copy/paste a "new" URL from your browser into Facebook/Twitter/Slack/whatever (without using "copy link" or share widget), then there *won't* be a generated index.html page and so you *won't* get the nice share preview image for that link. 😢   So we probably want to switch all of this to the new URLs.
2. If I navigate to a state/county via search, it doesn't use the new URLs (but maybe this doesn't matter? Google presumably just crawls the maps, so those links are the main thing that matters?).

P.S. Feel 1000% free to ignore this, but I realized I don't actually love the underscores for stateUrlName. 😬  While using "_" for spaces feels right (e.g. "north_dakota" instead of "north-dakota"), it doesn't feel as right for separating the state name from the state code.  So I think "colorado-co" looks better than "colorado_co". To keep things simple, I'd be okay with just going back to the version that used dashes.  But again, feel 1000% free to ignore.

mikelehen (approved):
Actually, I will go ahead and mark this approved, since I'm leaving.

I think this would be shippable as-is if we want to get it out before I'm back.  It'll break preview images for if you share a page via copy/paste of the URL, but that's not the end of the world.

Else I can do another review by Monday if we're not in a rush.  Thanks / sorry!

PR #1569, opened 2020-09-28
Add basic sentry config

ghop02:
Thought it might be worthwhile to add in sentry alerting for the website, since I don't think we have much visibility into any problems that might arise.  

This is the most basic setup - ideally we would also upload sourcemaps during builds to make the errors easy to trace, but that doesn't have to happen right away.  I'm more curious if there are currently a lot of edge cases that are bad experiences for a certain subset of users


mikelehen (approved):
Nice!  Let's give it a go!

PR #1572, opened 2020-09-29
Updating to new api

ghop02:
Updates all references to point to API v2 files.  

The biggest change is to a lot of the types in the new API.  A lot more values can be undefined. This was a bug in the original json spec, where it said that values would be required in the json schema spec, but they weren't actually required.  

Also, `stateName` was replaced with `state`, so the lookup goes the other direction now.  

I need to add a PR in covid data model to get the latest json-schema version on master (right now it points to a specific commit)


src/common/models/Projection.ts
```diff
@@ -136,14 +137,13 @@ export class Projection {
...
   readonly fips: string;
-  readonly dateOverwhelmed: Date | null;
 
   readonly icuHeadroomInfo: ICUHeadroomInfo | null;
 
   readonly currentCumulativeDeaths: number | null;
   readonly currentCumulativeCases: number | null;
   readonly currentContactTracerMetric: number | null;
-  readonly stateName: string;
+  readonly state: string;
```
mikelehen, line 146: FWIW- I try to use `stateCode` or `stateId` since `state` is kinda' ambiguous.

src/common/models/Projection.ts
```diff
@@ -194,22 +194,22 @@ export class Projection {
...
     this.isCounty = parameters.isCounty;
-    this.totalPopulation = actuals.population;
+    this.totalPopulation = summaryWithTimeseries.population;
     this.fips = summaryWithTimeseries.fips;
 
     // Set up our series data exposed via getDataset().
     this.rawDailyCases = this.deltasFromCumulatives(
       this.fillLeadingNullsWithZeros(
-        actualTimeseries.map(row => row && row.cumulativeConfirmedCases),
+        actualTimeseries.map(row => (row?.cases ? row.cases : null)),
```
mikelehen, line 205: As per before, this and all the similar constructs below will break when cases is `0`.

Not sure what prompted all these changes but if we need to tweak the types to use `undefined` (or even `undefined|null`) for missing values, that would probably be fine...

src/common/models/Projection.ts
```diff
@@ -258,45 +262,36 @@ export class Projection {
...
+      metrics?.contactTracerCapacityRatio &&
+      !DISABLED_CONTACT_TRACING.includes(this.fips)
         ? metrics.contactTracerCapacityRatio
         : null;
   }
 
   get currentContactTracers() {
     return (
-      CONTACT_TRACER_STATE_OVERRIDES[this.stateName] ||
+      CONTACT_TRACER_STATE_OVERRIDES[this.state] ||
```
mikelehen, line 294: Should we drop this override since it's not used, and it won't work right unless we override it in the backend and the frontend anyway, which is 🤮 ?

src/common/models/Projection.ts
```diff
@@ -435,7 +432,9 @@ export class Projection {
     // TODO(michael): I don't like hardcoding "County" into the name. We should
     // probably delete this code and get the location name from somewhere other
     // than the API (or improve the API value).
```
mikelehen, line 434: Can you remove this TODO?  It's outdated (at some point the API started including 'County' in the names...)

src/common/models/Projection.ts
```diff
@@ -435,7 +432,9 @@ export class Projection {
...
-    return s.countyName ? `${s.countyName}, ${s.stateName}` : s.stateName;
+    // TODO(chris): Fix state
+    // @ts-ignore
+    return s.county ? `${s.county}, ${STATES[s.state]}]` : STATES[s.state];
```
mikelehen, line 437: Perhaps add a getStateName() helper to locations.ts that either does the ts-ignore or `(STATES as any)[stateCode]`.

Alternatively, we have [getLocationNameForFips()](https://github.com/covid-projections/covid-projections/blob/062a2ac10fae43292e5c42eb34f48fbc69ab9f21/src/common/locations.ts#L119) we could use, but it only includes the state abbreviation.  We could change it to accept options or something like `getLocationNameForFips(s.fips, { useStateName: true })`

src/common/utils/model.ts
```diff
@@ -50,16 +48,12 @@ export function fetchAllStateProjections(snapshotUrl: string | null = null) {
...
-        summaryWithTimeseries => stateCode(summaryWithTimeseries) !== undefined,
-      )
-      .map(summaryWithTimeseries => {
-        return new Projections(
-          summaryWithTimeseries,
-          stateCode(summaryWithTimeseries)!,
-        );
-      });
+    return all.map(summaryWithTimeseries => {
+      return new Projections(
```
mikelehen, line 52: Hrm...  I think the filter you removed is what allowed us to add territories in the backend without breaking the frontend.

src/common/utils/model.ts
```diff
@@ -76,13 +70,13 @@ export function fetchAllCountyProjections(snapshotUrl: string | null = null) {
     );
     return all
       .filter(
-        summaryWithTimeseries => stateCode(summaryWithTimeseries) !== undefined,
+        summaryWithTimeseries => summaryWithTimeseries.state !== undefined,
```
mikelehen, line 73: ditto

mikelehen (changes requested):
Mostly looks good but a few nits and I think we ended up with some problematic `0` handling...

scripts/update_api_types.ts
```diff
@@ -38,7 +34,7 @@ const BANNER_COMMENT = `
 
 const OUTPUT_FOLDER = path.join(__dirname, '..', 'src', 'api', 'schema');
 const SCHEMAS_BASE_URL =
-  'https://raw.githubusercontent.com/covid-projections/covid-data-model/master/api/schemas/';
```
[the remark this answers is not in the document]
ghop02: fixed!

src/common/models/Projection.ts
ghop02: ah this makes sense.  Yeah - a few days ago I tried to tweak the types to use `undefined` and it turned into a lot bigger change set than originally intended.  My immediate reaction is just to use [isNil](https://lodash.com/docs/#isNil) to check for nil-ish values for all these cases. Does that seem alright? or happy to go down a different path also

src/common/utils/model.ts
ghop02: yeah that makes sense. I think I need to check if the state code is in the states object instead (because now you'll never have a null state code, but could run into problems elsewhere)

mikelehen (approved):
Mostly looks good but a few cases still need improved null/undefined handling I think.  I also didn't get a chance to test this, so please be careful.  Or if this isn't urgent, I can give it more attention when I'm back.

src/common/models/Projection.ts
```diff
@@ -222,25 +222,33 @@ export class Projection {
 
     this.rawHospitalizations = hospitalizationsDisabled
       ? []
-      : actualTimeseries.map(row => row && row.hospitalBeds.currentUsageCovid);
+      : actualTimeseries.map(row =>
+          row?.hospitalBeds?.currentUsageCovid !== undefined
```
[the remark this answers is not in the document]
ghop02: Yeah that makes sense. 

Ended up fixing the typing (https://github.com/covid-projections/covid-data-model/pull/710) here and simplifying this PR a lot

src/common/models/Projection.ts
```diff
@@ -258,45 +266,41 @@ export class Projection {
...
-    if (
-      moment(this.dateOverwhelmed).diff(moment(), 'days') >
-      PROJECTIONS_TRUNCATION_DAYS
-    ) {
-      this.dateOverwhelmed = null;
-    }
-
     this.currentCumulativeDeaths =
-      summaryWithTimeseries.actuals.cumulativeDeaths;
+      summaryWithTimeseries.actuals.deaths !== undefined
```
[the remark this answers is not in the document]
ghop02: yup!

src/common/models/Projection.ts
ghop02: this is fixed with better api typing!

src/common/models/Projection.ts
ghop02: Dropped!

src/common/models/Projection.ts
```diff
@@ -258,45 +266,41 @@ export class Projection {
...
+      metrics?.contactTracerCapacityRatio !== undefined &&
+      !DISABLED_CONTACT_TRACING.includes(this.fips)
         ? metrics.contactTracerCapacityRatio
         : null;
   }
 
   get currentContactTracers() {
     return (
-      CONTACT_TRACER_STATE_OVERRIDES[this.stateName] ||
+      CONTACT_TRACER_STATE_OVERRIDES[this.stateCode] ||
```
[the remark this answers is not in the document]
ghop02: removed

src/common/utils/snapshots.ts
```diff
@@ -14,13 +14,13 @@ export async function fetchMasterSnapshotNumber(): Promise<number> {
 
 export function snapshotFromUrl(url: string): number {
   assert(url, 'Empty URL provided');
-  const match = /(\d+)\/?$/.exec(url);
+  const match = /(\d+)\/v2\/?$/.exec(url);
```
mikelehen, line 17: I noticed that https://covid-projections-git-updating-to-new-api.covidactnow.vercel.app/internal/compare isn't working and I think it's because we end up getting the current snapshot URL from _master_ (see fetchMasterSnapshotNumber() above) which doesn't include /v2/...  I think the email alerts may do the same thing.  Should we make this regex accept either the old or new snapshot URLs?  Else, depending on how we deploy things we may have a temporary breakage?

mikelehen (approved):
Nice! This looks good with just a couple small tweaks.

src/assets/data/data_url.json
```diff
@@ -1,3 +1,3 @@
 {
-    "data_url": "https://data.covidactnow.org/snapshot/1089/"
+    "data_url": "https://data.covidactnow.org/snapshot/1089/v2/"
```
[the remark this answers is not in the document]
ghop02: Yup, was going to wait until today's build is merged before checking

PR #1582, opened 2020-09-30
add explore to the homepage

pnavarrc:


pnavarrc:
[Add the Explore chart to the Homepage](https://trello.com/c/wjCqXS6b/475-add-the-explore-chart-to-the-homepage)

src/components/Explore/Explore.tsx
```diff
@@ -80,14 +80,15 @@ function getLabelLength(series: Series, shortLabel: boolean) {
 }
 
 const Explore: React.FunctionComponent<{
-  fips: string;
-}> = ({ fips }) => {
+  fipsList: string[];
```
igorkofman, line 83: maybe "initial fips list" for clarity?  i was having trouble figuring out whether this was controlling the module or was just the initial list

igorkofman, line 83: initialFipsList?
Am I understanding it right that this what the explore module is intialized to before the user picks whatever locations they want?

src/components/Explore/Explore.tsx
```diff
@@ -116,17 +117,23 @@ const Explore: React.FunctionComponent<{
 
   const currentMetricName = getMetricName(currentMetric);
 
-  const currentLocation = useMemo(() => findLocationForFips(fips), [fips]);
-  const autocompleteLocations = useMemo(() => getAutocompleteLocations(fips), [
-    fips,
+  const currentLocations = useMemo(() => fipsList.map(findLocationForFips), [
```
igorkofman, line 120: i'm a little surprised this is helpful

src/components/Explore/Explore.tsx
```diff
@@ -116,17 +117,23 @@ const Explore: React.FunctionComponent<{
...
+    fipsList,
   ]);
+  const autocompleteLocations = useMemo(
+    () => getAutocompleteLocations(fipsList[0]),
```
igorkofman, line 124: in what situations does the fipsList change after the fact?

src/components/Explore/Explore.tsx
```diff
@@ -170,9 +177,9 @@ const Explore: React.FunctionComponent<{
 
   // Resets the state when navigating locations
   useEffect(() => {
-    setSelectedLocations([currentLocation]);
+    setSelectedLocations(currentLocations);
     setCurrentMetric(defaultMetric);
-  }, [currentLocation, defaultMetric]);
+  }, [currentLocations, defaultMetric]);
```
igorkofman, line 182: what does "navigating locations" mean?

pnavarrc: When users click other states or counties either in the compare table or the mini map, the page updates without doing a full reload, so we need to reset the selected locations and default metric so they don't see the selections of their previous location in the new location.

I will update the comment to clarify 👍

src/components/Explore/Explore.tsx
pnavarrc: The locations in the dropdown are sorted depending on the current location (this is the order if you don't type anything):
 - if you are in a state page, it shows you states first, then counties in the current state, then all the other counties
- if you are on a county page, it shows you other counties in the state, then states, and then other counties.

As soon as you start typing the matching selections will show up. Since we don't need the list to be re-ordered when other variables change (normalize, adding locations, etc), we use the `useMemo` hook to avoid the list from being recomputed.

src/components/Explore/Explore.tsx
pnavarrc: Good point, is probably not useful in this case, `findLocationForFips` is not an expensive computation, we don't need `useMemo` here.

src/components/Explore/Explore.tsx
pnavarrc: Correct. The parent component (`LocationPage` or `Homepage`) determines the initial set of locations for the component and then the user can add more or remove them. For the location page, the initial location corresponds to the page location, for the home page, is a random set of states.

pnavarrc: Makes sense, will update 👍

pnavarrc:
[[Explore] Move label above chart](https://trello.com/c/j7PkBPb2/439-explore-move-label-above-chart)

pnavarrc:
[Add CAN Explore Chart to homepage under the States tables. Call it "Cases, Deaths and Hospitalizations". On load, pick 1-2 random states and pre-fill with them.](https://trello.com/c/EMAkiem6/479-add-can-explore-chart-to-homepage-under-the-states-tables-call-it-cases-deaths-and-hospitalizations-on-load-pick-1-2-random-stat)

chasulin (approved):

PR #1602, opened 2020-10-05
add utility functions to get recommendations

pnavarrc:
This PR only adds the structure of the functions so it's not blocking while we implement the components.

### DIFF
--- a/src/common/utils/recommend.ts
+++ b/src/common/utils/recommend.ts
@@ -1,5 +1,5 @@
 import moment from 'moment';
-import { sum, partition } from 'lodash';
+import { sum } from 'lodash';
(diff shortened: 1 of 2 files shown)


src/common/utils/recommend.ts
```diff
@@ -0,0 +1,126 @@
...
+  HARVARD,
+}
+
+// Note: We need to change this interface to match the CMS data structure (or import it
+// from the CMS content directory). It's here just as a placeholder.
+export interface Recommendation {
+  copy: string;
+  source: Source;
+  level: FedLevel | HarvardLevel;
+}
```
pnavarrc, line 31: add `category`, `id`

chasulin (approved):

PR #1611, opened 2020-10-06
explore: fix homepage sharing url and image name

pnavarrc:
Fixes some issues when sharing the Explore chart from the homepage.

- When the user copies the link, the copied link will redirect the user to either the location or the home page, depending on the original location of the user. The link correctly updates the Explore chart and scrolls to the corresponding location on the page.
- When the user saves an image, the image filename will include the names of the locations currently selected, not the initial location (up to 2 locations, beyond that it will be named `multiple-locations-{METRIC}-{YYYY-MM-DD}.png`)

### DIFF
--- a/src/components/Explore/utils.ts
+++ b/src/components/Explore/utils.ts
@@ -24,6 +24,7 @@ import {
   isState,
   isCounty,
   belongsToState,
+  getRelativeUrlForFips,
 } from 'common/locations';
 import { share_image_url } from 'assets/data/share_images_url.json';
 import { SeriesType, Series } from './interfaces';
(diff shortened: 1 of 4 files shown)


src/components/Explore/utils.ts
```diff
@@ -303,9 +317,10 @@ export function getExportImageUrl(sharedComponentId: string) {
   return urlJoin(share_image_url, `share/${sharedComponentId}/export.png`);
 }
 
-export function getChartUrl(fips: string, sharedComponentId: string) {
+export function getChartUrl(sharedComponentId: string) {
+  const pathname = window.location.pathname;
   const redirectTo = urlJoin(
-    getRelativeUrlForFips(fips),
+    pathname.includes('us/') ? pathname : '/',
```
mikelehen, line 323: This is a little fragile / broken because the pathname may have additional stuff at the end (/compare/x, /chart/x, /explore/x, etc.)

As a simple example if I share an explore chart, I might end up with a URL like:
https://covid-projections-git-pablo-explore-homepage-share.covidactnow.vercel.app/us/north_dakota-nd/explore/149

If I go there and share it again, I end up at a broken URL:
https://covid-projections-git-pablo-explore-homepage-share.covidactnow.vercel.app/us/north_dakota-nd/explore/149/explore/150

Because /explore ends up appended twice....

mikelehen (approved):
I think there's a slight bug, but otherwise looks good.

pnavarrc:
[CAN Explore on Homepage sharing cleanup](https://trello.com/c/j21Ba14C/492-can-explore-on-homepage-sharing-cleanup)

mikelehen (approved):
Nice! Code looks good and seems to work as expected. 😁

PR #1622, opened 2020-10-08
Recommends: inline cont.

chasulin:
- Adds share buttons, share URL, /recommends route, scrollTo, more styling


src/components/SmallShareButtons/SmallShareButtons.tsx
```diff
@@ -0,0 +1,41 @@
...
+import LinkIcon from '@material-ui/icons/Link';
+import { CopyToClipboard } from 'react-copy-to-clipboard';
+
+//TODO (Chelsi): Add in final share quote
+const SmallShareButtons = (props: { shareUrl: string }) => {
+  const url = props.shareUrl;
+
+  const socialProps = {
+    url,
+    quote: '',
```
pnavarrc, line 20: We probably need the quote to be a prop for this component, as we will likely use it in other features

pnavarrc (approved):
Looks good!

PR #1624, opened 2020-10-08
recommend: update list of recommendations

pnavarrc:
Added Harvard recommendations for the modal and copy updates. The Fed recommendations for the modal are missing, will ship those on a separate PR.


src/cms-content/recommendations/index.ts
```diff
@@ -26,15 +26,30 @@ export enum HarvardLevel {
...
+  PUBLIC_INTERACTIONS_RED = 'PUBLIC_INTERACTIONS_RED',
+  MEDICAL_CONDITIONS_RED = 'MEDICAL_CONDITIONS_RED',
+  RESTAURANTS_RED = 'RESTAURANTS_RED',
+  BARS_RED = 'BARS_RED',
+  GATHERINGS_RED = 'GATHERINGS_RED',
+  MASKS_RED = 'MASKS_RED',
+  SCHOOLS_RED = 'SCHOOLS_RED',
+  SCHOOLS_ORANGE = 'SCHOOLS_ORANGE',
+  SCHOOLS_YELLOW = 'SCHOOLS_YELLOW',
+  SCHOOLS_GREEN = 'SCHOOLS_GREEN',
```
chasulin, line 52: this ID was originally intended to filter for sub-levels. if its just a combo of category+level, do we need it?

pnavarrc: Let's keep them a bit longer just in case. The sub-level was actually a recommendation for public officials, but we might want to tweak it so it's part of public messaging

chasulin (approved):

PR #1625, opened 2020-10-08
Recommends: add share tracking

chasulin:


src/components/Recommend/Recommend.tsx
```diff
@@ -43,6 +43,8 @@ const Footer: React.FC<{
   shareUrl: string;
   shareQuote: string;
 }> = ({ onClickOpenModal, shareUrl, shareQuote }) => {
+  const trackLabel = 'recommendations';
```
pnavarrc, line 46: The category is already recommendations, consider using the name of the social network as label instead so we can have more granular information in the reports.

pnavarrc (approved):

PR #1629, opened 2020-10-08
recommend: Add the full text of the Fed recommendations to the modal tabs

pnavarrc:


chasulin (approved):
looks good! a few style-related things: 
- the non-bold body copy should be `COLOR_MAP.GRAY_BODY_COPY`
- i feel like the italics is a little hard to read. but tbd- can get feedback when we send a more final preview link

pnavarrc:
I'm going to push some fixes and merge!

PR #1650, opened 2020-10-12
Disable MO counties test positivity.

mikelehen:


src/common/models/Projection.ts
```diff
@@ -13,14 +13,29 @@ import { ICUHeadroomInfo, calcICUHeadroom } from './ICUHeadroom';
 import { lastValue } from './utils';
 import { getStateName } from 'common/locations';
 
+/** Stores a list of FIPS or FIPS regex patterns to disable. */
+class DisabledFips {
+  constructor(private fipsPatterns: (string | RegExp)[]) {}
```
ghop02, line 18: Do you not have to initialize `fipsPatterns`?  Learning more about typescript :)

ghop02 (approved):

src/common/models/Projection.ts
mikelehen: This is shorthand for:
```
private fipsPatterns: (string|RegExp)[];
constructor(fipsPatterns: (string|RegExp)[]) {
  this.fipsPatterns = fipsPatterns;
}
```

So it a) declares a private member, b) declares a constructor that takes one parameter, c) assigns that parameter to the private member.

PR #1673, opened 2020-10-14
Recommends edits

chasulin:
- Replaces feedback modal with external link
- Removes recommendations for counties
- Adds tracking for feedback button

### DIFF
--- a/src/components/Recommend/Recommend.tsx
+++ b/src/components/Recommend/Recommend.tsx
@@ -54,6 +54,14 @@ const Footer: React.FC<{
 }> = ({ onClickOpenModal, shareUrl, shareQuote, feedbackFormUrl }) => {
   const trackLabel = 'recommendations';
 
+  const feedbackOnClick = () => {
+    trackEvent(
+      EventCategory.RECOMMENDATIONS,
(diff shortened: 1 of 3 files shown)


src/components/Recommend/Recommend.tsx
```diff
@@ -46,21 +49,28 @@ const Header = (props: {
...
           {footer.modalButtonLabel}
         </FooterLink>
-        <FooterLink onClick={onClickOpenFeedbackModal}>
-          {footer.feedbackButtonLabel}
-        </FooterLink>
+        {!isMobile && (
+          <ExternalLink href={feedbackFormLink}>
+            <FooterLink>{footer.feedbackButtonLabel}</FooterLink>
+          </ExternalLink>
+        )}
```
pnavarrc, line 73: Consider using `<Hidden xsDown>` from `@material-ui/core/Hidden` instead

pnavarrc: Should we add tracking here? Not urgent because we can track clicks on Typeform

pnavarrc (approved):
Looks good!

PR #1674, opened 2020-10-14
recommend: update modal layout

pnavarrc:


src/components/Recommend/RecommendTabs.tsx
```diff
@@ -31,7 +31,7 @@ const RecommendTabs: React.FC<{
   return (
     <Style.LevelContainer>
       <Style.Tabs
-        centered
+        centered={isMobile ? false : true}
```
chasulin, line 34: can this just be `centered={!isMobile}`?

src/components/Recommend/RecommendModal.style.ts
```diff
@@ -94,6 +104,30 @@ export const Tabs = styled(MuiTabs)`
...
+  text-align: left;
+  color: #00bfea;
+  text-decoration: none;
+  &:hover {
+    text-decoration: underline;
+  }
+`;
+
+export const ModalContents = styled.div`
+  border-bottom: solid 1px #f2f2f2;
```
chasulin, line 127: this is also `COLORS.LIGHTGRAY`

src/components/Recommend/RecommendModal.style.ts
```diff
@@ -94,6 +104,30 @@ export const Tabs = styled(MuiTabs)`
...
+
+export const ContentLink = styled.a`
+  font-family: Roboto;
+  font-size: 14px;
+  font-style: normal;
+  font-weight: 400;
+  line-height: 28px;
+  letter-spacing: 0em;
+  text-align: left;
+  color: #00bfea;
```
chasulin, line 119: nit: I think this is meant to be our blue (`COLOR_MAP.BLUE`). for some reason its sometimes #00bfea in the mocks, but that blue is slightly different

chasulin (approved):
Looks good! just a few misc comments, and 2 style-related things below:

I think we wanna nix this top margin from the header so its sits at the same level as the description container:
Screen Shot 2020-10-14 at 8 04 31 PM

Maybe we should either add space on the left/right or make the modal 100% width/height with no space
Screen Shot 2020-10-14 at 8 36 09 PM

PR #1679, opened 2020-10-15
recommend: fix Fed level calculation

pnavarrc:
Fixes the logic for the Fed level calculation. If we don't have weekly new cases or test positivity rate, we show the recommendations for Green (use masks), but we don't assert that the location is in the Green level

Locations on each level for QA:
- Red: Idaho
- Yellow: Nevada
- Green: Vermont
- Unknown: North Dakota


chasulin (approved):
this looks good! remove the unused imports and merge it in! woot

PR #1707, opened 2020-10-19
Learn glossary/faq cms structure

chasulin:


public/admin/config.yml
```diff
@@ -154,3 +154,66 @@ collections:
...
+          - label: Intro
+            name: intro
+            widget: markdown
+            default: true
+          - label: Terms
+            name: terms
+            widget: list
+            fields:
+              - { label: Term, name: term, widget: string }
+              - { label: Definition, name: definition, widget: markdown }
```
pnavarrc, line 199: Consider giving terms IDs, we will probably link to them as well

public/admin/config.yml
```diff
@@ -154,3 +154,66 @@ collections:
...
+  - name: learn
+    label: Learn
+    format: json
+    description: Learn content
+    slug: '{{slug}}'
+    editor:
+      preview: true
+    files:
+      - name: Learn-Landing
+        label: 'Learn: Landing page'
```
pnavarrc, line 166: Consider renaming to `Landing page` (without the "Learn:") prefix so it doesn't look so repetitive in the CMS

public/admin/config.yml
```diff
@@ -154,3 +154,66 @@ collections:
...
+            default: true
+          - label: Sections
+            name: sections
+            widget: list
+            fields:
+              - { label: Section title, name: sectionTitle, widget: string }
+              - { label: Section ID, name: sectionId, widget: string }
+              - { label: Description, name: description, widget: markdown }
+              - { label: Button CTA, name: buttonCta, widget: string }
+              - { label: Button redirect, name: buttonRedirect, widget: text }
```
pnavarrc, line 181: The home page seems to have only one link per section, and they are all links as far as I can tell

pnavarrc (approved):
Added some notes, looks good!

public/admin/config.yml
chasulin: yeah, my thought was that each button will need its own redirect. so for the example screenshot'ed below, the redirect will be '/glossary'

Screen Shot 2020-10-19 at 12 31 18 PM

chasulin: do you see it as needing a different setup?

PR #1715, opened 2020-10-19
transform the AppBar items to be links

pnavarrc:
Currently, the links on the application bar are not really links, they are styled divs that trigger navigation on click. This is not ideal 

- Accessibility: Link elements have a number of built-in accessibility features, such as enabling keyboard navigation and other assistive features.
- Search ranking: Having links to the sections on our page will help search engines understand the structure of our page and impact the search ranking of linked pages. This will be particularly important for the Learn feature. Having the Link page linked on the page will improve the time to index it.

This PR replaces the AppBar with the NavBar component. They have the same structure and functionality, but the NavBar component uses buttons and links instead of styled divs for navigation.


src/components/AppBar/NavBar.tsx
```diff
@@ -0,0 +1,81 @@
...
+  ['/alert_signup', '/compare'].includes(pathname);
+
+const NavBar: React.FC = () => {
+  const [isMenuOpen, setMenuOpen] = useState(false);
+  const closeMenu = () => setMenuOpen(false);
+
+  // We only fade the donate button on the home page on mobile, where the donate
+  // button doesn't appear until the banner is scrolled away.
+  const { pathname } = useLocation();
+  const MobileDonateButton = useMemo(() => {
```
chasulin, line 27: do we need useMemo here?

pnavarrc: Probably not, removing!

src/components/AppBar/NavBar.tsx
```diff
@@ -0,0 +1,81 @@
...
+    return isHomeOrLocationPage(pathname)
```
chasulin: this `isHomeOrLocationPage` mixes up the original logic a little bit. `DonateButtonWithFade` only applies to homepage paths (not location page paths). i think we can just keep the original logic? ::

```
  // Fade functionality for donate button only applies to homepage paths:
  const pathsToHomepage = ['/', '/alert_signup', '/compare'];
  const buttonWithFade =
    includes(pathsToHomepage, location.pathname) &&
    !location.pathname.includes('us');
```

src/components/AppBar/NavBar.tsx
```diff
@@ -0,0 +1,82 @@
...
+        <Link to="/">
+          <Logo />
+        </Link>
+        <Style.Spacer />
+        <Hidden smDown>
+          <Style.NavLink
+            to="/"
+            key="map"
+            activeClassName="active"
+            isActive={(match, { pathname }) => isHomeOrLocationPage(pathname)}
```
chasulin: i dont think this is meant to remain 'active' (underlined, etc.) when on a location page. current behavior on prod is that Map is active/underlined only when on the homepage, and nothing in the navbar is active/underlined when on a location page

src/components/AppBar/NavBar.tsx
```diff
@@ -0,0 +1,80 @@
...
+          </Style.NavLink>
+          <Style.TabLink href="https://blog.covidactnow.org" key="blog">
+            Blog
+          </Style.TabLink>
+          <Style.NavLink to="/contact" key="contact" activeClassName="active">
+            Contact Us
+          </Style.NavLink>
+          <DonateButtonWithoutFade />
+        </Hidden>
+        <Hidden mdUp>
```
chasulin, line 66: this breakpoint needs fixing (it should transition from mobile->desktop at 800px). this MUI breakpoint is 960px, but the `StyledMobileMenu` is set to `display: none` until 800px, so there is no navbar showing up between 800-960px:

Screen Shot 2020-10-20 at 1 34 46 PM

chasulin (commented):
looks good! the only thing is that there is some misc padding/spacing inconsistencies between this and the appbar on prod. it might be helpful to open them both in the same browser and toggle back and forth to compare. for instance, i think this padding to the left of the logo can go:

Screen Shot 2020-10-20 at 1 44 48 PM

src/components/AppBar/NavBar.tsx
pnavarrc: Good catch! I removed the media query since the mobile menu and button don't render for screens sized `md` and up, it works correctly now 👍

chasulin (approved):

PR #1716, opened 2020-10-20
WIP: cms data disclaimers

chasulin:
To test, the following 3 disclaimers should be active (in the location page header):

- "Disclaimer for all NY and CT counties"
- "Disclaimer for Harris County, TX and Philadelphia County, PA"
- "Disclaimer for CA state page"


public/admin/config.yml
```diff
@@ -218,3 +218,22 @@ collections:
...
+        label: Disclaimers
+        file: src/cms-content/data-disclaimers/disclaimers.json
+        fields:
+          - label: Disclaimers
+            name: disclaimers
+            widget: list
+            fields:
+              - { label: Disclaimer copy, name: disclaimerCopy, widget: markdown}
+              - { label: State codes, name: stateCodes, widget: string, required: false}
+              - { label: Fips codes, name: fipsCodes, widget: string, required: false}
```
pnavarrc, line 239: Consider using only FIPS codes to simplify the code. Also, consider adding a `comment` for each field explaining what it does (for states, will the disclaimer show up only on the state page or also for counties?). See https://www.netlifycms.org/docs/configuration-options/#fields

public/admin/config.yml
```diff
@@ -218,3 +218,22 @@ collections:
...
                   - { label: Answer, name: answer, widget: markdown }
+  - name: data-disclaimers
+    label: Data disclaimers
+    format: json
+    description: Data disclaimers
+    slug: '{{slug}}'
+    editor:
+      preview: true
+    files:
+      - name: Disclaimers
```
pnavarrc, line 229: Maybe it would be better to have 2 files in the collection, one for state-wide disclaimers and one more granular by FIPS code (if we add a state FIPS, it would apply to the state page only, but not to its counties), that way we can make the FIPS code mandatory and have more type safety (no possibility of a disclaimer without FIPS or state code). 

Consider also adding the `min` property set to `1` to ensure that the collection is never empty.

src/common/utils/netlify.ts
```diff
@@ -2,8 +2,12 @@
  * Utils for organizing and working with our netlify-cms-based content
  */
 
+// TODO in diff branch (Chelsi): transition Donate + About to
+// export content directly from their index.ts files
 import aboutPageContent from 'cms-content/about/about-page.json';
 import donateContent, { DonatePageContent } from 'cms-content/donate';
+import dataDisclaimers from 'cms-content/data-disclaimers';
+import { Disclaimer } from 'cms-content/data-disclaimers';
```
pnavarrc, line 10: You can probably merge these two imports in one statement

src/common/utils/netlify.ts
```diff
@@ -24,3 +28,22 @@ export function getPageContent(
...
+  stateCode: string,
+  countyId?: string,
+): Disclaimer[] {
+  const disclaimersContent = dataDisclaimers.disclaimers;
+  if (countyId) {
+    return disclaimersContent.filter(
+      disclaimer =>
+        disclaimer.stateCodes?.split(',').includes(stateCode) ||
+        disclaimer.fipsCodes?.split(',').includes(fips),
+    );
```
pnavarrc, line 44: I think we need to filter out invalid fips codes here (checking whether or not they are in the state or county fips list)

pnavarrc (changes requested):
Added some suggestions!

public/admin/config.yml
chasulin: good thought-- i split it into 2 files. as for the min property, its possible that the collection could be empty (if there are no disclaimers live at a given time)

chasulin:
No longer needed (for now)

PR #1719, opened 2020-10-20
explore: update the interpolation method for line series

pnavarrc:
[Trello card](https://trello.com/c/ReNU4QcC/521-removing-overshoot-artifact-from-explore-moving-averages) - This PR replaces the interpolation method used for the Explore charts to avoid artifacts. The new interpolation is [d3-shape#curveMonotoneX](https://github.com/d3/d3-shape#curveMonotoneX)

> “a smooth curve with continuous first-order derivatives that passes through any given set of data points without spurious oscillations. Local extrema can occur only at grid points where they are given by the data, but not in between two adjacent grid points.”

_Before_
image

_After_
image


BrettBoval (approved):
LGTM

PR #1721, opened 2020-10-21
homepage: adjusts the margins of the explore chart

pnavarrc:
The chart container for the Explore chart was adding negative margins to counteract the margins of the main container on each page. The negative margins were bigger than the main container margins, making the chart container to be wider than the viewport.

This PR uses the same margin as the main container, ensuring that the Home Page won't show the horizontal scrolling control. I also changed the number of characters to use the "narrow" margin to 4, to accommodate for the Native American abbreviation "NAMC" to be shown with narrow margins (more space for the chart)


mikelehen (approved):
Looks good!  Thanks!

PR #1722, opened 2020-10-21
analytics: tracking social sharing for explore and compare

pnavarrc:
Adding tracking for when users click "share" on Facebook, Twitter, and Linkedin for Compare and Explore. The metric charts use an old component and need to migrate that first.

### DIFF
--- a/src/components/ShareButtons/ShareButtonGroup.tsx
+++ b/src/components/ShareButtons/ShareButtonGroup.tsx
@@ -20,6 +20,7 @@ const ShareImageButtons: React.FC<{
   onCopyLink?: () => void;
   onShareOnFacebook: () => void;
   onShareOnTwitter: () => void;
+  onShareOnLinkedin: () => void;
 }> = ({
   imageUrl,
   imageFilename,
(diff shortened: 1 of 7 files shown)


src/components/ShareButtons/ShareButtonGroup.tsx
```diff
@@ -18,13 +18,19 @@ const ShareImageButtons: React.FC<{
   disabled?: boolean;
   onSaveImage?: () => void;
   onCopyLink?: () => void;
+  onShareOnFacebook: () => void;
+  onShareOnTwitter: () => void;
+  onShareOnLinkedin: () => void;
```
mikelehen, line 23: To be a bit less repetitive, you could have `onShare: (dest: string) => void` and pass 'Facebook', 'Twitter', or 'LinkedIn' as appropriate.

If you go that route, you could optionally create a string union type to make it self-documenting / typesafe.

`type ShareDestination = 'Facebook'|'Twitter'|'LinkedIn';`

But feel free to keep it as-is.  It's a minor cleanup at best.

mikelehen (approved):

src/components/ShareButtons/ShareButtonGroup.tsx
pnavarrc: Good idea! Will do on a separate PR

PR #1730, opened 2020-10-22
learn: implement the table of contents component (mobile)

pnavarrc:
Implements the TableOfContents component to show on mobile. It adds hash links for each section. Note that the content from the CMS needs to be sanitized before being used as links, we can remove this logic if we update the IDs on the CMS (or keep it, just to be safe)


src/cms-content/learn/utils.ts
```diff
@@ -0,0 +1,5 @@
+import { deburr, words } from 'lodash';
+
+export function sanitizeID(sectionId: string): string {
+  return deburr(words(sectionId).join('-')).toLowerCase();
```
chasulin, line 4: nice! i imagine `deburr` is intended for when we include the spanish translations?

chasulin (approved):

src/cms-content/learn/utils.ts
pnavarrc: it takes out any accents or other non-ascii characters (just in case)

PR #1756, opened 2020-10-25
Recommends: fix yellow

chasulin:
Fed task force's yellow zone is defined as:

```
Both:
- New cases between 10-100 per 100,000 population
- Diagnostic test positivity result between 5-10%

Or one of those two conditions and one condition qualifying as being in the “Red Zone”.
```

We weren't calculating this correctly:
- We were calculating the first half with `||` instead of `&&`
- We weren't capturing the second half/the 'Or...'

For reference (from the fed doc):

Screen Shot 2020-10-15 at 11 52 06 PM

To test:
Vermont, Maine = Fed green zone
Minnesota, Oregon = Fed yellow zone
South Dakota, Utah = Fed red zone

### DIFF
--- a/src/common/utils/recommend.ts
+++ b/src/common/utils/recommend.ts
@@ -90,16 +90,16 @@ export function getFedLevel(projection: Projection): FedLevel | null {
     return null;
   }
 
-  if (weeklyCasesPer100k > 100 && positiveTestRate > 0.1) {
+  const redPosTest = positiveTestRate > 0.1;
+  const redCases = weeklyCasesPer100k > 100;
+  const yellowTestPos = 0.05 < positiveTestRate && positiveTestRate <= 0.1;
+  const yellowCases = 10 < weeklyCasesPer100k && weeklyCasesPer100k <= 100;
+
+  if (redPosTest && redCases) {
     return FedLevel.RED;
-  } else if (weeklyCasesPer100k > 100 || positiveTestRate > 0.1) {
+  } else if ((redPosTest || redCases) && (yellowTestPos || yellowCases)) {
     return FedLevel.YELLOW;
-  } else if (
-    10 < weeklyCasesPer100k &&
-    weeklyCasesPer100k < 100 &&
-    0.05 <= positiveTestRate &&
-    positiveTestRate <= 0.1
-  ) {
+  } else if (yellowTestPos && yellowCases) {
     return FedLevel.YELLOW;
   } else {
     return FedLevel.GREEN;
--- a/src/common/utils/recommend.ts
+++ b/src/common/utils/recommend.ts
@@ -95,8 +95,10 @@ export function getFedLevel(projection: Projection): FedLevel | null {
   } else if (weeklyCasesPer100k > 100 || positiveTestRate > 0.1) {
     return FedLevel.YELLOW;
   } else if (
-    (10 < weeklyCasesPer100k && weeklyCasesPer100k < 100) ||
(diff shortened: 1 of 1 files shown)


src/common/utils/recommend.ts
```diff
@@ -92,9 +92,13 @@ export function getFedLevel(projection: Projection): FedLevel | null {
 
   if (weeklyCasesPer100k > 100 && positiveTestRate > 0.1) {
     return FedLevel.RED;
+  } else if (weeklyCasesPer100k > 100 || positiveTestRate > 0.1) {
+    return FedLevel.YELLOW;
   } else if (
-    (10 < weeklyCasesPer100k && weeklyCasesPer100k < 100) ||
-    (0.05 <= positiveTestRate && positiveTestRate <= 0.1)
+    10 < weeklyCasesPer100k &&
+    weeklyCasesPer100k < 100 &&
```
mikelehen, line 99: should probably be `<=` else there's an edge case where if weeklyCasesPer100k is 100, you'll end up green.

src/common/utils/recommend.ts
```diff
@@ -92,9 +92,13 @@ export function getFedLevel(projection: Projection): FedLevel | null {
...
+    0.05 <= positiveTestRate &&
+    positiveTestRate <= 0.1
```
mikelehen, line 101: In order to get to this `else if`, I think `weeklyCasesPer100k <= 100` and `positiveTestRate <= 0.1` both have to be true already (else you would have hit the previous `else if` block).  So if we want this can be simplified to:

`else if (10 < weeklyCasesPer100k && 0.05 <= positiveTestRate) {`

src/common/utils/recommend.ts
```diff
@@ -92,9 +92,13 @@ export function getFedLevel(projection: Projection): FedLevel | null {
 
   if (weeklyCasesPer100k > 100 && positiveTestRate > 0.1) {
     return FedLevel.RED;
+  } else if (weeklyCasesPer100k > 100 || positiveTestRate > 0.1) {
+    return FedLevel.YELLOW;
```
mikelehen, line 96: Hrm, isn't supposed to be one "red condition" (covered by your `if`) but also one "yellow condition" (not covered by your `if`)?

So by the fed rules, I think maybe positiveTestRate > 0.1 but weeklyCasesPer100k < 10 would be green, since it's only meeting 1 red condition and 0 yellow conditions.

But I may be confused! I haven't looked at the full fed guidelines.

mikelehen (approved):
I think this is an improvement, though there's a `<=` edge case and I'm not sure if it's still quite implementing the fed rules precisely the same.  I'd be tempted to try to write this so that it matches the exact logic of the fed rules, e.g.:

```
const redTestPos = positiveTestRate > 0.1;
const redCases = weeklyCasesPer100k > 100;
const yellowTestPos = 0.05 < positiveTestRate <= 0.1;
const yellowCases = 10 < weeklyCasesPer100k <= 100;

if (redTestPos && redCases) {
  return Red;
} else if ((redTestPos || redCases) && (yellowTestPos || yellowCases)) {
  return Yellow;
} else if (yellowTestPos && yellowCases) {
  return Yellow;
} else {
  return Green;
}
```

NOTE: Completely untested and probably buggy.

src/common/utils/recommend.ts
chasulin: ah no i just re-visited the fed rules and i think you're right. i wasn't factoring in that the non-red condition still needs to be yellow (and not green).

pnavarrc (approved):
👌

PR #1762, opened 2020-10-26
handle internal and external links on markdown content

pnavarrc:
The markdown component has a custom renderer for links. This PR differentiates internal and external links by parsing the `href` parameter as a [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) and checking if the link includes a hostname.

Internal links use the `HashLink` component, which scrolls to the specific link (let's say `/glossary#superspreader`, external links will open on a new tab (and have `rel="noopened noreferrer"`)

### DIFF
--- a/src/components/SharedComponents/Accordion.tsx
+++ b/src/components/SharedComponents/Accordion.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { AccordionDetails } from '@material-ui/core';
+import { AccordionDetails, AccordionProps } from '@material-ui/core';
 import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
 import {
   StyledMuiAccordion,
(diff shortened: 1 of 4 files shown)


chasulin (commented):
looks good so far!! 2 things:
-  the offset isnt exactly right. when clicking an internal link to another glossary term, it doesnt exactly scroll to the right place
- we can add this to the list, but it might be useful to use the [accordion's extended prop](https://material-ui.com/api/accordion/) to open the term that is clicked on/scrolled to.

src/components/SharedComponents/Accordion.tsx
```diff
@@ -1,20 +1,24 @@
...
+import { AccordionDetails, AccordionProps } from '@material-ui/core';
 import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
 import {
   StyledMuiAccordion,
   StyledAccordionSummary,
   MarkdownBody,
 } from './Accordion.style';
 
-const StyledAccordion = (props: {
+export interface StyledAccordionProps extends Omit<AccordionProps, 'children'> {
```
chasulin, line 10: does this update syntax serve a specific purpose in this case?

src/components/SharedComponents/Accordion.tsx
```diff
@@ -1,20 +1,24 @@
...
+}
 
+const StyledAccordion: React.FC<StyledAccordionProps> = ({
+  summaryText,
+  detailText,
+  ...otherProps
+}) => {
   return (
-    <StyledMuiAccordion>
+    <StyledMuiAccordion {...otherProps}>
```
chasulin, line 21: same as above q ^ :
does the updated syntax + addition of `{...otherProps}` serve a specific purpose here?

src/components/SharedComponents/Accordion.tsx
pnavarrc: yes! It allows us to pass the `expanded` and `onChange` props to the `StyledAccordion`, as well as any other `MuiAccordion` props

chasulin (approved):

PR #1776, opened 2020-10-27
fix custom link component on markdown

pnavarrc:
Fixes internal links on the markdown component


chasulin (approved):
looks good! do we want the links between CAN pages (ie an link in FAQ that links to glossary) to open in a new tab by default?

PR #1777, opened 2020-10-27
implement error boundary

pnavarrc:
An error boundary prevents the entire app from crashing when a component crashes. One potential problem with the CMS is that CMS users could put an invalid URL in the link, which would crash the app.

This PR uses the ErrorBoundary component provided by Sentry to wrap the markdown component. If the markdown component crashes for any reason, the error boundary will capture the error and render the fallback UI. It will also send the error report to Sentry

image

## How to Test

In local development, errors are shown with an overlay by Create React App, which hides the Error boundary. To test the error boundary, we need to use the production build

- Throw an error from the Markdown component (add `new URL('~~~')` somewhere in the component)
- Build the app locally (`yarn build`)
- Run the server using the build `serve -s build -l 4000` (I'm using https://www.npmjs.com/package/serve)
- The fallback component should render instead of the component, and the app shouldn't crash

### DIFF
--- a/src/components/ErrorBoundary/ErrorBoundary.tsx
+++ b/src/components/ErrorBoundary/ErrorBoundary.tsx
@@ -1,0 +1,11 @@
+import React from 'react';
+import * as Sentry from '@sentry/react';
+import FallbackAlert from './FallbackAlert';
+
+const ErrorBoundary: React.FC = ({ children }) => (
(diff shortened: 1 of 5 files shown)


chasulin (commented):
Didn't invalid links/broken markdown previously just fall back to regular/un-hyperlinked copy? Does this take the place of that?

chasulin (approved):

PR #1800, opened 2020-10-29
Adding geolocation to explore

pnavarrc:
🚧  _work in progress_

[Trello Card](https://trello.com/c/7rfWXRqB/494-geo-locate-user-and-use-the-location-to-pre-populate-explore-chart-on-homepage) 

## Idea

```ts
const [coordinates] = useGeolocation();

useEffect(() => {
  if (coordinates) {
    // find state, add to explore
  } else {
    // user didn't authorize, couldn't find the state (show alert, disable geolocation button)
  }
}, [coordinates])

```

## Note to self
- `geoContains` from `d3-geo` doesn't work with coordinates from our states dataset (two theories, spherical vs. planar coordinates, or clockwise vs. counterclockwise coordinates on the corresponding polygons). A workaround is to work with projected coordinates and use `contains` from `d3-polygon` , using any projection should work.
- Enable sensors on Chrome Developer Tools and select a US location during development, see [Override Geolocation in Chrome DevTools](https://developers.google.com/web/tools/chrome-devtools/device-mode/geolocation)

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "@types/d3-shape": "^1.3.2",
     "@types/d3-time": "^1.0.10",
     "@types/react-router-hash-link": "^1.2.1",
+    "@types/topojson": "^3.2.2",
     "@types/typeform__embed": "^0.22.2",
     "@types/uuid": "^8.3.0",
     "@vx/axis": "^0.0.196",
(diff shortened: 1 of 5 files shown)


pnavarrc:
Just a demo, closing for now

PR #1805, opened 2020-10-29
cms: setup config for case studies

pnavarrc:


public/admin/config.yml
```diff
@@ -255,3 +255,86 @@ collections:
         file: src/cms-content/learn/footer.json
         fields:
           - { label: Body, name: body, widget: markdown, default: true }
+      - name: Case Studies
+        file: 'src/cms-content/learn/case-studies.json'
+        label: Case Studies
+        fields:
+          - label: Title
```
chasulin, line 262: nit: just for consistency's sake between files, want to name this 'Header'?

public/admin/config.yml
```diff
@@ -255,3 +255,86 @@ collections:
...
+            widget: list
+            fields:
+              - {
+                  label: Category Name,
+                  name: title,
+                  widget: text,
+                  hint: Name of the category,
+                }
+              - {
+                  label: Anchor Link ID,
```
chasulin, line 281: I think just 'Category ID' would be good here (in line with term-id, etc.), rather than using 'anchor'/eng-language

public/admin/config.yml
```diff
@@ -255,3 +255,86 @@ collections:
...
+                  name: anchorId,
+                  widget: string,
+                  hint: 'String to use for anchor links (ex: public-health)',
```
chasulin, line 284: same as comment above. would nix dev-language (ie. string, anchor), and just write something like 'ID to use in URLs linking to this category'

chasulin: or rather, maybe the hint should just refer to the format and not the use. ie 'all lowercase category name, divided by dashes if multi-word'

chasulin (approved):

public/admin/config.yml
```diff
@@ -255,3 +255,86 @@ collections:
...
         fields:
           - { label: Body, name: body, widget: markdown, default: true }
+      - name: Case Studies
+        file: 'src/cms-content/learn/case-studies.json'
+        label: Case Studies
+        fields:
+          - label: Title
+            name: title
+            widget: text
+            hint: Title for the case studies landing page
```
chasulin, line 265: oooo hints!!! how fun!!

public/admin/config.yml
```diff
@@ -255,3 +255,86 @@ collections:
...
+                      hint: Full content of the case study,
+                    }
+                  - {
+                      label: Tags,
+                      name: tags,
+                      widget: list,
+                      hint: 'Optional tags for the case study. Separate with commas. (ex: api, embed, etc)',
+                    }
+                  - {
+                      label: Author,
```
chasulin, line 335: maybe move the author field up to live right below the short+long headers

public/admin/config.yml
```diff
@@ -255,3 +255,86 @@ collections:
...
+                      hint: Title for the case study page,
+                    }
+                  - {
+                      label: Short Title,
+                      name: shortTitle,
+                      widget: text,
+                      hint: Short title for the case studies landing page,
+                    }
+                  - {
+                      label: URL,
```
chasulin, line 304: nit: maybe we can call this ID or case study ID, rather than URL, and then edit the hint to say something like: "Added to the URL, formatted with no spaces, separated by dashes (ex: ibm-melanie-curry)"

src/cms-content/learn/case-studies.ts
```diff
@@ -0,0 +1,28 @@
+// TODO: Import the content from the JSON file
```
chasulin, line 1: do we not wanna keep everything in index and just document that file well?

public/admin/config.yml
```diff
@@ -255,3 +255,86 @@ collections:
...
+                      name: url,
+                      widget: string,
+                      hint: 'No spaces, separated by dashes (ex: ibm-melanie-curry)',
+                    }
+                  - {
+                      label: Logo,
```
chasulin, line 310: we also need a string field for the logo alt text

public/admin/config.yml
```diff
@@ -255,3 +255,86 @@ collections:
...
+                  - {
+                      label: Body,
+                      name: body,
+                      widget: markdown,
+                      hint: Full content of the case study,
+                    }
+                  - {
+                      label: Tags,
+                      name: tags,
+                      widget: list,
```
chasulin, line 338: i think this is meant to be string/text?

pnavarrc: yes, by default it will be a list of strings

chasulin: also i think if this field is going to be optional, we need to add `required: false`

PR #1808, opened 2020-10-29
Update to use newcases

ghop02:
Uses the new case series directly. Also updates the api types to latest


src/api/schema/AggregateRegionSummaryWithTimeseries.d.ts
```diff
@@ -76,6 +80,21 @@ export type NonCovidPatientsMethod =
...
+ * COVID Risk Level.
+ *
+ * ## Risk Level Definitions
+ *  *Low* - On track to contain COVID
+ *  *Medium* - Slow disease growth
+ *  *High* - At risk of outbreak
+ *  *Critical* - Active or imminent outbreak
+ *  *Unknown* - Risk unknown
+ */
+export type RiskLevel = 0 | 1 | 2 | 3 | 4;
```
mikelehen, line 97: FWIW- It looks like this is documented as if the risk levels are strings, not numeric.

On that note, strings would have probably been more friendly for API consumers (and less confusing if we add additional risk levels).  Too late to change? 😁

mikelehen (approved):
LGTM.

src/api/schema/AggregateRegionSummaryWithTimeseries.d.ts
ghop02: Ah I know.  I do regret this a lot.  I went this direction for a couple reasons:
 1. I thought the enum representation for the open api schema was better, so I mistakenly thought i could create a better interface for this that included the names.
 2. I was eventually hoping to add a `riskLevelTimeseries` and was thinking that ints would be a better representation for that. 

I think if we want to change it we should probably add it as a new field and potentially deprecate the current field.  I know we have some consumers using the current risk level field (at least resistbot).

PR #1811, opened 2020-10-29
Learn blockquote component

chasulin:


src/screens/Learn/Learn.style.tsx
```diff
@@ -45,7 +33,50 @@ export const PageHeader = styled.h1`
...
+
+  p {
+    &:not(:last-child) {
+      margin-bottom: 1rem;
+    }
+  }
+`;
+
+const BlockquoteStyles = css`
+  background-color: #fafafa;
```
pnavarrc, line 64: Maybe we should name this color, we have a couple of instances of it already

src/screens/Learn/Learn.style.tsx
```diff
@@ -45,7 +33,50 @@ export const PageHeader = styled.h1`
...
+  padding: 1.125rem;
```
pnavarrc, line 65: Consider using spacing from the theme (either rounding to 16 or 24). I'm actually not sure what's best here, using `rem` or theme spacing 🤔

src/screens/Learn/Learn.style.tsx
```diff
@@ -45,7 +33,50 @@ export const PageHeader = styled.h1`
...
+  display: inline-block;
```
pnavarrc, line 66: Why `inline-block` instead of block?

pnavarrc (approved):
Looks good!

src/screens/Learn/Learn.style.tsx
chasulin: the mocks had 20px of spacing so i kept it exact, but i can round down to 16 and use theme spacing

src/screens/Learn/Learn.style.tsx
chasulin: the small top/bottom margin i was adding on the blockquote wasn't working with `display: block`

PR #1822, opened 2020-11-01
update the donation embed (nov 2020)

pnavarrc:


src/screens/Donate/GiveButterEmbed.tsx
```diff
@@ -1,12 +1,12 @@
 import React from 'react';
 import { EmbedWrapper } from 'screens/Donate/Donate.style';
 
-const GiveButterEmbed = () => {
+const GiveButterEmbed: React.FC<{ embedUrl: string }> = ({ embedUrl }) => {
```
ghop02, line 4: What is the `React.FC`syntax?

ghop02 (approved):

src/screens/Donate/GiveButterEmbed.tsx
pnavarrc: It's an abbreviation of the type `React.FunctionComponent`, which adds type information common for react components (`defaultProps`, `children`, etc), and will show an error if you return something that is not a valid react component.

It's useful sometimes when you are passing the component to another component, that does check the type (usually when rendering props). For example, the following code would fail, because the `fallback` property needs to have type `React.FunctionalComponent`, and the `Fallback` function doesn't have it

```tsx
function Fallback() {
  return <div>Error</div>;
}

<Sentry.ErrorBoundary fallback={Fallback}>
  {/* */}
<Sentry.ErrorBoundary />
```

It's more useful when you are creating a custom component, because you can inherit the type of the native element, but customize some properties, while still accepting all the props that you would expect from the native element

```tsx
// Opens in a new tab by default, but allows the client code to override that behaviour
const CustomLink: React.FC<HTMLAnchorElement> = (props) => {
  <a target="_blank" rel="noopener noreferrer"  {...props} />
}
```

PR #1827, opened 2020-11-02
update the cap for the case incidence chart

pnavarrc:
Update the cap on the daily new cases chart. I moved the cap to 500 to account for counties in North Dakota that are unfortunately over 250 / 100k.

See [Wells County, ND](https://covid-projections-git-pablo-update-case-density-cap.covidactnow.vercel.app/us/north_dakota-nd/county/wells_county?s=1231303) for an example


mikelehen (approved):

pnavarrc:
[Case Incidence Chart is capped at 100 cases / 100k](https://trello.com/c/CwMfYJO0/544-case-incidence-chart-is-capped-at-100-cases-100k)

PR #1829, opened 2020-11-02
case studies - styling

pnavarrc:
Updates the styles of the case study (singular) page, and updates the accordion styles.

### DIFF
--- a/src/components/SharedComponents/Accordion.style.tsx
+++ b/src/components/SharedComponents/Accordion.style.tsx
@@ -32,6 +32,11 @@ export const MarkdownBody = styled(MarkdownContent)`
   line-height: 22px;
 
   p {
+    line-height: 1.4;
+    font-family: Roboto;
+    font-size: 14px;
+    font-weight: 400;
+    line-height: 22px;
     margin: 0;
     &:not(:last-child) {
       margin-bottom: ${theme.spacing(3)}px;
--- a/src/components/SharedComponents/Accordion.style.tsx
+++ b/src/components/SharedComponents/Accordion.style.tsx
@@ -23,20 +23,11 @@ export const StyledAccordionSummary = styled(AccordionSummary)`
   align-items: flex-start;
 `;
 
-export const MarkdownBody = styled(MarkdownContent)`
-  color: ${COLOR_MAP.GRAY_BODY_COPY};
-  line-height: 1.4;
-  font-family: Roboto;
-  font-size: 14px;
-  font-weight: 400;
-  line-height: 22px;
-
+export const AccordionMarkdownBody = styled(MarkdownContent)`
   p {
-    line-height: 1.4;
-    font-family: Roboto;
-    font-size: 14px;
-    font-weight: 400;
-    line-height: 22px;
+    color: ${COLOR_MAP.GRAY_BODY_COPY};
+    line-height: 1.6;
+    font-size: 16px;
     margin: 0;
     &:not(:last-child) {
       margin-bottom: ${theme.spacing(3)}px;
--- a/src/components/SharedComponents/Accordion.style.tsx
+++ b/src/components/SharedComponents/Accordion.style.tsx
@@ -3,6 +3,7 @@ import { Accordion, AccordionSummary } from '@material-ui/core';
 import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
 import { MarkdownContent } from 'components/Markdown';
 import { COLOR_MAP } from 'common/colors';
+import theme from 'assets/theme';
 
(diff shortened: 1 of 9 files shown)


src/screens/Learn/CaseStudies/CaseStudy.tsx
```diff
@@ -19,10 +20,13 @@ const CaseStudy: React.FC = () => {
   return (
     <Style.PageContainer>
       <Style.PageContent>
-        <Breadcrumbs
-          item={{ to: '/learn/case-studies', label: 'Case Studies' }}
-        />
-        <h1>{header}</h1>
+        <Style.BreadcrumbsContainer>
+          <Breadcrumbs
+            item={{ to: '/learn/case-studies', label: 'Case Studies' }}
```
chasulin, line 25: this should now change to `to: '/case-studies'`

chasulin (commented):
I would do a side by side comparison of preview link vs. prod. The spacing has has become a bit off

src/components/SharedComponents/Accordion.style.tsx
```diff
@@ -23,13 +25,21 @@ export const StyledAccordionSummary = styled(AccordionSummary)`
 
 export const MarkdownBody = styled(MarkdownContent)`
```
chasulin, line 26: suggestion: lets rename this something like `AccordionBodyMarkdown` so it doesnt get confused with the more global `MarkdownBody` component in `components/Markdown/Markdown.style.ts` (and be careful to swap out the right ones after the rename)

src/components/SharedComponents/Accordion.style.tsx
```diff
@@ -23,13 +25,21 @@ export const StyledAccordionSummary = styled(AccordionSummary)`
...
   color: ${COLOR_MAP.GRAY_BODY_COPY};
+  line-height: 1.4;
+  font-family: Roboto;
+  font-size: 14px;
+  font-weight: 400;
+  line-height: 22px;
```
chasulin, line 32: you only need to declare this block of styles once (in the `p`, as you've done below)

src/components/SharedComponents/Accordion.style.tsx
```diff
@@ -23,13 +25,21 @@ export const StyledAccordionSummary = styled(AccordionSummary)`
...
 
   p {
-    font-size: 1rem;
     line-height: 1.4;
+    font-family: Roboto;
+    font-size: 14px;
+    font-weight: 400;
+    line-height: 22px;
```
chasulin, line 39: line-height is declared here twice now (`line-height: 1.4` a few lines up). which is the correct one?

src/components/SharedComponents/Accordion.style.tsx
```diff
@@ -23,13 +25,21 @@ export const StyledAccordionSummary = styled(AccordionSummary)`
...
+  line-height: 1.4;
+  font-family: Roboto;
+  font-size: 14px;
+  font-weight: 400;
+  line-height: 22px;
 
   p {
-    font-size: 1rem;
     line-height: 1.4;
+    font-family: Roboto;
```
chasulin, line 36: i think it will default to roboto, no need to declare it

src/components/SharedComponents/Accordion.style.tsx
```diff
@@ -23,13 +25,21 @@ export const StyledAccordionSummary = styled(AccordionSummary)`
...
+    font-size: 14px;
+    font-weight: 400;
```
chasulin, line 38: same here- i think it will default to `font-weight: 400`, no need to declare it

src/screens/Learn/Learn.style.tsx
```diff
@@ -51,11 +47,14 @@ export const BreadcrumbsContainer = styled.div`
 export const BodyCopyStyles = css`
   color: ${COLOR_MAP.GRAY_BODY_COPY};
   line-height: 1.4;
-  letter-spacing: 1;
+  font-family: Roboto;
+  font-size: 14px;
+  font-weight: 400;
+  line-height: 22px;
```
chasulin, line 53: same set of comments re: these styles written above for `MarkdownBody`. an overall suggestion: maybe update/clean them up here in `BodyCopyStyles` and import `BodyCopyStyles` into `Accordion.style.tsx` and use them to declare styles for `MarkdownBody`, so we only need to edit them in one place going forward

src/components/SharedComponents/Accordion.style.tsx
pnavarrc: 1.6 is closest to what we have in Figma, updating

chasulin (approved):

PR #1833, opened 2020-11-03
case studies: add the section for more studies in the category

pnavarrc:


chasulin (approved):
Only blocking thing is that the footer links aren't working as expected, and i think its because the Link component is navigating to `/case-study/...` and not `/case-studies/...`

Looks good otherwise! Style-related nit (and also an accessibility consideration) - the map of `<li>`'s needs to be wrapped in a `<ul>` for the bullets to be indented properly. wrapping them in a `<ul>` and editing the `<ul>`'s `padding-inline-start` to ~1rem, that should do it

right now the text wraps below the bullet point:

Screen Shot 2020-11-02 at 11 47 07 PM

the above fix will indent them properly:


Screen Shot 2020-11-02 at 11 51 04 PM

src/screens/Learn/CaseStudies/CaseStudy.tsx
```diff
@@ -27,6 +39,20 @@ const CaseStudy: React.FC = () => {
         <Logo src={caseStudy.logoUrl} alt={caseStudy.logoAltText} />
         <BodyCopyMarkdown source={author} />
         <BodyCopyMarkdown source={body} />
+        {studyCategory && otherCaseStudies.length > 0 && (
+          <LearnMoreSection>
+            <LearnMoreTitle>{`More ${studyCategory.header.toLowerCase()} studies`}</LearnMoreTitle>
+            <LearnMoreBody>
+              {otherCaseStudies.map(study => (
+                <li key={study.caseStudyId}>
+                  <Link to={`/case-study/${study.caseStudyId}`}>
```
chasulin, line 48: i think this is meant to be `/case-studies/...`

PR #1834, opened 2020-11-03
case studies: embed

pnavarrc:


chasulin (approved):
This makes sense + works!! Feel free to merge it in and we can later on continue to explore how to make this less of a manual process ( / if thats possible)

PR #1850, opened 2020-11-05
Add CMS-specific update and disclaimer copy.

mikelehen:
This has no effect until we switch to a snapshot that uses CMS data.  So I created a separate branch with that preview.  See e.g. https://covid-projections-git-mikelehen-use-cms-demo.covidactnow.vercel.app/us/in/county/fayette_county?s=1231303

Update copy:
image

Disclaimer copy:
image


ghop02 (approved):
🥳 this looks good to me!

PR #1856, opened 2020-11-05
Learn sidebar edit

chasulin:


pnavarrc (approved):
Maybe we should add this to the CMS instead? I was thinking of adding "shortTitle" for category names

PR #1873, opened 2020-11-09
cms: add the Contact Us page 

pnavarrc:


public/admin/config.yml
```diff
@@ -47,6 +47,51 @@ collections:
...
+        file: src/cms-content/contact/contact.json
+        fields:
+          - label: Header
+            name: header
+            widget: text
+            hint: 'Page Header'
+          - label: Intro
+            name: intro
+            widget: markdown
+            default: ''
```
chasulin, line 68: i **think** this defaults to an empty string, so no need to declare it

src/cms-content/contact/index.ts
```diff
@@ -0,0 +1,55 @@
...
+    })),
+  },
+];
+
+interface Page<T> {
+  pageTitle: string;
+  pageDescription: string;
+  canonicalUrl: string;
+  sidebarItems: TocItem[];
+  content: T;
```
chasulin, line 46: what does T stand for here

src/screens/Contact/Contact.tsx
```diff
@@ -1,145 +1,31 @@
...
-  </Typography>
-);
-
-const ListItem: FunctionComponent = ({ children }) => (
-  <li>
-    <Paragraph>{children}</Paragraph>
-  </li>
+import { Heading1, Heading2, MarkdownContent } from 'components/Markdown';
+import { Anchor } from 'components/TableOfContents';
+import MainContent from 'components/PageContent';
```
chasulin, line 5: Thoughts on trying to keep naming consistent and calling this `PageContent` instead of `MainContent` (or renaming the component/file as `MainContent`)? Sometimes the inconsistent naming makes things read a bit confusing

src/cms-content/contact/index.ts
pnavarrc: oh, this was an experiment - The idea was to create a generic type for pages that we could reuse for all the CMS pages

```tsx
const ContactUsPage: Page<ContactUsContent> = {
  pageTitle: '...',
  // ...
  content: contactUsContent
}

const DonatePage: Page<DonateContent> = {
  pageTitle: '...',
  // ...
  content: donateContent
}
```

if we establish this as a standard for the pages, we could ensure that all the pages have metatags and sidebar items... but I think it might be overengineering a bit, what do you think?

src/screens/Contact/Contact.tsx
pnavarrc: agree, I think `PageContent` is better 👍

src/cms-content/contact/index.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export const ContactPage: Page<ContactUsContent> = {
+  pageTitle: 'Contact Us',
```
chasulin, line 50: why are we hardcoding this?

pnavarrc: Adding to the CMS!

chasulin (approved):
This looks good + feel free to merge it in. 

One thing/question that can be punted to a different PR if you wanna get this one merged:

The new `PageContent` styles+component are the exact same as the `LearnPageContainer` styles+component, except `PageContent` has `ShareBlock` appended to the bottom. Are we keeping these as separate components in case the structure of Learn evolves to include more things + we don't want them entangled? Moreover- is there a specific reason that you wanted to duplicate the styles and keep those separate?

PR #1880, opened 2020-11-09
fix some explore search quirks

pnavarrc:
[Trello](https://trello.com/c/fKbc9J9l/543-explore-search-has-sorting-quirks-for-states-and-especially-counties) - Explore search has sorting quirks for states and especially counties

This PR fixes the quirks, in particular:
- Sort the locations alphabetically, preserving the order of each group (i.e., aggregated, states, state counties, other counties),  (`Alabama` shows before `Alaska`)
- Only match strings from the start of the string (`king` will match `Kings County, TX` and `Kingfisher County, OR` but not `Hocking County, OH`)

The filter options are not super well documented, but the source code is fairly straightforward, see [material-ui/src/useAutocomplete/useAutocomplete.js](https://github.com/mui-org/material-ui/blob/4d50a9ed6a7c2007b9e5061d0842f38df191cfb2/packages/material-ui/src/useAutocomplete/useAutocomplete.js)

### DIFF
--- a/src/components/AutocompleteLocations/AutocompleteLocations.tsx
+++ b/src/components/AutocompleteLocations/AutocompleteLocations.tsx
@@ -1,8 +1,13 @@
 import React from 'react';
 import Autocomplete from '@material-ui/lab/Autocomplete';
+import { createFilterOptions } from '@material-ui/lab/useAutocomplete';
 import TextField from '@material-ui/core/TextField';
 import { Location } from 'common/locations';
-import { createFilterOptions } from '@material-ui/lab/useAutocomplete';
+/**
+ * `createFilterOptions` creates a configuration object that defines how the
+ * user input will be matched against the options in the Autocomplete
+ * component. See material-ui/lab/useAutocomplete.js for additional options.
+ */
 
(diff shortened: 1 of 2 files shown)


pnavarrc:
[Explore search has sorting quirks for states and especially counties.](https://trello.com/c/fKbc9J9l/543-explore-search-has-sorting-quirks-for-states-and-especially-counties)

src/components/AutocompleteLocations/AutocompleteLocations.tsx
```diff
@@ -31,6 +32,7 @@ const AutocompleteLocations: React.FC<{
     onChange={onChangeLocations}
     getOptionSelected={getOptionSelected}
     value={selectedLocations}
+    filterOptions={createFilterOptions({ matchFrom: 'start', trim: true })}
```
mikelehen, line 35: Maybe add a comment to document what this means?

mikelehen (approved):
Looks great!  If you have time, maybe add a comment to explain the filter options. But ✅  either way!  Thank you!

PR #1881, opened 2020-11-10
Learn products component structure

chasulin:


src/screens/Learn/Products/ProductPage.tsx
```diff
@@ -0,0 +1,56 @@
...
+    return null;
+  }
+
+  const {
+    productName,
+    productIntro,
+    metadataTitle,
+    metadataDescription,
+    sections,
+  } = pageContent[0];
```
pnavarrc, line 30: We probably need a map to fetch pageContent given a product ID (like we did for the case studies). We can do this on a follow-up PR to unblock content entry

pnavarrc (approved):
Just a small comment!

src/screens/Learn/Products/ProductPage.tsx
chasulin: that is already set up. `pageContent` is the return value of `getProductPageContent(productId)` a few lines up, which is the function I put in `learn/index` (image below). i used `.filter` so i just needed to grab the obj from the return array, hence the `[0]`

Screen Shot 2020-11-09 at 7 36 18 PM

pnavarrc: oh, how about using `find` instead then?

PR #1883, opened 2020-11-10
adjust ticks on metric charts

pnavarrc:
This PR adjusts the separation between the ticks on the x-axis for metric charts. It's just a stop gap while I work on a more definitive solution.


mikelehen (approved):
Thank you for jumping on this!  I hadn't realized things were getting squeezed.  Looks good!

PR #1891, opened 2020-11-10
products: add notes component

pnavarrc:


chasulin (approved):
i think some of the specifics of the styles will need tweaking to match the mocks more closely, but lets merge this in and tweak later!

PR #1901, opened 2020-11-12
Adding hospitalization aggregates for US

ghop02:
Adds aggregates for hospitalizations and icus for the US.


scripts/update_location_summaries.ts
```diff
@@ -104,11 +107,25 @@ const AGGREGATED_DATASETS: DatasetId[] = [
...
-function aggregate(allProjections: Projections[]) {
+const STATE_AGGREGATED_DATASETS: DatasetId[] = [
+  'rawDailyCases',
+  'smoothedDailyCases',
+  'rawDailyDeaths',
+  'smoothedDailyDeaths',
+  'rawHospitalizations',
+  'smoothedHospitalizations',
+  'rawICUHospitalizations',
+  'smoothedICUHospitalizations',
```
mikelehen, line 118: I assume you're aware but to flag:
1. Looks like we only have ICU data for 42 states. This means our "per 100k" calculation will be too low and comparing the US against any state will always favor the US in an unfair way.
2. Because the # of states included has changed over time, you end up with huge jumps in the chart that aren't indicative of actual increases in ICU usage.  (same issue for hospitalizations)

So this feels a little problematic...  but not quite problematic enough that I don't think we should ship it, I guess. 😁

mikelehen (approved):

scripts/update_location_summaries.ts
ghop02: Yeah I think that's a good point.  I would be in favor of removing the ICU calculation for the time being.  Having hospitalizations aggregated is a big step in the right direction.  

Would you prefer having ICUs in or not?

mikelehen: Removing sounds good for now.

PR #1904, opened 2020-11-12
rename the social image for the home page

pnavarrc:
[trello](https://trello.com/c/lgBfbQyI/578-rename-the-social-image-url) - Rename the social image filename for SEO

@mikelehen can I rely on both scripts to be run at the same date? We could remove the timestamp from the name if not.

### DIFF
--- a/scripts/generate_share_images/index.ts
+++ b/scripts/generate_share_images/index.ts
@@ -10,6 +10,7 @@ import fs from 'fs-extra';
 import _ from 'lodash';
 import path from 'path';
 import Pageres from 'pageres';
+import moment from 'moment';
 import urlJoin from 'url-join';
 import {
   fetchAllStateProjections,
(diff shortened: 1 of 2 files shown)


pnavarrc:
[Rename the social image URL](https://trello.com/c/lgBfbQyI/578-rename-the-social-image-url)

scripts/generate_share_images/index.ts
```diff
@@ -64,7 +70,7 @@ const BLACKLISTED_COUNTIES = [
   // Homepage share image.
   screenshots.push({
     url: '/',
-    filename: 'home',
+    filename: getHomeSocialImageFilename(),
```
mikelehen, line 73: Sorrrry...  I realized this is actually the wrong place.  We don't use this script anymore since we generate the images on-demand from our firebase functions code.  We should probably go ahead and delete the script at this point, since it's no longer used.

The right place to update is here: https://github.com/covid-projections/covid-projections/blob/0672c93b5752723e9d9d89e480eabc251eaae14c/firebase/functions/src/dynamic_web_content.ts#L33

And since that runs dynamically on-demand, as long as we make that `if` statement flexible enough, it should be fine to put the date in the filename.

mikelehen (changes requested):

mikelehen:
FYI- Sent you a PR to delete the no-longer-used script. https://github.com/covid-projections/covid-projections/pull/1912

scripts/generate_share_images/index.ts
pnavarrc: No problem! 😂  - I will create a new PR and update the right place, thanks!

pnavarrc:
Closing in favour of #1921

PR #1912, opened 2020-11-14
Remove generate_share_images script and related cruft.

mikelehen:
* Remove commented-out references from github workflow.
* Also remove page_screenshots/ script which we haven't used in ages.
* Remove pageres and os-utils dependencies since we no longer need them.


pnavarrc (approved):
Looks good! Consider updating [this comment](https://github.com/covid-projections/covid-projections/blob/fc789a23a85ef3f2fd1b006373ccf00b4a17c91a/src/components/Explore/utils.ts#L325) now that we are not using `generate_share_images` anymore.

PR #1928, opened 2020-11-17
cms setup: team directory

pnavarrc:
Most of the About page is a listing of current and former team members. This PR creates the setup for the team directory. I will add the new sections for the updated About page to the existing `about` collection and wait until we have all the updated content ready to implement the new page.

### DIFF
--- a/public/admin/config.yml
+++ b/public/admin/config.yml
@@ -47,6 +47,66 @@ collections:
                   - { label: 'Alt text', name: 'altText', widget: 'string' }
                   - { label: 'Image', name: 'image', widget: 'image' }
                   - { label: 'Url', name: 'url', widget: 'string' }
+
+  - name: team
+    label: Team Directory
+    description: Team directory
+    files:
+      - name: team
+        label: Team Structure
+        file: 'src/cms-content/team/core-team.json'
+        fields:
+          - label: Teams
+            name: teams
+            widget: list
+            hint: 'Product Management, etc'
+            fields:
+              - label: Team Name
+                name: teamName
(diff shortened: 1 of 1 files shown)


public/admin/config.yml
```diff
@@ -47,6 +47,66 @@ collections:
                   - { label: 'Alt text', name: 'altText', widget: 'string' }
                   - { label: 'Image', name: 'image', widget: 'image' }
                   - { label: 'Url', name: 'url', widget: 'string' }
+
+  - name: team
+    label: Team Directory
+    description: Team directory
+    files:
+      - name: team
+        label: Team Structure
```
chasulin, line 56: would maybe call this "Active - Team structure"

public/admin/config.yml
```diff
@@ -47,6 +47,66 @@ collections:
...
+            hint: 'Product Management, etc'
+            fields:
+              - label: Team Name
+                name: teamName
+                widget: string
+                hint: 'Product Management, Engineering, etc.'
+              - label: Team ID
+                name: teamId
+                widget: string
+                hint: 'ID for the URL (all lowercase, no spaces)'
```
chasulin, line 71: will we have URLs that link/scroll directly to certain teams?

chasulin: i guess either way IDs are always good to have

pnavarrc: there is a table of contents at the top (and maybe sidebar too), so we might need the IDs

pnavarrc:
Closing in favour of #1929

PR #1941, opened 2020-11-19
component library setup

pnavarrc:
This PR configures the component library and centralizes the high-level styling elements. This is intended to be a starting point, so it doesn't include all the variables and components that we will need. There are two main things implemented here:

- Create high-level elements on the theme (besides the low-level variables from Material UI)
- Setup Storybook with the accessibility add-on

## Theme and Style Variables

The Material UI theme offers low-level variables for styling, but those variables are not always easy to interpret or use. The Material UI theme variables are applied to all the Material UI components, so attempting to change those variables might break some components.

The idea is to keep the `theme` as the central place to define common styling primitives, but to add convenience variables so they can be reused across the app. For now, I only added 2 categories:

- **colors**: human-named colors to make them easier to use when styling components
- **fonts**: typography groups that better represent our typography units

## Component Library

I configured Storybook and added basic components as a demo. The idea is to have a central place to document the main variations of our components, and try to use those before implementing new variants.  To run Storybook:

- Install new dependencies `yarn`
- Run Storybook `yarn storybook`

The browser should open on http://localhost:6006

## Notes
- Consider installing the [MDX](https://marketplace.visualstudio.com/items?itemName=silvenon.mdx) extension if using VSCode to have syntax highlighting for stories
- Make sure to run `yarn` after changing to a different branch, as `react-scripts` will complain if it finds `babel-loader` to be installed but not included in the project dependencies

### DIFF
--- a/src/assets/theme/index.tsx
+++ b/src/assets/theme/index.tsx
@@ -28,12 +28,12 @@ interface ThemeFonts {
  */
 const fonts: ThemeFonts = {
   subtitle: css`
-    color: rgba(0, 0, 0, 0.7);
+    color: ${COLOR_MAP.GRAY_BODY_COPY};
     text-transform: uppercase;
-    letter-spacing: 0.02em;
-    opacity: 0.7;
-    font-size: 0.8125rem;
+    letter-spacing: 0.01em;
+    font-size: 0.875rem;
     line-height: 1.125;
+    margin-bottom: 0.75rem;
   `,
   disclaimer: css`
     font-size: 0.875rem;
--- a/src/assets/theme/index.tsx
+++ b/src/assets/theme/index.tsx
@@ -1,8 +1,56 @@
 import { createMuiTheme } from '@material-ui/core';
-
+import { css, FlattenSimpleInterpolation } from 'styled-components';
 import palette from './palette';
 import typography from './typography';
 import overrides from './overrides';
+import { COLOR_MAP } from 'common/colors';
+
+interface ThemeColors {
+  green: string;
+  greenDark: string;
+  lightBlue: string;
+}
+
+const colors: ThemeColors = {
+  green: COLOR_MAP.GREEN.BASE,
+  greenDark: COLOR_MAP.GREEN.DARK,
+  lightBlue: COLOR_MAP.BLUE,
+};
+
+interface ThemeFonts {
+  subtitle: FlattenSimpleInterpolation;
+  disclaimer: FlattenSimpleInterpolation;
+  link: FlattenSimpleInterpolation;
+}
+
+/**
+ * High-level typography groups
+ */
+const fonts: ThemeFonts = {
+  subtitle: css`
+    color: rgba(0, 0, 0, 0.7);
+    text-transform: uppercase;
+    letter-spacing: 0.02em;
+    opacity: 0.7;
+    font-size: 0.8125rem;
+    line-height: 1.125;
+  `,
+  disclaimer: css`
+    font-size: 0.875rem;
+    line-height: 1.6;
+  `,
+  link: css`
+    color: ${colors.lightBlue};
+  `,
+};
+
+declare module '@material-ui/core/styles/createMuiTheme' {
+  interface Theme {
+    colors: ThemeColors;
+    fonts: ThemeFonts;
+  }
+  interface ThemeOptions extends Theme {}
+}
 
 const theme = createMuiTheme({
   palette,
@@ -12,6 +60,8 @@ const theme = createMuiTheme({
     appBar: 1200,
     drawer: 1100,
   },
+  colors,
+  fonts,
 });
 
 export default theme;
--- a/src/assets/theme/index.tsx
+++ b/src/assets/theme/index.tsx
@@ -18,7 +18,7 @@ const colors: ThemeColors = {
 };
 
 interface ThemeFonts {
-  subtitle: FlattenSimpleInterpolation;
+  subtitle1: FlattenSimpleInterpolation;
   disclaimer: FlattenSimpleInterpolation;
   link: FlattenSimpleInterpolation;
 }
@@ -27,7 +27,7 @@ interface ThemeFonts {
  * High-level typography groups
  */
 const fonts: ThemeFonts = {
-  subtitle: css`
+  subtitle1: css`
     color: ${COLOR_MAP.GRAY_BODY_COPY};
     text-transform: uppercase;
     letter-spacing: 0.01em;
--- a/src/components/Typography/Typography.style.ts
+++ b/src/components/Typography/Typography.style.ts
@@ -2,7 +2,28 @@ import styled from 'styled-components';
 import Typography from '@material-ui/core/Typography';
 
 /**
- * https://material-ui.com/api/typography/
+ * Notes:
+ *
+ * 1. Use the Typography component from Material UI when possible, since it
+ *    has props that allows to customize some styling (align, gutterBottom, etc),
+ *    which can help us reduce the amount of style overrides.
+ *
+ *    Example:
+ *
+ *        <Subtitle1 align="center">how we determine risk factors</Subtitle1>
+ *
+ *    instead of
+ *
+ *      export const RiskFactorsSubtitle = styled(Subtitle1)`
+ *        text-align: center
+ *      `;
+ *
+ *     <RiskFactorsSubtitle>how we determine risk factors</RiskFactorsSubtitle>
+ *
+ * 2. The `theme` prop here comes from the theme provider from styled components
+ *    (see the `ScThemeProvider` wrapper in App.js. This would allow us to
+ *    dynamically change the theme on specific context (embeds, for example) or
+ *    at the application level (if we were to implement a dark theme).
  */
 
 export const Subtitle1 = styled(Typography).attrs(props => ({
--- a/src/components/Typography/Typography.style.ts
+++ b/src/components/Typography/Typography.style.ts
@@ -3,7 +3,7 @@ import theme from 'assets/theme';
 import Typography from '@material-ui/core/Typography';
 
 export const Subtitle = styled(Typography).attrs(props => ({
-  variant: 'p',
+  variant: 'body1',
 }))`
   ${theme.fonts.subtitle};
 `;
--- a/src/components/Typography/Typography.style.ts
+++ b/src/components/Typography/Typography.style.ts
@@ -1,0 +1,9 @@
+import styled from 'styled-components';
+import theme from 'assets/theme';
+import Typography from '@material-ui/core/Typography';
+
+export const Subtitle = styled(Typography).attrs(props => ({
+  variant: 'p',
+}))`
+  ${theme.fonts.subtitle};
+`;
--- a/src/components/Typography/Typography.style.ts
+++ b/src/components/Typography/Typography.style.ts
@@ -1,9 +1,12 @@
 import styled from 'styled-components';
-import theme from 'assets/theme';
 import Typography from '@material-ui/core/Typography';
 
-export const Subtitle = styled(Typography).attrs(props => ({
+/**
+ * https://material-ui.com/api/typography/
+ */
+
+export const Subtitle1 = styled(Typography).attrs(props => ({
   variant: 'body1',
 }))`
-  ${theme.fonts.subtitle};
+  ${props => props.theme.fonts.subtitle1};
 `;
--- a/src/screens/HomePage/HomePage.style.js
+++ b/src/screens/HomePage/HomePage.style.js
@@ -3,6 +3,7 @@ import { Typography, Box } from '@material-ui/core';
 import palette from 'assets/theme/palette';
 import { COLOR_MAP } from 'common/colors';
 import { mobileBreakpoint } from 'assets/theme/sizes';
+import { Subtitle } from 'components/Typography';
 
 export const SearchBarThermometerWrapper = styled(Box)`
   display: flex;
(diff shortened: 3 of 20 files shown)


src/screens/HomePage/HomePage.style.js
```diff
@@ -3,6 +3,7 @@ import { Typography, Box } from '@material-ui/core';
 import palette from 'assets/theme/palette';
 import { COLOR_MAP } from 'common/colors';
 import { mobileBreakpoint } from 'assets/theme/sizes';
```
mikelehen, line 5: Looks like you have some unused imports (Typography, COLOR_MAP)

src/assets/theme/index.tsx
```diff
@@ -1,8 +1,56 @@
...
+  subtitle: FlattenSimpleInterpolation;
+  disclaimer: FlattenSimpleInterpolation;
+  link: FlattenSimpleInterpolation;
+}
+
+/**
+ * High-level typography groups
+ */
+const fonts: ThemeFonts = {
+  subtitle: css`
```
mikelehen, line 30: FWIW- I worry that "subtitle" may be too generic...  We probably have subtitles on the site that don't use this style.  That said, we can always go with this for now and then rename when we have a clear collision.

Also, it looks like we already have typography.subtitle{1,2} defined in our theme: https://github.com/covid-projections/covid-projections/blob/30a17223005d159353db9aabd180cf0b741682c9/src/assets/theme/typography.tsx#L51

Should we be updating them and using them instead?  Or is there a distinction between `fonts` and `typography` that we want to keep?

On that note, maybe just having multiple numbered subtitle entries (`subtitle1`, `subtitle2`, etc.) is an okay answer for the collision I was worried about.

src/components/Typography/Typography.style.ts
```diff
@@ -0,0 +1,9 @@
+import styled from 'styled-components';
+import theme from 'assets/theme';
+import Typography from '@material-ui/core/Typography';
+
+export const Subtitle = styled(Typography).attrs(props => ({
```
mikelehen, line 5: Per my other comment, "Subtitle" feels like it might be too generic and lead to collisions.

src/components/Typography/Typography.style.ts
```diff
@@ -0,0 +1,9 @@
+import styled from 'styled-components';
+import theme from 'assets/theme';
```
mikelehen, line 2: should we be using `useTheme()` or something so that we actually benefit from the ThemeProvider infrastructure and can replace the theme dynamically as necessary?

mikelehen (approved):
Generally LGTM with a few questions / comments.

src/assets/theme/index.tsx
pnavarrc: For the name, makes sense, maybe `Subtitle1` leaves us more space to have variants in the future. Using fonts instead of the Typography directly allows us to do two things

- Have a better API to export and reuse styles
- Better communicate (in the code) what is the set of typography blocks we want to use

### Better API to export styles

The theme variables are a bit too low-level, and they don't directly provide a way to apply all of them at the same time with the flexibility that we would need. For the Markdown component, for example, we need to define the styling of the content by element type, it would be ideal to just import those styles instead of re-typing them each time:

```tsx
import ReactMarkdown from 'react-markdown';
import theme from 'assets/theme';

export const MarkdownBlock = styled(ReactMarkdown)`
  {/* using an exported css function */}
  h1 {
    ${themes.fonts.heading1};
  }

  {/* using theme variables */}
  h2 {
    font-size: ${themes.typography.h2.fontSize};
    font-family: ${themes.typography.h2.fontFamily};
    {/* etc... */}
  }
`
```

The Heading2 and Link component, for example, is used both inside markdown and outside, and having css functions to grab the styles directly help us centralize and reuse them in both contexts easily.

### Better communicate our typography building blocks

Having the CSS functions being exported as part of the theme allows us to still use the Typography component (to keep the good things, like the ability to set alignment using props).

```tsx
export const Heading1 = styled(Typography)
   .attrs(props => ({ component: 'h1' }))`
      ${theme.fonts.heading1};
   `;
```

which is more explicit (and less verbose) in the client code than using Typography directly.

```tsx
<Heading1 align="center">Title</Heading1>

{/* something like can be confusing */}
<Typography align="center" component="h1" variant="h2">
Title
</Typography>
```

Ideally, we want to have a small-ish set of font styles to be used across the app, and using this structure will help us be more thoughtful when adding a new variant. It would be nice to have components for: Heading1, Heading2, Heading3, Subtitle1, Paragraph, ChartCaption, ChartDisclaimer, Blockquote, Link among others. 

Neither of these arguments is super strong against using Typography or typography variables from the theme directly, it just makes is a bit more convenient (in my opinion). I'm open to other solutions as well, let me know if this makes sense!

src/components/Typography/Typography.style.ts
pnavarrc: We can't use the `useTheme` hook directly on styled components, but we can access the theme variables directly since we have the app wrapped in the styled components theme provider

```tsx
import theme from 'assets/theme';

<ThemeProvider theme={theme}>
  <ScThemeProvider={theme}>
    {/* components here*/}
  <ScThemeProvider/>
</ThemeProvider>
```

in function components, we can use the `useTheme` hook if we need access to theme variables, but for styled components, we should use the direct access to the theme

```ts
// The theme is accessible from ScThemeProvider
const Subtitle1 = styled(Typography).attrs(props => ({ component: 'body1' }))`
  ${props => props.theme.fonts.subtitle1};
`
```

so yes, we should use the theme provider in both cases, but in different ways, will update!

src/assets/theme/index.tsx
pnavarrc: As a side note, subtitles are a bit tricky, because semantically, they belong to the header, but they are styled differently and not always present. 

Using headings to implement them is not correct (Material UI uses H2) and can break the hierarchy of the document, I think that using a block element is probably fine, but I haven't found best practices for them.

ghop02: Quick question on the typography example:
```html
{/* something like can be confusing */}
<Typography align="center" component="h1" variant="h2">
```

Why would variant h2 be remapped to h1 there? I guess I would expect that to just be:
```html
<Typography variant="h1" align="center">
```

pnavarrc: We could, but defining that at the theme level would apply that to all the h1 elements in the app.

src/assets/theme/index.tsx
```diff
@@ -1,8 +1,56 @@
...
+interface ThemeFonts {
+  subtitle1: FlattenSimpleInterpolation;
+  disclaimer: FlattenSimpleInterpolation;
+  link: FlattenSimpleInterpolation;
+}
+
+/**
+ * High-level typography groups
+ */
+const fonts: ThemeFonts = {
```
chasulin, line 29: im confused about this but it might be better discussed on a call tomorrow. i thought the original goal we landed on was to have a set of fonts that are identified by their styling (ie. `monospaceBold`, `regularBold`, `regularItalic` -- regular being our main font ie. Roboto), rather than defining them by their use case (`subtitle1`, `disclaimer`). starting from the point of defining them by their use case feels limiting/not all that flexible

src/assets/theme/index.tsx
ghop02: I think that makes sense! 

Did we explore generating typography compliant code from React Markdown? There seems to be an [example of generating typography elements](https://github.com/mui-org/material-ui/blob/master/docs/src/pages/getting-started/templates/blog/Markdown.js) from markdown (albeit using a slightly different package).  The [thread](https://github.com/mui-org/material-ui/issues/12290) goes into a bit more detail (with a couple alternatives).

My overall thinking here is that it seems like it's a bit more confusing to have to maintain both ways of defining an `h1` (and other similar elements).  But i'm just getting a better sense of the implications of either choice, lmk what I'm missing :)

mikelehen: It sorta' makes sense (it would probably make a lot more sense if I actually understood material-ui theming 😁 )!  I guess `theme.typography` is a builtin material-ui feature that can be used via the builtin `<Typography>` component?  Whereas `theme.fonts` is our own invention that lets us be more explicit and direct.

That seems fine to me, I think...  Though perhaps we should at least add a comment to make it clearer what parts of `theme` are "built-in" material-ui variables and what parts are just our own invention.  We could also use a naming convention...  like `theme.can.*` (e.g. `theme.can.fonts` and `theme.can.colors`) would be our own inventions.

Beyond making that distinction clearer, I do wonder if it makes sense to try to still leverage `typography` and other low-level builtin material-ui theming concepts where it makes sense.  Like we move as much of the subtitle properties as we can directly into `theme.typography.subtitle1` and then reference that from our definition for `theme.fonts.subtitle1` so we don't end up with two unrelated and divergent "definitions" for what our subtitles should look like.  But again, I only sort of understand what's going on, so not 100% sure if this makes sense.

src/assets/theme/index.tsx
```diff
@@ -1,8 +1,56 @@
...
 import overrides from './overrides';
+import { COLOR_MAP } from 'common/colors';
+
+interface ThemeColors {
+  green: string;
+  greenDark: string;
+  lightBlue: string;
+}
+
+const colors: ThemeColors = {
```
ghop02, line 14: I'm not quite sure I understand the limitations of using the colors in the palette rather than defined and included in the theme as an explicit colors property.

Also, looking at our palette, I'm not sure I understand why the values are what they are:
```
primary: {
    contrastText: black,
    dark: white,
    main: white,
    light: white,
  },
  secondary: {
    contrastText: white,
    dark: '#009e53',
    main: '#00d07d',
    light: '#61ffb0',
  },
```
here, primary is just white (? that feels incorrect) and secondary is the green color. 

It feels like the green is the primary color and blue would be the secondary color.  

I mostly bring this up because I think fixing that would simplify some of the other components and add some flexibility.  I.e. for almost all of the material ui components, you could then use the color key without having to manually add `color:` in a styled component.

pnavarrc: The Material design system is supposed to be customized as a whole, in a manner that is consistent with your brand (see [Crane](https://material.io/design/material-studies/crane.html) and other Material design studies). For most projects on their early stages, it's hard to have a comprehensive design system right from the start, and at this point, we don't have one.

The Material UI color variables are applied to all the Material UI components, and can be used by other components as well. Right now, it would be a bit risky to change the value of a color in the theme, because we don't know the repercussions that that would have in the components we use (if we change secondary to be light blue, will that change the color on the MiniMap tabs?).

The variables are also not super readable, so having color variables with better names help us use them more consistently in our components. If at some point we decide to restyle our site, we can style at the theme level and change the values of named colors, but we would need the design system first.

ghop02: > The Material design system is supposed to be customized as a whole, in a manner that is consistent with your brand (see [Crane](https://material.io/design/material-studies/crane.html) and other Material design studies). For most projects on their early stages, it's hard to have a comprehensive design system right from the start, and at this point, we don't have one.
> 

I think Crane is a great example of how the design system really shines, but I don't think you have to be as structured and complex to take advantage of it.  We do have a primary/secondary color that we use consistently throughout the site so I think it fits in fairly nicely with what we have. 

> The Material UI color variables are applied to all the Material UI components, and can be used by other components as well. Right now, it would be a bit risky to change the value of a color in the theme, because we don't know the repercussions that that would have in the components we use (if we change secondary to be light blue, will that change the color on the MiniMap tabs?).

I might be missing something, but I don't think that it would be too much work to clean up the colors as we have them.  There are definitely some changes, but for the most part we don't rely on the default colors used from the theme because we never really set it up properly - I made a [quick attempt at it](https://github.com/covid-projections/covid-projections/compare/pablo/component-library-setup...ghop02/component-library-but-more-centralized-colors?expand=1) and while it may take a slightly closer look I don't think it's a very drastic change. 

> The variables are also not super readable, so having color variables with better names help us use them more consistently in our components. If at some point we decide to restyle our site, we can style at the theme level and change the values of named colors, but we would need the design system first.

I'm not sure I quite understand this point? Do you have a couple examples of where we'd have to step outside of the palette to choose a color? haha I think that would be helpful so I make sure I'm not overlooking any huge issues with it

chasulin: i dont know if crane is a super applicable example. its quite theme-y/template-y, whereas our frontend is comprised of a lot of customized components that all kind of use colors differently.

re: the notion of primary/secondary, i dont think we currently have a convention that fits this exactly. green is the color of our logo, the footer, and the 'low' threshold, but thats about it? likewise our use of blue is sometimes a little disorganized.

i feel like if we want to use palette (primary, secondary, etc), we should have an index file of our own making in which we define/name these colors (green, blue, etc), and then use them to set primary/etc in palette. ie. primary.main: theme.blue

chasulin (commented):
@pnavarrc @mikelehen @ghop02 - let's move this convo out of PR comments and into a meeting if that's cool with everyone. i'll send an invite for tomorrow

src/assets/theme/index.tsx
pnavarrc: we probably need to do both, right now there are more high-level elements repeated (headings, disclaimers, etc) than low-level elements (regularBold)

src/assets/theme/index.tsx
ghop02: Very true re notion of primary/secondary and it's not really obvious which one you would use.  (Sidenote but it's really fun looking at some of the other color schemes on that material design page.  [Rally](https://material.io/design/material-studies/rally.html#color) and [owl](https://material.io/design/material-studies/owl.html#color) do some fun things with color.  Rally feels closest to us, where they just have a primary color and then a couple others to choose from)

I definitely agree and also I'm realizing that I'm not putting a ton of semantic weight on what primary and secondary mean - I think the `COLOR_MAP` constant is a bit chaotic right now, and I like the fact that the theme palette is constrained more than anything.

Here's an idea (that we can talk about in the meeting today):

* change `ThemeColors` to be an object of `[colorName]: PaletteColor` (PaletteColor is a [material ui type](https://material-ui.com/customization/palette/#customization).  That way, we can customize the material-ui palette with the colors we have.  For instance here's a quick snippet of it:

```ts
const colors: ThemeColors = {
  green: {
    main: '#00D474',
    dark: '#0A3D31',
    light: '...'
  },
  blue: {
    main: '#3BBCE6',
    light: '...'
  },
  orange: {
    ...
  },
   ...
};

// This would be used in the mui theme
const palette = {
  primary: colors.green,
  secondary: colors.blue,
  warning: colors.orange,
};
```

`colors` would essentially replace `COLOR_MAP`. My thinking is that giving each color a "main", "light", and "dark" will encourage the most reusability while giving flexibility.

PR #1952, opened 2020-11-21
small accessibility fixes

pnavarrc:
Found while running an audit, there is a couple of empty links and a missing alt text on an image


mikelehen (approved):
👍 
Thanks!

PR #1964, opened 2020-11-23
Fix extra rendering of map on hover tooltip change.

mikelehen:
Fixes https://trello.com/c/FnNibaOg/392-tooltip-causing-home-page-map-to-be-a-bit-laggy cc/ @ghop02 

I was playing with adding counties back to the homepage map and it was super slow without this...

### DIFF
--- a/src/components/Map/Map.js
+++ b/src/components/Map/Map.js
@@ -24,29 +24,35 @@ function Map({
   const history = useHistory();
   const [content, setContent] = useState('');
 
-  const goToStatePage = page => {
-    window.scrollTo(0, 0);
-    history.push(page);
-  };
+  const goToStatePage = React.useCallback(
+    page => {
+      window.scrollTo(0, 0);
+      history.push(page);
+    },
+    [history],
+  );
 
-  const handleClick = stateName => {
-    // externally provided click handler
-    if (onClick) {
-      return onClick(stateName);
-    }
+  const handleClick = React.useCallback(
+    stateName => {
+      // externally provided click handler
+      if (onClick) {
+        return onClick(stateName);
+      }
 
-    const stateCode = REVERSED_STATES[stateName];
+      const stateCode = REVERSED_STATES[stateName];
 
-    goToStatePage(`/us/${stateCode.toLowerCase()}`);
+      goToStatePage(`/us/${stateCode.toLowerCase()}`);
 
-    if (setMapOption) {
-      setMapOption(MAP_FILTERS.STATE);
-    }
+      if (setMapOption) {
+        setMapOption(MAP_FILTERS.STATE);
+      }
 
-    if (setMobileMenuOpen) {
-      setMobileMenuOpen(false);
-    }
-  };
+      if (setMobileMenuOpen) {
+        setMobileMenuOpen(false);
+      }
+    },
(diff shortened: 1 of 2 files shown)


src/components/Map/Map.js
```diff
@@ -24,29 +24,35 @@ function Map({
...
-    if (setMobileMenuOpen) {
-      setMobileMenuOpen(false);
-    }
-  };
+      if (setMobileMenuOpen) {
+        setMobileMenuOpen(false);
+      }
+    },
+    [onClick, goToStatePage, setMapOption, setMobileMenuOpen],
+  );
```
pnavarrc, line 55: I don't think this code is used anymore now that the states are links

src/components/Map/USACountyMap.js
```diff
@@ -34,81 +34,83 @@ const MarianaIslands = ({ fill, onMouseEnter, onMouseLeave }) => {
...
+                        <Link
+                          key={stateCode}
+                          to={stateUrl}
+                          aria-label={locationName}
+                        >
+                          <MarianaIslands
+                            key={geo.rsmKey}
+                            onMouseEnter={() => setTooltipContent(name)}
+                            onMouseLeave={onMouseLeave}
+                            onClick={() => stateClickHandler(name)}
```
pnavarrc, line 81: I think we can get rid of the click handler

src/components/Map/Map.js
```diff
@@ -24,29 +24,35 @@ function Map({
...
-  const goToStatePage = page => {
-    window.scrollTo(0, 0);
-    history.push(page);
-  };
+  const goToStatePage = React.useCallback(
+    page => {
+      window.scrollTo(0, 0);
+      history.push(page);
+    },
+    [history],
```
pnavarrc, line 32: This might not work as intended, because the [history object](https://reactrouter.com/web/api/history) is a mutable object which makes detecting changes to it a bit unpredictable. We might not need this function anyways, so that's probably okay (we do listen to `history.location` in other places)

pnavarrc (commented):
Added some notes, we can probably get rid of the click handler all together.

src/components/Map/Map.js
mikelehen: Thanks for the flag.  I think it's okay if it mutates... it just means we'll re-generate this callback, which isn't harmful (except it causes the map to rerender unnecessarily, but that's okay as long as it's not frequent).

src/components/Map/Map.js
mikelehen: Oooh, interesting!  Unfortunately I tried to remove this, but got a bit nervous when I looked at the usages.  I _think_ it'd probably be fine to remove from the location page MiniMap, but our embed uses the onClick handler to redirect to the CAN homepage (rather than the state you clicked on), so I think that would break unless we rework that.

So for now I'm adding a TODO, but this really would be nice to clean up.  I just don't want to break something right now. 😬

src/components/Map/USACountyMap.js
mikelehen: keeping it for now. 😢

mikelehen:
This got merged as part of the 5th color work.

PR #1975, opened 2020-11-24
Demo of adding 5th color (purple) to maps and case density metric.

mikelehen:
* Added purple ("Super Critical") level which is only used by case incidence and the overall risk level (and therefore map coloring).
* Case density threshold is 75 cases/day.

image

image


src/common/metrics/interfaces.ts
```diff
@@ -7,5 +7,7 @@ export interface MetricDefinition {
   metricName: string;
   extendedMetricName: string;
   metricNameForCompare: string;
+
+  // TODO(michael): This seems to be unused. We should probably remove it!
```
pnavarrc, line 11: This will be used on the metric explainers, we implemented the component for each metric in advance

pnavarrc (approved):
Looks great!

pnavarrc:
[Change the role of geography elements on the map (Location page)](https://trello.com/c/yDnX0w3S/620-change-the-role-of-geography-elements-on-the-map-location-page)

PR #1989, opened 2020-11-25
Learn updates

chasulin:
- Rebrands 'learn' --> 'explained'
- Launches 'deep dives'
- Adds 3rd wave article to homepage banner + location page header


src/cms-content/articles/can-compare.json
```diff
@@ -0,0 +1,7 @@
...
+}
```
pnavarrc, line 7: Is it intentional that the CAN Analytics links go to Explore in California? Also, not sure if we want to keep that name for the feature

src/App.js
```diff
@@ -107,10 +106,11 @@ export default function App() {
                 path="/us/:stateId/county/:countyId/compare/:sharedComponentId?"
                 component={LocationPage}
               />
-              <Route exact path="/learn" component={Landing} />
+              <Route exact path="/explained" component={Landing} />
```
pnavarrc, line 109: We need to add a redirect for `/learn` to ensure that links in the wild are not broken

pnavarrc (approved):
Looks good!

PR #1995, opened 2020-11-25
Use summary text instead of enum name in alert email alt text.

mikelehen:
While testing the new thermometer images, I noticed that our alt text was non-ideal.

*Before:* Image depicting that District of Columbia went from level HIGH to SUPER_CRITICAL
*After:* Image depicting that District of Columbia went from "At risk of outbreak" to "Severe outbreak"


pnavarrc (approved):
Looks good!

PR #2000, opened 2020-11-26
explore: minor fixes

pnavarrc:
from Igor on Slack. This PR moves the checkbox and description of the Native American population feature below the explore chart and simplifies a bit the copy, to make the chart more prominent. I also reduced the number of preselected states to 5 so the search box takes less vertical space on Desktop.

image

We also want to move "Learn more about our methodology or view our observations." to the grey box below, but that involves moving the modal logic to the Explore component, I prefer to do that on a separate PR.


mikelehen:
[Update Email Footer to Include 5th Level Language](https://trello.com/c/ejdEJAFW/627-update-email-footer-to-include-5th-level-language)

mikelehen (approved):
LGTM except when you click "our methodology" and dismiss the modal, the explore chart ends up scrolled out of view.  We maybe want to at least add an offset or something to try to keep explore visible.

As an aside, have we looked into using [Dialog](https://material-ui.com/components/dialogs/) instead of Modal?  I notice that the Modal comments say "If you are creating a modal dialog, you probably want to use the Dialog component rather than directly using Modal." ...  so I'm wondering how we ended up using Modal directly and whether Dialog would save us any of these scrolling headaches.  cc/ @chasulin

pnavarrc:
> As an aside, have we looked into using Dialog instead of Modal?

Yes, we do have a `Dialog` component for this. Before this PR, it had the defect of losing the scroll position when closing, which was due to a bad implementation of scroll lock in the component. The last few commits use the `Dialog` component, add it to Storybook and update the Explore component to use it.

src/components/Dialog/Dialog.tsx
```diff
@@ -48,12 +48,19 @@ const Dialog: React.FC<DialogProps> = props => {
       onClose={dismissDialog}
       PaperComponent={Style.StyledPaper}
       fullScreen={fullScreen}
+      aria-labelledby="dialog-title"
+      disableScrollLock={false}
       {...otherProps}
     >
       <LockBodyScroll />
       <ButtonCloseBox onClickClose={closeDialog} />
```
mikelehen, line 56: FWIW- The close button feels too close to the edges to me, and it's actually pushing the content of the dialog down in a way that doesn't seem desirable.

It seems better if I change the [ButtonContainer](https://github.com/covid-projections/covid-projections/blob/753cdf547a48f31749f095d29e36107f34d7e266/src/components/Dialog/Dialog.style.ts#L12) style to:
```
  position: absolute;
  top: 10px;
  right: 10px;
```

Before: 
image

After:
image

Feel free to leave as-is if you disagree though. I guess we've been using this as-is in the recommends modal for a while.

mikelehen (approved):
LGTM though we lost some padding around the close button. I gave a potential fix, but not sure if it's ideal.  Feel free to punt for now.

mikelehen:
🙏

PR #2002, opened 2020-11-27
Fix inconsistent tracking of switch component clicks due to extra event callback.

mikelehen:
SwitchComponent had two separate callbacks (onChange and onGridClick) and we were only tracking analytics events when you hit the onChange handler (concretely, if you hit the actual switch GUI then `onChange` would fire and we'd track it; but if you clicked on the "States" or "Counties" labels, then `onGridClick` would fire instead and we wouldn't track it). 

I reworked it to only have a single event callback and also added it to storybook and did some minor cleanup.

### DIFF
--- a/src/components/Compare/Filters.tsx
+++ b/src/components/Compare/Filters.tsx
@@ -97,9 +97,8 @@ const Filters = (props: {
     },
   ];
 
-  const switchHandleChange = () => {
+  const switchHandleChange = (newViewAllCounties: boolean) => {
     if (props.isHomepage && props.setViewAllCounties) {
-      const newViewAllCounties = !props.viewAllCounties;
       props.setViewAllCounties(newViewAllCounties);
       trackCompareEvent(
         EventAction.SELECT,
@@ -143,16 +142,11 @@ const Filters = (props: {
   const theme = useTheme();
   const isMobile = useMediaQuery(theme.breakpoints.down('xs'));
 
-  const gridOnClick = (on: boolean) => {
-    props.setViewAllCounties && props.setViewAllCounties(on);
-  };
-
   const switchProps = {
     leftLabel: 'States',
     rightLabel: 'Counties',
-    checkedValue: props.viewAllCounties,
+    checked: props.viewAllCounties,
     onChange: switchHandleChange,
-    gridOnClick,
     isModal: props.isModal,
   };
 
--- a/src/components/SharedComponents/SwitchComponent.stories.mdx
+++ b/src/components/SharedComponents/SwitchComponent.stories.mdx
@@ -1,0 +1,10 @@
+import { Meta, Story } from '@storybook/addon-docs/blocks';
+import SwitchComponent from './SwitchComponent';
+
+<Meta title="Shared Components/Switch Component" argTypes={{ onChange: { action: 'changed' } }} />
+
+export const Template = (args) => <SwitchComponent {...args} />
+
+<Story name="Default" args={{leftLabel: 'Left', rightLabel: 'Right'  }}>
+  {Template.bind({})}
+</Story>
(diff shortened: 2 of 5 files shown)


src/components/Compare/Filters.tsx
```diff
@@ -143,16 +142,11 @@ const Filters = (props: {
...
 
-  const gridOnClick = (on: boolean) => {
-    props.setViewAllCounties && props.setViewAllCounties(on);
-  };
-
   const switchProps = {
     leftLabel: 'States',
     rightLabel: 'Counties',
-    checkedValue: props.viewAllCounties,
+    checked: props.viewAllCounties,
```
mikelehen, line 148: Renamed this just for consistency with the underlying component we're wrapping.

src/components/SharedComponents/SwitchComponent.stories.mdx
```diff
@@ -0,0 +1,10 @@
+import { Meta, Story } from '@storybook/addon-docs/blocks';
+import SwitchComponent from './SwitchComponent';
+
+<Meta title="Shared Components/Switch Component" argTypes={{ onChange: { action: 'changed' } }} />
```
mikelehen, line 4: FWIW- I saw that you have some global argTypesRegex thing: https://github.com/covid-projections/covid-projections/blob/f9c78b6eb942f585dddccdcb6aed3f486125b79a/.storybook/preview.js#L11

Not sure if I can/should be integrating with that somehow.

src/components/SharedComponents/SwitchComponent.style.tsx
```diff
@@ -7,24 +7,24 @@ export const SwitchLabel = styled(Grid)`
 `;
 
 export const SwitchGrid = styled(Grid)<{
-  viewAllCounties: boolean | undefined;
+  checked: boolean | undefined;
```
mikelehen, line 10: I assume `viewAllCounties` was from before we made this a shared component.  Renamed to `checked` so it's more generic. 😁

src/components/SharedComponents/SwitchComponent.tsx
```diff
@@ -5,40 +5,33 @@ import { SwitchGrid, SwitchLabel } from './SwitchComponent.style';
 const SwitchComponent = (props: {
   leftLabel: string;
   rightLabel: string;
-  checkedValue: boolean | undefined;
-  onChange: any;
-  gridOnClick: any;
+  checked: boolean | undefined;
+  onChange: (newValue: boolean) => void;
```
mikelehen, line 9: Removed `gridOnClick` and changed `onChange` so it takes a boolean so it can fill the role.

src/components/SharedComponents/SwitchComponent.style.tsx
```diff
@@ -7,24 +7,24 @@ export const SwitchLabel = styled(Grid)`
 `;
 
 export const SwitchGrid = styled(Grid)<{
-  viewAllCounties: boolean | undefined;
-  isModal?: boolean;
+  $checked: boolean | undefined;
+  $isModal?: boolean;
```
mikelehen, line 11: Made these into [transient props](https://styled-components.com/docs/api#transient-props) to avoid a React warning about invalid props being passed to a DOM component.

src/components/SharedComponents/SwitchComponent.stories.mdx
pnavarrc: This is added by default by Storybook, I'm not sure if it does anything 🤷‍♂️

pnavarrc (approved):
Looks good, thanks for improving the component! I think we need to provide a way to attach a label to this component (and fix the focused state) to make it more accessible, but is a good start!

PR #2010, opened 2020-11-28
add isModal to the Switch story

pnavarrc:
not the ideal solution, but it works for now.

### DIFF
--- a/src/components/SharedComponents/SwitchComponent.stories.mdx
+++ b/src/components/SharedComponents/SwitchComponent.stories.mdx
@@ -1,10 +1,22 @@
 import { Meta, Story } from '@storybook/addon-docs/blocks';
 import SwitchComponent from './SwitchComponent';
+import { withDarkBackground } from 'components/storybook';
 
-<Meta title="Shared Components/Switch Component" argTypes={{ onChange: { action: 'changed' } }} />
+<Meta
+  title="Shared Components/Switch Component"
+  component={SwitchComponent}
+  argTypes={{ onChange: { action: 'changed' } }}
+  args={{ leftLabel: 'Left', rightLabel: 'Right' }}
+/>
 
-export const Template = (args) => <SwitchComponent {...args} />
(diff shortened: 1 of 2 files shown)


src/components/SharedComponents/SwitchComponent.stories.mdx
```diff
@@ -1,10 +1,22 @@
...
-<Meta title="Shared Components/Switch Component" argTypes={{ onChange: { action: 'changed' } }} />
+<Meta
+  title="Shared Components/Switch Component"
+  component={SwitchComponent}
+  argTypes={{ onChange: { action: 'changed' } }}
+  args={{ leftLabel: 'Left', rightLabel: 'Right' }}
+/>
 
-export const Template = (args) => <SwitchComponent {...args} />
+export const Template = args => <SwitchComponent {...args} />;
```
mikelehen, line 12: On a tangent, I notice you removed the parens around `(args)`...  normally we wouldn't have to think about that sort of thing, since prettier takes care of formatting automatically.  But I guess since this is `mdx` it doesn't.

I am a bit curious if we think we'll actually use markdown enough to make using `mdx` worthwhile.  All of our tooling would probably work better if we just used normal typescript (prettier would work, you wouldn't need an extra extension in VS Code, the error messages storybook prints out might be simpler?).  And then our code would look more like the example code in the "Writing Stories" docs (https://storybook.js.org/docs/react/writing-stories/introduction).  I was a bit confused at first when I was trying to read their docs.

I don't care strongly though.  We can stick with MDX if we think it'll be useful.

mikelehen (approved):
Ah, nice! Thanks for figuring this out.

src/components/SharedComponents/SwitchComponent.stories.mdx
pnavarrc: That makes sense, I can migrate the files on a separate PR (we can still use MDX if we need to write Markdown in some cases, but most times we don't)

PR #2019, opened 2020-11-30
Remove college tag

chasulin:


src/components/Compare/ModalCompare.tsx
```diff
@@ -11,8 +11,7 @@ import {
 import Filters from 'components/Compare/Filters';
 
 const BodyScrollLock = createGlobalStyle`
-  body {
-    height: 100vh;
+  html {
```
pnavarrc, line 14: Consider importing the component `LockBodyScroll` from `components/Dialog` (we might need to export it) to avoid code duplication (also for `ModalFaq`)

pnavarrc (approved):
Looks good!

PR #2021, opened 2020-11-30
Recommendations updates

chasulin:
- Adds travel recommendation
- Divides recommendations into columns to solve for wonky spacing/gaps between recommendations of different lengths


pnavarrc (approved):
Looks good!

PR #2022, opened 2020-12-01
Alert when the location is on a hospitalization peak

pnavarrc:
It adds the `defaultMetric` parameter to the Explore component to allow the container component to set the initially selected metric. Examples:

- [New Mexico](https://covid-projections-git-pablo-hospitalization-peak.covidactnow.vercel.app/us/nm?s=1380364) - showing the alert
- [Montana](https://covid-projections-git-pablo-hospitalization-peak.covidactnow.vercel.app/us/montana-mt?s=1380364) - not showing the alert


src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -60,9 +60,14 @@ const ChartsHolder = (props: {
   const theme = useTheme();
   const isMobile = useMediaQuery(theme.breakpoints.down('xs'));
 
-  const { pathname } = useLocation();
+  const { pathname, hash } = useLocation();
   const isRecommendationsShareUrl = pathname.includes('recommendations');
 
+  const defaultMetric =
+    hash === '#explore-chart'
```
mikelehen, line 67: Should this be "#explore-hospitalizations" maybe?  Else add a comment about how this hash is used to make this code clearer?

And rename "defaultMetric" to "defaultExploreMetric" ?

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -60,9 +60,14 @@ const ChartsHolder = (props: {
...
+      ? ExploreMetric.HOSPITALIZATIONS
+      : ExploreMetric.CASES;
```
mikelehen, line 69: FWIW- We actually have existing navigation / scrolling set up for URLs like https://covidactnow.org/us/wa/explore/hospitalizations (this still works from the original version of explore chart sharing), so if we thought it was useful we could leverage that.  But since you have this working, I assume it makes sense to just ship it!

src/components/LocationPage/utils.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { Fragment } from 'react';
+import moment from 'moment';
+import { maxBy } from 'lodash';
+import { Projection, Column } from 'common/models/Projection';
+import { HashLink } from 'react-router-hash-link';
+import { scrollWithOffset } from 'components/TableOfContents';
+
+const numDays = 7;
```
mikelehen, line 8: Can we get a more descriptive name or else move this inside isHospitalizationsPeak()?  Maybe HOSPITALIZATIONS_PEAK_LOOKBACK_DAYS or something?

src/components/LocationPage/utils.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const { locationName } = projection;
+  const peak = findHospitaliztionsPeak(projection);
+
+  if (!peak || !peak.y) {
+    return null;
+  }
+
+  return (
+    <Fragment>
+      {locationName} COVID hospitalizations are at an all-time peak -{' '}
```
mikelehen, line 35: FWIW- It feels a little weird to have this copy buried in a utils.tsx file.  I think of utils as a little more general-purpose than that.

src/components/LocationPage/LocationPageHeader.tsx
```diff
@@ -47,6 +48,12 @@ const NewFeatureCopy = (props: {
   );
 };
 
+const HospitalizationsAlert: React.FC<{ projection: Projection }> = ({
+  projection,
+}) => {
+  return <Copy>{getHospitalizationsAlert(projection)}</Copy>;
```
mikelehen, line 54: Not really sure what the point of this component is if it basically just defers to `getHospitalizationsAlert()`.  I feel like this could be refactored a bit (probably to pull the logic from getHospitalizationsAlert() into this component?).

src/components/LocationPage/LocationPageHeader.tsx
```diff
@@ -110,6 +117,8 @@ const LocationPageHeader = (props: {
   const lastUpdatedDateString =
     lastUpdatedDate !== null ? formatUtcDate(lastUpdatedDate) : '';
 
+  const isIcuPeak = isHospitalizationsPeak(props.projections.primary);
```
mikelehen, line 120: not icu anymore.

mikelehen (approved):
Sorry for being slow. This looks good to me! Thanks for knocking it out so quickly.  I left some nitty comments, but feel free to punt.

PR #2026, opened 2020-12-01
About page CMS config updates

chasulin:


public/admin/config.yml
```diff
@@ -47,7 +47,40 @@ collections:
...
+          - {
+              label: 'Who we serve: Header',
+              name: 'whoWeServeHeader',
+              widget: 'string',
+            }
+          - {
+              label: 'Who we serve: Content',
+              name: 'whoWeServeContent',
+              widget: 'markdown',
+            }
```
pnavarrc, line 59: Consider adding a `sectionID` to each section on this page (we might need to add a table of contents/sidebar here).

chasulin: good call

chasulin: update: punting this for now

PR #2036, opened 2020-12-01
use h1 in the location page

pnavarrc:
The state or county name in the location page looks like a heading, but is a styled div. This makes the page more annoying to navigate for users that need assistive technologies and hurts SEO.

This PR updates the Location page header to use the H1 element. It should be visually identical than the exiting title. I also removed the link on the state code in the header, I don't think that that feature was useful, as it was invisible for users.

### References
- https://web.dev/headings-and-landmarks/

### Notes
- There are other headings to add as well (chart titles should be H2, etc), will work on those on separate PRs

### DIFF
--- a/src/assets/theme/index.tsx
+++ b/src/assets/theme/index.tsx
@@ -37,10 +37,12 @@ const fonts: ThemeFonts = {
     margin-bottom: 0.75rem;
   `,
   heading1: css`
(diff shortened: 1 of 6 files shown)


pnavarrc:
Closing for now, will probably use aria-labels and roles instead until we implement the new designs.

PR #2043, opened 2020-12-02
centralize data structures and methods for locations

pnavarrc:
_work in progress_

### Centralize data access for a given location

We don't have a unified data structure for locations, but we do have a unified UI for counties and states, which forces us to handle both cases in multiple places, notably when we want to render the location name, grab the FIPS code or need the canonical URL for the page. The idea is to have shared methods that we can rely on independent of the type of location

```ts
region.fullName()
// Washington (State)
// King County, WA (County)
// Austin, TX (for MSAs)
```

### Filtering locations

The idea is to have a couple of methods as shortcuts for the most common queries, and leave the rest to the client code. Example:

```ts
const allRegions = regions.all();

const states = regions.states();

const texasCounties = regions.counties()
  .filter(county => county.state.stateCode === 'TX');

const collegeCounties = regions.counties()
  .filter(county => county.isCollegeCounty);

const top50Counties = chain(counties)
  .sortBy(county => county.population)
  .takeRight(50)
  .value();
```

### Finding locations by URL parameters

Idea: create separate routes for each location type and create a region provider using React context. This eliminates the need of having to handle the case where the location is not found, and make the region information available to all the inner components

```tsx
<Route path="/us/:stateId" component={StatePage} />
<Route path="/us/:stateId/county/:countyId" component={CountyPage} />
<Route path="/us/metro/:msaId" component={MsaPage} />
```

and then have a wrapper to handle the case where the `msaId` is not valid. This will allow us to ensure the existence of the location for all the inner components, avoiding having to deal with variations or missing fields from that point on

```tsx
import regions from 'common/regions';

const MsaPage = () => {
  const { msaId } = useParams<{ msaId: string}>();
  const msa = regions.findMsaByUrlParam(msaId);
  if (!msa) {
    // redirect to /
  }  

  return (
    <RegionProvider region={region}>
      <SharedLocationPage />
    </RegionProvider>
  );
}
```

and then inside `SharedLocationPage`, we can use the hook `useRegion` and rely on the fields and methods common to all region types (including `regionType`, in case we need to customize the behaviour of a component)

```tsx
const region = useRegion();

// ...
<Heading1>{region.fullName()}</Heading1>
```


src/common/regions.ts
```diff
@@ -0,0 +1,146 @@
+import { chain, fromPairs } from 'lodash';
+import urlJoin from 'url-join';
+import US_STATE_DATASET from 'components/MapSelectors/datasets/us_states_dataset_01_02_2020.json';
+import countyAdjacencyMsa from './data/county_adjacency_msa.json';
+const { state_dataset, state_county_map_dataset } = US_STATE_DATASET;
+
+export type FipsCode = string;
+
+export enum RegionType {
```
ghop02, line 9: It would be great if the values of this could match (or just be) [AggregationLevel](https://github.com/covid-projections/covid-projections/blob/develop/src/api/schema/RegionSummary.d.ts#L26) so it matches the values from the API

src/common/regions.ts
```diff
@@ -0,0 +1,146 @@
...
+}
+
+export class County extends Region {
+  constructor(
+    public name: string,
+    public urlName: string,
+    public fipsCode: FipsCode,
+    public population: number,
+    public state: State,
+    public cities: string[],
```
ghop02, line 58: small, but `cityNames` might be more clear here.  Based on `state` above being an object, my first thought was that `cities` might also be Region objects

src/common/regions.ts
```diff
@@ -0,0 +1,146 @@
...
+import US_STATE_DATASET from 'components/MapSelectors/datasets/us_states_dataset_01_02_2020.json';
+import countyAdjacencyMsa from './data/county_adjacency_msa.json';
+const { state_dataset, state_county_map_dataset } = US_STATE_DATASET;
+
+export type FipsCode = string;
+
+export enum RegionType {
+  COUNTY = 'County',
+  STATE = 'State',
+  MSA = 'MSA',
```
ghop02, line 12: This is a technicality, but the region we use from the API is cbsa, not msa.  They're close to each other, but would rather get in the habit of using CBSA since we consistently use it on the API side

src/common/regions.ts
```diff
@@ -0,0 +1,146 @@
...
+}
+
+abstract class Region {
+  constructor(
+    public name: string,
+    public urlName: string,
```
ghop02, line 18: for some reason my first thought is that url name would be the full relative url - I'm not sure if this is something other people think also or if it's just me :)

src/common/regions.ts
```diff
@@ -0,0 +1,146 @@
...
+      countyInfo.full_fips_code,
+      countyInfo.population,
+      state,
+      countyInfo.cities || [],
+      adjacentCounties || [],
+    );
+  })
+  .value();
+
+class RegionDB {
```
ghop02, line 117: I'm not 100% sure, but using a class like this feels like it might not be as flexible as using a hook to filter by different regions.  

My thinking is that there are cases where we'd want to work with subsets of regions, and having selectors that will update if the subset of regions changes would be useful. 

I guess if this is the global list of regions, that should be fine, but it might be something to think about for different access patterns.

ghop02 (commented):
Looks great! Just had a couple thoughts reading through :)

src/common/regions.ts
pnavarrc: The URL segment is loaded from the JSON file, it's easier to set the segment and construct the URL than to update the JSON file

src/common/regions.ts
pnavarrc: I reviewed the access patterns, these are the ones I could find:
- all counties
- all states
- all counties in a state
- all metro counties in a state
- all non-metro counties in a state
- all counties adjacent to a given county

let me know if you find missing queries. I think that this is enough for now, as we can always filter directly using `.all()` or with the convenience methods by region type

src/common/regions.ts
pnavarrc: Makes sense, will update!

src/common/regions.ts
pnavarrc: Good catch, will update!

src/common/regions.ts
ghop02: oh right.  maybe `urlSegment` actually? but i'm probably overthinking it :)

pnavarrc: I'm going to make it private, `urlSegment` is a bit better I think

pnavarrc:
[Improve location data structure to prepare for CBSAs](https://trello.com/c/70MemmzX/640-improve-location-data-structure-to-prepare-for-cbsas)

pnavarrc:
[Improve location data structure to prepare for CBSAs](https://trello.com/c/70MemmzX/640-improve-location-data-structure-to-prepare-for-cbsas)

src/common/regions.ts
```diff
@@ -0,0 +1,146 @@
...
+    protected urlSegment: string,
+    public fipsCode: FipsCode,
+    public population: number,
+    public regionType: RegionType,
+  ) {}
+
+  abstract fullName(): string;
+  abstract relativeUrl(): string;
+  canonicalUrl() {
+    return urlJoin('https://covidactnow.org', this.relativeUrl());
```
ghop02, line 27: Do we have some sort of environment variables for this instead of hardcoding the url?

src/common/regions.ts
mikelehen: I'm not 100% sure what "using a hook to filter by different regions" refers to.  If we're talking about a react hook, that feels like it could easily built on top of this, so this still feels like a good primitive.

mikelehen (changes requested):
This looks like a great basis for cleaning up the region code!  I left some comments but very much on-board with this direction.

src/common/regions.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  test('counties() returns all the counties', () => {
+    const counties = regions.counties();
+    expect(counties).toHaveLength(3224);
+  });
+
+  test('findByFipsCode returns a state for a state FIPS', () => {
+    const dc = regions.findByFipsCode('11');
+    expect(dc).toBeTruthy();
+    expect(dc?.regionType === RegionType.STATE);
```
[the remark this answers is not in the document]
pnavarrc: oops! that's a mistake, fixing!

src/common/regions.ts
mikelehen: FWIW- We have a common/urls.ts file where I have been haphazardly trying to unify url handling...  Note that we're currently using `window.location.origin` instead of hardcoding the base URL at least in some cases.  This allows it to work on vercel preview links, etc.
https://github.com/covid-projections/covid-projections/blob/6849cf1245480ab97979ba93be2f3fd732047f36/src/common/urls.ts#L41

Also, FWIW- we do have infrastructure set up to detect whether we're on prod/staging/dev via 
https://github.com/covid-projections/covid-projections/blob/6849cf1245480ab97979ba93be2f3fd732047f36/src/common/utils/environment.ts#L7

We could use that to decide what URL to use, but as alluded to, that won't work for vercel preview links...

pnavarrc:
> I'm not 100% sure what "using a hook to filter by different regions" 

That's weird, I probably didn't finish editing that sentence 😂 , I'm proposing 2 ideas:

- Having a React hook to access the region object from any component inside the `RegionProvider`
- Having a `regions` instance that can return lists of locations with some convenience methods for the most common filters (`.all()`, `.states()`, `.counties()` and maybe others)

src/common/regions.ts
```diff
@@ -0,0 +1,146 @@
...
+// TODO: Aggregated regions
+
+const statesByFips = fromPairs(states.map(state => [state.fipsCode, state]));
+
+interface AdjacencyInfo {
+  adjacent_counties: FipsCode[];
+  msa_code?: string;
+}
+
+const adjacency: { [fipsCode: string]: AdjacencyInfo } =
```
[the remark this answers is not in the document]
pnavarrc: That was just giving types to the content of the file with adjacency data [common/data/county_adjacency_msa.json](https://github.com/covid-projections/covid-projections/blob/develop/src/common/data/county_adjacency_msa.json). That file also tells wether or not the county belongs to an MSA, which we use for the metro / non-metro filter.

I'm not sure if I actually need to create types for it, I might be able to just use the content to attach the adjacency county list to each county without creating an interface.

src/common/regions.test.ts
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  test('canonicalUrl', () => {
+    expect(state?.canonicalUrl()).toBe('https://covidactnow.org/us/alaska-ak');
+  });
+});
+
+describe('County', () => {
+  const county = regions.findByFipsCode('53033');
+  test('fullName', () => {
```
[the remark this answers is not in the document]
pnavarrc: yeah, having multiple versions of location names is a bit annoying, how about this?

| method | county | state |
| --- | --- | --- |
| fullName | King County, Washington | Washington State | 
| shortName | King County, WA | Washington |
| abbreviatedName | King Co., WA  | WA |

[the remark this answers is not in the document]
mikelehen: We also use the raw county name in the location page header. 😬   It almost feels like we need "name" (e.g. "King County" or "King Co.") and "fully qualified name" (e.g. "King County, Washington" or "King Co., WA") as separate concepts.

image

That said, I feel like we're not going to "get this right" until we enumerate all the usages and can see the full taxonomy of what we use where and why... and FWIW- CBSAs are probably going to be even weirder!

So it probably makes sense to just take a stab at it like you are doing, and be sure to document the methods really well in terms of what they return for counties, states, cbsas...  and then we can tweak / iterate on them as we begin to understand the full taxonomy better.

mikelehen:
Oh. "using a hook to filter by different regions" was actually from @ghop02's [comment](https://github.com/covid-projections/covid-projections/pull/2043#discussion_r534557991).  I wasn't sure what that was referring to.

But what you are proposing seems reasonable to me.

src/common/regions.test.ts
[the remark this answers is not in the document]
chasulin: ditto to @mikelehen 's comment. there is a handful of different abbreviations (county-->co, parish-->par., etc) that i just implement manually in a utility func. we should probably aim to steamline the names used (ie. choosing between king county, washington vs. king county, WA), but all in all, i think these pain points will become clearer as we implement these changes and see what else is needed

mikelehen (approved):

PR #2050, opened 2020-12-03
Metric explainers CMS config

chasulin:
- Adds config + types for metric explainers
- Removes config + types for product pages (section nix'ed, no longer needed)


pnavarrc (approved):
Looks good!

PR #2052, opened 2020-12-03
About page follow ups

chasulin:
Includes:
- About page UI updates
- About page CMS content updates
- Metric explainers section content/json file (but no UI)


src/screens/About/About.style.tsx
```diff
@@ -29,3 +37,9 @@ export const AboutHeading3 = styled(Heading3)`
   highlighted when the section's content is in view but its header is not.
 */
 export const HashWrapper = styled.div``;
+
+export const PartnersSectionWrapper = styled.div`
```
chasulin, line 41: a quick but not ideal way to add needed spacing... _throw a div around it_

src/cms-content/learn/index.ts
```diff
@@ -4,7 +4,6 @@ import glossary from './learn-glossary.json';
 import landing from './learn-landing.json';
 import caseStudies from './learn-case-studies.json';
 import productsLanding from './products-landing.json';
-import productPages from './products-full-pages.json';
```
mikelehen, line 7: What’s up with products-full-pages.json?  Looks like we removed some references to it.  Should the file be deleted?

mikelehen (approved):

src/cms-content/learn/index.ts
chasulin: yes!

PR #2068, opened 2020-12-07
Fix horizontal scrollbar on iPhone

pnavarrc:
[Trello](https://trello.com/c/Kpu34lhS/603-horizontal-scrollbar-on-iphone-after-using-search-box) - Horizontal scrollbar on iPhone after using search box

This is a bug in Safari, if the font size on an input element is smaller than 16px, Safari will zoom in when the user focus the input element, but it won't zoom out once the user leaves the input area. The font size of our search input element is 14px 

The best fix I could find is to increase the font-size to 16px (1rem) and scale the input element so the text _looks_ 14px, so by a factor of (14 / 16 = 0.875). I adjusted the spacing around too so it matches (and remove the border radius, which is not necessary since borders are not visible).

## Testing
- Open the page with Safari on iPhone
- Click on the search box, the browser should not zoom-in
- Test in production, the browser will zoom-in

### DIFF
--- a/src/components/MapSelectors/GlobalSelector.style.js
+++ b/src/components/MapSelectors/GlobalSelector.style.js
@@ -70,12 +70,20 @@ export const StyledInputWrapper = styled.div`
 `;
 
 export const StyledInput = styled.input`
-  border-radius: ${props => (props.isOpen ? '4px 4px 0 0' : '4px')};
-  font-size: 16px;
+  font-size: 1rem;
   line-height: 1.5rem;
   border: none;
   width: 100%;
   padding: 0 1rem 0 0;
+
+  // When the font-size of an input element is smaller than 16px, Safari will
+  // zoom in when the user enters the input, and won't zoom out when the user
+  // leaves the input area. The fix is increasing the font size from 14px to 16px
+  // and scaling the input element by 14/16 = 0.875.
+  // https://trello.com/c/Kpu34lhS/603-horizontal-scrollbar-on-iphone-after-using-search-box
+  transform: scale(0.875);
+  transform-origin: left top;
+  margin-top: 0.25rem;
 
   &:focus {
     outline: none;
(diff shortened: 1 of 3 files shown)


pnavarrc:
[Horizontal scrollbar on iPhone after using search box.](https://trello.com/c/Kpu34lhS/603-horizontal-scrollbar-on-iphone-after-using-search-box)

src/components/MapSelectors/GlobalSelector.style.js
```diff
@@ -70,13 +70,21 @@ export const StyledInputWrapper = styled.div`
...
 export const StyledInput = styled.input`
-  border-radius: ${props => (props.isOpen ? '4px 4px 0 0' : '4px')};
-  font-size: 14px;
+  font-size: 1rem;
   line-height: 1.5rem;
   border: none;
   width: 100%;
   padding: 0 1rem 0 0;
 
+  // When the font-size of an input element is smaller than 16px, Safari will
```
ghop02, line 79: fwiw I think it happens on chrome too?

chasulin (changes requested):
per our convo - fixes the bug on the location page (woooo!), but still zooms slightly on the homepage and aligns the search bar to the left of the screen

ghop02:
> per our convo - fixes the bug on the location page (woooo!), but still zooms slightly on the homepage and aligns the search bar to the left of the screen

Could this be because of the weird banner issue, where the width of the banner seems to be wider than the nav bar

src/components/MapSelectors/GlobalSelector.style.js
pnavarrc: Oh that's good to know, I'm going crazy trying to debug this in Safari (Safari Developer tools keep crashing when I connect to my iPhone and Safari mobile insists on caching the page forever)

pnavarrc:
@ghop02  that's a good point, I adjusted that and it's a bit better

chasulin (approved):

PR #2075, opened 2020-12-07
Adds timestamp to glossary+FAQ

chasulin:
- The timestamps are fields in the glossary+FAQ's CMS UIs
- Whoever deploys cms changes for either of those pages needs to be sure to update the respective timestamp to `now` before shipping

### DIFF
--- a/public/admin/config.yml
+++ b/public/admin/config.yml
@@ -360,6 +360,12 @@ collections:
           - label: Intro
             name: intro
             widget: markdown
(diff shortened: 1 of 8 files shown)


pnavarrc (approved):
I think it makes more sense to export a `Date` from the CMS index file and format it as string where it gets rendered, that way we keep presentational concerns closer to the render and data structures closer to the CMS

PR #2076, opened 2020-12-08
Region API: Use region context to access location properties

pnavarrc:
This PR builds on top of  #2043 and uses the proposed region context to get location information from within location pages. It doesn't migrate all the instances yet.

- Adds the methods `findCountyByUrlParams` and `findStateByUrlParams` to `RegionDB` (and corresponding test cases)
- Implements `RegionContext` and `useRegion` to allow accessing the `region` object from anywhere under a location page.

When creating the `RegionContext`, I passed a stand-in region as default value. This allows TypeScript to understand that `useRegion` will always return an instance of `Region` (otherwise we will have to null-check everywhere). Open to suggestions!

### DIFF
--- a/src/common/regions.test.ts
+++ b/src/common/regions.test.ts
@@ -2,66 +2,66 @@ import regions, { RegionType } from './regions';
 
 describe('regions', () => {
   test('states() returns all the states', () => {
-    const states = regions.states();
+    const states = regions.states;
     expect(states).toHaveLength(53);
   });
 
   test('counties() returns all the counties', () => {
-    const counties = regions.counties();
+    const counties = regions.counties;
     expect(counties).toHaveLength(3224);
   });
 
   describe('findByFipsCode', () => {
     test('returns a state when passing a valid state FIPS', () => {
       const dc = regions.findByFipsCode('11');
-      expect(dc).toBeTruthy();
+      expect(dc).not.toBeNull();
       expect(dc?.regionType).toBe(RegionType.STATE);
     });
 
     test('returns a county when passing a valid county FIPS', () => {
       const county = regions.findByFipsCode('53033');
-      expect(county).toBeTruthy();
+      expect(county).not.toBeNull();
       expect(county?.regionType).toBe(RegionType.COUNTY);
     });
 
     test('returns null if the FIPS doesn’t exist', () => {
       const notFound = regions.findByFipsCode('xx');
-      expect(notFound).toBe(null);
+      expect(notFound).toBeNull();
     });
   });
 
   describe('findStateByUrlParams', () => {
-    test(`returns the correct state when passing a state code"`, () => {
+    test('returns the correct state when passing a state code', () => {
       const state = regions.findStateByUrlParams('OR');
-      expect(state).toBeTruthy();
+      expect(state).not.toBeNull();
       expect(state?.fipsCode).toBe('41');
     });
 
     test('returns the correct state when passing a valid URL segment', () => {
       const state = regions.findStateByUrlParams('oregon-or');
-      expect(state).toBeTruthy();
+      expect(state).not.toBeNull();
       expect(state?.fipsCode).toBe('41');
     });
 
     test('returns null for invalid params "xxx"', () => {
       const state = regions.findStateByUrlParams('xxx');
-      expect(state).toBe(null);
+      expect(state).toBeNull();
     });
   });
 
   describe('findCountyByUrlParams', () => {
-    test('returns the correct county for valid state and county IDs ', () => {
+    test('returns the correct county for valid state and county IDs', () => {
       const county = regions.findCountyByUrlParams(
         'washington-wa',
         'king_county',
       );
-      expect(county).toBeTruthy();
+      expect(county).not.toBeNull();
       expect(county?.fipsCode).toBe('53033');
     });
 
     test('returns the correct county when using state code as stateId', () => {
       const county = regions.findCountyByUrlParams('wa', 'king_county');
-      expect(county).toBeTruthy();
+      expect(county).not.toBeNull();
       expect(county?.fipsCode).toBe('53033');
     });
 
@@ -72,7 +72,7 @@ describe('regions', () => {
 
     test('returns null for invalid stateId and countyID', () => {
       const county = regions.findCountyByUrlParams('xxx', 'xxx');
-      expect(county).toBe(null);
+      expect(county).toBeNull();
     });
   });
 });
--- a/src/common/regions.test.ts
+++ b/src/common/regions.test.ts
@@ -11,15 +11,69 @@ describe('regions', () => {
     expect(counties).toHaveLength(3224);
   });
 
-  test('findByFipsCode returns a state for a state FIPS', () => {
-    const dc = regions.findByFipsCode('11');
-    expect(dc).toBeTruthy();
-    expect(dc?.regionType).toBe(RegionType.STATE);
+  describe('findByFipsCode', () => {
+    test('returns a state when passing a valid state FIPS', () => {
+      const dc = regions.findByFipsCode('11');
+      expect(dc).toBeTruthy();
+      expect(dc?.regionType).toBe(RegionType.STATE);
+    });
+
+    test('returns a county when passing a valid county FIPS', () => {
+      const county = regions.findByFipsCode('53033');
+      expect(county).toBeTruthy();
+      expect(county?.regionType).toBe(RegionType.COUNTY);
+    });
+
+    test('returns null if the FIPS doesn’t exist', () => {
+      const notFound = regions.findByFipsCode('xx');
+      expect(notFound).toBe(null);
+    });
   });
 
-  test('findByFipsCode returns null if the FIPS doesn’t exist', () => {
-    const notFound = regions.findByFipsCode('xx');
-    expect(notFound).toBe(null);
+  describe('findStateByUrlParams', () => {
+    test(`returns the correct state when passing a state code"`, () => {
+      const state = regions.findStateByUrlParams('OR');
+      expect(state).toBeTruthy();
+      expect(state?.fipsCode).toBe('41');
+    });
+
+    test('returns the correct state when passing a valid URL segment', () => {
+      const state = regions.findStateByUrlParams('oregon-or');
+      expect(state).toBeTruthy();
+      expect(state?.fipsCode).toBe('41');
+    });
+
+    test('returns null for invalid params "xxx"', () => {
+      const state = regions.findStateByUrlParams('xxx');
+      expect(state).toBe(null);
+    });
+  });
+
+  describe('findCountyByUrlParams', () => {
+    test('returns the correct county for valid state and county IDs ', () => {
+      const county = regions.findCountyByUrlParams(
+        'washington-wa',
+        'king_county',
+      );
+      expect(county).toBeTruthy();
+      expect(county?.fipsCode).toBe('53033');
+    });
+
+    test('returns the correct county when using state code as stateId', () => {
+      const county = regions.findCountyByUrlParams('wa', 'king_county');
+      expect(county).toBeTruthy();
+      expect(county?.fipsCode).toBe('53033');
+    });
+
+    test('returns null for invalid stateId and valid countyID', () => {
+      const county = regions.findCountyByUrlParams('xxx', 'king_county');
+      expect(county).toBe(null);
+    });
+
+    test('returns null for invalid stateId and countyID', () => {
+      const county = regions.findCountyByUrlParams('xxx', 'xxx');
+      expect(county).toBe(null);
+    });
   });
 });
 
@@ -41,7 +95,7 @@ describe('State', () => {
 describe('County', () => {
   const county = regions.findByFipsCode('53033');
   test('fullName', () => {
-    expect(county?.fullName()).toBe('King County, WA');
+    expect(county?.fullName()).toBe('King County, Washington');
   });
 
   test('relativeUrl', () => {
--- a/src/common/regions.ts
+++ b/src/common/regions.ts
@@ -141,13 +141,9 @@ class RegionDB {
       return null;
     }
 
-    const stateCounties = this.counties().filter(
-      county => county.state.fipsCode === foundState.fipsCode,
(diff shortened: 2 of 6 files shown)


src/components/LocationPage/LocationPageHeading.tsx
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  if (region instanceof County) {
+    return (
+      <Fragment>
+        <strong>{region.name}, </strong>
+        {region.state.stateCode}
+      </Fragment>
+    );
+  }
```
pnavarrc, line 16: Note: The old code had a link on the state code, I think that we don't need that, but I can add it back if we think that that's useful.

src/common/regions.test.ts
```diff
@@ -11,15 +11,69 @@ describe('regions', () => {
...
+  describe('findByFipsCode', () => {
+    test('returns a state when passing a valid state FIPS', () => {
+      const dc = regions.findByFipsCode('11');
+      expect(dc).toBeTruthy();
+      expect(dc?.regionType).toBe(RegionType.STATE);
+    });
+
+    test('returns a county when passing a valid county FIPS', () => {
+      const county = regions.findByFipsCode('53033');
+      expect(county).toBeTruthy();
```
mikelehen, line 23: nit: I think `.not.toBe(null)` would better capture the intention of these tests and better complement the tests that *are* expecting null and therefore use `.toBe(null)`.  Feel free to disagree and punt though.

src/common/regions.ts
```diff
@@ -116,27 +121,65 @@ const counties: County[] = chain(state_county_map_dataset)
...
+}
+
 const regions = new RegionDB([...states, ...counties]);
 export default regions;
+
+// TODO: Is there a better way to do this? If I pass `null` as default value
+// for the context, we will need to check that the region is not null
+// everywhere (we are checking for that in the LocationRouter)
+const standInRegion: State = new State('name', 'name', '99', 0, '99');
+export const RegionContext = createContext<Region>(standInRegion);
```
pnavarrc, line 183: Creating the context with a non-null default value makes TS understand that the value cannot be `null` or `undefined`. If we initialize with `null`, then we need to check that the value is not null everywhere.

src/common/regions.test.ts
```diff
@@ -11,15 +11,69 @@ describe('regions', () => {
...
+    });
+
+    test('returns null for invalid params "xxx"', () => {
+      const state = regions.findStateByUrlParams('xxx');
+      expect(state).toBe(null);
+    });
+  });
+
+  describe('findCountyByUrlParams', () => {
+    test('returns the correct county for valid state and county IDs ', () => {
```
mikelehen, line 53: nit: trailing space in test name.

src/common/regions.ts
```diff
@@ -116,27 +121,65 @@ const counties: County[] = chain(state_county_map_dataset)
...
-  private filterByType(regionType: RegionType) {
-    return this.all().filter(region => region.regionType === regionType);
-  }
-
   findByFipsCode(fipsCode: FipsCode): Region | null {
     const region = this.regions.find(region => region.fipsCode === fipsCode);
     return region || null;
   }
 
+  findStateByUrlParams(stateId: string): State | null {
```
mikelehen, line 129: Should this be `stateUrlSegment` or something?  `stateId` is a bit general and in our codebase I'd normally assume that's the 2-letter state code.

src/common/regions.ts
```diff
@@ -116,27 +121,65 @@ const counties: County[] = chain(state_county_map_dataset)
...
+    const foundState = this.states().find(
+      state =>
+        equalLower(state.urlSegment, stateId) ||
+        equalLower(state.stateCode, stateId),
```
mikelehen, line 133: Perhaps add a comment that this supports "legacy" URLs that just included the state code in the URL?

src/common/regions.ts
```diff
@@ -116,27 +121,65 @@ const counties: County[] = chain(state_county_map_dataset)
...
+    );
+    return foundState || null;
+  }
+
+  findCountyByUrlParams(stateId: string, countyId: string): County | null {
```
mikelehen, line 138: Again, can we use clearer names?  `stateUrlSegment`, `countyUrlSegment`?

src/common/regions.ts
```diff
@@ -116,27 +121,65 @@ const counties: County[] = chain(state_county_map_dataset)
...
+      (region): region is State => region.regionType === RegionType.STATE,
+    );
   }
 
-  counties(): Region[] {
-    return this.filterByType(RegionType.COUNTY);
+  counties(): County[] {
+    return this.all().filter(
+      (region): region is County => region.regionType === RegionType.COUNTY,
+    );
```
mikelehen, line 164: Don't we already have `states` and `counties` constants defined in this file? Do we need to recalculate these?

FWIW- I really like to make method/function names reflect the performance characteristics.  I'd intuitively expect `RegionsDB.states()` to be a very cheap method call, but as-is it's going to iterate over 3000+ entries and construct a new list of the state entries, I think.  So we should either make sure the list is cached (so it _is_ cheap) or else name this `filterToStates()` or something to make it clearer that you shouldn't call it repeatedly.

src/common/regions.ts
```diff
@@ -116,27 +121,61 @@ const counties: County[] = chain(state_county_map_dataset)
...
+
+  findCountyByUrlParams(stateId: string, countyId: string): County | null {
+    const foundState = this.findStateByUrlParams(stateId);
+    if (!foundState) {
+      return null;
+    }
+
+    const foundCounty = this.counties()
+      .filter(county => county.state.fipsCode === foundState.fipsCode)
+      .find(county => equalLower(county.urlSegment, countyId));
```
mikelehen, line 146: nit: Is there a reason to do this as two passes (a filter and then a find) rather than combining?

const foundCounty = this.counties().find(county => 
      county.state.fipsCode === foundState.fipsCode && 
      equalLower(county.urlSegment, countyId)
);

src/common/regions.ts
mikelehen: Yeah, I think this is okay.  The other option would be to lie to TypeScript and do something like:

```
// We are careful to never use `useRegion()` in components that aren't nested inside a
// <RegionContext.Provider> so we cheat and pretend the value is non-nullable.
export const RegionContext = createContext<Region>(null as any);
```

src/common/regions.ts
```diff
@@ -116,27 +121,61 @@ const counties: County[] = chain(state_county_map_dataset)
...
 const regions = new RegionDB([...states, ...counties]);
 export default regions;
+
+// TODO: Is there a better way to do this? If I pass `null` as default value
+// for the context, we will need to check that the region is not null
+// everywhere (we are checking for that in the LocationRouter)
+const standInRegion: State = new State('name', 'name', '99', 0, '99');
+export const RegionContext = createContext<Region>(standInRegion);
+
+export const useRegion = () => useContext(RegionContext);
```
mikelehen, line 181: If we think we are going to stick with our "location page" terminology for a while I'd be pretty tempted to name this `useLocationPageRegion()` or similar to make it a bit more explicit what region it is talking about.

Else maybe `useUrlRegion` or something? Idunno.

src/components/LocationPage/LocationPageHeader.tsx
```diff
@@ -124,7 +104,7 @@ const LocationPageHeader = (props: {
           <HeaderSection>
             <LocationCopyWrapper>
               <HeaderTitle isEmbed={isEmbed}>
-                <LocationPageHeading projections={projections} />
+                <LocationPageHeading region={region} />
```
mikelehen, line 107: Hrm. Is it weird that we're passing in the region instead of using `useRegion()` within the component?  I guess either way is fine. 🤷

src/components/LocationPage/LocationPageHeading.tsx
```diff
@@ -0,0 +1,21 @@
+import React, { Fragment } from 'react';
+import { Region, State, County } from 'common/regions';
+
+const LocationPageHeading: React.FC<{ region: Region }> = ({ region }) => {
+  if (region instanceof State) {
+    return <strong>{region.name}</strong>;
+  }
+
+  if (region instanceof County) {
```
mikelehen, line 9: nit: `else if` ?

src/components/LocationPage/LocationPageHeading.tsx
mikelehen: I definitely use it sometimes.  So I'd keep it unless there's a specific reason you're removing it.

At some point it'd be nice to have breadcrumbs. 😁 

USA > Washington > King County

mikelehen (changes requested):
Basically LGTM but a few requests!

ghop02 (commented):
Looks good! i'm wondering what the advantages vs disadvantages are of creating a context here vs just passing region to `LocationPage` and trying to use that more heavily.  If we were doing more to change or update the region used I'd probably agree with this path, but it's pretty static the whole time.

The `Projections` objects is pretty heavily used throughout the codebase and one thought I have is to add `region` to the projections object/query projections by a region.  If `useProjections` took a region instead of a `stateId` and `countyId` I think that'll be a big step forward in unifying the API.  The less we have to work with stateId and County ID the easier i think our lives will be.

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -46,6 +47,8 @@ const ChartsHolder = (props: {
   const { chartId } = props;
   const projection = props.projections.primary;
 
+  const region = useRegion();
```
ghop02, line 50: Now that we have region here, can we remove `countyId`, `stateId`, and `county`?

src/common/regions.ts
pnavarrc: I find it a bit cleaner this way, but no strong feelings either way, updating!

src/components/LocationPage/LocationPageHeader.tsx
pnavarrc: In general, using React context makes components less reusable and binds them to a specific ... eh ... context. In this example, if we don't pass `region` to `LocationPageHeading`, then the component can only be used inside a `LocationPage` (inside `RegionContext.Provider` to be precise).

My preference is to use React context when:
- components at multiple levels need the same information
- but some components in the middle don't really need them

In this case, `LocationPageHeader` already has the `region`, so it makes sense to just pass it down. It doesn't really matter in this case though because the component is really tailored to this page, but that's the reason.

src/components/LocationPage/LocationPageHeading.tsx
pnavarrc: adding it back!

src/common/regions.ts
pnavarrc: Updated, passing states and counties in the constructor, they can be accessed directly now (faster)

src/common/regions.ts
pnavarrc: changed to `useLocationPageRegion`

src/components/LocationPage/ChartsHolder.tsx
pnavarrc: Eventually! That would be a big refactor, as `stateId` is used by `Compare`, the share buttons, etc - I prefer to clean up in smaller PRs to make it easier to review & ship

src/screens/LocationPage/LocationRouter.tsx
```diff
@@ -10,15 +10,22 @@ interface LocationPageUrlParams {
...
-  // redirect the user to the homepage
-  const isValidLocation = state && !(countyId && !countyOption);
+  const state = regions.findStateByUrlParams(stateId);
+  const county = countyId
+    ? regions.findCountyByUrlParams(stateId, countyId)
+    : null;
+  const region = county || state;
 
-  return isValidLocation ? <LocationPage /> : <Redirect to="/" />;
+  if (!region) {
```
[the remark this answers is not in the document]
pnavarrc: oh yeah, that would be handy. I think that a page that lists all the locations sorted by state as fallback would be even better.

pnavarrc:
> Looks good! i'm wondering what the advantages vs disadvantages are of creating a context here vs just passing region to LocationPage and trying to use that more heavily.

### Advantages
React context is useful when you have a multiple levels of React components that need access to the same information, but when not all the components in the middle need the information. Compare:

```tsx
<RegionProvider region={region}>
  <LocationPage>
    <LocationHeader>
      <LocationTitle /> // <- needs the region, will call `useRegion`
    </LocationHeader>
  </LocationPage>
</Region>
```

with passing the `region` down as prop through multiple levels that don't necessarily need the `region` data. This is known as _prop-drilling`, and causes unnecessary re-renders in the components in the middle

```tsx
<RegionProvider region={region}>
  <LocationPage region={region}>
    <LocationHeader region={region}>
      <LocationTitle region={region} />
    </LocationHeader>
  </LocationPage>
</Region>
```

 ### Disadvantages

Components that rely on a context cannot be used outside that context, which makes less reusable, since they require to be inside the context provider to work. If we wanted to use the `LocationTitle` to render the alert emails, the second option would be better.

I think that while it would be useful to embed the `region` in the `projection`, I wouldn't make the projection the only access point for the location information, because that would block rendering things that don't necessarily need the full projection (location summaries, location name, etc).

Using the context also helps us transition to using the `regions` without forcing us to do everything in one PR, and implementing CBSAs will show us what we need to change as we go.

mikelehen (approved):
LGTM!

FWIW- Reading your and Chris's discussion, I'm finding myself also going back and forth on whether the context seems like the best choice.  In general I do like explicit code (passing parameters in explicitly, etc.).  And I could imagine it makes our lives easier in some cases (e.g. if we're adding components to storybook, ideally they could be constructed via properties and not need a RegionContext around them).

That said, it does seem like it'll save us some property plumbing and I'm enough of a React novice, I'm happy to go with the flow.  Plus, given that the region is just derived from URL params (which are already available directly to any component), this isn't much of a deviation from the current situation.

So I'm still down to give this a try. 👍

pnavarrc:
I have to say, I'm a bit on the fence on this too. I will merge for now and see how easy would be to get rid of it as we implement CBSAs.

PR #2078, opened 2020-12-08
Cdc testing data changes

ghop02:
This will prepare the frontend to properly take in cdc testing data.  

Note that the API update here was manual - because we haven't updated since CBSAs were introduced, there are a handful of type issues that will be resolved when we rely less on state codes in the `Projection` and `Projections` object.


src/common/metrics/positive_rate.tsx
```diff
@@ -142,19 +142,32 @@ function renderStatus(projections: Projections) {
...
           aggregated by the Centers for Medicare and Medicaid Services. It is
           calculated as a 14-day trailing average and may take a week or more to
           update.
         </React.Fragment>
-      ) : (
+      )}
+      {usingCDCData && (
+        <React.Fragment>
+          {METRIC_NAME} for {projections.locationName} comes from data
+          aggregated by the Center for Disease Control and Prevention. It is
```
mikelehen, line 165: Center => Centers

mikelehen (approved):
Nice!

PR #2083, opened 2020-12-09
implement MetroArea class and add it to the regions API

pnavarrc:
- Implements a script to parse and format the MSA data
- Implement the `MetroArea` class and import the metro areas
- Add metro areas to the region DB

# Divisions and Terminology

I'm writing this down mostly for my own benefit, will probably add it to the script folder and add a comment on the regions file.

**Metropolitan Statistical Areas (MSA)**: Have at least one urbanized area of 50,000+ population plus adjacent territory with strong social and economic ties to the urban center. There are 362 in total as of June 6, 2003.
**Micropolitan Statistical Areas (µSA):** Have between 10k - 50k population plus adjacent territory with strong social and economic ties to the urban center. There are 560 in total as of June 6, 2003.

Both metropolitan and micropolitan statistical areas are defined in terms of whole counties, and are referred collectively as Core-Based Statistical Areas.

image
Source: [Metropolitan and Micropolitan Statistical Areas of the United States and Puerto Rico (pdf)](https://www2.census.gov/geo/maps/metroarea/us_wall/Sep2018/CBSA_WallMap_Sep2018.pdf)

A **Combined Statistical Area (CSA)** is a combination of adjacent MSA and µSA with economic ties, so it can be larger than a MSA. For example, the Seattle-Tacoma CSA is composed of the following metropolitan and micropolitan statistical areas:

- Mount-Vernon-Anacortes (MSA)
- Seattle-Tacoma-Bellevue (MSA)
- Olumpia-Lacey-Tumwater (MSA)
- Bremerton-Silverdale-Port Orchard (MSA)
- Shelton (µSA)
- Oak Harbor (µSA)
- Centralia (µSA)

image

Legend: MSA (dark green areas), µSA (light green areas), CSA border: dark green line

We want to create pages for Metropolitan Statistical Areas. 

### References
- [About CBSA - Census.gov](https://www.census.gov/topics/housing/housing-patterns/about/core-based-statistical-areas.html)


scripts/cbsa/parse-metro-areas.ts
```diff
@@ -0,0 +1,92 @@
...
+    cbsaTitle: row[3],
+    statisticalAreaType: row[4] as StatisticalAreaType,
+    metropolitanDivisionTitle: row[5],
+    csaTitle: row[6],
+    countyEquivalent: row[7],
+    stateName: row[8],
+    fipsStateCode: row[9],
+    fipsCountyCode: row[10],
+    centralOutlyingCounty: row[11],
+  };
```
pnavarrc, line 87: I'm not using all the fields here, but I think we might want to keep them in case we want to add more information to the MSAs

scripts/cbsa/parse-metro-areas.ts
```diff
@@ -0,0 +1,92 @@
...
+  const csvText = fs.readFileSync(path.join(__dirname, 'mar2020.csv'));
+  const [, ...rows] = await csvParse(csvText);
+  const items: CsvRow[] = rows
+    .map(parseRow)
+    .filter((item: CsvRow) => item.csaCode);
+
+  const groupedByCbsa = splitBy(items, item => item.cbsaCode);
+  const msaItems = groupedByCbsa
+    .map(formatGroup)
+    .filter(metroArea => metroArea.countyFips.length > 1);
```
pnavarrc, line 42: I'm not 100% sure that we should filter metro areas with only one county out of the list 🤔

src/common/regions.ts
```diff
@@ -119,16 +141,62 @@ const counties: County[] = chain(state_county_map_dataset)
...
+    .map(part => deburr(words(part).join('-')))
+    .join('_');
+}
+
+const metroAreas: MetroArea[] = msaDefinitions.msa_regions.map(metroArea => {
+  // TODO: Handle counties with FIPS: 36005, 360047, 36081, 36085, those
+  // are part of MSAs, but not in the countiesByFips map
+  const metroAreaCounties: County[] = metroArea.countyFips
+    .map(countyFips => countiesByFips[countyFips])
+    .filter(county => county);
```
pnavarrc, line 162: TODO: Investigate why we don't have those counties in our list of counties and either add them or remove them from the metro areas file (via the script)

pnavarrc:
[CBSA: Add MSAs to the Region API](https://trello.com/c/hsoRqSbf/651-cbsa-add-msas-to-the-region-api)

scripts/cbsa/parse-metro-areas.ts
```diff
@@ -0,0 +1,92 @@
...
+  fipsCountyCode: string;
+  centralOutlyingCounty: string;
+}
+
+async function main() {
+  const csvText = fs.readFileSync(path.join(__dirname, 'mar2020.csv'));
+  const [, ...rows] = await csvParse(csvText);
+  const items: CsvRow[] = rows
+    .map(parseRow)
+    .filter((item: CsvRow) => item.csaCode);
```
mikelehen, line 37: Am I reading this code right that we are using only CBSAs that are within a Combined Statistical Area?  

Can you add a comment or two to this code to make the business logic clearer?  

And is this logic covered by the PRD?  I didn't realize we weren't using all CBSAs, and I don't understand CSAs and CBSAs well enough to understand why we aren't using all CBSAs or just using CSAs directly.  🤔

scripts/cbsa/parse-metro-areas.ts
```diff
@@ -0,0 +1,92 @@
...
+  );
+}
+
+function splitBy(items: CsvRow[], keyFn: (item: CsvRow) => string) {
+  return values(groupBy(items, keyFn));
+}
+
+function formatGroup(groupItems: CsvRow[]) {
+  const firstItem = groupItems[0];
+  const [name] = firstItem.cbsaTitle.split(',');
```
mikelehen, line 60: style nit: Why are we using two very different syntaxes for getting the first item of a list? 😁   I'd probably stick with;
```
const name = firstItem.cbsaTitle.split(',')[0];
```

scripts/cbsa/parse-metro-areas.ts
```diff
@@ -0,0 +1,92 @@
...
+  return {
+    cbsaCode: firstItem.cbsaCode,
+    cbsaTitle: firstItem.cbsaTitle,
+    csaCode: firstItem.csaCode,
+    csaTitle: firstItem.csaTitle,
+    name,
+    countyFips: groupItems.map(
```
mikelehen, line 67: nit: Maybe `countyFipsCodes` so it's more clearly an array of "county FIPS" instead of a single "county FIPS"?

src/common/regions.ts
```diff
@@ -1,17 +1,18 @@
...
+import msaDefinitions from './data/msa-data.json';
 const { state_dataset, state_county_map_dataset } = US_STATE_DATASET;
 
 export type FipsCode = string;
 
 export enum RegionType {
   COUNTY = 'county',
   STATE = 'state',
-  CBSA = 'CBSA',
+  MSA = 'MSA',
```
mikelehen, line 15: I might have mentioned this in a different review but is it clear that "MSA" is the right term for us to use throughout our code as opposed to "CBSA" I guess?  Or maybe we need some other term if it's not strictly an MSA or CBSA (MetroArea?)?

Maybe you or somebody else just needs to explain to me exactly what we mean by "MSA" and this will all make sense though. 😁

src/common/regions.ts
```diff
@@ -119,16 +141,62 @@ const counties: County[] = chain(state_county_map_dataset)
   })
   .value();
 
+const countiesByFips = fromPairs(
+  counties.map(county => [county.fipsCode, county]),
+);
+
+// Seattle-Tacoma, WA -> seattle-tacoma_wa
```
mikelehen, line 148: @ghop02 just added our location pages to our normal API endpoints I believe.  What's the status for CBSA location pages in the API?

Should we put these urls into the json file and then share the json file with the backend API code?

src/common/regions.ts
mikelehen: I think those are the New York City boroughs that we aggregate into a single fake county thing.  Tom and I are working on unraveling that this week.  I'll probably send you a PR tonight or tomorrow.

mikelehen (approved):
LGTM with a few minor things!

src/common/regions.ts
ghop02: https://github.com/covid-projections/covid-data-public/blob/master/scripts/update_can_location_urls.py We update them here

ghop02: fwiw I think this would be helpful to have in the json file both for that and so we don't accidentally change them in the future.

src/common/regions.ts
```diff
@@ -119,16 +141,62 @@ const counties: County[] = chain(state_county_map_dataset)
...
+    metroAreaCounties.map(county => county.population),
+  );
+
+  return new MetroArea(
+    metroArea.csaTitle,
+    urlFormat(metroArea.csaTitle),
+    metroArea.cbsaCode,
+    totalPopulation,
+    metroAreaCounties,
+  );
```
ghop02, line 174: Could we have this code in the file that produces the json? Feels a bit unnecessary that we have to regenerate this every time, when its essentially a static calculation

src/common/regions.ts
ghop02: I think part of the confusion here is that we will be showing a subset of CBSAs on the site, they might be the same as all of the MSAs.  
I don't have a super strong preference, but would lean towards referring to them as CBSAs even though its a subset - for example in the future if we wanted to include micropolitan areas, I don't think the data structure would look any different

From a [good blog post](https://www.inc.com/encyclopedia/metropolitan-statistical-area-msa.html#:~:text=Core%2DBased%20Statistical%20Area,core%20and%20the%20outlying%20areas.&text=The%20CBSA%20is%20a%20generic,difference%20being%20core%20population%20size.)
> A CBSA is one or more counties with an urbanized cluster of at least 10,000 people. The area as a whole is defined by the interaction between the core and the outlying areas. This interaction, measured by commuting, means that at least 25 percent of people in outlying areas are working in the core. The CBSA is a generic definition of MICROs and MSAs, the difference being core population size.

scripts/cbsa/parse-metro-areas.ts
pnavarrc: I updated the PR description to clarify. We actually need all the Metro Areas, not CSAs, updating the script!

scripts/cbsa/parse-metro-areas.ts
pnavarrc: I'm going to keep metro areas with one county, there is no harm and there are not that many with a single county.

scripts/cbsa/parse-metro-areas.ts
pnavarrc: I personally prefer `const [first, ...rest] = items;` when I can use it, in this case, I need all the items later to generate the list of FIPS, so using that will force me to concatenate the first and remaining items.

Updating so both use the same syntax

src/common/regions.ts
pnavarrc: Let's keep it as MSA for now, we can always define CBSA as a union type if we ever add µSA to the site.

src/common/regions.ts
pnavarrc: I will add this to the script so they are hard-coded in the JSON file

pnavarrc:
Closing in favour of #2099

PR #2090, opened 2020-12-09
Remove NYC hacks.

mikelehen:
This removes all the code that remaps all 5 NYC boroughs onto NYC (fips=36061).

DO NOT SUBMIT. This is using a snapshot built from a non-main branch.  We need to get that merged and in a normal snapshot before we merge this.


ghop02 (approved):
Looks great to me! One last thing - do we want to change the alert for these counties announcing the change? Don't think it's necessary but would be a nice to have!

mikelehen:
I'm merging this as-is because I have to run for a bit and I want to make sure we are unblocked from shipping future snapshots, but I'll try to get alert copy in place tonight or first thing tomorrow. 🙏

PR #2097, opened 2020-12-09
Refactoring regions

ghop02:
* Splits up the existing regions file into a few separate files
* adds a few hooks for easily accessing regions from url params
* convert simple functions to properties on `Region` object

### DIFF
--- a/src/common/regions/index.ts
+++ b/src/common/regions/index.ts
@@ -1,9 +1,11 @@
-import { createContext, useContext } from 'react';
+import { createContext } from 'react';
 import { chain, fromPairs } from 'lodash';
 import urlJoin from 'url-join';
-import { assert } from 'common/utils';
 import US_STATE_DATASET from 'components/MapSelectors/datasets/us_states_dataset_01_02_2020.json';
-import countyAdjacencyMsa from './data/county_adjacency_msa.json';
+import countyAdjacencyMsa from 'common/data/county_adjacency_msa.json';
+import { getAbbreviatedCounty } from 'common/utils/compare';
+import RegionDB from 'common/regions/region_db';
+
 const { state_dataset, state_county_map_dataset } = US_STATE_DATASET;
 
 export type FipsCode = string;
@@ -23,10 +25,12 @@ export abstract class Region {
     public readonly regionType: RegionType,
   ) {}
 
-  abstract fullName(): string;
-  abstract relativeUrl(): string;
-  canonicalUrl() {
-    return urlJoin('https://covidactnow.org', this.relativeUrl());
+  abstract get fullName(): string;
+  abstract get abbreviation(): string;
+  abstract get relativeUrl(): string;
+
+  get canonicalUrl() {
+    return urlJoin('https://covidactnow.org', this.relativeUrl);
   }
 }
 
@@ -45,11 +49,14 @@ export class State extends Region {
    * State: Washington
    * County: King County, Washington
    */
-  fullName() {
+  get fullName() {
     return this.name;
   }
 
-  relativeUrl() {
+  get abbreviation() {
+    return this.stateCode;
+  }
+  get relativeUrl() {
     return `us/${this.urlSegment}`;
   }
 }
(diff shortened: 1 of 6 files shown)


src/common/regions/index.ts
```diff
@@ -0,0 +1,176 @@
...
+
+export const getStateCode = (region: Region): string | null => {
+  if (region.regionType === RegionType.COUNTY) {
+    return (region as County).state.stateCode;
+  }
+  if (region.regionType === RegionType.STATE) {
+    return (region as State).stateCode;
+  }
+  return null;
+};
```
pnavarrc, line 114: I don't think we should implement the two functions above as they won't work for MSAs, and will generate more work for us as we refactor. Having state code and state name separated might also hide useful abstractions that we might need.

For example, if we need `King County, Washington` (`fullName`), `King County, WA` (`shortName`?) and `King Co. WA.` (`abbreviation`), using state name and code will hide the need for `shortName`.

src/common/regions/region_db.ts
```diff
@@ -0,0 +1,66 @@
...
+  private regions: Region[];
+
+  constructor(public states: State[], public counties: County[]) {
+    this.regions = [...states, ...counties];
+  }
+
+  findByFipsCode(fipsCode: FipsCode): Region | null {
+    const region = this.regions.find(region => region.fipsCode === fipsCode);
+    return region || null;
+  }
```
pnavarrc, line 15: I think I did this in one of my throwaway branches, but maybe we should import `statesByFips` and `countiesByFips` and add them as private properties for fast search by FIPS?

```ts
private regionsByFips: {[fipsCode:string]: Region};

// ...
this.regionsByFips = {...statesByFips, ...countiesByFips };
```
Also, it would be nice to have properties to access states and counties directly, without having to filter by type (since we already have them in the other file)

```
private states: State[];
private counties: Counties[];

// ...
this.states = states;
this.counties = counties;
```

src/common/regions/region_db.ts
```diff
@@ -0,0 +1,66 @@
...
+        county.state.fipsCode === foundState.fipsCode &&
+        equalLower(county.urlSegment, countyUrlSegment),
+    );
+
+    return foundCounty || null;
+  }
+
+  all(): Region[] {
+    return this.regions;
+  }
```
pnavarrc, line 52: This can be probably an accessor if we add states and counties as properties

src/common/regions/region_hooks.ts
```diff
@@ -0,0 +1,50 @@
...
+) => {
+  if (countyFipsId) {
+    return regions.findByFipsCode(countyFipsId);
+  }
+  const state = regions.findStateByUrlParams(stateId);
+  const county = countyId
+    ? regions.findCountyByUrlParams(stateId, countyId)
+    : null;
+  return county || state;
+};
```
pnavarrc, line 28: I would recommend not implementing these hooks as part of the refactor, they might be convenient but I don't think is going to help us with MSAs

pnavarrc (changes requested):
Looks good in general, I have some recommendations.
- Maybe it would be better to split in 3 parts:
  - a file with classes and enums
  - a file that parses the data and creates `states`, `counties`, `statesByFips` and `countiesByFips`
  - the regions db
- I would recommend not to implement functionality based on state code and state name (or transitional hooks), that might hurt us as we try to refactor. Let's integrate MSAs before we start refactoring to ensure that we do it in a way that works for MSA

ghop02:
> Looks good in general, I have some recommendations.
> 
> * Maybe it would be better to split in 3 parts:
>   
>   * a file with classes and enums
>   * a file that parses the data and creates `states`, `counties`, `statesByFips` and `countiesByFips`
>   * the regions db
> * I would recommend not to implement functionality based on state code and state name (or transitional hooks), that might hurt us as we try to refactor. Let's integrate MSAs before we start refactoring to ensure that we do it in a way that works for MSA

I think these are necessary to get ourselves out of the holes we're in now.  These functions are used sparingly and will help us consolidate the logic for some of these state edge cases.  Though I do think with `useRegionFromParams` we don't need the legacy location id logic nearly as much. 

I guess I think that we won't be able to integrate MSAs until most of the code is using regions - these helper functions will help keep the refactors small and allow us to make incremental progress.  The idea is that we'll have another PR that will allow us to rip out these functions at the end.

src/common/regions/region_db.ts
ghop02: ooo maybe i'll take your function that you had.  if it's alright, I think we can handle that in another PR - it'll be easy to swap that in or out at any point!

pnavarrc (approved):

PR #2099, opened 2020-12-09
implement a script to parse metro areas

pnavarrc:
Split the PR of parsing and adding MSAs to the region API into parsing only and addressing the feedback for #2083

### DIFF
--- a/scripts/cbsa/parse-metro-areas.ts
+++ b/scripts/cbsa/parse-metro-areas.ts
@@ -1,0 +1,114 @@
+import fs from 'fs';
+import path from 'path';
+import { groupBy, values, words, deburr, sum } from 'lodash';
+import regions from '../../src/common/regions';
+const promisify = require('util').promisify;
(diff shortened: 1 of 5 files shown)


scripts/cbsa/parse-metro-areas.ts
```diff
@@ -0,0 +1,114 @@
...
+
+function formatUrl(name: string) {
+  return name
+    .toLowerCase()
+    .split(', ')
+    .map(part => deburr(words(part).join('-')))
+    .join('_');
+}
+
+function formatGroup(groupItems: CsvRow[]) {
```
ghop02, line 66: It might be nice to define an interface for the return object just to make it explicit what fields we expect to be in the json file

ghop02 (approved):
LGTM!

scripts/cbsa/parse-metro-areas.ts
pnavarrc: I'm going to update once we merge your reorganization!

PR #2102, opened 2020-12-09
CMS updates

chasulin:
- Learn + About page content updates
- Replaces a headshot whose naming syntax was making git unhappy


pnavarrc (approved):
Thanks for fixing this!

PR #2103, opened 2020-12-09
Add a couple use cases of useProjectionsForRegion

ghop02:
Added a hook for getting the projection from a region object + a couple of example uses of it.  Need to clean up some of the comments in model.ts, but the structure should be good to go!


src/screens/internal/ShareImage/ChartShareImage.tsx
```diff
@@ -12,29 +12,24 @@ import {
...
-import { Projections } from 'common/models/Projections';
 import { MetricChart } from '../../../components/Charts';
 import { ALL_METRICS, getMetricNameExtended } from 'common/metric';
 import { Metric } from 'common/metric';
-import { findCountyByFips } from 'common/locations';
-import { useProjections } from 'common/utils/model';
-import { Projection } from 'common/models/Projection';
+import { useProjectionsFromRegion } from 'common/utils/model';
 import { SCREENSHOT_CLASS } from 'components/Screenshot';
+import { useRegionFromParams } from 'common/regions/region_hooks';
```
pnavarrc, line 20: It's better to import from `common/regions`, that way we don't leak implementation details of the module (if we decide to move the function, for example)

src/screens/internal/ShareImage/ShareCardImage.tsx
```diff
@@ -1,17 +1,16 @@
...
 import {
   ShareCardWrapper,
   TitleWrapper,
   LastUpdatedWrapper,
 } from './ShareCardImage.style';
 import { DarkScreenshotWrapper } from './ShareImage.style';
 import { formatLocalDate } from 'common/utils';
-import { findCountyByFips } from 'common/locations';
 import { ScreenshotReady, SCREENSHOT_CLASS } from 'components/Screenshot';
+import { useRegionFromParams } from 'common/regions/region_hooks';
```
pnavarrc, line 12: same here

src/screens/internal/ShareImage/ShareCardImage.tsx
```diff
@@ -50,21 +44,24 @@ const Header = (props: { isHomePage?: Boolean }) => {
...
-  const [countyOption] = useState(
-    countyFipsId && findCountyByFips(countyFipsId),
-  );
-  stateId = stateId || countyOption.state_code;
-  projections = useProjections(stateId!, countyOption) as any;
+interface LocationShareCardProps {
+  region: Region;
+}
+
+const LocationShareCard = ({ region }: LocationShareCardProps) => {
```
pnavarrc, line 63: Maybe this component should use a projection instead? It doesn't seem that the component uses the region for anything other than getting the projections. We can do that later

pnavarrc (approved):
Added some notes, looks good otherwise!

src/screens/internal/ShareImage/ShareCardImage.tsx
ghop02: Yeah I think ideally it would, probably will keep it as is for now

PR #2104, opened 2020-12-09
More uses of useProjectionsFromRegion

ghop02:
Replaces a handful of other useProjections calls.  This replaces most of them, but there is one left in `LocationPage` which I'm planning to tackle separately


src/screens/Embed/Embed.js
```diff
@@ -26,39 +26,12 @@ import {
   EmbedSubheader,
 } from './Embed.style';
 import SocialLocationPreview from 'components/SocialLocationPreview/SocialLocationPreview';
-import {
-  getCountyByUrlName,
-  getStateByUrlName,
-  getCanonicalUrl,
-} from 'common/locations';
+import { useRegionFromParams } from 'common/regions/region_hooks';
```
pnavarrc, line 29: we can import from `common/regions`

src/screens/Embed/Embed.js
```diff
@@ -69,21 +42,14 @@ function LocationEmbed() {
...
     <EmbedContainer>
       <EmbedWrapper>
         <EmbedHeaderWrapper>
           <HeaderText>
-            <EmbedHeader>
-              {projections.countyName
-                ? `${projections.countyName}, ${projections.stateCode}`
-                : projections.stateName}
-            </EmbedHeader>
+            <EmbedHeader>{region.fullName}</EmbedHeader>
```
pnavarrc, line 52: beautiful!

src/screens/internal/AllStates/AllStates.js
```diff
@@ -1,18 +1,20 @@
...
 import { MetricChart } from 'components/Charts';
+import regions from 'common/regions';
 
 function AllStates() {
-  return Object.keys(STATES).map(stateId => (
-    <State key={stateId} stateId={stateId} />
-  ));
+  return Object.keys(STATES).map(stateCode => {
+    return <State key={stateCode} stateCode={stateCode} />;
+  });
```
pnavarrc, line 12: It might be better to pass the `state` objects directly instead of passing state codes, getting the region and then getting the projection

```tsx
import regions  from 'common/regions';

function allStates() {
  return regions.states.map(state => (
    <State key={state.fipsCode} state={state} />
  );
}

// ...
function State({ state }) {
  const projections = useProjectionsFromRegion(state);
  // ...
}
```

src/common/regions/region_db.ts
```diff
@@ -14,6 +14,11 @@ class RegionDB {
     return region || null;
   }
 
+  findStateByStateCode(stateCode: string): State | null {
+    const region = this.states.find(state => state.stateCode === stateCode);
+    return region || null;
+  }
```
pnavarrc, line 20: If you like my proposal below, we might not need this function yet (but we can keep it)

pnavarrc (approved):
Looks good, added some comments!

src/screens/internal/AllStates/AllStates.js
ghop02: I'll give it a shot and see if it takes much effort!

ghop02: Question - are the sets of states the same between the two? I guess I can find that out pretty easily

PR #2105, opened 2020-12-09
Update charts holder to take region

ghop02:
This refactors just the charts holder to rely on region more.  Children components still use stateId (or state code) and countyId (which is the county url segment).  It uses values from `region` where possible and recreates `stateCode` and `countyId` from the region


src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -141,80 +142,71 @@ const ChartsHolder = (props: {
...
+    projections.getMetricValues(),
   );
 
   // TODO(pablo): Create separate refs for signup and share
   return (
     <>
-      {!projection ? (
-        <NoCountyDetail
-          countyId={props.county?.county_url_name}
-          stateId={props.stateId}
```
ghop02, line 159: I removed this because projection is never null here.  I may re-add it back though just for the sake of the diff

ghop02: re-added it :D

pnavarrc (approved):
🚢

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -124,12 +126,12 @@ const ChartsHolder = (props: {
...
   const recommendsShareUrl = getRecommendationsShareUrl(
-    props.projections.primary.fips,
+    projections.primary.fips,
   );
 
-  const alarmLevel = props.projections.getAlarmLevel();
+  const alarmLevel = projections.getAlarmLevel();
   const recommendsShareQuote = getShareQuote(
-    props.projections.locationName,
+    projections.locationName,
```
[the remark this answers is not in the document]
ghop02: hmm, I think this has some additional logic in the projections object to handle nyc counties, so might leave this here for now

PR #2106, opened 2020-12-09
Add metro areas to the regions API

pnavarrc:
This PR adds the metro areas to the regions API and fixes a bug (feature?) that was causing us to miscalculate the population of the metro area "New York-Newark-Jersey City, NY-NJ-PA".

I organized the code a bit more and split the logic in 3 parts. I'm also importing internally from the corresponding files instead of importing from the `index` file to avoid circular references (I'm not really sure if this is a problem). 
  - `types.ts` with enums and class definitions
  - `regions_data.ts` with reading and parsing data, exporting dictionaries for counties, states and metro areas
  - `region_db.ts` with the regions class, it exports a single instance with all the data already loaded.

# Notes
There are 4 counties in New York State that share the same FIPS code (`full_fips_code`), which was breaking a test and causing the total population for "New York-Newark-Jersey City, NY-NJ-PA" to be short by ~600k from its actual number. I'm not sure if we should treat these counties as one or split them.


src/common/regions/region_db.ts
```diff
@@ -48,7 +61,7 @@ class RegionDB {
   }
 
   all(): Region[] {
-    return this.regions;
+    return [...this.states, ...this.counties, ...this.metroAreas];
```
ghop02, line 64: I don't think it'll really be an issue, but maybe better to not include metro areas until they're really supported elsewhere

src/common/regions/regions_data.ts
```diff
@@ -0,0 +1,127 @@
+import { chain, Dictionary, fromPairs, groupBy, values } from 'lodash';
+import US_STATE_DATASET from 'components/MapSelectors/datasets/us_states_dataset_01_02_2020.json';
```
ghop02, line 2: Looks like you have a couple lint errors ^^

ghop02 (approved):

pnavarrc:
[CBSA: Add MSAs to the Region API](https://trello.com/c/hsoRqSbf/651-cbsa-add-msas-to-the-region-api)

PR #2108, opened 2020-12-10
implement autocomplete regions component

pnavarrc:
Similar to `AutocompleteLocations`, but for regions. I want to update the email alert form to use this component, but will do that on a separate PR.

I added the `ariaLabelledBy` prop so we can label the input correctly


ghop02 (approved):
lgtm!

pnavarrc:
[CBSA: Update the AutocompleteLocations component to support MSAs](https://trello.com/c/rV2G0vIK/652-cbsa-update-the-autocompletelocations-component-to-support-msas)

PR #2109, opened 2020-12-10
refactor location page + make it typescript!

ghop02:


src/screens/LocationPage/LocationPage.tsx
```diff
@@ -7,41 +7,33 @@ import MiniMap from 'components/MiniMap';
...
+import { getStateCode, Region } from 'common/regions';
+import { assert } from 'common/utils';
 
-function LocationPage({ region }) {
-  let { stateId, countyId, chartId } = useParams();
+interface LocationPageProps {
+  region: Region;
+}
+
+function LocationPage({ region }: LocationPageProps) {
```
pnavarrc, line 20: Assigning the `React.FC` type to function components is slightly better than using a function with implicit return types because some components only accept children with the `React.FC` type. Not a big deal though!

```tsx
const LocationPage: React.FC<{LocationPageProps}> = ({}) => {

  return ();
}
```

pnavarrc (approved):
:shipit:

PR #2112, opened 2020-12-10
add fail import

ghop02:


src/common/utils/model.ts
```diff
@@ -14,6 +14,7 @@ import moment from 'moment';
 import { assert } from '.';
 import { getSnapshotUrlOverride } from './snapshots';
 import { Region, County, State, getStateCode } from 'common/regions';
+import { fail } from 'assert';
```
mikelehen, line 17: I think we actually want to import it from 'common/utils' which is our own implementation (which among other things plays better with TypeScript) instead of the node.js one.

mikelehen (approved):

PR #2113, opened 2020-12-10
Return NYC from regions database

ghop02:
Queries to regions from either fips codes or regions were returning the rest of the nyc counties.  we just want to return a single nyc county for all nyc boroughs


mikelehen (approved):
LGTM.  Thanks for adding tests. 😁 

If you want to be awesome, you can merge develop into https://github.com/covid-projections/covid-projections/pull/2090 and delete all this code. 😂

PR #2116, opened 2020-12-10
Share buttons with regions

ghop02:
This pushes regions into the share section.  I've tested locally and it seems to work, but may need a closer set of eyes on this

### DIFF
--- a/src/components/LocationPage/ShareButtons.tsx
+++ b/src/components/LocationPage/ShareButtons.tsx
@@ -109,8 +109,6 @@ const InnerContent = ({
     <ClickAwayListener onClickAway={() => setShowShareIcons(false)}>
       <ClickAwayWrapper>
         <SaveOrShareContainer>
-          {/* HELP: Why the type error? */}
-          {/* 
           <SaveOrShareButton
             onClick={() => {
               setShowShareIcons(false);
@@ -127,7 +125,6 @@ const InnerContent = ({
           >
             Share
           </SaveOrShareButton>
-          */}
         </SaveOrShareContainer>
         <SocialButtonsWrapper
           onClick={() => {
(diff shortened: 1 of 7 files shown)


src/components/LocationPage/ShareButtons.tsx
```diff
@@ -65,9 +74,8 @@ const InnerContent = ({
 
     // @ts-ignore
     const chartType = chartDownloadType[chartIdentifier];
-    const location = countyId
-      ? `${countyId}_${stateId.toLowerCase()}`
-      : `${stateId.toLowerCase()}`;
+    // TODO Make this better capture
+    const location = region.fullName;
```
pnavarrc, line 78: We might need so simplify the name, consider 

```ts
import { deburr, words } from 'lodash';
const location = deburr(words(region.fullName).join('_')).toLowerCase();
```

src/components/ShareBlock/ShareBlock.tsx
```diff
@@ -28,28 +27,27 @@ import {
   SocialTextAreaWrapper,
   EmbedPrompt,
 } from './ShareBlock.style';
-import { STATES } from 'common';
 import { trackShare } from 'components/Analytics';
+import { Region } from 'common/regions';
+import { STATES } from 'common';
```
pnavarrc, line 32: Consider using `regions.states.map(state => state.stateCode)` instead of `STATES`

pnavarrc (commented):
Added some notes, you might need to rebase from the base branch!

src/components/ShareBlock/ShareBlock.tsx
ghop02: 💯

ghop02: Oh this is actually just for the location page match, used as a type for the keys

src/components/ShareBlock/ShareModelBlock.tsx
```diff
@@ -1,64 +1,62 @@
...
         projections={projections}
         onClickEmbed={() => setShowEmbedPreviewModal(true)}
         stats={stats}
       />
+      {/* todo: add region to this. county={county} */}
       <EmbedPreview
         open={showEmbedPreviewModal}
-        county={county}
         onClose={() => setShowEmbedPreviewModal(false)}
+        county={county}
```
pnavarrc, line 41: The county here is supposed to be a "location" county, not a region, this will probably fail

pnavarrc (approved):
Looks good!

PR #2117, opened 2020-12-10
implement the MiniMap for regions (including MSA)

pnavarrc:
This PR implements a map for Metro Areas. Is not incorporated yet with the MiniMap component, but it can be used in MiniMap to show metro areas.

- Creates a lower-resolution TopoJSON file for states and counties to make rendering faster
- Implement a `MetroAreaMap` component that renders a metro area

image

## Notes
- It doesn't have labels for the states yet, positioning for them is likely to be challenging, I will try once we are closer to be done.


src/components/RegionMap/MetroAreaMap.tsx
```diff
@@ -0,0 +1,161 @@
...
+        width={width}
+        projectionConfig={projectionConfig}
+        data-tip=""
+      >
+        <Geographies key="states" geography={statesTopoJson}>
+          {({ geographies }) =>
+            geographies.map(geo => (
+              <Link
+                key={geo.rsmKey}
+                to={getRelativeUrlByFips(geo.id)}
```
ghop02, line 48: thoughts on loading the region once? This strategy kinda depends on all of the regions being there, but they should be, right?
```
geo => {
  const region = regions.getRegionsByFips(geo.id);
  assert(region);
  return (
   <Geographies> ... (then you don't need the helper functions)
  )
}

```

ghop02 (approved):
🚢 

I think we may want to zoom out a bit - a bit difficult to see exactly what borders are around but I say merge it!

PR #2118, opened 2020-12-10
Regions on projection

ghop02:
Alrighty! Onto projections. This cleans up projections to use region.


src/common/models/Projections.ts
```diff
@@ -15,37 +15,36 @@ import { RegionSummaryWithTimeseries } from 'api/schema/RegionSummaryWithTimeser
...
   }
 
   populateCounty(county: any) {
     if (!county) {
       return;
     }
+  }
+
+  get fips(): string {
+    return this.primary.fips;
```
pnavarrc, line 40: Should we return `region.fipsCode` instead?

ghop02: I don't think it matters too much here, they should never be different

pnavarrc (approved):

PR #2119, opened 2020-12-10
Adding in regions in compare table

ghop02:
This changes the guts of the compare table.  I need to do some more verification - especially around sharing.  

To be clear, I haven't changed the external ways of calling the compare table just yet, but this is a big step in the right direction.


src/components/Compare/CompareTableRow.tsx
```diff
@@ -55,12 +56,16 @@ const CompareTableRow = (props: {
     return 4;
   }
 
-  const fipsCode =
-    location?.locationInfo?.full_fips_code ||
-    location?.locationInfo.state_fips_code;
-  const locationLink = `/${getCanonicalUrl(fipsCode)}`;
+  const fipsCode = location?.region.fipsCode;
```
chasulin, line 59: i think this `?` can go

src/components/Compare/CompareTable.tsx
```diff
@@ -96,7 +94,7 @@ const CompareTable = (props: {
   }
 
   const getPopulation = (location: SummaryForCompare) =>
-    location?.locationInfo?.population;
+    location?.region?.population;
```
chasulin, line 97: hrmm same here- i think both of these `?` can go

src/components/Compare/CompareMain.tsx
```diff
@@ -206,7 +208,7 @@ const CompareMain = (props: {
     setShowFaqModal,
     createCompareShareId,
   };
-
+  console.log('HI');
```
chasulin, line 211: 👋

chasulin (approved):
some new `!`, but it looks good and seems to work just fine. i qa’ed a bunch but maybe you can click around a bit more for sanity.

also not sure what you mean by you “haven’t changed the external ways of calling compare yet”?

PR #2124, opened 2020-12-10
Remove useProjections and fetchProjections

ghop02:


src/components/Explore/utils.ts
```diff
@@ -12,7 +12,7 @@ import {
 } from 'lodash';
 import { color } from 'd3-color';
 import { schemeCategory10 } from 'd3-scale-chromatic';
-import { fetchProjections } from 'common/utils/model';
+import { fetchProjectionsRegion } from 'common/utils/model';
```
mikelehen, line 15: FWIW- "fetchProjections" is a very antiquated name.  I didn't get a chance to review the PR that added fetchProjectionsRegion, but we would ideally be moving towards better naming, e.g. `fetchRegionData()` or `fetchLocationData()`. We can punt for now though.

src/screens/internal/CompareSnapshots/CompareSnapshots.tsx
```diff
@@ -418,7 +418,7 @@ function useProjectionsSet(
           ),
         );
       } else if (locations === Locations.TOP_COUNTIES_BY_POPULATION) {
-        const topCounties = topCountiesByPopulation(COUNTIES_LIMIT);
+        const topCounties = regions.topCountiesByPopulation(COUNTIES_LIMIT);
```
mikelehen, line 421: Can we delete the version of this method in common/locations now?

mikelehen (approved):

src/screens/internal/CompareSnapshots/CompareSnapshots.tsx
ghop02: oo probably, i'll check

src/components/Explore/utils.ts
ghop02: `fetchRegionTimeseries` might be good too

ghop02: Actually, i'd rather change the name alongside changing the `Projections` class name, will 🦶 🏈 for now

mikelehen: SG.  I'd keep "timeseries" out of it since it includes a bunch of timeseries and non-timeseries data.

FWIW- It also includes helper methods for getting risk level color, etc. so it's really quite broad. 😁

PR #2125, opened 2020-12-10
Updating urls to use region where possible

ghop02:
This updates a handful of url functions to use region.  one thing to think about is we might want to use `window.location.origin` in `canonicalUrl` so that all the links point to links specific to the domain we're on.

A couple functions didn't have any callers with county and state, so I just removed those arguments in favor of cutting code, can re add if we think we'll use those in the future.


chasulin (approved):
looks good! +1 to using window.location.origin so it points wherever you are

PR #2127, opened 2020-12-10
update location page metatags to use region

pnavarrc:


src/screens/LocationPage/utils.ts
```diff
@@ -1,27 +1,18 @@
...
-    ? `${countyName}, ${stateName} (${stateCode})`
-    : `${stateName} (${stateCode})`;
-  return locationName;
-}
-
-export function getPageTitle(projections: Projections): string {
-  const locationName = getLocationName(projections.fips);
-  return `${locationName} - COVID Risk Map & Key Metrics`;
+export function getPageTitle(region: Region): string {
+  return `${region.shortName} - COVID Risk Map & Key Metrics`;
```
ghop02, line 7: Is short name the correct one to use? Feels like we should use fullName + the title will change a bit

chasulin, line 7: look how efficient that is !

chasulin (approved):

src/screens/LocationPage/utils.ts
chasulin: ie.-

 Kimball County, Nebraska (NE) - COVID Risk Map & Key Metrics

Nebraska (NE) - COVID Risk Map & Key Metrics

pnavarrc: right, not quite the same - I guess I would handle it type by type with a good fallback, good catch!

PR #2128, opened 2020-12-10
draft: MSA location page

pnavarrc:
This is a draft PR to test which parts of the location page work for MSAs, do not merge!


src/screens/LocationPage/LocationRouter.tsx
```diff
@@ -2,19 +2,37 @@ import React from 'react';
...
 interface LocationPageUrlParams {
-  stateId: string;
+  stateId: string | undefined;
   countyId: string | undefined;
+  metroAreaUrlSegment: string | undefined;
 }
 
 const LocationRouter: React.FC = () => {
-  let { stateId, countyId } = useParams<LocationPageUrlParams>();
+  let { stateId, countyId, metroAreaUrlSegment } = useParams<
```
ghop02, line 14: I would do:
```
let region = useRegionFromParams();
```
and update the hook to query the new function you used.  then you can do:
```
if (region instance of MetroArea) {
 
} else {
return (
  <RegionContext.Provider value={metroArea}>
    <RegionPage region={metroArea} />
  </RegionContext.Provider>
);
```

ghop02:
Overall looks good. I think the NYC metro area is a bit weird because of nyc aggregation (i think it should include most of the counties in the middle). 

Also, the rotation is off - is that an easy fix?

PR #2136, opened 2020-12-11
Ghop02/region page even more of a draft

ghop02:


src/components/LocationPage/ShareButtons.tsx
```diff
@@ -32,6 +32,11 @@ const getShareImageUrl = (region: Region, chartIdentifier: number): string => {
       imageBaseUrl +
       `states/${state.stateCode.toLowerCase()}/chart/${chartIdentifier}/export.png`
     );
+  } else if (region instanceof MetroArea) {
+    return (
+      imageBaseUrl +
+      `metros/${region.fipsCode}/chart/${chartIdentifier}/export.png`
```
ghop02, line 38: Note that I don't actually think this will work on the backend yet

PR #2137, opened 2020-12-11
add msa to the minimap component

pnavarrc:


pnavarrc:
[CBSA: Mini Map](https://trello.com/c/sqTs2KfL/656-cbsa-mini-map)

src/components/MiniMap/MiniMap.tsx
```diff
@@ -25,16 +25,12 @@ const MiniMap: FunctionComponent<MiniMapProperties> = ({
...
-  const stateName = getStateName(region);
-  assert(stateName);
-  assert(stateCode);
-  const showState = stateCode !== MAP_FILTERS.DC;
-
   const onSelectCounty = () => {
     setMobileMenuOpen(false);
   };
 
+  const showState = region.fipsCode !== '11';
```
ghop02, line 32: maybe `region.fipsCode.startswith('11')` (or whatever it is in typescript haha) since there can be dc state and counties (though we should change this)

src/components/MiniMap/MiniMap.tsx
```diff
@@ -43,7 +39,7 @@ const MiniMap: FunctionComponent<MiniMapProperties> = ({
             selected={mapOption === MAP_FILTERS.STATE}
             onClick={() => setMapOption(MAP_FILTERS.STATE)}
           >
-            {stateName}
+            {region.abbreviation}
```
ghop02, line 42: This will take it from the full state name to the abbreviation - this should probably be `fullName`

ghop02 (commented):
Couple small points, but looks good!

src/components/MiniMap/MiniMap.tsx
pnavarrc: I think it should be `shortName`, `fullName` might be too long for MSAs

ghop02 (approved):

PR #2144, opened 2020-12-11
Guard against missing counties

ghop02:
This guards against the case where there is a new county in a snapshot.  

If a county is missing in a snapshot, `fetchProjectionsRegion` will fail an assertion. passing in the summaries map will let us check that the fips exists in the snapshot before attempting to fetch.


mikelehen (changes requested):
I think this makes the compare tool even more reliant on summary files existing, which I'd ideally like to avoid.  Originally the summary files were just an optimization to make it faster if they existed.  But then when the API files got too big to load from the browser, I think we ended up in a world where only comparing states and top 100 population counties worked.  With this PR I think nothing will work without summary files?

So I think my preference would be to just try/catch the failure instead.

mikelehen:
Propose merging https://github.com/covid-projections/covid-projections/pull/2150 instead.

PR #2149, opened 2020-12-11
Deploy snapshot 1433 with HHS state hospital data, metric explainers copy tweak, and compare fix

BrettBoval:
Update ICU headroom used copy to reflect the fact that we use better data sources now.


pnavarrc (approved):
Looks good!

BrettBoval:
Holding off until we flip the switch in covid-data-model too.

PR #2160, opened 2020-12-14
Fix aggregations to deal with non-uniform start date from API.

mikelehen:
We just manually start at 2020-03-01 in all cases now.


mikelehen:
[USA and NAMC aggregations are broken in website.](https://trello.com/c/PMnZ3DqX/665-usa-and-namc-aggregations-are-broken-in-website)

PR #2162, opened 2020-12-14
Reenable email alerts report and add subscriptions by date.

mikelehen:
I haven't hooked up AWS engagement stats, but we at least get the location stats and the (new) "subscribers by date" stats.

Context: Gaia had asked for a graph of subscribers over time.

### DIFF
--- a/scripts/alert_emails_report/generate-report.ts
+++ b/scripts/alert_emails_report/generate-report.ts
@@ -1,13 +1,10 @@
-// TODO(michael): This script no longer works since we switched to AWS Simple Email Service. :~(
-
 import path from 'path';
+import moment from 'moment';
 import _ from 'lodash';
-import CampaignMonitor, {
-  CampaignMonitorStats,
-} from '../alert_emails/campaign-monitor';
 import {
   getFirestore,
   fetchAllAlertSubscriptions,
+  Subscription,
 } from '../alert_emails/firestore';
 import GoogleSheets, { Cell } from '../common/google-sheets';
 import {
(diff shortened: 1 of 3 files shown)


scripts/alert_emails_report/generate-report.ts
```diff
@@ -81,6 +84,39 @@ async function updateSubscriptionsByLocation() {
...
+async function updateSubscriptionsByDate(subscriptions: Subscription[]) {
+  const gsheets = new GoogleSheets(keyFile);
+  const spreadsheetId = getSpreadsheetId();
+
+  const sortedSubscriptions = _.chain(subscriptions)
+    .filter(s => s.subscribedAt !== undefined)
+    .sortBy(s => s.subscribedAt)
+    .value();
+
+  const dateTotals = [];
```
pnavarrc, line 96: Maybe this can be simplified a bit?

```
const dateTotals = _.chain(subscriptions)
  .filter(s => s.subscribedAt !== undefined)
  .groupBy(s => moment(s.subscribedAt).format('YYY-MM-DD'))
  .map((subscriptionList, dateKey) => [dateKey, subscriptionList.length])
  .value()
```

pnavarrc (approved):
Looks good, added some suggestions.

scripts/alert_emails_report/generate-report.ts
mikelehen: While this is much cleaner (thank you!), it ends up computing daily deltas instead of daily cumulatives which means there's an extra step involved to turn it into the data we can graph. So I've opted to leave the existing code for now, since I haven't gotten a chance to revisit this. Thanks / sorry!

PR #2167, opened 2020-12-14
update location header to support MSAs

pnavarrc:
Update the `LocationPageHeading` component to support MSAs. I'm not adding a link on the state codes because some MSAs go to more than one state.

image

I tested this on the `pablo/region-page-draft` branch, but didn't commit the changes there because this can be merged to `develop`.

### DIFF
--- a/src/components/LocationPage/LocationPageHeading.tsx
+++ b/src/components/LocationPage/LocationPageHeading.tsx
@@ -15,7 +15,9 @@ const LocationPageHeading: React.FC<{
       </Styles.Container>
     );
   } else if (region instanceof County) {
-    const stateUrl = `${isEmbed ? '/embed' : ''}/us/${region.state.urlSegment}`;
+    const stateUrl = `${isEmbed ? '/embed' : ''}/us/${
+      region.state.relativeUrl
+    }`;
     return (
       <Styles.Container>
         <Styles.HeaderTitle $isEmbed={isEmbed}>
--- a/src/components/LocationPage/LocationPageHeading.tsx
+++ b/src/components/LocationPage/LocationPageHeading.tsx
@@ -1,5 +1,6 @@
 import React, { Fragment } from 'react';
-import { Region, State, County } from 'common/regions';
+import { Region, State, County, MetroArea } from 'common/regions';
+import { HeaderStateCode } from './LocationPageHeader.style';
 
 const LocationPageHeading: React.FC<{ region: Region; isEmbed: boolean }> = ({
   region,
@@ -10,10 +11,21 @@ const LocationPageHeading: React.FC<{ region: Region; isEmbed: boolean }> = ({
   } else if (region instanceof County) {
     return (
       <Fragment>
-        <strong>{region.name}, </strong>
-        <a href={`${isEmbed ? '/embed' : ''}/us/${region.state.urlSegment}`}>
-          {region.state.stateCode}
-        </a>
+        <strong>{region.name}</strong>
+        {', '}
+        <HeaderStateCode>
+          <a href={`${isEmbed ? '/embed' : ''}/us/${region.state.urlSegment}`}>
+            {region.state.stateCode}
+          </a>
+        </HeaderStateCode>
+      </Fragment>
+    );
+  } else if (region instanceof MetroArea) {
+    return (
+      <Fragment>
+        <strong>{region.name}</strong>
+        {', '}
+        <HeaderStateCode>{region.stateCodes}</HeaderStateCode>
       </Fragment>
     );
   } else {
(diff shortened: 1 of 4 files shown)


src/components/LocationPage/LocationPageHeading.tsx
```diff
@@ -10,10 +11,21 @@ const LocationPageHeading: React.FC<{ region: Region; isEmbed: boolean }> = ({
...
     return (
       <Fragment>
-        <strong>{region.name}, </strong>
-        <a href={`${isEmbed ? '/embed' : ''}/us/${region.state.urlSegment}`}>
-          {region.state.stateCode}
-        </a>
+        <strong>{region.name}</strong>
+        {', '}
+        <HeaderStateCode>
+          <a href={`${isEmbed ? '/embed' : ''}/us/${region.state.urlSegment}`}>
```
ghop02, line 17: I realize you didn't change this, but I think this href logic is more clear as a variable above rather than here

src/components/LocationPage/LocationPageHeading.tsx
```diff
@@ -10,10 +11,21 @@ const LocationPageHeading: React.FC<{ region: Region; isEmbed: boolean }> = ({
...
+          </a>
+        </HeaderStateCode>
+      </Fragment>
+    );
+  } else if (region instanceof MetroArea) {
+    return (
+      <Fragment>
+        <strong>{region.name}</strong>
+        {', '}
+        <HeaderStateCode>{region.stateCodes}</HeaderStateCode>
```
ghop02, line 28: is state codes a string? Taking a look at the `MetroArea` object, I think the `stateCodes` + `stateCodesList` property might be unnecessary.  Joining them with a `-` seems a bit arbitrary and I think i would be more clear if `stateCodes` was a list of state codes and the logic on how to present them just lives here

src/components/LocationPage/LocationPageHeading.tsx
```diff
@@ -10,10 +11,21 @@ const LocationPageHeading: React.FC<{ region: Region; isEmbed: boolean }> = ({
   } else if (region instanceof County) {
     return (
       <Fragment>
```
ghop02, line 13: A bit of a orthogonal comment, but what are your thoughts on having:

```jsx
<HeaderTitle isEmbed={isEmbed}>
  <LocationPageHeading region={region} isEmbed={isEmbed} />
</HeaderTitle>
```
vs 
```
<LocationPageHeading region={region} isEmbed={isEmbed} />
```
where the `HeaderTitle` is basically in the `<LocationPageHeading />`.

src/components/LocationPage/LocationPageHeading.tsx
pnavarrc: That makes sense, however, I think it makes sense to define a standard for rendering those so we don't have to join them each time and also to avoid having too many versions of the same thing (separated by comma, separated by dashes, etc).

src/components/LocationPage/LocationPageHeading.tsx
pnavarrc: good point!

src/components/LocationPage/LocationPageHeading.tsx
ghop02: Definitely! Maybe even just formattedStateNames or something similar would be more clear. Definitely a nit and not a big deal though

src/components/LocationPage/LocationPageHeading.tsx
```diff
@@ -1,20 +1,41 @@
-import React, { Fragment } from 'react';
-import { Region, State, County } from 'common/regions';
+import React from 'react';
+import { Region, State, County, MetroArea } from 'common/regions';
+import * as Styles from './LocationPageHeading.style';
```
ghop02, line 3: ooo i like this renaming!

src/components/LocationPage/LocationPageHeading.tsx
```diff
@@ -1,20 +1,41 @@
...
-    return <strong>{region.name}</strong>;
+    return (
+      <Styles.Container>
+        <Styles.HeaderTitle $isEmbed={isEmbed}>
+          <strong>{region.name}</strong>
+        </Styles.HeaderTitle>
+      </Styles.Container>
+    );
   } else if (region instanceof County) {
+    const stateUrl = `${isEmbed ? '/embed' : ''}/us/${region.state.urlSegment}`;
```
ghop02, line 18: `${isEmbed ? '/embed' : ''}/${region.state.relativeUrl}` maybe?

ghop02 (approved):
Thanks for moving that!

PR #2168, opened 2020-12-14
Search bar updates

chasulin:


src/components/Search/SearchAutocomplete.tsx
```diff
@@ -0,0 +1,61 @@
...
+  // Todo (chelsi) - theres prob a better/built-in way to grab the county url
+  const onSelect = (e: any, value: any) => {
+    if (value.regionType === RegionType.STATE) {
+      const stateUrl = value.canonicalUrl;
+      window.location.href = stateUrl;
+    } else if (value.regionType === RegionType.COUNTY) {
+      const countyUrl = `${value.state.canonicalUrl}/county/${value.urlSegment}`;
+      window.location.href = countyUrl;
+    }
+  };
```
pnavarrc, line 32: All the region elements have the `relativeUrl` property, we don't need to have special handling

src/components/Search/SearchAutocomplete.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const SearchAutocomplete = (props: {
+  locations: Region[];
+  region?: Region;
+}) => {
+  const { locations, region } = props;
+
+  function getOptionLabel(location: any) {
+    return location.name;
+  }
```
pnavarrc, line 21: `getOptionLabel` can take a `Region` and use `name` or `fullName`

src/components/Search/SearchAutocomplete.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { Autocomplete } from '@material-ui/lab';
+import { createFilterOptions } from '@material-ui/lab/useAutocomplete';
+import TextField from '@material-ui/core/TextField';
+import { getFilterLimit } from './utils';
+import { RegionType, Region } from 'common/regions';
+import MenuItem from './MenuItem';
+
+function getOptionSelected(option: any, selectedOption: any) {
+  return option.fipsCode === selectedOption.fipsCode;
+}
```
pnavarrc, line 11: Both `option` and `selectedOption` should have `Region` type

src/common/regions/index.ts
```diff
@@ -29,8 +29,10 @@ export abstract class Region {
   abstract get abbreviation(): string;
   abstract get relativeUrl(): string;
 
+  // Todo (Chelsi) - switch back before merging (or maybe keep this change)
   get canonicalUrl() {
-    return urlJoin('https://covidactnow.org', this.relativeUrl);
+    // return urlJoin('https://covidactnow.org', this.relativeUrl);
+    return urlJoin(window.location.origin, this.relativeUrl);
```
pnavarrc, line 35: I _think_ we can keep this change (we will need to test it with embeds)

src/components/Search/utils.ts
```diff
@@ -0,0 +1,67 @@
+import regions, { RegionType, County, getStateFips } from 'common/regions';
+import { sortBy, partition } from 'lodash';
+import { countyFipsToZips } from 'components/MapSelectors/datasets';
+import { stateColor, countyColor } from 'common/colors';
+
+// Todo (Chelsi): replace the many anys
+
+function getCountyZips(countyFips: string): string[] {
+  return countyFipsToZips[countyFips];
+}
```
pnavarrc, line 10: We should probably add the zip codes to the `County` class and add them on `common/regions/regions_data` to keep things centralized

src/components/StateSvg/StateCircleSvg.js
```diff
@@ -1,17 +1,11 @@
...
-  actionBackgroundFill,
-  fillColor,
-  hasAction = false,
-}) => {
+const StateCircleSvg = ({ state, ratio, fillColor, hasAction = false }) => {
   return (
     <Circle
       ratio={ratio}
-      actionBackgroundFill={actionBackgroundFill}
+      actionBackgroundFill="#FFFFFF"
```
ghop02, line 8: Is there a constant for white we could use?

src/components/Search/SearchAutocomplete.tsx
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  // Todo (chelsi) - theres prob a better/built-in way to grab the county url
+  const onSelect = (e: any, value: any) => {
+    if (value.regionType === RegionType.STATE) {
+      const stateUrl = value.canonicalUrl;
+      window.location.href = stateUrl;
+    } else if (value.regionType === RegionType.COUNTY) {
+      const countyUrl = `${value.state.canonicalUrl}/county/${value.urlSegment}`;
+      window.location.href = countyUrl;
```
ghop02, line 30: Is this doing the navigation? Should we be using react router to do the navigation instead?

src/components/Search/SearchAutocomplete.tsx
```diff
@@ -0,0 +1,61 @@
...
+      noOptionsText="No location found"
+      options={locations}
+      getOptionLabel={getOptionLabel}
+      disableListWrap
+      disableClearable
+      onChange={onSelect}
+      getOptionSelected={getOptionSelected}
+      filterOptions={createFilterOptions({
+        matchFrom: 'start',
+        limit: getFilterLimit(region),
```
ghop02, line 45: random idea: could we use `config.stringify` from [createFilterOptions](https://material-ui.com/components/autocomplete/#createfilteroptions-config-filteroptions) to add various properties to the string that's matched? i.e. `stringify: (option) => `${option.region.fullName} ${option.region.zipCodes.join(' ')`

> config.stringify (Func [optional]): Controls how an option is converted into a string so that it can be matched against the input text fragment.

src/components/Search/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/* 
+  Determines amount of locations that show in dropdown menu when clicking into searchbar.
+  If on homepage, should only show states.
+  If on location page, should only show counties within state.
+*/
+export function getFilterLimit(region?: any) {
+  if (region) {
+    const stateFips = getStateFips(region);
+    const countiesInState = regions.counties.filter(county =>
```
ghop02, line 61: maybe this could be a function on `RegionDB` `regions.getCountiesForState(region: State)`

src/components/Search/SearchAutocomplete.tsx
chasulin: solid. yeah feel free to ignore the `any`s for now

ghop02 (commented):
Looks good! had a couple questions/ideas.  Is there a list of things we need to finish up before merging in?

A couple things:
- I noticed the styling is subtly different. do we need to match that before deploying to master?
- looks like we need a bit more work for zip codes

Also if it's faster we could continue using the old GlobalSelector and continue to iterate on this guy.  Would be great to have it on develop sooner rather than later so we could use it in a feature branch for metros, but don't have super strong opinions

src/components/Search/utils.ts
chasulin: k can add that in

src/components/Search/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+  return [...stateCounties, ...sortedStates, ...otherCounties];
+}
+
+/* To get color of location icon in dropdown menu:  */
+export function getStateIconFillColor(region?: any) {
+  if (region.regionType === RegionType.STATE) {
+    return stateColor(region.stateCode);
+  }
+  return countyColor(region.fipsCode, stateColor(region.state.stateCode));
+}
```
pnavarrc, line 37: Consider simplifying the logic a bit by separating the cases more explicitly 

```ts
function getSearchAutocompleteLocations(region: Region | null) {
  // Homepage
  if (!region) {
    return regions.all;
  }

  if (region instanceof State | region instanceof County) {
    return [...stateCounties, ...stateMSAs, ...states, ...otherCounties]; 
  } else if (region instanceof MetroArea) {
    return [...states, ...metroAreas, ...counties];
  } else {
    return [];
  }

}
```

src/components/Search/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+    sortedCountiesWithZips,
+    county => belongsToState(county, stateFips),
+  );
+
+  return [...stateCounties, ...sortedStates, ...otherCounties];
+}
+
+/* To get color of location icon in dropdown menu:  */
+export function getStateIconFillColor(region?: any) {
+  if (region.regionType === RegionType.STATE) {
```
pnavarrc, line 33: I would recommend writing this function in terms of a LocationSummary, so it doesn't depend on the specifics of the region

```ts
getFillColor(locationLevel: Level) {
  return LOCATION_SUMMARY_LEVELS[locationLevel].color
}
```

src/components/Search/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+export function getFilterLimit(region?: any) {
+  if (region) {
+    const stateFips = getStateFips(region);
+    const countiesInState = regions.counties.filter(county =>
+      belongsToState(county, stateFips),
+    );
+    return countiesInState.length;
+  }
+  return regions.states.length;
+}
```
pnavarrc, line 53: Does this help? We can probably set a fix number if this is only done for performance (100 or so)

src/components/StateSvg/StateCircleSvg.js
chasulin: i actually think this entire `actionBackgroundFill` can come out. it looks like a remnant from a `hasAction` prop that doesnt seem like its used anywhere anymore

src/screens/HomePage/HomePage.js
```diff
@@ -115,10 +102,13 @@ export default function HomePage() {
...
-                      handleChange={handleSelectChange}
-                      extendRight={undefined}
-                    />
+                    <ParentSize>
+                      {() => (
+                        <SearchAutocomplete
+                          locations={getSearchAutocompleteLocations()}
+                        />
+                      )}
+                    </ParentSize>
```
pnavarrc, line 111: We might not need `ParentSize`, we should be able to set the styles so it expands to take the width of its parent. Using `ParentSize` should be used when we _really_ need the size in pixels, because `ParentSize` needs to render a couple of times to get the size in pixels

src/components/Header/SearchHeader.tsx
```diff
@@ -62,10 +44,14 @@ const SearchHeader = ({
...
-                extendRight={true}
-                handleChange={handleSelectChange}
-              />
+              <ParentSize>
+                {() => (
+                  <SearchAutocomplete
+                    locations={getSearchAutocompleteLocations(region)}
+                    region={region}
+                  />
+                )}
```
pnavarrc, line 53: I think that the props `locations` and `region` are a bit redundant here, since the `locations` can be determined from the `region`. Consider either moving the logic to get the locations inside the component or not passing the `region` (I think that this might be better)

src/components/Search/utils.ts
chasulin: yeah it makes sure only states show when the textfield is in focus but nothing is yet typed. feels like it makes the most sense like this, rather than having an arbitrary # (which might include all states and then a random # of counties)

chasulin: ditto to location pages - makes the default dropdown selection include all counties within the state

pnavarrc (approved):
Added some notes, I can test again when we have styling!

src/components/Search/SearchAutocomplete.tsx
```diff
@@ -0,0 +1,89 @@
...
+    // }
+    else {
+      window.location.href = '/';
+    }
+  };
+
+  const zipCodeInput = isZip ? input : '';
+
+  return (
+    <Autocomplete
```
ghop02, line 62: Should we use the [freeSolo](https://material-ui.com/components/autocomplete/#free-solo) property to remove the triangle? It seems like that's built for our use case:

src/screens/HomePage/HomePage.js
chasulin: going to punt this for after merging it into msa preview link (as we discussed- still gotta work on properly sizing it on mobile)

src/components/StateSvg/StateCircleSvg.js
chasulin: punting this to old-code-cleanup-time

src/components/Search/utils.ts
chasulin: also punting and adding to todo list

src/common/regions/index.ts
chasulin: reverting for now

src/components/Search/utils.ts
chasulin: adding to list/punting for now for the sake of merging this into preview link asap

src/components/Header/SearchHeader.tsx
```diff
@@ -13,39 +9,28 @@ import {
   MenuBarWrapper,
   SearchHeaderWrapper,
 } from './SearchHeader.style';
+import SearchAutocomplete from 'components/Search';
+import { ParentSize } from '@vx/responsive';
+import { Region } from 'common/regions';
+import {
+  getSearchAutocompleteLocations,
+  getFilterLimit,
+} from 'components/Search/utils';
```
pnavarrc, line 18: Maybe there should be exported from `components/Search/index` so you can merge the imports?

src/components/Search/MenuItem.tsx
```diff
@@ -0,0 +1,43 @@
...
+              state={getStateCode(region)}
+            />
+          </IconWrapper>
+        )}
+        <div>
+          <LocationName>{region.shortName}</LocationName>
+          {zipCodeInput && <Zipcode>contains zipcode {zipCodeInput}</Zipcode>}
+          <div>
+            <span>{ShortNumber(region.population)} residents</span>
+          </div>
```
pnavarrc, line 36: we probably don't need the span here

src/components/Search/SearchAutocomplete.tsx
```diff
@@ -0,0 +1,74 @@
...
+  };
+
+  const stringifyOption = (option: Region) => {
+    if (isZip) {
+      if ((option as County).zipCodes) {
+        return `${(option as County).zipCodes.join(' ')}`;
+      }
+      return option.name;
+    }
+    return option.name;
```
pnavarrc, line 37: Consider simplifying here

```tsx
if (isZip && (option as County).zipCodes) {
  // ...
}
return option.name;
```

src/components/Search/SearchAutocomplete.tsx
```diff
@@ -0,0 +1,74 @@
...
+  };
+
+  const onSelect = (e: any, value: Region) => {
+    window.location.href = `/${value.relativeUrl}`;
```
pnavarrc, line 41: Maybe we should fix this in `regions_data` instead? we always want it with the leading `/`

src/components/Search/SearchAutocomplete.tsx
```diff
@@ -0,0 +1,74 @@
...
+function getOptionSelected(option: Region, selectedOption: Region) {
+  return option.fipsCode === selectedOption.fipsCode;
+}
+
+const SearchAutocomplete: React.FC<{
+  locations: Region[];
+  filterLimit: number;
+}> = ({ locations, filterLimit }) => {
+  const [input, setInput] = useState('');
+  const [isZip, setIsZip] = useState(false);
```
pnavarrc, line 17: maybe `isZipCode`?

pnavarrc (approved):
Looks good!

src/components/Search/MenuItem.tsx
chasulin: span sets the color to gray (all spans in the menuItem are gray)

PR #2170, opened 2020-12-14
Remove no county detail from charts holder

ghop02:
Just removing the `NoCountyDetail` element because it's not reachable code (we always have a projection if `ChartsHolder` is called)


src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -145,74 +144,63 @@ const ChartsHolder = ({ projections, region, chartId }: ChartsHolderProps) => {
   // TODO(pablo): Create separate refs for signup and share
   return (
     <>
-      {!projection ? (
-        <NoCountyDetail
-          countyId={
-            region.regionType === RegionType.COUNTY ? region.urlSegment : ''
-          }
+      <ChartContentWrapper>
```
pnavarrc, line 147: Consider removing other conditionals on projection above https://github.com/covid-projections/covid-projections/pull/2170/files#diff-074b0a67b501ec0a6b4ea0047e8af2f238c0e7ffb9b70533ed0ba791d6ccedf9R104

pnavarrc (approved):
Looks good! Consider cleaning up some `projections.primary` → `projection` and `projections.getMetricValues()` → `stats`

src/components/LocationPage/ChartsHolder.tsx
ghop02: ooo good call, will add that in my next PR (that's coming pretty much immediately :) )

PR #2175, opened 2020-12-15
Explore as regions

ghop02:
I think this should be good to go! Curious if my solution for the aggregated regions seems reasonable - happy to try something else!


src/common/locations.ts
```diff
@@ -159,14 +159,6 @@ export function getLocationNameForFips(fips: string): string {
   }
 }
 
-export function getRelativeUrlForFips(fips: string): string {
```
pnavarrc, line 162: Unused dependency above! (`urlJoin`)

src/components/Explore/utils.ts
```diff
@@ -396,19 +396,18 @@ function truncateCountyName(countyName: string) {
...
+  if (location.fipsCode && location.fipsCode.startsWith('00')) {
+    return location.name;
   } else {
-    const { county: countyName, state_code } = location;
-    return countyName
-      ? `${truncateCountyName(getAbbreviatedCounty(countyName))}`
-      : state_code.toUpperCase();
+    return location.regionType === RegionType.COUNTY
+      ? `${truncateCountyName(location.abbreviation)}`
+      : location.name.toUpperCase();
```
pnavarrc, line 405: I think we can just use `abbreviation` here (at least for non-county)

src/components/Explore/utils.ts
```diff
@@ -426,12 +425,12 @@ export function getLocationNames(
...
   normalizeData: boolean,
-  locations: Location[],
+  regions: Region[],
 ) {
   const textPer100k = normalizeData ? 'per 100k population' : '';
-  return locations.length === 0
+  return regions.length === 0
     ? 'Select states or counties to explore trends'
-    : `${metricName} ${textPer100k} in ${getLocationNames(locations)}`;
+    : `${metricName} ${textPer100k} in ${getLocationNames(regions)}`;
```
pnavarrc, line 433: We might need to update the placeholder to include metro areas 🎉

pnavarrc (commented):
Looks good!

src/components/Explore/Explore.tsx
```diff
@@ -152,11 +150,12 @@ const Explore: React.FunctionComponent<{
...
 
   const initialLocations = useMemo(
-    () => initialFipsList.map(findLocationForFips),
+    () => initialFipsList.map(fipsCode => regions.findByFipsCode(fipsCode)!),
     [initialFipsList],
   );
+
   const indigeneousPopulationsLocations = useMemo(
-    () => ['00001', '00002'].map(findLocationForFips),
+    () => ['00001', '00002'].map(fips => regions.findByFipsCode(fips)!),
```
pnavarrc, line 158: Unrelated, we probably don't need to use memo anymore since we have the regions by FIPS in a dictionary already

src/components/Explore/utils.ts
```diff
@@ -396,19 +396,14 @@ function truncateCountyName(countyName: string) {
...
-    return location.state_code;
-  } else {
-    const { county: countyName, state_code } = location;
-    return countyName
-      ? `${truncateCountyName(getAbbreviatedCounty(countyName))}`
-      : state_code.toUpperCase();
-  }
+function getShortLocationLabel(location: Region) {
+  return location.regionType === RegionType.COUNTY
+    ? `${truncateCountyName(location.abbreviation)}`
```
pnavarrc, line 401: we don't need the string interpolation here

```ts
return location.regionType === RegionType.COUNTY
  ? truncateCountyName(location.abbreviation)
  : location.abbreviation
```

pnavarrc (approved):
Looks good! Note that on the MSA page the autocomplete doesn't work as well because we haven't included MSAs on the AutocompleteRegion component

src/components/Explore/utils.ts
ghop02: truu

src/components/Explore/Explore.tsx
ghop02: Looks like this removing this instance of useMemo was necessary - may need a slightly larger pass at removing useMemo in general here

PR #2177, opened 2020-12-15
Filters out DC county

chasulin:
DC currently has both a state page and a county page (noticed via DC state page's compare table, which was showing the non-matching county page's data - screenshot below). This PR filters out DC county when building regions in `regions_data.ts`.

[Trello item](https://trello.com/c/tIqXApGD/667-dc-has-two-location-pages-state-and-county-with-different-data)

Screen Shot 2020-12-10 at 2 34 57 PM

### DIFF
--- a/src/common/regions/regions_data.ts
+++ b/src/common/regions/regions_data.ts
@@ -54,34 +54,38 @@ function buildCounties(
   statesByFips: Dictionary<State>,
   countyAdjacency: { [fipsCode: string]: AdjacencyInfo },
 ): County[] {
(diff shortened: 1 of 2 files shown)


ghop02 (approved):
This seems okay as a temp fix, but could we fix it more prominently by removing the dc county where we [build the regions](https://github.com/covid-projections/covid-projections/blob/develop/src/common/regions/regions_data.ts#L53-L56).  Thoughts on adding a `filter` in that function?

ghop02 (approved):

PR #2178, opened 2020-12-15
More metro area in name

ghop02:
This uses `<first city name> metro area` as the name instead of the full cbsa title.  I think we're going to need to add some more context somewhere, but my thinking is that this is a better starting point than the cbsa title.

Also, in this process I realized that our names are a bit confusing on `Region`.  We have `name`, `fullName`, `shortName`, and `abbrevation`.  As we go forward we may want to be more principled as to what those mean exactly.

### DIFF
--- a/src/common/regions/regions_data.ts
+++ b/src/common/regions/regions_data.ts
@@ -94,7 +94,7 @@ function buildMetroAreas(countiesByFips: Dictionary<County>): MetroArea[] {
       const [name, statesText] = metro.cbsaTitle.split(', ');
       const stateCodes = statesText.split('-');
       return new MetroArea(
-        name,
+        `${name.split('-')[0]} metro area`,
         metro.urlSegment,
         metro.cbsaCode,
         metro.population,
--- a/src/common/regions/types.ts
+++ b/src/common/regions/types.ts
@@ -110,7 +110,7 @@ export class MetroArea extends Region {
   }
 
   get fullName() {
-    return `${this.name}, ${this.stateCodes}`;
+    return `${this.name}`;
   }
 
   get shortName() {
--- a/src/components/LocationPage/LocationPageHeading.tsx
+++ b/src/components/LocationPage/LocationPageHeading.tsx
@@ -33,9 +33,7 @@ const LocationPageHeading: React.FC<{
     return (
       <Styles.Container>
         <Styles.HeaderTitle $isEmbed={isEmbed}>
-          <strong>{region.name}</strong>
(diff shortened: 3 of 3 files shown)


chasulin (commented):
Skimmed through the names and they all look good. Do we maybe want the urlSegment to match this naming convention?

src/common/regions/regions_data.ts
```diff
@@ -94,7 +94,7 @@ function buildMetroAreas(countiesByFips: Dictionary<County>): MetroArea[] {
       const [name, statesText] = metro.cbsaTitle.split(', ');
       const stateCodes = statesText.split('-');
       return new MetroArea(
-        name,
+        `${name.split('-')[0]} metro area`,
```
pnavarrc, line 97: Maybe it would be better to store the name as is, and have an accessor method that returns the name of the main city with "metro area"?

```ts
const metro = new MetroArea('New York-Newark-Jersey City', ... );

// New York Metro Area
metro.metroName;
```

Some of the mocks had the names of all the cities under the location title, I think we might need them. Can you confirm capitalization of "New York Metro Area" with Karen? I'm not sure if we want `metro area` or `Metro Area` (that can be fixed in the rendering logic as well)

src/common/regions/types.ts
```diff
@@ -110,7 +110,7 @@ export class MetroArea extends Region {
   }
 
   get fullName() {
-    return `${this.name}, ${this.stateCodes}`;
+    return `${this.name}`;
```
pnavarrc, line 113: I think the designs call for "New York-Newark-Jersey City, NY-NJ-PA" in the location title, we might need to keep this

src/components/LocationPage/LocationPageHeading.tsx
```diff
@@ -33,9 +33,7 @@ const LocationPageHeading: React.FC<{
     return (
       <Styles.Container>
         <Styles.HeaderTitle $isEmbed={isEmbed}>
-          <strong>{region.name}</strong>
-          {', '}
-          <Styles.HeaderStateCode>{region.stateCodes}</Styles.HeaderStateCode>
+          <strong>{region.fullName}</strong>
```
pnavarrc, line 36: and this 😃

pnavarrc (changes requested):
I think that an accessor method would be better to provide a shorter version of the metro area name, but we probably need something like this!

ghop02:
Closing this for now in favor of waiting for feedback from others

PR #2185, opened 2020-12-15
Remove additional push to history on click of state from state map

ghop02:
This removes most of the logic in the `goToStatePage` click handler when navigating to a state from the map.  It was causing some weirdness in the history by including an old url.  

Now that we are using links in the map, we don't need to control as much state from places that we click.  For instance, in this PR I move the logic to close the mobile map to simply change when the region changes on a location page.


mikelehen (approved):
🎉 🎉 🎉 🎉 

Thanks for untangling this!

PR #2186, opened 2020-12-15
Fix index page generation.

mikelehen:
The script was silently failing during deploys due to trying to read CBSA fips codes from county API endpoints.

> Error: INTERNAL ASSERTION FAILED: Failed to find county 10180

This is a minimal fix to only read counties and to and makes sure that the script doesn't silently fail in the future.


ghop02 (approved):
LGTM!

pnavarrc (approved):

PR #2187, opened 2020-12-15
update the alert email signup form to support MSA

pnavarrc:
- Update the `Newsletter` component to use `AutocompleteRegions`
- Changed the component to use `setState` instead of using class attributes to store state
- Moved the email validation logic to a more reusable place

### Testing

- Subscribe to alerts to one or more locations
- Check that you are actually subscribed on Firebase (dev)

### DIFF
--- a/src/components/Newsletter/Newsletter.js
+++ b/src/components/Newsletter/Newsletter.js
@@ -1,18 +1,16 @@
 import React from 'react';
-import { StyledNewsletter, InputHolder, InputError } from './Newsletter.style';
-import { getLocationNames } from 'common/locations';
-import { EventAction, EventCategory, trackEvent } from 'components/Analytics';
 import regions, { County, getStateCode } from 'common/regions';
+import { isValidEmail } from 'common/utils';
+import { EventAction, EventCategory, trackEvent } from 'components/Analytics';
 import { AutocompleteRegions } from 'components/AutocompleteLocations';
 import { getDefaultRegions, subcribeToLocations } from './utils';
-import { isValidEmail } from 'common/utils';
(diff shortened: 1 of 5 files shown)


src/components/Newsletter/Newsletter.js
```diff
@@ -1,33 +1,33 @@
...
 
 class Newsletter extends React.Component {
-  constructor() {
+  constructor(props) {
     super();
     this.form = null;
     this.emailInput = null;
-    this.alertsSelectionArray = [];
-    this.defaultValues = [];
     this.autocompleteOptions = getLocationNames();
```
ghop02, line 15: is this used anymore?

src/components/Newsletter/Newsletter.js
```diff
@@ -145,41 +117,12 @@ class Newsletter extends React.Component {
...
-                variant="outlined"
-                {...params}
-                placeholder="+ Add alert locations"
-              />
-            )}
+          <AutocompleteRegions
+            regions={regions.all()}
+            selectedRegions={selectedRegions}
+            onChangeRegions={onChangeRegions}
+            ariaLabelledBy=""
```
ghop02, line 124: should this be empty?

ghop02 (approved):
Looks good!

src/components/Newsletter/Newsletter.js
pnavarrc: good catch!

src/components/Newsletter/Newsletter.js
pnavarrc: not really 😅  - I added a default label so it should be good even if we don't have a label for the element

pnavarrc:
updated!

PR #2190, opened 2020-12-16
migrate generate-index-pages to regions

pnavarrc:
This PR migrates the `scripts/generate_index_pages.ts` to use `regions`. It also remove some of the custom logic and black lists and they are incorporated in regions already.


scripts/generate_index_pages.ts
```diff
@@ -151,37 +144,31 @@ async function main() {
...
     );
-  }
+  });
+
+  regions.counties
+    .filter(hasLocationSummary)
+    .forEach(async (county: County) => {
+      const stateCode = county.stateCode.toLowerCase();
+      const relativeSiteUrl = `/us/${stateCode}/county/${county.urlSegment}`;
+      const relativeImageUrl = `counties/${county.fipsCode}`;
```
ghop02, line 164: This might not be specific to this, but why is the relative image url for county `counties/...` but the relative image url for states is `/states/...`? I guess it's using url join so shouldn't matter

pnavarrc: It doesn't matter, we use `urlJoin` to join `relativeImageUrl` with the rest :)

ghop02 (approved):
This looks good to me!

scripts/generate_index_pages.ts
```diff
@@ -151,37 +144,31 @@ async function main() {
...
-    }
-    const county = findCountyByFips(fips);
-    assert(county, 'Failed to find county ' + fips);
-    const stateCode = county.state_code;
-    const locationName = `${county.county}, ${(STATES as any)[stateCode]}`;
-    const relativeSiteUrl = `/us/${stateCode.toLowerCase()}/county/${
-      county.county_url_name
-    }`;
-    const relativeImageUrl = `counties/${fips}`;
+  regions.states.forEach(async (state: State) => {
```
mikelehen, line 147: Hrm.  I think forEach() isn't promise-aware, which means this is going to end up doing all states in parallel, not wait for them to complete (but of course node won't end the process until they do), and silently ignore any promise failures.

So I'd just do a normal `for (const state of regions.states) {` loop...  Or if you really do want to do them in parallel, I guess:

```
await Promise.all(regions.states.map(async state => { 
 ...
}));
```

scripts/generate_index_pages.ts
```diff
@@ -151,37 +144,31 @@ async function main() {
...
       relativeImageUrl,
-      locationName,
+      state.fullName,
     );
-  }
+  });
+
+  regions.counties
+    .filter(hasLocationSummary)
+    .forEach(async (county: County) => {
```
mikelehen, line 161: same here

mikelehen (commented):
Nice! I think we need to tweak the `Promise` handling a bit, but this is much cleaner!

scripts/generate_index_pages.ts
pnavarrc: Oh that's good to know, I'm changing it back to use the `for` loop!

PR #2191, opened 2020-12-16
fix: filter out county projections if the county is not in the regions instance

pnavarrc:
The job [update-data-snapshot](https://github.com/covid-projections/covid-projections/runs/1560828963?check_suite_focus=true#step:10:24) failed because because the API returns the projections for the county with FIPS `11001` (District of Columbia), but we don't have it in `regions` as a county (we have it as a state, with FIPS code "11").

This PR filters out the county projections that don't have a county.

### DIFF
--- a/src/common/utils/model.ts
+++ b/src/common/utils/model.ts
@@ -75,6 +75,8 @@ export function fetchAllCountyProjections(snapshotUrl: string | null = null) {
           getStateName(summaryWithTimeseries.state) !== undefined,
       )
       .filter(summaryWithTimeseries => {
+        // We don't want to return county projections for counties that we
+        // don't have in the regions instance
         const region = regions.findByFipsCode(summaryWithTimeseries.fips);
         return region instanceof County;
       })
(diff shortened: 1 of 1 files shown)


src/common/utils/model.ts
```diff
@@ -74,6 +74,12 @@ export function fetchAllCountyProjections(snapshotUrl: string | null = null) {
...
       )
+      .filter(summaryWithTimeseries => {
+        // We don't want to return county projections for counties that we
+        // don't have in the regions instance
+        const region = regions.findByFipsCode(summaryWithTimeseries.fips);
+        return region instanceof County;
+      })
       .map(summaryWithTimeseries => {
         const region = regions.findByFipsCode(summaryWithTimeseries.fips);
         assert(region, 'Failed to find region ' + summaryWithTimeseries.fips);
```
ghop02, line 85: Since you're making sure we have a region, you can call `regions.findByFipsCodeStrict` above

ghop02 (approved):

mikelehen (approved):
Thank you!

PR #2192, opened 2020-12-16
Embed with regions

ghop02:
Separates out the `isEmbed` logic into it's own hook as it's a pretty different use case, and this felt more clean to separate the `useEmbed` to rely solely on the region passed into it rather than url params.

Also modified the embed so that all embed requests hit `/embed/fips/<fipsCode` instead of having separate sections for state and county.  This will let it work seamlessly with Metros.  As a result, I had to change the public Embed script to handle the old version of the embed (with state code + optionally fips) set and the new version (with just fips).

I don't have a ton of experience with the embeds so i might be missing something, but it seems to work on my tests!

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -183,7 +183,7 @@ export default function App() {
                 path="/embed/us/county/:countyFipsId"
                 component={Embed}
               />
-
+              <Route exact path="/embed/us/fips/:fipsCode" component={Embed} />
               {/* <Route path="/donate" component={ComingSoon} /> */}
               {/* /model, /contact, and /about are deprecated in favor of /faq */}
               <Route path="/model">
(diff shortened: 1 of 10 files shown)


src/App.js
```diff
@@ -183,7 +183,7 @@ export default function App() {
                 path="/embed/us/county/:countyFipsId"
                 component={Embed}
               />
-
+              <Route exact path="/embed/us/fips/:fipsCode" component={Embed} />
               {/* <Route path="/donate" component={ComingSoon} /> */}
```
chasulin, line 187: random but this `/donate` comment can go

chasulin (approved):
Looks good. I know there are a bunch of instances of `isEmbed` around the code so just double check that all of them are taken care of and nothing ended up wonky

PR #2193, opened 2020-12-16
Removes locationName from projection and cleans up usage

ghop02:
This removes `locationName` from projection and standardizes on using a region based fullName throughout the code


pnavarrc (approved):
Looks good!

PR #2201, opened 2020-12-16
Fix type issue for explore share images

ghop02:
This fixes the type issue for explore share image generation. I wasn't actually able to test locally, but properly adding types brought up the type issues that are causing the mismatch


pnavarrc (approved):
Looks good!

ghop02:
[Explore share images broken](https://trello.com/c/reS2pyXo/684-explore-share-images-broken)

PR #2205, opened 2020-12-16
First attempt at changing MSA name

ghop02:
This is the beginning of changing the MSA name to include "metro area".  

There are still some unresolved issues.  One issue is that we use `region.name` throughout the code and I'm not 100% clear what it refers to, but it'll pop up as the uglier version of the Metro name.

### DIFF
--- a/src/common/regions/types.ts
+++ b/src/common/regions/types.ts
@@ -112,11 +112,11 @@ export class MetroArea extends Region {
   }
 
   get fullName() {
-    return `${this.name}, ${this.stateCodes}`;
+    return `${this.name.split('-')[0]} metro area`;
   }
 
   get shortName() {
-    return this.name;
+    return `${this.name.split('-')[0]} metro`;
   }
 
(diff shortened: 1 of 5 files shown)


pnavarrc:
`region.name` is intended to be the name attribute passed in the constructor, maybe we should make it private? The intent for `fullName`, `shortName` and `abbreviation` was to be used as needed, hoping that we won't need more than 3 versions of a location name.

pnavarrc:
Note: I'm basing the style of the city names on the screenshots from the [original PRD](https://docs.google.com/document/d/1daAU_ubCQ7vEcWgyd3fxqqY8tFYPthNR76TSEQrmwyE/edit#) (they are not in the mocks)

image

pnavarrc:
Note: Updated the fullName to use title case based on the [APA Style Guide: Title Case Capitalization](https://apastyle.apa.org/style-grammar-guidelines/capitalization/title-case)

PR #2210, opened 2020-12-17
fix metric name on embed

pnavarrc:
The embed preview has some layout issues, but I'm not fixing those here. The metric name on the header has "per 100k" on a different element, so I'm using the function that returns the names for compare.

### DIFF
--- a/src/components/SummaryStats/SummaryStats.tsx
+++ b/src/components/SummaryStats/SummaryStats.tsx
@@ -1,6 +1,10 @@
 import React from 'react';
 import { Metric } from 'common/metric';
-import { getMetricName, getLevelInfo } from 'common/metric';
+import {
+  getMetricNameForCompare,
+  formatValue,
+  getLevelInfo,
+} from 'common/metric';
 import {
   SummaryStatsWrapper,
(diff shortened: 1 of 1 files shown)


pnavarrc:
[Add "per 100k" to incidence on embed & share preview](https://trello.com/c/IEegi9Mr/511-add-per-100k-to-incidence-on-embed-share-preview)

src/components/SummaryStats/SummaryStats.tsx
```diff
@@ -87,7 +69,7 @@ const SummaryStat = ({
         {value == null ? null : (
           <>
             <StatValueText condensed={condensed} isEmbed={isEmbed}>
-              {formatValueForChart(chartType, value)}
+              {formatValue(chartType, value, '')}
```
mikelehen, line 72: The old code used `-` as the "null value".  Also, I'm a fan of inline comments for documenting literal values passed to functions, e.g.:

```
{formatValue(chartType, value, /*nullValueCopy=*/'-')}
```

mikelehen (approved):
Thank you!! One minor optional nit, but looks good.

PR #2211, opened 2020-12-17
update the email unsubscribe form to use regions

pnavarrc:
Updates the `AlertUnsubscribe` component to work with regions (it's broken in the feature branch) to use regions instead of locations.

### Testing
- Go to `http://localhost:3000/alert_unsubscribe?email=pnavarrc@gmail.com` (the component doesn't work without the `email` param)
- You should see your alert subscriptions
- Remove some locations, unsubscribe


ghop02 (approved):
LGTM! I'd rather merge this into develop if possible! (I think this should just work right?)

pnavarrc:
> LGTM! I'd rather merge this into develop if possible! (I think this should just work right?)

It should and it makes sense, I will test against `develop` tomorrow and merge if it's all good!

pnavarrc:
Closing in favour of #2212

PR #2214, opened 2020-12-17
update Map component to use regions

pnavarrc:


src/components/Map/USACountyMap.js
```diff
@@ -91,45 +88,41 @@ const USACountyMap = React.memo(
               <Geographies geography={geoStates}>
                 {({ geographies }) =>
                   geographies
-                    .filter(geo => isStateFips(geo.id))
+                    .filter(geo => stateFipsCodes.includes(geo.id))
                     .map(geo => {
-                      const { name } = geo.properties;
                       const fipsCode = geo.id;
-                      const stateCode = getStateCode(name);
-                      const stateUrl = `/${getCanonicalUrl(fipsCode)}`;
```
ghop02, line 99: I think you can remove `getCanonicalUrl` + a couple other functions that `getCanonicalUrl` uses now! (I only say this because I have a very similar change stashed away somewhere :) )

src/components/Map/USACountyMap.js
```diff
@@ -91,45 +88,41 @@ const USACountyMap = React.memo(
...
-                            to={stateUrl}
-                            aria-label={locationName}
+                            key={state.stateCode}
+                            to={state.relativeUrl}
+                            aria-label={state.fullName}
                           >
                             <MarianaIslands
                               key={geo.rsmKey}
-                              onMouseEnter={() => setTooltipContent(name)}
+                              onMouseEnter={() => setTooltipContent(state.name)}
```
ghop02, line 106: Nit: Here's a chance where maybe using `state.fullName` makes sense? I still find `name` to be a bit ambiguous

src/components/Map/USACountyMap.js
```diff
@@ -91,45 +88,41 @@ const USACountyMap = React.memo(
...
                       // accessibility due to the small size.
-                      if (stateCode === 'MP') {
+                      if (state.stateCode === 'MP') {
                         return (
                           <Link
-                            key={stateCode}
-                            to={stateUrl}
-                            aria-label={locationName}
+                            key={state.stateCode}
+                            to={state.relativeUrl}
```
ghop02, line 101: this needs to be `/${state.relativeUrl}` I believe

src/components/Map/USACountyMap.js
```diff
@@ -91,45 +88,41 @@ const USACountyMap = React.memo(
...
                         );
                       }
-                      return stateCode ? (
+                      return state ? (
                         <Link
-                          key={stateCode}
-                          to={stateUrl}
-                          aria-label={locationName}
+                          key={state.stateCode}
+                          to={state.relativeUrl}
```
ghop02, line 117: `/${state.relativeUrl}`

ghop02 (approved):
Few small comments that we should probably address before merging, but after those are addressed, lgtm!

src/components/Map/USACountyMap.js
pnavarrc: I'm going to update this in the class on a separate PR

src/components/Map/USACountyMap.js
pnavarrc: for states is the same thing, but I prefer to use `fullName`, is more explicit

src/components/Map/USACountyMap.js
pnavarrc: will do the cleanup on a separate PR

PR #2220, opened 2020-12-17
Remove unused state+setState variables

chasulin:
`viewAllCounties` and `setViewAllCounties` controlled the state<->county switch that is no longer used (replaced by state/city/county slider)


ghop02 (approved):
Yay!

PR #2223, opened 2020-12-17
Compare on MSA pages

chasulin:


src/common/utils/compare.ts
```diff
@@ -63,6 +70,10 @@ export function getAllMetroAreas(): SummaryForCompare[] {
   return regions.metroAreas.map(getLocationObj);
 }
 
+export function getAllCountiesOfMetroArea(region: Region): SummaryForCompare[] {
```
pnavarrc, line 73: You can probably just change the type of the input to `region: MetroArea`

src/common/utils/compare.ts
```diff
@@ -345,13 +358,32 @@ export function getShareQuote(
...
-  } else if (!currentLocation && stateName) {
-    return stateShareCopy;
-  } else if (currentLocation) {
-    return countyShareCopy || stateShareCopy;
   } else {
-    return stateShareCopy;
+    if (region instanceof MetroArea) {
+      return `Compare COVID metrics between counties in ${
+        (region as MetroArea).name
+      } with @CovidActNow.`;
```
pnavarrc, line 367: nit: I don't think you need to add the type here, all the regions have `name`

pnavarrc (approved):
Looks good!

PR #2224, opened 2020-12-17
metro: update the order or results for search

pnavarrc:


src/common/regions/region_db.ts
```diff
@@ -88,7 +95,7 @@ class RegionDB {
   }
 
   all(): Region[] {
-    return [...this.states, ...this.counties, ...this.metroAreas];
+    return this.allRegions;
```
ghop02, line 98: do you think there's much advantage in having `allRegions` internally but `all()` publicly?  Seems like all could be a property

src/components/Search/utils.ts
```diff
@@ -15,47 +14,8 @@ function belongsToState(county: County, stateFips: string | null) {
...
-    return [
-      ...stateCounties,
-      ...sortedStates,
-      ...sortedMetros,
-      ...otherCounties,
-    ];
-  } else {
-    return [];
-  }
+export function getSearchAutocompleteLocations() {
```
ghop02, line 17: Won't this break the behavior where it shows specific locations for the current state? Im not sure we want to be doing that?

PR #2226, opened 2020-12-17
update api types

ghop02:


ghop02:
This was fixed in ICU utilization change

PR #2228, opened 2020-12-18
sitemap generation

pnavarrc:
Implement a script to generate the sitemap and integrate it in the build process.


pnavarrc:
[SEO: Add a site-map to surface metro pages](https://trello.com/c/1khF13CJ/696-seo-add-a-site-map-to-surface-metro-pages)

scripts/generate_sitemap/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+  const homeMapUrl = urlJoin(shareImageBaseUrl, 'home.png');
+
+  const items = [
+    {
+      url: '/',
+      img: [
+        {
+          url: homeMapUrl,
+          caption: 'Realtime US COVID Risk Map & Key Metrics',
+          title: 'Realtime US COVID Risk Map & Key Metrics',
```
mikelehen, line 73: Should we be unifying around our new homepage title ("U.S. COVID Map & Risk Levels") ?

scripts/generate_sitemap/index.ts
```diff
@@ -0,0 +1,38 @@
...
+import { concat } from 'lodash';
+import { simpleSitemapAndIndex } from 'sitemap';
+import {
+  getLearnPageItems,
+  getTopLevelPageItems,
+  getLocationPageItems,
+  getHomeImageItems,
+} from './utils';
+
+const destinationDir = path.join(__dirname, '../../public');
```
mikelehen, line 11: Are you intending that we check this file in?  Or do we generate it as part of our deploy process?  If the latter, I think it should probably go in `../../build` instead?

scripts/generate_sitemap/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+    '/terms',
+    '/privacy',
+  ];
+  return relativeUrlList.map(relativeUrl => ({
+    url: relativeUrl,
+  }));
+}
+
+export function getHomeImageItems(): SitemapItemLoose[] {
+  const homeMapUrl = urlJoin(shareImageBaseUrl, 'home.png');
```
mikelehen, line 64: Oooh nice!  Though this is the sharing image but I think we probably want to use the one I made to put in our `<meta name='image'>` tag instead: https://github.com/covid-projections/covid-projections/blob/40c34030f374e2da845a79bd4f06bd6721c53e6e/src/common/urls.ts#L48

scripts/generate_sitemap/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+  const items = [
+    {
+      url: '/',
+      img: [
+        {
+          url: homeMapUrl,
+          caption: 'Realtime US COVID Risk Map & Key Metrics',
+          title: 'Realtime US COVID Risk Map & Key Metrics',
+          geoLocation: 'United States',
+          // license
```
mikelehen, line 75: what is this? should it just be removed?

scripts/generate_sitemap/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export function getLearnPageItems(): SitemapItemLoose[] {
+  const topLevelRelativeUrls = [
+    '/learn',
+    '/glossary',
+    '/faq',
+    '/case-studies',
+    '/deep-dives',
+    '/covid-risk-levels-metrics',
+  ];
```
mikelehen, line 27: Is there a way we can avoid hard-coding these so it doesn't get out-of date?  Else, is there a way we can create "cross-referenced" comments between here and our router / whatever to help us remember to keep these in sync?  Or somehow we can have a test that catches if things get out of sync?

That said, I guess I don't know for sure what happens if we get out-of-sync.  I assume google will still crawl any URL it finds on our site, so maybe it doesn't matter too much.  In that case, maybe we don't even need to bother including everything in the sitemap?  I'm not sure...  Any idea?

mikelehen (commented):
🎉  

Couple suggestions, but looks good!

Seems like we'll want to integrate this `yarn build` so this happens automatically as part of deploys. Is that what you're thinking?

scripts/generate_sitemap/utils.ts
ghop02: This may be overkill, but looks like there's an npm package that helps with it: https://github.com/kuflash/react-router-sitemap

scripts/generate_sitemap/index.ts
pnavarrc: I think we need to generate it as part of the deploy process - it needs to be available under `https://covidactnow.org/` for it to work, so I think it needs to go in `../../build`

scripts/generate_sitemap/utils.ts
pnavarrc: I investigated `react-router-sitemap` - there is a simpler way, just export the router config for each family of pages and use it to generate the sitemap. This will help us keep the router and sitemap in sync 👌

scripts/generate_sitemap/utils.ts
pnavarrc: we can add the license, but we don't have to, removing!

mikelehen:
Let me know if/when you want a final review.

pnavarrc:
I'm going to have something cleaner tomorrow morning!

scripts/generate_sitemap/utils.ts
pnavarrc: Actually I looked into this a bit more, in order to centralize the config we will need to load all the possible URL params in the router config, which we probably don't want to do.

scripts/generate_sitemap/utils.ts
pnavarrc: The package doesn't generate exactly the sitemap markup needed by Google, I'm removing the image part for now to avoid delaying the metro areas.

scripts/generate_sitemap/index.ts
```diff
@@ -0,0 +1,31 @@
...
+function main() {
+  console.log('Generating the sitemap');
+  const topLevelPageItems = getTopLevelPageItems();
+  const locationPageItems = getLocationPageItems();
+  const learnPageItems = getLearnPageItems();
+
+  const urlList = concat(topLevelPageItems, locationPageItems, learnPageItems);
+  console.log(`Total URLs: ${urlList.length}`);
+
+  simpleSitemapAndIndex({
```
mikelehen, line 21: This returns a promise but we're ignoring it, which probably means:
1. We'll log 'Done' before the sitemap is generated.
2. If there are any errors, they won't fail the script.

I think we need to add:

```
.catch(err => {
  console.error(err);
  process.exit(-1);
}).then(() => { 
  console.log('Done.');
});
```

or else rework it to use async/await.  Note that if we use async / await we'll still need a top-level `catch` handler that calls `process.exit`, similar to https://github.com/covid-projections/covid-projections/blob/267ff9ea4155fb17f6eb78012c5b7f95c12283bd/scripts/generate_index_pages.ts#L263

src/App.js
```diff
@@ -152,12 +152,12 @@ export default function App() {
...
-                exact
-                path="/state/:stateId/county/:countyId"
-                component={LocationPage}
-              />
+              <Route exact path="/state/:stateId">
+                <Redirect to="/us/:stateId" />
+              </Route>
+              <Route exact path="/state/:stateId/county/:countyId">
+                <Redirect to="/us/:stateId/county/:countyId" />
+              </Route>
```
mikelehen, line 160: 1. This seems unrelated to sitemap, so just want to confirm this is intentional.
2. Won't this redirect to our "old" URLs?  (i.e. `/us/wa` instead of `/us/washington-wa/`)?  Is that useful?

mikelehen (approved):

src/App.js
pnavarrc: One way to tell Google which page is the canonical is to redirect to the canonical version. I'm thinking of doing this in 2 steps:

1. redirect `state/wa` → `us/wa`
2. redirect `us/wa` → `us/washington-wa` (on a separate PR)

The second redirect is a bit more tricky, because we need to write the code to find the region by state code and redirect, the first one is straightforward. I can remove this change from this PR and implement both things on a separate PR if that's better!

pnavarrc: I implemented the redirect on a follow up PR https://github.com/covid-projections/covid-projections/pull/2279

PR #2241, opened 2020-12-21
ICU Headroom => ICU Capacity

mikelehen:


src/common/metrics/hospitalizations.tsx
```diff
@@ -107,28 +119,14 @@ function renderStatus(projections: Projections): React.ReactElement {
...
   const totalICUBeds = formatInteger(icu.totalBeds);
-  const nonCovidUsedBeds = formatInteger(icu.nonCovidPatients);
-  const nonCovidUsedBedsPercent = formatPercent(
-    icu.nonCovidPatients / icu.totalBeds,
-  );
-  const remainingICUBeds = formatInteger(icu.totalBeds - icu.nonCovidPatients);
+  const nonCovidICUPatients = formatInteger(icu.nonCovidPatients);
   const covidICUPatients = formatInteger(icu.covidPatients);
+  const totalICUPatients = formatInteger(
+    icu.nonCovidPatients + icu.covidPatients,
```
ghop02, line 125: This seems fine because we explicitly break it out in the icu headroom metric - maybe in the future it would be better to just use the `actuals.icuBeds.currentUsageTotal` so we're not relying on the icu headroom code?

src/common/metrics/hospitalizations.tsx
```diff
@@ -138,54 +136,39 @@ function renderStatus(projections: Projections): React.ReactElement {
...
+    <Fragment>
+      ICU data sourced from the{' '}
+      <ExternalLink href="https://healthdata.gov/dataset/covid-19-reported-patient-impact-and-hospital-capacity-state-timeseries">
+        Department of Health and Human Services (HHS)
+      </ExternalLink>
+      . As of December 21, we use "{ICUHeadroomMetric.extendedMetricName}"
+      instead of "ICU headroom used" as our primary ICU metric. Learn more about
+      our{' '}
+      <ExternalLink href="/covid-risk-levels-metrics#icu-capacity-used">
+        ICU Capacity methodology
```
ghop02, line 168: should capacity be capitalized? my first thought is it should be a lowercase c

src/components/Header/HomePageHeader.tsx
```diff
@@ -30,7 +30,7 @@ const HomePageHeader = (props: { indicatorsLinkOnClick: () => void }) => {
               Our map shows risk levels for <strong>50 states</strong> and{' '}
               <strong>3,200 counties</strong> using{' '}
               <ClickableCopy onClick={props.indicatorsLinkOnClick || noop}>
-                5 key metrics
+                3 key metrics
```
ghop02, line 33: Do we want to have a note about this changing? I'm not sure if we explicitly have a blog post or similar explainer about why it went from 5 to 3

src/common/metrics/hospitalizations.tsx
```diff
@@ -138,54 +136,39 @@ function renderStatus(projections: Projections): React.ReactElement {
...
-          our data sources
-        </ExternalLink>
-        .
-      </Fragment>
-    );
-  }
+  return (
+    <Fragment>
+      ICU data sourced from the{' '}
+      <ExternalLink href="https://healthdata.gov/dataset/covid-19-reported-patient-impact-and-hospital-capacity-state-timeseries">
```
ghop02, line 161: Just a note that if we have any county data, this will be a different dataset

ghop02 (approved):
LGTM! Couple small comments but nothing blocking

src/common/metrics/hospitalizations.tsx
mikelehen: Yep, will do in a subsequent PR!

src/common/metrics/hospitalizations.tsx
mikelehen: It was capitalized before, but I think you're right.  Fixed!

src/components/Header/HomePageHeader.tsx
mikelehen: I have mentioned this to Debbie.

PR #2246, opened 2020-12-21
Use h2 for location page sections

chasulin:
[Trello item](https://trello.com/c/b3HkNl1i/491-seo-update-the-location-page-to-use-the-correct-markup)


pnavarrc (approved):
👍

PR #2248, opened 2020-12-21
add metros to explore location search

pnavarrc:
- Centralized the logic with the global location search

### DIFF
--- a/src/common/regions/utils.ts
+++ b/src/common/regions/utils.ts
@@ -7,6 +7,14 @@ export function belongsToState(county: County, stateFips: string | null) {
   return county.state.fipsCode === stateFips;
 }
 
+/**
+ * Returns a list of regions in the order that is most relevant given the
+ * region passed as argument
+ *
+ * Homepage (region is undefined): states, metros, counties
+ * Metro: counties in metro, states, other counties, metros
+ * State and County: counties in the state, states, metros, other counties
+ */
 export function getAutocompleteRegions(region?: Region) {
   const { states, metroAreas, counties } = regions;
 
--- a/src/common/regions/utils.ts
+++ b/src/common/regions/utils.ts
@@ -7,7 +7,7 @@ export function belongsToState(county: County, stateFips: string | null) {
   return county.state.fipsCode === stateFips;
 }
 
-export function getAutocompleteLocations(region?: Region) {
+export function getAutocompleteRegions(region?: Region) {
   const { states, metroAreas, counties } = regions;
 
   // Homepage
--- a/src/common/regions/utils.ts
+++ b/src/common/regions/utils.ts
@@ -1,0 +1,33 @@
+import { concat, partition } from 'lodash';
+import regions from './region_db';
+import { getStateFips } from './regions_data';
+import { County, State, Region, MetroArea } from './types';
+
+export function belongsToState(county: County, stateFips: string | null) {
+  return county.state.fipsCode === stateFips;
+}
+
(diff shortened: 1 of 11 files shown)


src/components/Explore/utils.ts
```diff
@@ -429,42 +420,9 @@ export function getSubtitle(
...
-        ...sortedStateCounties,
-        ...sortedOtherCounties,
-      ]
-    : [
-        ...regions.customAreas,
-        ...sortedStateCounties,
-        ...sortedStates,
-        ...sortedOtherCounties,
-      ];
+  return currentLocation ? getSearchAutocompleteLocations(currentLocation) : [];
```
ghop02, line 425: This feels a bit like it's tying two different areas together - thought on making a generic `getAutoCompleteLocations(region?: Region)` that both use?

pnavarrc: Yeah, I can move the function to a more central place, as the logic is the same for both

src/common/regions/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+export function belongsToState(county: County, stateFips: string | null) {
+  return county.state.fipsCode === stateFips;
+}
+
+export function getAutocompleteLocations(region?: Region) {
+  const { states, metroAreas, counties } = regions;
+
+  // Homepage
+  if (!region) {
+    return concat<Region>(states, metroAreas, counties);
```
ghop02, line 15: we want to include USA and NAMC here right?

src/common/regions/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+    const [countiesInMetro, otherCounties] = partition(counties, county =>
+      region.counties.includes(county),
+    );
+    return concat<Region>(countiesInMetro, states, otherCounties, metroAreas);
+  } else if (region instanceof State || region instanceof County) {
+    const stateFips = getStateFips(region);
+    const [stateCounties, otherCounties] = partition(counties, county =>
+      belongsToState(county, stateFips),
+    );
+    return concat<Region>(stateCounties, states, metroAreas, otherCounties);
```
ghop02, line 29: Happy to fix this up in a separate PR, but I would love to partition the metros to pull out metros that have counties in the state

src/common/regions/utils.ts
pnavarrc: No, they don't have location pages - I'm adding them back on the autocomplete for explore

src/common/regions/utils.ts
```diff
@@ -0,0 +1,33 @@
+import { concat, partition } from 'lodash';
+import regions from './region_db';
+import { getStateFips } from './regions_data';
+import { County, State, Region, MetroArea } from './types';
+
+export function belongsToState(county: County, stateFips: string | null) {
+  return county.state.fipsCode === stateFips;
+}
+
+export function getAutocompleteLocations(region?: Region) {
```
ghop02, line 10: nit: Would prefer this to be `getAutocompleteRegions` as we have the `Location` and `Region` object. Also, will type inference work properly here? 

Also, maybe a name like `getAllAutocompleteRegions` or `getAllAutocompleteRegionsSorted` might be more clear? Though I don't love those names usggestions.  I think a comment here might be useful to indicate what the `region` argument does.  it's not clear from the function signature that region changes the sort order of the list.

ghop02 (approved):
whoops meant to approve last time

chasulin (approved):
Pending the comment re: updating `getFilterLimit` in slack, looks good

PR #2249, opened 2020-12-21
Compare sharing fix pt. 1

chasulin:
Share links from compare on MSA location pages were redirecting to the homepage because `getComparePageUrl` didn't use regions/didn't account for metros. This fixes that

### DIFF
--- a/src/common/urls.ts
+++ b/src/common/urls.ts
@@ -1,6 +1,5 @@
 import ShareImageUrlJSON from 'assets/data/share_images_url.json';
 import { assert } from 'common/utils';
-import { County } from './locations';
 import urlJoin from 'url-join';
 import * as QueryString from 'query-string';
 import moment from 'moment';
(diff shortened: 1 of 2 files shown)


src/common/urls.ts
```diff
@@ -80,11 +64,13 @@ export function getComparePageUrl(
...
   let params: { [key: string]: unknown } = {};
   ensureSharingIdInQueryParams(params);
-  params['redirectTo'] = urlJoin(
-    getPagePath(stateId, county),
-    'compare',
-    compareShareId,
-  );
+  if (region) {
+    params['redirectTo'] = urlJoin(
+      region.relativeUrl,
```
ghop02, line 69: do you need the initial `/`?

ghop02 (approved):
One tiny comment, but otherwise love to see fewer `County` instances!

src/common/urls.ts
chasulin: doesnt seem like it

PR #2250, opened 2020-12-21
show exposure notifications for metro areas in states where enabled

pnavarrc:
- It add states to the `MetroArea` class
- Enables exposure notification alert if the metro area belongs in a state where that's enabled

### DIFF
--- a/src/common/regions/regions.test.ts
+++ b/src/common/regions/regions.test.ts
@@ -8,7 +8,7 @@ describe('regions', () => {
 
   test('counties() returns all the counties', () => {
     const counties = regions.counties;
-    expect(counties).toHaveLength(3224);
+    expect(counties).toHaveLength(3223);
   });
 
   describe('findByFipsCode', () => {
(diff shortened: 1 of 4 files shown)


src/common/regions/regions.test.ts
```diff
@@ -8,7 +8,7 @@ describe('regions', () => {
 
   test('counties() returns all the counties', () => {
     const counties = regions.counties;
-    expect(counties).toHaveLength(3224);
+    expect(counties).toHaveLength(3223);
```
pnavarrc, line 11: we removed DC

src/common/regions/regions.test.ts
```diff
@@ -134,11 +134,11 @@ describe('County', () => {
 describe('Metro Area', () => {
   const metro = regions.findByFipsCode('35620');
   test('fullName', () => {
-    expect(metro?.fullName).toBe('New York-Newark-Jersey City, NY-NJ-PA');
+    expect(metro?.fullName).toBe('New York metro area');
   });
 
   test('shortName', () => {
-    expect(metro?.shortName).toBe('New York-Newark-Jersey City');
+    expect(metro?.shortName).toBe('New York metro');
```
pnavarrc, line 141: These tests are not running on CI, I will fix that after we launch

src/common/regions/types.ts
```diff
@@ -100,15 +100,18 @@ export class County extends Region {
  * Metropolitan Statistical Areas
  */
 export class MetroArea extends Region {
+  private stateCodeList: string[];
```
ghop02, line 103: Do we need this? I'd rather just change `stateCodes` to use `states`

ghop02 (approved):

PR #2251, opened 2020-12-21
Compare share images fix

chasulin:


src/components/Compare/CompareMain.tsx
```diff
@@ -163,9 +164,9 @@ const CompareMain = (props: {
     }
   }
 
-  const locations = getFinalLocations(props.region);
+  const locations = getFinalLocations(region);
 
-  const viewMoreCopy = !props.region
+  const viewMoreCopy = !region
```
pnavarrc, line 169: Maybe you can call `getLocationPageViewMoreCopy` here instead so we have fewer conditionals on `region`?

```ts
const viewMoreCopy = region
  ? getLocationPageViewMoreCopy(/* ... */)
  : homepageViewMoreCopy;

```

pnavarrc (approved):
Works great!

PR #2252, opened 2020-12-21
Add share image urls + index page generation

ghop02:
This adds metros to the share image urls and generates metro level index files.


pnavarrc (approved):
Looks good!

PR #2271, opened 2020-12-22
Remove location and global selector code

ghop02:
Removes a bunch of location/global selector code.  There are a few lingering county instances that we need to clean up before we can remove the rest of location code


pnavarrc (approved):
Looks good! 🧹

PR #2272, opened 2020-12-22
Compare edit

chasulin:
Adds check for `locations`

Quick fix for [this trello item](https://trello.com/c/ZbTTsIiD/710-investigate-compare-sentry-error)


ghop02 (approved):
Seems alright as a guard - can you add a return type to `getFinalLocations` that would've caught this?

PR #2274, opened 2020-12-22
Removing county object from CompareMain

ghop02:
Just a bit of compare table cleanup to start to remove the `County` from `locations` from the compare table, relying more on the region object


src/common/utils/compare.ts
```diff
@@ -106,25 +101,16 @@ export function getStateNonMetroCounties(
...
-    .map(getLocationObj);
-  const currentCounty = getCountyObj(countyFips);
-  return [...adjacentCounties, ...currentCounty];
+  const county = regions.findByFipsCodeStrict(countyFips) as County;
+  const adjacentCounties = county.adjacentCountyFips.map(fips => {
+    const region = regions.findByFipsCodeStrict(fips);
+    return getLocationObj(region);
+  });
+
+  return [...adjacentCounties, ...[getLocationObj(county)]];
```
pnavarrc, line 113: It might be simpler to do:

```ts
return [...adjacentCounties, getLocationObj(county)];
```

(no need to expand an new array with a single element)

src/common/utils/compare.ts
```diff
@@ -106,25 +101,16 @@ export function getStateNonMetroCounties(
...
-  const adjacentCounties = regions.counties
-    .filter((region: Region) => isNeighboringCounty(region, countyFips))
-    .map(getLocationObj);
-  const currentCounty = getCountyObj(countyFips);
-  return [...adjacentCounties, ...currentCounty];
+  const county = regions.findByFipsCodeStrict(countyFips) as County;
+  const adjacentCounties = county.adjacentCountyFips.map(fips => {
+    const region = regions.findByFipsCodeStrict(fips);
+    return getLocationObj(region);
+  });
```
pnavarrc, line 111: We can do this later, but here is an idea. The `County` objects have a `state` property with a `State` object, we could do the same for `adjacentCounties` to avoid having to look up by FIPS each time.

pnavarrc (approved):
Looks good!

PR #2278, opened 2020-12-22
National summary paragraph

chasulin:
- [x] UX signoff
- [x] QA numbers
- [x] Fix `any` and the many uses of `!`


src/components/NationalText/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+  const lastTwoWeeks = series.slice(-14);
+  const difference = last(lastTwoWeeks)! - lastTwoWeeks[0]!;
+  const percentChange = difference / lastTwoWeeks[0]!;
+  return percentChange;
+};
+
+const getDescriptiorCopy = (percentChange: number): string => {
+  if (percentChange > 0) {
+    return 'increased ';
+  } else return 'decreased ';
```
pnavarrc, line 18: Consider handling the case of no change too, here it would be "decreased"

src/components/NationalText/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+import { formatPercent, formatDecimal } from 'common/utils';
+
+// Fix all uses of !
+
+const getPercentChange = (series: any[]): number => {
+  const lastTwoWeeks = series.slice(-14);
+  const difference = last(lastTwoWeeks)! - lastTwoWeeks[0]!;
+  const percentChange = difference / lastTwoWeeks[0]!;
+  return percentChange;
+};
```
pnavarrc, line 13: I would recommend implementing a `percentChange` function that just takes 2 numbers so we can use it in other places

```ts
function percentChange(a: number, b: number): number | null {
  const validParams = isFinite(a) && isFinite(b) && b !== 0;
  return validParams ? a / b : null;
}
```

src/components/NationalText/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const getDescriptiorCopy = (percentChange: number): string => {
+  if (percentChange > 0) {
+    return 'increased ';
+  } else return 'decreased ';
+};
+
+const getTotalCasesCopy = (summedRawCases: number): string => {
+  const divided = summedRawCases / 1000000;
```
pnavarrc, line 22: Consider defining a constant for `1000000` to make this easier to read, or use `1_000_000` (which is another way to write 1000000, I found out about this recently!)

```ts
const ONE_MILLION = 1000000;

// this also works:
const divided = summedRawCases / 1_000_000;
```

src/components/NationalText/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+    rawDailyDeaths,
+    smoothedDailyCases,
+    smoothedDailyDeaths,
+    dates,
+  } = usaAggregation;
+
+  const totalCases = rawDailyCases.reduce(
+    (acc: number, curr: number) => acc + curr,
+    0,
+  );
```
pnavarrc, line 39: Consider using `sum` from `lodash`

```ts
const totalCases = sum(rawDailyCases);
```

src/components/NationalText/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+    return 'increased ';
+  } else return 'decreased ';
+};
+
+const getTotalCasesCopy = (summedRawCases: number): string => {
+  const divided = summedRawCases / 1000000;
+  return `${formatDecimal(divided, 1)} million`;
+};
+
+export function getNationalText(): string {
```
pnavarrc, line 26: We might want to return a React element here in case we want to highlight the numbers, add links and the like. I don't think that's being required here, but it has happened a couple of times that the requirement comes in later :)

src/components/NationalText/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+import moment from 'moment';
+import { formatPercent, formatDecimal } from 'common/utils';
+
+// Fix all uses of !
+
+const getPercentChange = (series: any[]): number => {
+  const lastTwoWeeks = series.slice(-14);
+  const difference = last(lastTwoWeeks)! - lastTwoWeeks[0]!;
+  const percentChange = difference / lastTwoWeeks[0]!;
+  return percentChange;
```
pnavarrc, line 12: Consider finding the point that matches the date instead of relying on the series having all the points and being sorted by date, this will help us in case there are holes in the data

```ts
const currPoint = last(sortedSeries);
const prevDate = moment(currPoint.x).subtract(14, 'days').toDate();
const prevPoint = sortedSeries.find(p => isSameDay(p.x, prevDate));

return prevPoint && currPoint ? precentChange(prevPoint.y, currPoint.y) : null;

// ...

// maybe it would be better to compare `moment(date).format('YYYY-MM-DD')`
function isSameDay(dateA: Date, dateB: Date) {
  return moment(dateA).startOf('day') === moment(dateB).startOf('day');
}

```

pnavarrc (commented):
Added some comments, looks generally good!

src/components/NationalText/utils.tsx
```diff
@@ -0,0 +1,80 @@
...
+  return (
+    <Fragment>
+      As of {lastDateFormatted}, there have been roughly{' '}
+      {getTotalCasesCopy(totalCases)} cases and{' '}
+      {formatEstimate(totalDeaths, 3).toLocaleString()} deaths from COVID in the
+      United States.{' '}
+      {renderPercentChangeSentence
+        ? `Over the last 14 days, daily new
+    cases have ${getChangeDescriptorCopy(
+      percentChangeSmoothedCases!,
```
chasulin, line 74: even with the `renderPercentChangeSentence` check, this still yells at me when theres not a `!`. same with `percentChangeSmoothedDeaths!` below

src/common/utils/index.ts
```diff
@@ -119,3 +119,15 @@ export function isValidEmail(emailAddress: string) {
     emailAddress && emailAddress.length > 1 && EMAIL_REGEX.test(emailAddress)
   );
 }
+
+// Returns unformatted percent in decimal form
+export function getPercentChange(numA: number, numB: number): number | null {
+  const validParams = _.isNumber(numA) && _.isNumber(numB) && numA !== 0;
```
pnavarrc, line 125: you need `_.isFinite`, because `_.isNumber(NaN)` is `true` (also `_.isNumber(Infinity)` and `_.isNumber(-Infinity)`)

src/components/NationalText/utils.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const percentChangeSmoothedCases = getTwoWeekPercentChange(
+    smoothedDailyCases,
+  );
+  const percentChangeSmoothedDeaths = getTwoWeekPercentChange(
+    smoothedDailyDeaths,
+  );
+
+  const renderPercentChangeSentence =
+    isNumber(percentChangeSmoothedCases) &&
+    isNumber(percentChangeSmoothedDeaths);
```
pnavarrc, line 51: Also here, `isFinite` is better

src/components/NationalText/utils.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const lastDate = last(dates);
+  const lastDateFormatted: string = moment(lastDate).format('MMMM Do, YYYY');
+
+  const getChangeDescriptorCopy = (percentChange: number): string => {
+    const changeByCopy = `by about ${formatPercent(Math.abs(percentChange))}`;
+    if (percentChange > 0.01) {
+      return `increased ${changeByCopy}`;
+    } else if (percentChange < 0) {
+      return `decreased ${changeByCopy}`;
+    } else return 'remained about consistent each day';
```
pnavarrc, line 62: The copy is probably not correct, since the change is comparing a date and 2 weeks before that date. The metric could have change during that period, but it just so happens that the first and last day of the interval have the same value.

src/components/NationalText/utils.tsx
pnavarrc: That's ok, we can keep the `!` 🤷

pnavarrc (approved):
Looks good, let me know if you need another pass!

src/components/NationalText/utils.tsx
chasulin: ah yes good point

src/common/utils/index.ts
chasulin: i kind of feel like numA as the denominator (and therefore representing the initial # ) makes more sense? gonna keep as is for now

src/components/NationalText/utils.tsx
```diff
@@ -0,0 +1,80 @@
...
+import moment from 'moment';
+import {
+  formatPercent,
+  formatDecimal,
+  getPercentChange,
+  formatEstimate,
+} from 'common/utils';
+
+const getTwoWeekPercentChange = (series: any[]): number | null => {
+  const lastTwoWeeks = series.slice(-14);
```
mikelehen, line 13: I think you need `.slice(-15)` to get the two week delta since to get a 1-day delta you'd actually have to do `.slice(-2)` [which gives you the last 2 values (yesterday and today)].

src/components/NationalText/utils.tsx
```diff
@@ -0,0 +1,80 @@
...
+    isFinite(percentChangeSmoothedDeaths);
+
+  const lastDate = last(dates);
+  const lastDateFormatted: string = moment(lastDate).format('MMMM Do, YYYY');
+
+  const getChangeDescriptorCopy = (percentChange: number): string => {
+    const changeByCopy = `by about ${formatPercent(Math.abs(percentChange))}`;
+    if (percentChange > 0.01) {
+      return `increased ${changeByCopy}`;
+    } else if (percentChange < 0) {
```
mikelehen, line 60: nit: do we want `< 0.01` so that we don't get "decreased 0%" text?

src/components/NationalText/utils.tsx
mikelehen: FWIW- TypeScript has no way to know that if the lodash `isFinite()` function returns true then the value you passed wasn't null.  If you wanted to get rid of the `!` here, you could do explicit null checks:

```
  const renderPercentChangeSentence =
    percentChangeSmoothedCases !== null &&
    percentChangeSmoothedDeaths !== null;
```

And then TypeScript wouldn't complain that it might be null.  

Feel free to leave as-is though.

mikelehen (approved):
I think the 14-day % change has an off-by-one error.  While I was in here, I also left a couple nits, but you can ignore.

src/components/NationalText/utils.tsx
chasulin: if the change were 0%, it would go into the else ("remained about the same"), no?

mikelehen: Oops, sorry. I meant `< -0.01`.  Right now, if the percentChange is e.g. `-0.004` then it'll round to `0%` and it'll say "decreased 0%", right?

PR #2287, opened 2020-12-24
Improve USA cumulative cases / deaths numbers.

mikelehen:
For the national summary text, rather than sum rawDailyCases / rawDailyDeaths which may be missing early data (we start at 2020-03-01) or have outliers removed (maybe? not 100% sure), sum the cumulative cases/deaths from the API.

This bumps today's aggregated cases from 18.1M to 18.2M which is still shy of the 18.4M that NYT [reports](https://www.nytimes.com/interactive/2020/us/coronavirus-us-cases.html) for today, but it's a little closer. 🤷


pnavarrc (approved):
Looks good!

PR #2322, opened 2021-01-04
SEO: structured data for FAQ

chasulin:
[Trello](https://trello.com/c/n43p1uU9/661-seo-implement-structured-data-on-our-homepage-to-help-google-see-the-map)

Implementing [FAQpage structured data type](https://developers.google.com/search/docs/data-types/faqpage) using [react-structured-data](https://www.npmjs.com/package/react-structured-data)

### DIFF
--- a/src/screens/Learn/Faq/Faq.tsx
+++ b/src/screens/Learn/Faq/Faq.tsx
@@ -15,6 +15,7 @@ import Section from './Section';
 import { useScrollToTopButton } from 'common/hooks';
 import ScrollToTopButton from 'components/SharedComponents/ScrollToTopButton';
 import { EventCategory } from 'components/Analytics';
+import FaqStructuredData from './FaqStructuredData';
 
 const Faq: React.FC = () => {
   const {
@@ -47,6 +48,7 @@ const Faq: React.FC = () => {
         pageTitle={metadataTitle}
         pageDescription={`${date} ${metadataDescription}`}
       />
+      <FaqStructuredData />
       <PageContent sidebarItems={learnPages}>
         <BreadcrumbsContainer>
           <Breadcrumbs item={{ to: '/learn', label: 'Learn' }} />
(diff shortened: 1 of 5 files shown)


src/screens/Learn/Faq/FaqStructuredData.tsx
```diff
@@ -0,0 +1,36 @@
...
+  FaqSection,
+  Question as QuestionItem,
+} from 'cms-content/learn';
+
+const FaqStructuredData: React.FC = () => {
+  const { sections } = faqContent;
+  return (
+    <JSONLD>
+      <GenericCollection type="FAQPage">
+        {sections.forEach((section: FaqSection) => {
```
pnavarrc, line 20: We need to use `map` instead, because [`forEach`](http://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) always returns `undefined` (it just iterates over the elements of the array). That's why the `script` element is almost empty in the DOM.

image

src/screens/Learn/Faq/Faq.tsx
```diff
@@ -47,6 +48,7 @@ const Faq: React.FC = () => {
         pageTitle={metadataTitle}
         pageDescription={`${date} ${metadataDescription}`}
       />
+      <FaqStructuredData />
```
pnavarrc, line 51: I think we need to use `Helmet` here (see [AppMetaTags.js#L24](https://github.com/covid-projections/covid-projections/blob/develop/src/components/AppMetaTags/AppMetaTags.js#L24)) to make sure that the script is in the `head` element in the page instead of inside the body.

pnavarrc (changes requested):
Added some notes, we need to move the script to the header and use `map` instead of `forEach` for it to work!

chasulin:
Closing this in favor of #2327

PR #2323, opened 2021-01-04
fix redirect for deprecated location routes

pnavarrc:
The redirect for `/state/:stateId` and `/state/:stateId/county/:countyId` was not implemented correctly (apparently, the children component of a route doesn't have access to the URL parameters). 

This PR fixes that by using `from` and `to` in the `Redirect` component, and updates the syntax other instances of nested redirects for consistency.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -157,12 +157,12 @@ export default function App() {
               />
 
               {/* /state/ routes are deprecated but still supported. */}
-              <Route exact path="/state/:stateId">
-                <Redirect to="/us/:stateId" />
-              </Route>
-              <Route exact path="/state/:stateId/county/:countyId">
-                <Redirect to="/us/:stateId/county/:countyId" />
-              </Route>
+              <Redirect exact from="/state/:stateId" to="/us/:stateId" />
+              <Redirect
+                exact
+                from="/state/:stateId/county/:countyId"
+                to="/us/:stateId/county/:countyId"
+              />
 
               <Route path="/about" component={About} />
               <Route path="/tools" component={Tools} />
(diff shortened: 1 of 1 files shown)


src/App.js
```diff
@@ -157,12 +157,12 @@ export default function App() {
...
-                <Redirect to="/us/:stateId" />
-              </Route>
-              <Route exact path="/state/:stateId/county/:countyId">
-                <Redirect to="/us/:stateId/county/:countyId" />
-              </Route>
+              <Redirect exact from="/state/:stateId" to="/us/:stateId" />
+              <Redirect
+                exact
+                from="/state/:stateId/county/:countyId"
+                to="/us/:stateId/county/:countyId"
```
ghop02, line 164: How do we handle going from the state code (say "/mo" -> "/us/missouri-mo")?

pnavarrc: That is done in [LocationRouter.tsx#L17](https://github.com/covid-projections/covid-projections/blob/c80e6f6d62448bac9e670745fe80bb1fd20cc3e2/src/screens/LocationPage/LocationRouter.tsx#L17)

ghop02: makes sense! Is there a world in which we could update the stateId here? or is it a problem that it's doing two redirects as opposed to consolidating into one?

ghop02 (approved):
Looks fine to me! it would be cool if we could update stateId here, but understand that might not be possible :)

pnavarrc:
[Follow-up: Fix broken share images for direct location based sharing](https://trello.com/c/At4Ak5Gq/478-follow-up-fix-broken-share-images-for-direct-location-based-sharing)

pnavarrc:
[Follow-up: Fix broken share images for direct location based sharing](https://trello.com/c/At4Ak5Gq/478-follow-up-fix-broken-share-images-for-direct-location-based-sharing)

PR #2326, opened 2021-01-04
update generate-index-page

pnavarrc:
The `generate-index-pages` script was using the old URL structure for sharing, even when the sharing component was using new URLs (`https://covidactnow.org/us/washington-wa`).

I also discovered that the `index.html` pages for sharing have the wrong canonical URL and image URL (for example, see `og:url` in https://covidactnow.org/us/alabama-al/county/baldwin_county/chart/0?s=1485441)

image

I fixed the canonical URL, but might need help with the `image` tag (I think that the regular expression is ignoring meta tags with ` data-react-helmet="true"`).

### DIFF
--- a/scripts/generate_index_pages.ts
+++ b/scripts/generate_index_pages.ts
@@ -145,15 +145,10 @@ async function main() {
   }
 
   for (const state of regions.states) {
-    const stateCode = state.stateCode.toLowerCase();
-    // TODO(chris): Change this url to be the updated url (use `state.relativeUrl`). Currently
-    // any location page link shared from the new URL structure will try to access an index page at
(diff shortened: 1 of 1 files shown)


mikelehen (changes requested):

pnavarrc:
It does work for me, can you double check? I'm looking at `build/us/alabama-al/index.html` and `build/us/alabama-al/charts/0/index.html`, both have the correct canonical

```html
<link data-react-helmet="true" rel="canonical" href="https://covidactnow.org/us/alabama-al/index.html" />
```

and 

```html
<link data-react-helmet="true" rel="canonical" href="https://covidactnow.org/us/alabama-al/chart/0/index.html" />
```

PR #2341, opened 2021-01-05
Update alert/newsleter signup to not store alert subscribers in Campaign Monitor.

mikelehen:
For historical reasons, when somebody signed up for alerts or the newsletter, we were using a Campaign Monitor signup form that would:
1. Always add the user to the "Alert Signups" list in Campaign Monitor.
2. Optionally (depending on whether they checked the newsletter checkbox) add them to the "Full Sign Up List_Scrubbed" list in Campaign Monitor.

But we don't actually use Campaign Monitor at all for alert emails anymore, so we don't need to sign them up to the first list, and it was costing us extra money.

So I've regenerated the form so that it only signs them up for "Full Sign Up List_Scrubbed" and we only actually submit the form if they check the newsletter checkbox.

Since we only *sometimes* submit the Campaign Monitor signup form now, I had to invent a new subscription confirmation UI if they sign up for alerts but not the newsletter, so I added a little UI to change the "Sign up" button to say "Subscribed!" for 3 seconds after they sign up for alerts (but not the newsletter).

I've manually verified that checking the box and clicking "sign up" correctly results in the email getting added to "Full Sign Up List_Scrubbed" in Campaign Monitor as desired.

### DIFF
--- a/src/components/Newsletter/Newsletter.js
+++ b/src/components/Newsletter/Newsletter.js
@@ -4,7 +4,12 @@ import { isValidEmail } from 'common/utils';
 import { EventAction, EventCategory, trackEvent } from 'components/Analytics';
 import AutocompleteRegions from 'components/AutocompleteRegions';
 import { getDefaultRegions, subscribeToLocations } from './utils';
-import { StyledNewsletter, InputHolder, InputError } from './Newsletter.style';
+import {
+  StyledNewsletter,
+  SubscribeButton,
+  InputHolder,
+  InputError,
+} from './Newsletter.style';
 
 class Newsletter extends React.Component {
   constructor(props) {
@@ -12,10 +17,11 @@ class Newsletter extends React.Component {
     this.form = null;
     this.emailInput = null;
     this.state = {
-      checked: true,
+      dailyDownloadChecked: true,
       email: '',
       errorMessage: '',
       selectedRegions: props.region ? getDefaultRegions(props.region) : [],
+      showConfirmationText: false,
     };
     this.submitForm = this.submitForm.bind(this);
   }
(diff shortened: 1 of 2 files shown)


src/components/Newsletter/Newsletter.js
```diff
@@ -4,18 +4,24 @@ import { isValidEmail } from 'common/utils';
...
+} from './Newsletter.style';
 
 class Newsletter extends React.Component {
   constructor(props) {
     super();
     this.form = null;
     this.emailInput = null;
     this.state = {
-      checked: true,
+      dailyDownloadChecked: true,
```
mikelehen, line 20: I just renamed this for readability.

src/components/Newsletter/Newsletter.js
```diff
@@ -83,36 +101,36 @@ class Newsletter extends React.Component {
...
-        We hide some of the information users don't need to enter and add some form fields that the api doesn't
-        require (i.e the alert-loctions autocomplete).
          */}
         <form
           ref={f => (this.form = f)}
           className="js-cm-form"
           id="subForm"
           method="post"
-          data-id="2BE4EF332AA2E32596E38B640E905619E90CD5DAC48A878CDEBFFE3B420D8CD24E4AEABAB52A4CE3526219C7A966AE965B84F99C823C89EFF1F01B28DE4F975E"
+          data-id="2BE4EF332AA2E32596E38B640E90561930C9C3A433D015D9D4BD88E99175E51395EF5EBFFD527179E032AC15455BB1208D87A6CE87843E524B0EA520CBFF446E"
```
mikelehen, line 116: All of these changed IDs just come from the campaign monitor form builder. 🤷

src/components/Newsletter/Newsletter.style.tsx
```diff
@@ -58,33 +58,35 @@ export const StyledNewsletter = styled.div`
...
-      padding: 0.25rem 1.25rem;
-      line-height: 1rem;
-      text-transform: uppercase;
-      transition: 0.3s ease background-color;
-      background-color: ${palette.info.main};
-      border: 1px solid ${palette.info.dark};
-      color: ${palette.white};
-      font-weight: 700;
-      font-family: 'Roboto', 'Helvetica', 'Arial', sans-serif;
+export const SubscribeButton = styled.button<{ confirmation: boolean }>`
```
mikelehen, line 64: I made this its own styled component so I could pass the `confirmation` prop to it.  I assume that's the right thing to do?

pnavarrc: yes, this should do it!

pnavarrc (approved):
Looks good!

PR #2345, opened 2021-01-05
Fixes autocomplete bug

chasulin:
Addresses bug mentioned in [this trello card](https://trello.com/c/PhgcJBxK/723-search-priority-inconsistent-across-pages-align-on-homepage-search-order):
"For example, another question i have is: should searching metro pull up the metros? Also, right now if you search "new york metro area" it actually doesn't show the result (seems like a bug)."

Search was matching against the wrong location name variation. This fixes (now matches against the same name variation used throughout the dropdown menu).


ghop02 (approved):
Nice!

PR #2349, opened 2021-01-06
refactor the minimap to not use state topojson

pnavarrc:
One of the big chunks of the main bundle are the TopoJSON files that we have for each state. The Map for metro areas uses the list of counties in the metro area, but we can use the same logic to show any set of counties, including counties in a state, which allows us to remove the TopoJSON files for each state.

- Update the MetroAreaMap component to work for states, counties and metros
- Remove the CountyMap component and all the state TopoJSON files

**State Page**
image

**Metro Area**
image

**County Page**
image

The MiniMap has now a similar design for state, county and metro areas. I adjusted the scale up a bit to make the state and county maps a bit bigger.

**Before**
```
  1.16 MB (+301.04 KB)  build/static/js/main.4618c95c.chunk.js
  410.89 KB (-288 B)    build/static/js/2.9ef820c9.chunk.js
  778 B                 build/static/js/runtime-main.7e3eb5a3.js
  750 B (+1 B)          build/static/css/main.9383c564.chunk.css
```

**After**
```
  885.58 KB (-300.85 KB)  build/static/js/main.e3db1757.chunk.js
  411.17 KB (+288 B)      build/static/js/2.29ec59fe.chunk.js
  778 B                   build/static/js/runtime-main.7e3eb5a3.js
  749 B (-1 B)            build/static/css/main.7d34e3e9.chunk.css
```


pnavarrc:
[Performance - Remove state by state TopoJSON files](https://trello.com/c/bMV9dqY8/739-performance-remove-state-by-state-topojson-files)

mikelehen (approved):
Looks good other than refactoring to combine MetroAreaMap / RegionMap.

FWIW:
1. I notice that this makes the state map a tiny bit laggy as I drag my mouse cursor around.  I'm not sure why.
2. The nice thing about separate topo files per state is that in theory we could optimize the location page to only load the topo file(s) needed by that location page.  Which means e.g. we'd load 1/50 as much data for a typical state.  But we weren't doing that before, and we don't have concrete plans to, so this seems fine for now.  But I could imagine we end up revisiting eventually if we get super serious about minimizing bundle sizes, etc.

pnavarrc:
I fixed the lagging, we were doing extra calculation on each tooltip update. It would be super nice to generate separated TopoJSON files for each case, we can do that if we need it later! The MiniMap is also a candidate for lazy loading, since we don't need it in all the pages and we only need it when it becomes visible.

pnavarrc:
[Performance - Remove state by state TopoJSON files](https://trello.com/c/bMV9dqY8/739-performance-remove-state-by-state-topojson-files)

PR #2356, opened 2021-01-06
Appends 'city' to New York metro principal city

chasulin:
[Trello item](https://trello.com/c/pPCFELxY/725-seo-some-metros-could-be-more-easily-discoverable-with-multiple-names)


ghop02 (approved):
LGTM!

PR #2357, opened 2021-01-06
New york metro URL segment update

chasulin:
Updates NY metro area's URL segment to include 'city'

[prod ny metro url](https://covidactnow.org/us/metro/new-york-newark-jersey-city_ny-nj-pa/?s=1490446) --> [preview ny metro url](https://covid-projections-git-casulin-city-in-ny-metro-url.covidactnow.vercel.app/us/metro/new-york-city-newark-jersey-city_ny-nj-pa?s=1490446)

### DIFF
--- a/scripts/cbsa/parse-metro-areas.ts
+++ b/scripts/cbsa/parse-metro-areas.ts
@@ -80,6 +80,11 @@ function formatGroup(groupItems: CsvRow[]) {
 
   const totalPopulation = sum(counties.map(county => county?.population || 0));
 
+  const nameForFormatUrl =
+    firstItem.cbsaTitle === 'New York-Newark-Jersey City, NY-NJ-PA'
+      ? 'New York City-Newark-Jersey City, NY-NJ-PA'
+      : firstItem.cbsaTitle;
+
   return {
     cbsaCode: firstItem.cbsaCode,
     cbsaTitle: firstItem.cbsaTitle,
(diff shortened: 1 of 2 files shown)


scripts/cbsa/parse-metro-areas.ts
```diff
@@ -80,14 +80,19 @@ function formatGroup(groupItems: CsvRow[]) {
 
   const totalPopulation = sum(counties.map(county => county?.population || 0));
 
+  const nameForFormatUrl =
+    firstItem.cbsaTitle === 'New York-Newark-Jersey City, NY-NJ-PA'
+      ? 'New York City-Newark-Jersey City, NY-NJ-PA'
+      : firstItem.cbsaTitle;
```
mikelehen, line 86: Curious if we considered doing this in the original CSV file? (scripts/cbsa/mar2020.csv)

Or is the CSV file in a pristine state from some external source and we want to keep it that way?

chasulin: do you mean updating the json file directly? updating the csv would mean updating a lot of instances of 'New York'. all in all i just figured i should update via the script to be safe, but can do it another way!

mikelehen: Oh, hrm. I don't know how the metros data works well enough to know which json file you're referring to.  But yeah, I'm basically wondering if we want to do these sorts of fix-ups via patches in code or actually patch our underlying data file.  

It sounds like it's easier to do it in code.  That seems fine, though I could imagine we'll have more of these in the future so we might want to eventually make the code more general-purpose (have a METRO_RENAMES mapping dictionary or something). But we can start with this and evolve it as necessary

mikelehen (approved):

PR #2358, opened 2021-01-06
add script to optimize static images

pnavarrc:
This PR adds a script to resize images so they match the size that we use to render them. The script needs us to add images and their target sizes one by one.

We might be able to batch-resize certain images (team headshots, for example) since they all have the same size, but I will do that on a separate PR.


pnavarrc:
[Performance - Optimize static images (non-CMS)](https://trello.com/c/8wxDkQTC/744-performance-optimize-static-images-non-cms)

scripts/optimize_images/README.md
```diff
@@ -0,0 +1,36 @@
+# Optimizing Static Images
+
+Static images can take a long time to load on web pages. In order to make this
+faster, we can optimize them in two ways:
+
+- Reduce the image resolution to be appropiate to most devices
```
mikelehen, line 6: Can you explain the difference between reducing resolution and reducing dimensions?

Also, appropiate => appropriate 😛

scripts/optimize_images/optimize-images.ts
```diff
@@ -0,0 +1,44 @@
...
+    if (fs.existsSync(backupImagePath)) {
+      fs.unlinkSync(backupImagePath);
+    }
+  }
+
+  console.log('Done.');
+}
+
+if (require.main === module) {
+  main();
```
mikelehen, line 43: Can you add `.catch(err => { console.log(err); process.exit(-1); }` so errors are handled?

pnavarrc:
[Performance - Optimize static images (non-CMS)](https://trello.com/c/8wxDkQTC/744-performance-optimize-static-images-non-cms)

scripts/optimize_images/optimize-images.ts
```diff
@@ -0,0 +1,44 @@
...
+    fs.renameSync(inputPath, backupImagePath);
+
+    // Resize the image preserving the aspect ratio
+    console.log(`Resizing ${imagePath}`);
+    await sharp(backupImagePath).resize(sizeParams).toFile(inputPath);
+
+    // Remove the backup image
+    if (fs.existsSync(backupImagePath)) {
+      fs.unlinkSync(backupImagePath);
+    }
```
mikelehen, line 36: It's tempting to keep the original files checked into source control so if we ever need to resize an image bigger we don't have to go chase down the original again.

But maybe I'm over-thinking it. WDYT?

mikelehen (approved):

scripts/optimize_images/optimize-images.ts
pnavarrc: yeah, let's keep them in case we need to resize them again!

scripts/optimize_images/README.md
pnavarrc: Updated - we are just reducing the dimensions in the script, CRA doesn't do anything else actually.

scripts/optimize_images/optimize-images.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    // Rename the image `image.png` -> `image-original.png` (sharp doesn't allow
+    // the same input and output path).
+    const { dir: imageDir, name: fileName, ext } = path.parse(inputPath);
+    const backupName = `${fileName}-original${ext}`;
+    const backupImagePath = path.join(imageDir, backupName);
+
+    if (fs.existsSync(backupImagePath)) {
+      console.log(`Backup image ${backupName} found, skipping.`);
+      continue;
```
mikelehen, line 24: I think instead of skipping the image entirely, we could maybe just skip the rename but still run sharp again.

That way if e.g. you need to resize an image, you can update the size in `static-images.ts` and just re-run `yarn optimize-images` and it'll do the right thing.  WDYT?

mikelehen (approved):

scripts/optimize_images/optimize-images.ts
pnavarrc: Good point!

PR #2360, opened 2021-01-06
remove the axios library

pnavarrc:
It looks like we are not using axios anymore, removing!


mikelehen (approved):
I was just about to do this. Thanks!

Issue #70, opened 2020-03-21
Replace Signup Link with Mailjet Signup


zjrosen1:
- Add signup component replace on ModelPage with current signup section
- Add signup component to bottom of home page
- Use existing text with an input field for email address

PR #2361, opened 2021-01-07
remove iframe-resizer

pnavarrc:
Remove an unused script (see https://github.com/covid-projections/covid-projections/issues/70), it was added to autosize an iframe for email signups with Mailjet back in March.


mikelehen (approved):
LOL, nice! 😁

PR #2367, opened 2021-01-07
optimize cms images (part 1)

pnavarrc:
Update the script that resizes static images to resize CMS images based on how they are used.

The `yarn optimize-images` script resizes the images so they match their rendered size, keeping a copy of the original in case we need a higher resolution later. For static images from `assets/images`, we can store the originals in the same directory, but for CMS images, is better to store the copy on a different directory.

Currently, the CMS images are stored in the `public/images_cms` folder, and all the URLs of the images point to that location. The build process copies any images found in the `public` directory unchanged to `build/static/media`, all those images are uploaded to S3 during the deployment process. In order to save on S3 storage and image transfers, we are going to store the originals under `src/assets/original_cms_images`, which are not uploaded to S3, that way we will still have them.

This first PR resizes the CMS images for:
- profile pictures
- government logos
- partners logos
- case study logos

That covers all the images that are not part of markdown content, the next update will be to resize all the remaining images to be 450px (900px to account for retina displays), which matches the body of markdown content, the easier way is to resize all remaining images to have that size.

The rest of the images will be resized on a separate PR. The first two commits have the code changes, the last one is resizing the images.


pnavarrc:
[Performance · Optimize CMS images](https://trello.com/c/XRyWn6eI/730-performance-optimize-cms-images)

scripts/optimize_images/static-images.ts
```diff
@@ -1,28 +1,100 @@
...
+  isRasterImage,
+  joinPublicFolder,
+  cmsBackupPath,
+} from './utils';
+
+const staticImages: ImageInfo[] = [
   {
-    path: 'src/assets/images/ghss.png',
+    inputPath: 'src/assets/images/ghss.png',
+    backupPath: 'src/assets/images/ghss-original.png',
```
mikelehen, line 15: `backupPath` makes it sound like it's just a backup copy or something.  Perhaps `originalPath` ?

mikelehen (approved):
LGTM other than some images getting bigger. 😁 

Also, can you double check if your README needs any updates to cover the CMS images?

chasulin (approved):
other than solving for the images that sized up, looks good!

do we want to apply this to the rest of the cms images (images in case studies/metric explainers/deep dives)? this is a nice improvement as is, so feel free to punt

pnavarrc:
Two images are still slightly larger than the original (despite the checks I added), I will dig into that and fix the remaining images on a follow-up PR.

PR #2371, opened 2021-01-08
Enable ICU Capacity Used for metros.

mikelehen:
NOTE: Only ~half of top metros end up with an ICU Capacity used value right now, I think due to some counties not having it available.  I need to dig into this more, but the frontend changes should be good to go.


src/common/models/Projection.ts
```diff
@@ -226,19 +235,45 @@ export class Projection {
...
+      !DISABLED_ICU.includes(this.fips) &&
+      region.regionType !== RegionType.COUNTY
+    ) {
+      const metricSeries = metricsTimeseries.map(
+        row => row && row.icuCapacityRatio,
       );
-    } else {
-      this.icuHeadroomInfo = null;
+      const metricValue = metrics.icuCapacityRatio;
+      // TODO(michael): We get this from the timeseries since sometimes actuals.icuBeds.capacity can contain a different value.
```
mikelehen, line 251: FWIW- This seems like a potential backend/API bug.

ghop02 (approved):
LGTM!

PR #2372, opened 2021-01-08
resize remaining CMS images

pnavarrc:
- Instead of renaming the original image, copy it to the "originals" folder
- If a CMS image is not found in the list of images to process from the CMS (images that we know how they will be used), resize them to be 600px (x2 for retina), which is a bit bigger than what we need (450px) to have a safety margin.


pnavarrc:
Note: Some images are resized slightly up, that seems to be an artifact of processing them

pnavarrc:
Also, it seems that sometimes the images get retouched: 👨‍🎨 

image

scripts/optimize_images/static-images.ts
```diff
@@ -91,11 +94,31 @@ const caseStudyLogos = chain(allCaseStudies)
...
+  .readdirSync(absolutePath('public/images_cms'))
+  .filter(isRasterImage)
+  .map(imgFilename => path.join('public/images_cms', imgFilename));
+
+const remainingCmsImages = difference(
+  allCmsImages,
+  cmsStructuredImageFilenames,
+);
+
+const cmsRemainingImages = remainingCmsImages.map((imagePath: string) => ({
```
mikelehen, line 118: Can you add a comment (perhaps borrowing from the PR description) about these images and the chosen 600px size?

mikelehen (approved):

pnavarrc:
[Performance · Optimize CMS images](https://trello.com/c/XRyWn6eI/730-performance-optimize-cms-images)

PR #2391, opened 2021-01-11
Performance - defer loading and evaluating 3rd party scripts

pnavarrc:
According to [PageSpeed Insights](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fcovidactnow.org&tab=mobile), we need to optimize loading third party code.

image

Using `async` instead of `defer` because I imagine that the page load event is important to event trackers

- [Loading third-party JavaScript](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/loading-third-party-javascript)

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -119,7 +119,7 @@
 
   <!-- Global site tag (gtag.js) - Google Ads: 527465415 -->
   <script async src="https://www.googletagmanager.com/gtag/js?id=AW-527465415"></script>
-  <script async>
+  <script>
     window.dataLayer = window.dataLayer || [];
     function gtag() { dataLayer.push(arguments); }
     gtag('js', new Date());
@@ -127,7 +127,7 @@
   </script>
 
   <!-- Event snippet for Website traffic conversion page -->
-  <script async>
+  <script>
     gtag('event', 'conversion', { 'send_to': 'AW-527465415/lRyqCM2a9-kBEMf3wfsB' });
   </script>
 
--- a/public/index.html
+++ b/public/index.html
@@ -61,33 +61,33 @@
     rel="stylesheet">
 
   <!-- fullstory -->
-  <script>
-  window['_fs_debug'] = false;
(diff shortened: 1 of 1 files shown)


public/index.html
```diff
@@ -119,15 +119,15 @@
 
   <!-- Global site tag (gtag.js) - Google Ads: 527465415 -->
   <script async src="https://www.googletagmanager.com/gtag/js?id=AW-527465415"></script>
-  <script>
+  <script async>
```
mikelehen, line 122: Hrm...  this seems especially questionable since Google tells you to use `async` for the external script reference but not for this inline script.  Should we really be diverging from their example?

mikelehen (commented):
I am slightly nervous about using `async` scripts when that’s not how google/twitter/facebook/fullstory/etc. tell you to do it.  It _hopefully_ works fine, but if this is the best practice, why don't they prescribe it?  And in general I get nervous about unintended consequences with this sort of thing (which could affect ordering scripts execute, etc.).

I think moving the scripts to the bottom of our `<body>` tag is the more standard solution to make sure they don't block page rendering and that probably works better in old browsers, etc. anyway  https://stackoverflow.com/a/10808243

mikelehen:
Also note https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/loading-third-party-javascript says:

> In Blink-based browsers, async and defer currently lower the priority of the network request for resources so it can cause it to load significantly later than it would as a blocking script. This is useful to know in particular for analytics scripts.

mikelehen (approved):
That said, we can try this and see if it causes any issues.  I guess we were already using `async` for the main google analytics load, which is almost certainly the most important one.  So maybe it's fine.

public/index.html
pnavarrc: That's a good point - I'm going to restore GA to what they recommend, but keep the async for the remaining scripts. FullStory is particularly heavy, and it's mostly used to record user behaviour.

PR #2395, opened 2021-01-12
update page metatags to calculate them with summaries

pnavarrc:
The page meta tags don't really depend on the projections, we can calculate them using location summaries, which should be faster.


pnavarrc:
[Performance / SEO - Update page title and canonical before loading the projection (location page)](https://trello.com/c/5UHi8ib9/762-performance-seo-update-page-title-and-canonical-before-loading-the-projection-location-page)

mikelehen (approved):

PR #2407, opened 2021-01-13
Geolocation

chasulin:
Geolocates user using [IP-API](https://ip-api.com/). If successful, ranks user's regions at the top of the search bar dropdown menu


src/common/regions/utils.ts
```diff
@@ -56,3 +60,77 @@ export function getAutocompleteRegions(region?: Region) {
...
+
+// TODO (Chelsi): fix the any
+export function getMetroRegionFromZipCode(zipCode: string): any {
+  const metroFromZip = find(regions.metroAreas, (region: MetroArea) =>
+    find(region.counties, (region: County) =>
+      countyIncludesZip(region, zipCode),
+    ),
+  );
+  return metroFromZip;
+}
```
pnavarrc, line 95: Consider finding the county first using your function from above, and then use that to find the metro area.

src/common/regions/utils.ts
```diff
@@ -56,3 +60,77 @@ export function getAutocompleteRegions(region?: Region) {
...
+
+// TODO (chelsi): move elsewhere?
+interface GeolocatedRegions {
+  county?: Region;
+  metroArea?: Region;
+  state?: Region;
+}
```
pnavarrc, line 102: Maybe we don't need a type for this? we could just return an array of `Region` in order

src/common/regions/utils.ts
```diff
@@ -56,3 +60,77 @@ export function getAutocompleteRegions(region?: Region) {
...
+
+/**
+ * If we are able to geolocate the user, we rank their regions (state, county, metro if applicable) first
+ * in the searchbar dropdown menu. This sorts the Region[] accordingly.
+ */
+export function getAutocompleteRegionsWithGeolocation(
+  geolocation: GeolocationInfo,
+  locations: Region[],
+): Region[] {
+  const geolocatedRegions = getGeolocatedRegions(geolocation);
```
pnavarrc, line 126: Returning an array of `Region` object (with 0, 1, 2, or 3 items) will eliminate the need of sanitizing

pnavarrc (commented):
Added some suggestions, will take a second pass tomorrow morning!

src/common/regions/utils.ts
chasulin: Per our chat, gonna keep it in obj format

pnavarrc (approved):
Looks good from my location!

PR #2416, opened 2021-01-14
analytics: add tracking for the search box

pnavarrc:
Adds tracking for:
- Focus event (by clicking and keyboard navigation)
- Navigation event. 

For the navigation event, it includes the full name of the location, and the string that the user entered. I also added some missing keys


src/components/Search/SearchAutocomplete.tsx
```diff
@@ -73,8 +79,10 @@ const SearchAutocomplete: React.FC<{
       renderOption={option => {
         return <MenuItem region={option} zipCodeInput={zipCodeInput} />;
       }}
+      getOptionLabel={getOptionLabel}
```
pnavarrc, line 82: There was an error when selecting metro areas, this makes the error go away 🤷

pnavarrc:
[Analytics - Track events on the search bar](https://trello.com/c/JunCwMbn/771-analytics-track-events-on-the-search-bar)

src/components/Search/SearchAutocomplete.tsx
chasulin, line 82: adding `getOptionLabel` makes it so that when you select a location, the name of that location populates the searchbar for a second before navigating to the next page. it might look smoother without this/how it is currently on prod

chasulin: oh interesting. also- didnt see this comment [before writing mine below](https://github.com/covid-projections/covid-projections/pull/2416#discussion_r556987450). feel free to ignore

chasulin: making getOptionLabel return an empty string could nix the location name flicker, but not sure if that nixes the error you were seeing for metros

chasulin: making `getOptionLabel` return an empty string could nix the location-name-flicker, but not sure if that nixes the error you were seeing for metros

chasulin (approved):

PR #2417, opened 2021-01-14
Enable county ICU Capacity Used metric for counties with >= 15 beds.

mikelehen:
This threshold prevents us from displaying charts that are *too* crazy due to low bed counts and still covers enough counties for 80% of the US population. 

For states with too few beds, you get:

image


src/common/metrics/hospitalizations.tsx
```diff
@@ -129,7 +132,7 @@ function renderStatus(projections: Projections): React.ReactElement {
     'This suggests there is likely enough capacity to absorb a wave of new COVID infections',
     'This suggests some ability to absorb an increase in COVID cases',
     'This suggests hospitals may not be well positioned to absorb a wave of new COVID infections without substantial surge capacity. Caution is warranted',
-    'This suggests hospitals cannot absorb a wave of new COVID infections without substantial surge capacity. Aggressive action urgently needed',
+    'This suggests hospitals cannot absorb a wave of new COVID infections without substantial surge capacity.',
```
mikelehen, line 135: I tweaked this copy since there are still a lot of counties in this bucket and I think the copy may be overly sensational.

cc/ @BrettBoval

BrettBoval: SGTM

BrettBoval (approved):
Cutoff threshold and working are okay with me.

ghop02 (approved):

PR #2422, opened 2021-01-14
Covid projections: master to main

ghop02:
Replaces all uses of master with main, should be merged after main branch goes live.  Will require a bit of extra coordination with merging to develop then doing the merge into master/main


bin/deploy.sh
```diff
@@ -1,7 +1,7 @@
-git branch | grep "* master" > /dev/null
+git branch | grep "* main" > /dev/null
```
mikelehen, line 1: this script is unused and I have a PR to delete it. https://github.com/covid-projections/covid-projections/pull/2453

mikelehen (approved):
I think this looks good, but we need to be sure everybody who does deploys is aware of this change in advance.

PR #2425, opened 2021-01-14
Learn CE landing page card

chasulin:
This PR includes:
- `LandingPageCard` component + styles, used for covid explained landing + CAN updates landing
- Accessibility fix for card component (keyboard navigation wasn't focusing on the component, fix: wrapped the `card` in the `link`, rather than the other way around)
- Adds CAN updates section to the learn sidebar
- 1 of 2 share-quotes (Explained share quote is in, still waiting on Updates copy)


src/screens/Learn/Learn.style.tsx
```diff
@@ -52,18 +58,131 @@ export const LearnHeading1 = styled(Heading1)`
...
+  margin-top: 0.75rem;
+  &:last-child {
+    margin-bottom: 0;
+  }
+`;
+
+/*
+ TODO (Chelsi): we're almost always removing the underline
+ manually. Lets put this styled Link somewhere global ::
+*/
```
pnavarrc, line 93: Actually, the best practice is to have links underlined for accessibility (they don't have enough contrast), we can probably revisit this with Victoria (but true, we might as well style this globally)

src/screens/Learn/CaseStudies/CaseStudyCard.tsx
```diff
@@ -15,8 +15,8 @@ const CaseStudyCard = (props: { cardContent: CaseStudy; url: string }) => {
   const { cardContent, url } = props;
   const { shortTitle, summary, caseStudyId } = cardContent;
   return (
-    <StyledCard>
-      <StyledLink to={`${url}/${caseStudyId}`}>
+    <StyledLink to={`${url}/${caseStudyId}`}>
```
pnavarrc, line 18: We need to add `aria-label` set to the short title here. Screen readers determine how to read links checking the following properties on the link element (in order of precedence, more or less)

- `aria-label`
- `title`
- text of the link (what's inside `<a>` and `</a>`)

As it is right now, the text of the link includes the card title, the alt text of the logo, the summary and "Learn more", so the screen reader will read it as:

> **California Department of Public Health** California Department of Public Health logo California’s CalCAT tool uses CAN data to analyze current COVID spread and project the impact of policy interventions. Learn How

The `aria-label` usually takes precedence over the link text when set, so if we add `aria-label` to the styled link

```tsx
<StyledLink to={} aria-label={shortTitle}>
  {/* ... */}
</StyledLink>
```

the screen reader will announce just the short title:

>  **California Department of Public Health***

In reality, it's a bit more complex than that, [different screen readers announce slightly different things](https://www.deque.com/blog/text-links-practices-screen-readers/), but this will probably be an improvement.

pnavarrc (approved):
Looks great! Thanks for fixing the focus issue! 🙌

src/screens/Learn/CaseStudies/CaseStudyCard.tsx
chasulin: on it!

PR #2466, opened 2021-01-19
Geolocation updates

chasulin:
- Geolocates only on homepage
- Upgrades our use of ip-api to pro (unlimited lookups)

### DIFF
--- a/src/common/hooks/useGeolocation.ts
+++ b/src/common/hooks/useGeolocation.ts
@@ -22,9 +22,9 @@ export default function useGeolocation(): GeolocationInfo | undefined {
         .then(response => response.json())
         .then(data =>
           setIpData({
-            zipCode: data.postal,
-            stateCode: data.region_code,
-            country: data.country_name,
+            zipCode: data.zip,
+            stateCode: data.region,
+            country: data.country,
           }),
         )
         .catch(e => console.error(e));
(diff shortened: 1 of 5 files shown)


src/common/hooks/useGeolocation.ts
```diff
@@ -4,6 +4,7 @@
 */
 
 import { useEffect, useState } from 'react';
+import { IP_API } from 'common/ip-api';
```
pnavarrc, line 7: Consider renaming this as `IP_API_KEY` instead

src/common/ip-api.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Key for our pro subscription of ip-api (https://ip-api.com/) - used for
+ * geolocating and pre-populating user's locations on the homepage.
+ *
+ * Keeping the key here is fine (origin restrictions make it usable only for CAN)
```
pnavarrc, line 5: Maybe add a note saying that it won't work locally or in preview links so we don't forget about that

src/common/regions/utils.ts
```diff
@@ -140,15 +140,15 @@ export function getAutocompleteRegionsWithGeolocation(
  * If not, returns autocomplete results sorted for pagetype.
  */
 export function getFinalAutocompleteLocations(
-  geolocation: GeolocationInfo | undefined,
+  geolocation?: GeolocationInfo | undefined,
```
pnavarrc, line 143: Isn't `undefined` implied if you add the `?`? 🤔

pnavarrc (approved):
Looks good, added some notes!

PR #2469, opened 2021-01-20
Update vaccination chart

pnavarrc:
[Mocks](https://www.figma.com/file/oqCyg93qwqNLdQLa8xYdry/CAN-Vaccinate?node-id=1%3A217)

Adding initiated and completed vaccinations to the vaccinations chart. It shows the series for:

- 1st shot
- 2nd shot

I followed the style of axes, grid lines, and other elements to match the most recent chart (Explore).

image

[Feedback paper](https://www.dropbox.com/scl/fi/tr5udqg76sej83v104yc1/CAN-Vaccinate-Kickoff.paper?dl=0&rlkey=88j803zmtb12cvqstxetjzwwd)


src/components/Charts/ChartVaccinations.tsx
```diff
@@ -1,37 +1,245 @@
...
+  seriesList: Series[];
+  left: (d: Column) => number;
+  top: (d: Column) => number;
+  subtext: string;
+}> = ({ seriesList, left, top, date }) => {
+  const [seriesInitiated, seriesCompleted] = seriesList;
+  const pointCompleted = findPointByDate(seriesCompleted.data, date);
+  const pointInitiated = findPointByDate(seriesInitiated.data, date);
+
+  return pointCompleted && pointInitiated ? (
```
ghop02, line 43: I think this would be a bit more concise as: 

```
if (!pointCompleted || !pointInitiated) {
  return null;
}

return (
...
)
```

src/components/Charts/ChartVaccinations.tsx
```diff
@@ -1,37 +1,245 @@
...
+  );
 };
 
-const ChartVaccinations = ({
-  columnData,
-  height = 400,
-}: {
-  columnData: Column[];
+const ChartVaccinationsAutosize: React.FC<{
+  seriesList?: Series[];
```
ghop02, line 229: Thoughts on making this not a list? Just more explicit if you pass in the series as either an interface:
```
interface VaccinationSeries {
  iniated?: Series
  completed?: Series
}
```
or their own arguments

src/components/Charts/ChartVaccinations.tsx
```diff
@@ -1,37 +1,245 @@
...
+          {isNumber(pointCompleted.y)
+            ? formatPercent(pointCompleted.y, 1)
+            : '-'}
+        </Styles.TooltipValue>
+      </Styles.TooltipLine>
+    </Tooltip>
+  ) : null;
+};
+
+const DataHoverMarkers: React.FC<{
```
ghop02, line 74: I might recommend just making this: `DataHoverMarker` and have it just take a `Series`, pushing the `map` up to the calling component.  I think it makes the code a bit more declarative and gives you more flexibility with how you display this

src/components/Charts/ChartVaccinations.tsx
```diff
@@ -1,37 +1,245 @@
...
+  // Shows the year if the tick is in January (0) or December (11)
+  const dateFormat =
+    momentDate.month() === 0 || momentDate.month() === 11 ? 'MMM YYYY' : 'MMM';
+  return momentDate.format(dateFormat);
+};
+
+const VaccinesTooltip: React.FC<{
+  date: Date;
+  seriesList: Series[];
+  left: (d: Column) => number;
```
ghop02, line 35: fwiw I didn't know what left and top were here until i read the calling code - maybe `marginLeft` and `marginTop`?

src/components/Charts/ChartVaccinations.tsx
```diff
@@ -1,37 +1,245 @@
...
+  top: (d: Column) => number;
+  subtext: string;
+}> = ({ seriesList, left, top, date }) => {
+  const [seriesInitiated, seriesCompleted] = seriesList;
+  const pointCompleted = findPointByDate(seriesCompleted.data, date);
+  const pointInitiated = findPointByDate(seriesInitiated.data, date);
+
+  return pointCompleted && pointInitiated ? (
+    <Tooltip
+      width={'160px'}
```
ghop02, line 45: Is this the right place for this constant? (don't necessarily have anything better in mind, just flagging)

ghop02 (commented):
Overall looks good! Couple of style questions but nothing too crazy hopefully :)

src/components/Charts/ChartVaccinations.tsx
pnavarrc: I'm trying to unify the data model of charts, the array of series works well for Explore and this chart, I prefer to keep it

chasulin (approved):
Other than the little nits I slacked you about, looks good to me!

src/components/Charts/ChartVaccinations.tsx
pnavarrc: Not really, because those are `top` and `left` to absolutely position the tooltip

src/components/Charts/ChartVaccinations.tsx
pnavarrc: That might not work in general, because sometimes the marker might have other elements. In this case, for example, we are adding a "tracker line", a vertical line that goes between the 2 points (doing that on a separate PR)

src/components/Charts/ChartVaccinations.tsx
pnavarrc: I think so, other tooltips need more space. Each chart can have customized tooltips, and those might need more or less space, so it makes sense to pass the width as parameter

PR #2475, opened 2021-01-20
fix timezone on explore chart

pnavarrc:
The grid line is not aligned with the y-axis because of the timezone difference. This PR uses the `scaleUtc` to make sure they match.

**Before**
image

**After**
image


ghop02 (approved):
lgtm!

PR #2480, opened 2021-01-20
Geolocation updates, cont.

chasulin:
- Adds `useGeolocation` back into location pages (since we now have unlimited lookups)
- Re-orders geolocated regions (metro, county, state)

### DIFF
--- a/src/common/regions/utils.ts
+++ b/src/common/regions/utils.ts
@@ -1,4 +1,12 @@
-import { concat, partition, sortBy, find, values, isNull } from 'lodash';
+import {
+  concat,
+  partition,
+  sortBy,
+  find,
+  values,
+  remove,
+  isNull,
+} from 'lodash';
 import regions from './region_db';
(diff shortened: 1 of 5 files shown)


src/common/regions/utils.ts
```diff
@@ -120,18 +134,24 @@ export function getAutocompleteRegionsWithGeolocation(
...
-    (region: Region | undefined) => region instanceof Region,
-  ); // filters out undefined
-  const [usersRegions, otherRegions] = partition(
+
+  const geolocatedRegionValues = values(geolocatedRegions);
+  const sanitizedGeolocatedRegions: Region[] = filterUndefinedRegions(
+    geolocatedRegionValues,
+  );
+
+  const otherRegions: Region[] = remove(
```
pnavarrc, line 147: Maybe using `_.without` is better?

pnavarrc (approved):
Looks good!

PR #2483, opened 2021-01-21
vaccination: copy updates 2

pnavarrc:
[Copy doc](https://docs.google.com/document/d/14d2VfKcl0eqTYUyvuEzR0-cxb95tmdARJ7rceaRwAvg/edit#) + comments from Igor on Slack.

### Changes
- Update the name of the vaccination metric on:
  - Location header
  - Compare
  - Location page embed
  - Vaccination chart
- Change the level of the vaccination metric to unknown so the dot is grey 
- Add tracking for clicks on the CDC link (vaccination disclaimer)
- Remove _Beta_ from ICU Capacity
- Add _Beta_ to Vaccinations

### DIFF
--- a/src/common/metric.tsx
+++ b/src/common/metric.tsx
@@ -46,6 +46,11 @@ export function getLevelInfo(metric: Metric, value: number | null): LevelInfo {
 }
 
 export function getLevel(metric: Metric, value: number | null): Level {
+  // Vaccinations don't have levels
+  if (metric === Metric.VACCINATIONS) {
+    return Level.UNKNOWN;
+  }
+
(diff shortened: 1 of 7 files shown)


chasulin (approved):
Two things I noticed:
- The circle icons in the compare table are still picking up a level color
- The % in the compare table isn't the most recent datapoint + doesn't match the % in the location's header (it looks like compare is showing the Jan 15th %)

The changes in this PR look good--approving in case you want to merge this in and address the things above separately!

pnavarrc:
Thanks for reviewing! I'm going to merge the PR and fix the circle icons on a separate PR. The data in the compare table might need to be updated, I need to check if the script to generate location summaries adds vaccinations yet or not.

PR #2488, opened 2021-01-21
Updated region name to state name in notification link

faihegberg:
Currently the link in the "Notifications" section under "Recommended Actions" section in the locations page uses the general region name instead of state name. This PR changes the link to use state name instead.

image


chasulin (approved):
Looks good!

PR #2489, opened 2021-01-21
update the vaccination tag to be "new"

pnavarrc:
Replace the _Beta_ tag with _New Indicator_

image


chasulin (approved):
Do we want to add the tag next to the chart header as well?

PR #2490, opened 2021-01-21
vaccination: remove metric detail

pnavarrc:
We don't need the summary since this metric is not graded

image


chasulin (approved):
just need to remove the unused consts!

PR #2491, opened 2021-01-21
Vaccines announcement

chasulin:


src/components/Compare/CompareMain.tsx
```diff
@@ -276,6 +277,7 @@ const CompareMain = (props: {
   return (
     <Fragment>
       <DivForRef ref={tableRef}>
+        <Anchor id="compare" />
```
pnavarrc, line 280: I think we can just give the ID to `DivForRef` element

src/screens/HomePage/HomePage.tsx
```diff
@@ -108,6 +109,7 @@ export default function HomePage() {
                   justify="flex-end"
                 >
                   <SelectorWrapper>
+                    <Anchor id="search" />
```
pnavarrc, line 112: Also here, we can set the id in the `SelectorWrapper` element

pnavarrc (approved):
Added some notes, looks good!

pnavarrc (approved):

PR #2495, opened 2021-01-22
Banner-vaccination data

chasulin:


src/components/Banner/VaccinationsBanner.style.tsx
```diff
@@ -0,0 +1,88 @@
...
+  &:last-of-type {
+    background-color: ${COLOR_MAP.LIGHTGRAY_BG};
+    color: ${COLOR_MAP.BLUE};
+    box-shadow: none;
+    margin-left: .5rem;
+    &:hover {
+        text-decoration: underline;
+    }
+}
+  `;
```
pnavarrc, line 42: It might be better to use 2 separate buttons and style them independently - the second button looks like it could use `variant="text"` https://material-ui.com/components/buttons/#text-buttons

pnavarrc (approved):
Looks good!

PR #2497, opened 2021-01-22
Use vaccines from metrics

ghop02:
Update the data snapshot, api types, and code to use metrics for api values


src/common/models/Projection.ts
```diff
@@ -364,36 +369,47 @@ export class Projection {
...
 
-    return null;
+    const peopleVaccinated = actuals.vaccinationsCompleted;
+    const peopleInitiated = actuals.vaccinationsInitiated;
+    const percentInitiated = metrics.vaccinationsInitiatedRatio;
+    const percentVaccinated = metrics.vaccinationsCompletedRatio;
+
+    if (
+      peopleVaccinated === null ||
+      peopleVaccinated === undefined ||
```
ghop02, line 387: There's probably a better way to do this, suggestions included :)

pnavarrc: I would use `Number.isFinite` instead, it returns `false` for: `NaN`, `-Infinity` and `+Infinity`, `null` and `undefined`

pnavarrc (approved):
Looks good! Added a suggestion for the validation

src/common/models/Projection.ts
ghop02: ah this doesn't quite work with typescript.  will attempt to do later

PR #2499, opened 2021-01-22
Disable ohio vaccines

ghop02:


src/common/models/Projection.ts
```diff
@@ -39,7 +39,9 @@ const DISABLED_ICU = new DisabledFips([
   '53', // https://trello.com/c/1IkmUuhw/
 ]);
 
-const DISABLED_VACCINATIONS = new DisabledFips([]);
+const DISABLED_VACCINATIONS = new DisabledFips([
+  /39.../, // Disabling Ohio counties - vaccinations showing 0%
```
BrettBoval, line 43: Just noting that we know this is caused by a mistake in the scrapers treating a daily metric as a cumulative.

BrettBoval (approved):

src/common/models/Projection.ts
ghop02: yeah i'll replace this with the trello ticket

PR #2519, opened 2021-01-25
Geolocation in homepage explore

chasulin:
[Trello item](https://trello.com/c/idum9uby/821-make-explore-reflect-the-users-location)

### DIFF
--- a/src/common/hooks/useGeolocationInExplore.ts
+++ b/src/common/hooks/useGeolocationInExplore.ts
@@ -9,11 +9,12 @@
  */
 
 import { useEffect, useState } from 'react';
-import { isNull, values, chain } from 'lodash';
+import { isNull, values } from 'lodash';
 import {
   getGeolocatedRegions,
   filterUndefinedRegions,
   FipsCode,
+  Region,
 } from 'common/regions';
 import { getLargestMetroFipsForExplore } from 'screens/HomePage/utils';
 import { GeolocationInfo } from './useGeolocation';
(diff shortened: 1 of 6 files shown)


src/common/hooks/useGeolocationInExplore.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  useEffect(() => {
+    if (geolocation) {
+      const userRegions = getGeolocatedRegions(geolocation);
+      if (!isNull(userRegions)) {
+        const regionValues = values(userRegions);
+        const filteredUserRegions = filterUndefinedRegions(regionValues);
+        const userRegionsFips = chain(filteredUserRegions)
+          .map((region: any) => region.fipsCode)
+          .value();
```
pnavarrc, line 37: It might be better to call `map` directly since there is only one operation in the `chain` call (also, maybe add the `Region` type? it might not be necessary to add the type explicitly).

src/screens/HomePage/HomePage.tsx
```diff
@@ -47,6 +46,7 @@ export default function HomePage() {
   const indicatorsRef = useRef(null);
 
   const geolocation = useGeolocation();
+  const initialFipsForExplore = useGeolocationInExplore(5, geolocation);
```
pnavarrc, line 49: It might be better to simplify `useGeolocationInExplore` to make it always return user locations by the case of `geolocation` being `null` outside the function

```tsx
const initialFipsForExplore = geolocation
  ? useGeolocationInExplore(geolocation)
  : defaultExploreFipsLocations;
```

pnavarrc (approved):
Added some suggestions, looks good

src/screens/HomePage/HomePage.tsx
chasulin: per our chat, keeping all the logic+checks contained in `useGeolocationInExplore`

PR #2521, opened 2021-01-25
better alignment for the explore labels

pnavarrc:
The solution is pretty simple, it just adds a margin on the next label if the y coordinate of the current label is too close to the previous label, and keep the position otherwise. A potential drawback of this solution is that the labels could go off-screen, but since the labels are outside the `RectClipGroup`, there is some extra space below the x-axis that could be used by the labels.

**Before**
image

**After**
image

### DIFF
--- a/src/components/Explore/ExploreChart.tsx
+++ b/src/components/Explore/ExploreChart.tsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import { sortBy, last } from 'lodash';
 import { Series } from './interfaces';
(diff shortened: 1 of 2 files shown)


chasulin (approved):

src/components/Explore/ExploreChart.tsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import { sortBy, last } from 'lodash';
```
chasulin, line 2: remove before merging!

PR #2522, opened 2021-01-25
CMS content

chasulin:
Working branch for content entry


BrettBoval (approved):

src/components/Markdown/MarkdownLink.tsx
```diff
@@ -12,6 +12,14 @@ import ExternalLink from 'components/ExternalLink';
  * it in the current page. If the link is invalid, we just render the text of
  * the link.
  */
+
+/* Implementing lazy loading requires a short timeout to make scrollWithOffset scroll the page to the correct location */
```
ghop02, line 16: Is there an easy way to scroll on elements loaded? I'm guessing no

ghop02 (approved):

src/components/Markdown/MarkdownLink.tsx
chasulin: @faihegberg is actually working on that!

PR #2531, opened 2021-01-26
transform the header buttons into real buttons

pnavarrc:
Changes the location page header buttons to be real buttons instead of wrapped divs to make them accessible. I also added a focused style to match the focused state of links. The buttons can be operated with the keyboard only and have a visual indicator of focus now.

image


chasulin (changes requested):
looks like this change introduces an unintended gray background on hover. other than that, looks good!

chasulin (approved):

PR #2542, opened 2021-01-26
implement vaccination block component

pnavarrc:
Implement the vaccination block component - [mocks](https://www.figma.com/file/8BG3s3peEdA2LsaHsnZtm3/CAN-Vaccinate?node-id=1%3A387). 

Note that there are some components that look different (the buttons are missing the arrow) and some pending items, but it would be good to merge this to avoid blocking the integration with vaccine data.


chasulin (approved):
LGTM

PR #2543, opened 2021-01-26
Ghop02/add vaccination info to cms

ghop02:
* adds vaccine info to cms with fips code (state only)
* adds existing vaccine data into cms format
* adds regions to cms (not quite the most efficient way I think)


src/components/VaccinationBlock/utils.ts
```diff
@@ -1,14 +1,43 @@
...
+    return [region];
+  }
+
+  if (region instanceof County) {
+    const county = region as County;
+    const metro = regions.metroAreas.find(metro =>
+      metro.counties.map(county => county.fipsCode).includes(county.fipsCode),
+    );
+    if (metro) {
+      return [region, metro, region.state];
```
pnavarrc, line 25: Maybe `[county, metro, region.state` would be more clear?

src/components/VaccinationBlock/utils.ts
```diff
@@ -1,14 +1,43 @@
...
+    } else {
+      return [region, region.state];
```
pnavarrc, line 27: same `[county, county.state]` might be better since we have the variable

src/cms-content/vaccines/index.ts
```diff
@@ -35,15 +35,19 @@ export const useRegionTypeVaccineData = (regionType: RegionType) => {
   return data;
 };
 
-export const useRegionVaccineData = (region: Region) => {
-  const [data, setData] = useState<RegionVaccinationInfo>();
-  const regionTypeData = useRegionTypeVaccineData(region.regionType);
+export const useRegionVaccineData = (
+  regionType: RegionType,
+  regions: Region[],
```
pnavarrc, line 40: Can we determine what to fetch based on the `regions` argument? I'm not sure we really need to use the `regionType` argument here

public/admin/config.yml
```diff
@@ -644,3 +644,70 @@ collections:
...
+  - label: Locations
+    name: locations
+    label_singular: Location
+    description: description
+    format: json
+    create: true
+    files:
+      - label: States
+        name: states
+        file: src/cms-content/locations/state.json
```
pnavarrc, line 657: This file won't be available when we build, because we are not importing it in the app.   We will need to save it in `public` for it to be copied during the build process

public/admin/config.yml
```diff
@@ -644,3 +644,70 @@ collections:
...
+                file_name: states.json
+                value_field: regions.*.fips
+                search_fields: ["regions.*.locationName", "regions.*.fips"]
+              - label: Hidden
+                name: hidden
+                widget: boolean
+                required: false
+              - label: Eligibility Information URL
+                name: eligibilityInfoUrl
+                widget: string
```
pnavarrc, line 709: I would make `eligibilityInfoUrl` required, there isn't much point on adding the data item without at least this

public/cms-content/locations/massachussetts.json
```diff
@@ -0,0 +1,6 @@
+{
+  "locationName": "Massachussetts",
+  "locationType": "state",
+  "fips": "25",
+  "stateCode": "MA"
+}
```
pnavarrc, line 6: Do we need to keep this file?

scripts/update_cms_locations.ts
```diff
@@ -0,0 +1,33 @@
...
+    stateCode: '',
+  }));
+
+  const outputPath = path.join(
+    __dirname,
+    '..',
+    'src',
+    'cms-content',
+    'locations',
+    `state.json`,
```
pnavarrc, line 20: nit: you can add a subpath here to make it more readable

```ts
// or split the path at one point or two
path.join(__dirname, "../src/cms-content/locations/state.json")
path.join(__dirname, "../src/cms-content/locations", "state.json")
```

src/cms-content/locations/72.json
```diff
@@ -0,0 +1 @@
+{"locationName":"Puerto Rico","locationType":"state","fips":"72","stateCode":""}
```
pnavarrc, line 1: I'm not sure if we need this file

src/cms-content/locations/state.json
```diff
@@ -0,0 +1 @@
...Code":""},{"locationName":"Oregon","locationType":"state","fips":"41","stateCode":""},{"locationName":"Pennsylvania","locationType":"state","fips":"42","stateCode":""},{"locationName":"Puerto Rico","locationType":"state","fips":"72","stateCode":""},{"locationName":"Rhode Island","locationType":"state","fips":"44","stateCode":""},{"locationName":"South Carolina","locationType":"state","fips":"45","stateCode":""},{"locationName":"South Dakota","locationType":"state","fips":"46","stateCode":""},{"locationName":"Tennessee","locationType":"state","fips":"47","stateCode":""},{"locationName":"Texas","locationType":"state","fips":"48","stateCode":""},{"locationName":"Utah","locationType":"state","fips":"49","stateCode":""},{"locationName":"Vermont","locationType":"state","fips":"50","stateCode":""},{"locationName":"Virginia","locationType":"state","fips":"51","stateCode":""},{"locationName":"Washington","locationType":"state","fips":"53","stateCode":""},{"locationName":"West Virginia","locationType":"state","fips":"54","stateCode":""},{"locationName":"Wisconsin","locationType":"state","fips":"55","stateCode":""},{"locationName":"Wyoming","locationType":"state","fips":"56","stateCode":""}]}
```
pnavarrc, line 1: Same than before, since this file is not imported, is not going to be copied during the build process, we need to move it to `public`

src/cms-content/recommendations/recommendations-main.json
```diff
@@ -70,34 +70,6 @@
...
-      "id": "SCHOOLS_YELLOW",
-      "body": "**Schools:** If [safe learning spaces](https://globalepidemics.org/wp-content/uploads/2020/07/pandemic_resilient_schools_briefing_72020.pdf) exist, phase re-opening: 1st priority: preK-5 and special education services for preK-8. 2nd priority: grades 6-8 and special education services for grades 9-12. 3rd priority: grades 9-12 on a hybrid schedule."
-    },
-    {
-      "body": "**Schools:** In person education is appropriate only if conditions for [safe learning spaces](https://globalepidemics.org/wp-content/uploads/2020/07/pandemic_resilient_schools_briefing_72020.pdf) are in place.",
-      "source": "HARVARD",
-      "category": "SCHOOLS",
-      "level": "GREEN",
-      "id": "SCHOOLS_GREEN"
-    },
```
pnavarrc, line 100: 🤔

pnavarrc (commented):
Added some notes! I think it needs some cleanup, but overall looks good!

src/cms-content/locations/state.json
ghop02: It's fine if this isn't imported - it's only used in the CMS

public/admin/config.yml
ghop02: I think the default value is true, so it shouldn't be necessary

src/cms-content/vaccines/index.ts
```diff
@@ -0,0 +1,66 @@
+import { useState, useEffect } from 'react';
+import { getStateCode, Region, RegionType, State } from 'common/regions';
+
+export interface RegionVaccinationInfo {
```
pnavarrc, line 4: There are some linting warnings here ☝️

public/admin/config.yml
```diff
@@ -644,3 +644,35 @@ collections:
           - label: Metadata description
             name: metadataDescription
             widget: text
+
+  - label: Vaccination Information
+    name: vaccinationInfo
+    label_singular: Region Vaccine information
+    description: description
```
pnavarrc, line 651: Maybe update the description so it helps people using the CMS

src/cms-content/vaccines/index.ts
```diff
@@ -0,0 +1,66 @@
...
+) => {
+  const [data, setData] = useState<RegionVaccinationInfo[]>();
+  const regionTypeData = useRegionTypeVaccineData(regionType);
+  useEffect(() => {
+    const matching =
+      regionTypeData?.filter(item =>
+        regions.map(region => region.fipsCode).includes(item.fips),
+      ) || [];
+    setData(matching);
+  }, [regionTypeData, regions]);
```
pnavarrc, line 50: It might be easier to create a map with the vaccination information by flips to avoid iterating through the `regionTypeData` array for each lookup

src/cms-content/vaccines/index.ts
```diff
@@ -0,0 +1,66 @@
...
+  return data;
+};
+
+export const useRegionVaccineData = (region: Region) => {
+  const [data, setData] = useState<RegionVaccinationInfo>();
+  const regionTypeData = useRegionTypeVaccineData(region.regionType);
+  useEffect(() => {
+    const matching = regionTypeData?.find(
+      item => item.fips === region.fipsCode,
+    );
```
pnavarrc, line 61: Also here

pnavarrc (approved):
Remember to restore the cms brach name before merging !

src/cms-content/vaccines/index.ts
ghop02: I think i might end up redoing how this is called, so going to punt this for now!

src/cms-content/vaccines/index.ts
ghop02: It's a bit tricky because of the way the hooks work right now.  I think we'll be able to clean this up a bit but can deal with it when we add other region types

PR #2545, opened 2021-01-27
vaccinations: button fixes + analytics

pnavarrc:
Add the chevron icon + tracking for clicks

### DIFF
--- a/src/components/VaccinationBlock/LinkButton.tsx
+++ b/src/components/VaccinationBlock/LinkButton.tsx
@@ -1,0 +1,18 @@
+import React from 'react';
+import { Button, ChevronRightIcon } from './VaccinationBlock.style';
+
+type ButtonProps = React.ComponentProps<typeof Button>;
+
+const LinkButton: React.FC<ButtonProps> = ({
+  href,
+  children,
+  ...otherProps
+}) => {
+  return (
+    <Button href={href} endIcon={<ChevronRightIcon />} {...otherProps}>
+      {children}
+    </Button>
+  );
+};
+
+export default LinkButton;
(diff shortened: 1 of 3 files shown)


src/components/VaccinationBlock/LinkButton.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { Button, ChevronRightIcon } from './VaccinationBlock.style';
+
+type ButtonProps = React.ComponentProps<typeof Button>;
+
+const LinkButton: React.FC<ButtonProps> = ({
```
pnavarrc, line 6: This is kind of new, I didn't know that you can grab the type of the component that you are importing and then use it. This is handy because then the `LinkButton` can receive the same props than `Button` (which in turn, can receive the same props than `MuiButton`)

chasulin: ooo interesting

src/components/VaccinationBlock/LinkButton.tsx
```diff
@@ -0,0 +1,18 @@
...
+  href,
+  children,
+  ...otherProps
+}) => {
+  return (
+    <Button href={href} endIcon={<ChevronRightIcon />} {...otherProps}>
```
chasulin, line 12: ooooo `endIcon`! fun

src/components/VaccinationBlock/VaccinationBlock.tsx
```diff
@@ -56,17 +61,29 @@ const VaccinationBlock: React.FC<{ region: Region }> = ({ region }) => {
 const VaccinationLinksBlock: React.FC<{
   title: string;
   links: VaccinationLink[];
-}> = ({ title, links }) => (
+  trackingLinkPrefix?: string;
```
chasulin, line 64: do we need the `?` ? seems like a prefix is always declared

chasulin (approved):

PR #2547, opened 2021-01-27
vaccinations: implement vaccination block for states

pnavarrc:
[Mocks](https://www.figma.com/file/8BG3s3peEdA2LsaHsnZtm3/CAN-Vaccinate?node-id=1%3A283)

Note: We are not adding the second line on the see options link on this first version, Material UI buttons can't take a second line.

**Desktop**
image

**Mobile**


src/components/RegionVaccinationBlock/RegionVaccinationBlock.tsx
```diff
@@ -0,0 +1,16 @@
...
+import StateVaccinationBlock from './StateVaccinationBlock';
+import { Region, County, State, MetroArea } from 'common/regions';
+
+const RegionVaccinationBox: React.FC<{ region: Region }> = ({ region }) => {
+  if (region instanceof County || region instanceof MetroArea) {
+    return <VaccinationBlock region={region} />;
+  } else if (region instanceof State) {
+    return <StateVaccinationBlock region={region} />;
+  } else {
+    return null;
```
ghop02, line 12: thoughts on failing so it's more obvious if we add another location type?

src/components/RegionVaccinationBlock/StateVaccinationBlock.tsx
```diff
@@ -0,0 +1,46 @@
...
+        get on a waitlist.
+      </Paragraph>
+      <Grid container spacing={1}>
+        <Grid item xs={12} md={6} key="eligibility">
+          <StateLinkButton
+            href=""
+            target="_blank"
+            rel="noopener noreferrer"
+            onClick={() =>
+              trackVaccinationLink(`Eligibiligy: ${region.fullName}`)
```
ghop02: typo - also is there a way to include both values? I guess we probably don't have that level of specificity, but would be nice if we were able to have the hierarchy of `[vaccination, click_link, eligibility, region_name]`.  Just curious how we're planning on aggregating this event

src/components/RegionVaccinationBlock/StateVaccinationBlock.tsx
```diff
@@ -0,0 +1,60 @@
...
+import FeedbackBox from './FeedbackBox';
+import { StateLinkButton } from './StateVaccionationBlock.style';
+import { trackVaccinationLink } from './utils';
+
+const StateVaccinationBlock: React.FC<{ region: Region }> = ({ region }) => {
+  // TODO: Update with the eligibility and options links
+  const eligibilityUrl = '';
+  const optionsUrl = '';
+
+  if (!eligibilityUrl || !optionsUrl) {
```
ghop02, line 14: Not having the options url is potentially expected right?

ghop02 (approved):

src/components/RegionVaccinationBlock/StateVaccinationBlock.tsx
pnavarrc: No, GA only has category, action and label

src/components/RegionVaccinationBlock/StateVaccinationBlock.tsx
pnavarrc: right, I meant `&&` 😅

PR #2552, opened 2021-01-27
Fixes markdown links, pt. 2

chasulin:
#2539 added an `isInternalUrlWithCan` util, which broke markdown links that include covidactnow.org but aren't actually internal links (ie. apidocs.covidactnow.org, blog.covidactnow.org...). This PR fixes


ghop02 (approved):
🙏 thank you looks good!

PR #2554, opened 2021-01-27
Recommendations update: remove hghi as source

chasulin:
Removes hghi/school related content from inline recommendations content + recommendations sources and methodology modal


pnavarrc (approved):
Looks good! Let's also remove it from the CMS when we have time !

PR #2557, opened 2021-01-27
workaround to fix useEffect bug

ghop02:


chasulin (commented):
how about moving these hooks to their own files in `common/hooks`?

PR #2558, opened 2021-01-27
Increase timeout in scrollWithTimeout

chasulin:
The timeout needed a bump from 350->400 in order for the scroll to work properly when page's content isn't yet lazy-loaded


pnavarrc (approved):
Looks good!

PR #2559, opened 2021-01-27
Vaccinations in bundle

ghop02:
This moves the vaccination data into the bundle and simplifies accessors


src/cms-content/vaccines/index.ts
```diff
@@ -14,53 +16,17 @@ export interface RegionTypeVaccinationInfo {
...
 }
 
-const fetchRegionTypeVaccineData = (regionType: RegionType) => {
-  if (regionType !== RegionType.STATE) {
-    return;
-  }
-  return fetch(`/cms-content/vaccines/${regionType}.json`).then(
-    res => res.json() as Promise<RegionTypeVaccinationInfo>,
-  );
+const VACCINATION_DATA_BY_FIPS: { [fips: string]: RegionVaccinationInfo } = {
```
ghop02, line 19: This is more complicated than we need right now, but planning on adding more blocks like this for metro and county data as well

pnavarrc: I think you can use `fromPairs` directly, no need to wrap it with an object and destructuring

```tsx
const VACCINATION_DATA_BY_FIPS: Type = fromPairs(/* ... */);
```

pnavarrc (approved):
Perfect! 💯

PR #2561, opened 2021-01-27
cms: allows adding embeds in the markdown blocks

pnavarrc:
This PR enables CMS users to add our embeds to markdown content inline by adding the URL of the embed as a link on the markdown content.

I also fixed the embed a bit because it wasn't showing the borders correctly, however, this fix will be noticeable only once we deploy this to production (we can test it by replacing the url with the local embed URL (`http://localhost:3000/embed/us/fips/19100`, for example)

### DIFF
--- a/src/components/Markdown/MarkdownLink.tsx
+++ b/src/components/Markdown/MarkdownLink.tsx
@@ -1,7 +1,12 @@
 import React, { Fragment } from 'react';
 import { HashLink } from 'react-router-hash-link';
 import { scrollWithOffset } from 'components/TableOfContents';
-import { isValidURL, isInternalLink, formatInternalLink } from './utils';
+import {
+  isValidURL,
+  isInternalLink,
+  isInternalEmbed,
+  formatInternalLink,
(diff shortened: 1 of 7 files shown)


src/components/Markdown/MarkdownLink.tsx
```diff
@@ -37,8 +43,23 @@ const MarkdownLink: React.FC<{
...
 
-  // HACK to track clicks on the exposure notification recommendation
+  if (isInternalEmbed(href)) {
+    return (
+      <CenterEmbed>
+        <iframe
+          src={href}
+          title="Covid Act Now"
+          width="400"
+          height="440"
```
chasulin, line 53: these are the dimensions of the homepage embed. do we need to have different iframes/dimensions depending on homepage vs location page embed?

chasulin: actually, i guess just using the larger of the sets of dimensions should work for both

chasulin (approved):

PR #2566, opened 2021-01-27
New searchbar component

chasulin:
Still a wip, but would like to merge it in as is for a preliminary preview link


src/components/Search/Homepage/StyledPopper.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { Popper } from '@material-ui/core';
```
pnavarrc, line 2: nit: It's better to import the component with the full path, that reduces the bundle size

```tsx
import Popper from '@material-ui/core/Popper;
```

src/assets/theme/customStyleBlocks/getSearchAutocompleteStyles.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Autocomplete styles specific to homepage searchbar (removes gray hover from dropdown option items)
+ */
+import { makeStyles } from '@material-ui/core/styles';
+import { COLOR_MAP } from 'common/colors';
+
+export const getSearchAutocompleteStyles = makeStyles(() => ({
+  option: {
+    '&[data-focus="true"]': {
+      backgroundColor: 'transparent',
+      color: `${COLOR_MAP.BLUE}`,
```
pnavarrc, line 11: nit: I think you don't need the string template here (`COLOR_MAP.BLUE` is already a string)

pnavarrc (approved):
Looks good!

PR #2568, opened 2021-01-27
Add first scrape of CA county data

ghop02:
This is the output of a scrape from CA Vaccines and adds in all CA counties to the CMS


ghop02:
Was merged in with vaccine info feature branch

PR #2569, opened 2021-01-28
update the vaccination copy and feedback link

pnavarrc:
Update the copy of the vaccination block to match the new copy (see the [Mocks](https://www.figma.com/file/8BG3s3peEdA2LsaHsnZtm3/CAN-Vaccinate?node-id=0%3A1) under "new"). I also updated the feedback form using the link from the [Phase 2 paper doc](https://www.dropbox.com/scl/fi/3iq9uzrtvmjpdi71qd6wq/-Phase-2-Launch-Plan.paper?dl=0&rlkey=gou6giiggng27u8m7eboalc53)

### DIFF
--- a/src/components/RegionVaccinationBlock/FeedbackBox.tsx
+++ b/src/components/RegionVaccinationBlock/FeedbackBox.tsx
@@ -20,7 +20,7 @@ function trackFeedbackClick() {
   trackEvent(
     EventCategory.VACCINATION,
     EventAction.CLICK_LINK,
-    'Suggest and improvement',
+    'Suggest an improvement',
   );
 }
 
--- a/src/components/RegionVaccinationBlock/FeedbackBox.tsx
+++ b/src/components/RegionVaccinationBlock/FeedbackBox.tsx
@@ -1,13 +1,27 @@
 import React from 'react';
 import { FeedbackBox as BoxContainer } from './RegionVaccinationBlock.style';
+import ExternalLink from 'components/ExternalLink';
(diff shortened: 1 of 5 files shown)


src/components/RegionVaccinationBlock/utils.ts
```diff
@@ -39,3 +39,8 @@ export const getVaccinationRegions = (region: Region): Region[] => {
 export function trackVaccinationLink(label: string) {
   trackEvent(EventCategory.VACCINATION, EventAction.CLICK_LINK, label);
 }
+
+export const title = 'Getting vaccinated';
+
+export const introText =
+  'Vaccination procedures vary by location. Below are official resources on when and how you can get vaccinated.';
```
ghop02, line 46: Was there a reason you put this in `utils` as opposed to component?

src/components/RegionVaccinationBlock/FeedbackBox.tsx
```diff
@@ -1,13 +1,27 @@
...
+      </ExternalLink>
     </BoxContainer>
   );
 };
 
+function trackFeedbackClick() {
+  trackEvent(
+    EventCategory.VACCINATION,
+    EventAction.CLICK_LINK,
+    'Suggest and improvement',
```
ghop02, line 23: *suggest an improvement

ghop02 (approved):
Looks good! couple small comments

src/components/RegionVaccinationBlock/utils.ts
pnavarrc: Maybe a component is better, updating!

PR #2577, opened 2021-01-28
vaccination block: remove special case for states

pnavarrc:
This PR removes the special case for states. I'm keeping the component just in case, but will remove and rename after we launch.

### DIFF
--- a/src/components/RegionVaccinationBlock/RegionVaccinationBlock.tsx
+++ b/src/components/RegionVaccinationBlock/RegionVaccinationBlock.tsx
@@ -1,20 +1,3 @@
-import React from 'react';
 import VaccinationBlock from './VaccinationBlock';
-import StateVaccinationBlock from './StateVaccinationBlock';
-import { fail } from 'common/utils';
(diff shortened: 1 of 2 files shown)


src/components/RegionVaccinationBlock/RegionVaccinationBlock.tsx
```diff
@@ -1,20 +1,3 @@
...
-  } else {
-    fail(
-      `RegionVaccinationBox: missing implementation for region with fips ${region.fipsCode}`,
-    );
-    return null;
-  }
-};
-
-export default RegionVaccinationBox;
+export default VaccinationBlock;
```
ghop02, line 3: We don't need this anymore right?

ghop02 (approved):

src/components/RegionVaccinationBlock/RegionVaccinationBlock.tsx
pnavarrc: We don't, but I want to keep it until we launch just in case the decision is reversed - I will clean up early next week I promise!

PR #2579, opened 2021-01-28
Add useScrollToTerm function

faihegberg:
Currently, when the user navigates to a term on the glossary page _using a new tab/window_, the page doesn't scroll to the term. This PR fixes the scroll location so even when the user navigates to a term in a new tab/window, the page will automatically scroll to the term's location. Some points to consider:
- Perhaps the `useScrollToTerm` function can be placed in a common location so it can be reused elsewhere. Is `common/hooks` an appropriate place? (If this function is moved, it will be renamed accordingly).
- Glossary term is moved out of the `return` clause to acknowledge them as their own objects. @pnavarrc suggests to have the scroll-to-term functionality within each term object (correct me if I misunderstand though!), although I'm not sure whether we should be mixing an object with page navigation functionalities, and if a scroll-to-object method exists then one can just call it at appropriate points in the page?

### DIFF
--- a/src/screens/Learn/Glossary/Glossary.tsx
+++ b/src/screens/Learn/Glossary/Glossary.tsx
@@ -14,7 +14,7 @@ import { learnPages } from 'cms-content/learn';
 import Breadcrumbs from 'components/Breadcrumbs';
 import { formatMetatagDate, formatNumericalDate } from 'common/utils';
 import ScrollToTopButton from 'components/SharedComponents/ScrollToTopButton';
-import { useScrollToTopButton } from 'common/hooks';
+import { useScrollToElement, useScrollToTopButton } from 'common/hooks';
 import Footer from 'screens/Learn/Footer/Footer';
 import ExternalLink from 'components/ExternalLink';
 
@@ -30,6 +30,15 @@ function getGlossaryFooter(): React.ReactElement {
     </Fragment>
   );
 }
+
+const GlossaryTerm: React.FC<{ term: Term }> = ({ term }) => {
+  return (
+    <Fragment key={`glossary-term-${term.termId}`}>
+      <SectionName id={term.termId}>{term.term}</SectionName>
+      <MarkdownContent source={term.definition} />
+    </Fragment>
+  );
+};
 
 const Glossary: React.FC = () => {
   const {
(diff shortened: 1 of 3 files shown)


src/screens/Learn/Glossary/Glossary.tsx
```diff
@@ -31,6 +32,27 @@ function getGlossaryFooter(): React.ReactElement {
   );
 }
 
+function useScrollToTerm(termToScrollTo: string) {
+  // Only scroll to the term when the page first loads.
+  const [scroll, setScroll] = useState(false);
```
chasulin, line 37: i think we can nix this state variable

src/screens/Learn/Glossary/Glossary.tsx
```diff
@@ -31,6 +32,27 @@ function getGlossaryFooter(): React.ReactElement {
...
+  useEffect(() => {
+    const element = document.querySelector(termToScrollTo);
+    if (element && !scroll) {
+      setScroll(true);
+      element.scrollIntoView();
```
chasulin, line 42: we need to account for the height of the navbar when scrolling to a location on the page. right now, the page scrolls to the correct glossary term, but part of the content is covered by the navbar (image below, you'll notice that the term header **Presymptomatic spread*** isn't visible). replacing `element.scrollIntoView()` with [scrollWithOffset](https://github.com/covid-projections/covid-projections/blob/develop/src/components/TableOfContents/utils.ts) would solve for that

Screen Shot 2021-01-31 at 9 06 05 PM

chasulin (changes requested):
nice!! this definitely moves towards a fix. a few things:
- this breaks when there is no `#` in the url, ie. just [/glossary](- https://covid-projections-git-faihegberg-fix-scroll-on-glossary.covidactnow.now.sh/glossary). in this case, `useLocation().hash` is resolving to an empty string (and so you end up calling `document.querySelector()` on an empty string). maybe lets move the useLocation() call into the useScrollToTerm to consolidate everything in the hook, and add a check for `termToScrollTo` there:
```
const termToScrollTo = useLocation().hash
const element = termToScrollTo ? document.querySelector(termToScrollTo) : null
if (element) {
   ........
 }
```

lets also move this into its own file in`common/hooks` and use it throughout learn!

chasulin (approved):

PR #2580, opened 2021-01-28
Add loading state to useGeolocation

chasulin:
Adds a loading state to be used in new homepage UI

### DIFF
--- a/src/common/hooks/useGeolocation.ts
+++ b/src/common/hooks/useGeolocation.ts
@@ -41,7 +41,7 @@ export default function useGeolocation(): UseGeolocationReturn {
         .finally(() => setIsLoading(false));
     };
     fetchIpData();
-  }, []);
+  }, [setIpData, setIsLoading]);
 
   const geolocationReturnObj: UseGeolocationReturn = {
     geolocationData: ipData,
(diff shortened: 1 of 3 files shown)


src/common/hooks/useGeolocation.ts
```diff
@@ -30,10 +37,16 @@ export default function useGeolocation(): GeolocationInfo | undefined {
             country: data.country,
           }),
         )
-        .catch(e => console.error(e));
+        .catch(e => console.error(e))
+        .finally(() => setIsLoading(false));
     };
     fetchIpData();
   }, []);
```
pnavarrc, line 44: `useEffect` need to have `setIpData` and `setIsLoading` as dependencies, also, we probably don't want to fetch the IP data if we have it already. If we don't add the dependencies the code might run more often that cause issues that are hard to debug

```ts
useEffect(() => {
  const fetchIpData = /* ... */;
  if (ipData) {
    fetchIpData();
  }
}, [setIpData, ipData, setIsLoading]);
```

pnavarrc (commented):
Added a comment, can test after it's fixed!

pnavarrc (approved):

PR #2593, opened 2021-01-29
Homepage updates

chasulin:
New above-the-fold UI on the homepage (and A/B experiment)! [Mocks](https://www.figma.com/file/JGqemgZwlG9e5DQbZpdKgy/Homepage-Above-the-fold?node-id=10%3A4741)


src/App.js
```diff
@@ -60,7 +62,9 @@ export default function App() {
             <ErrorBoundary>
               <Suspense fallback={<SuspenseFallback />}>
                 <Switch>
-                  <Route exact path="/" component={HomePage} />
+                  {/* <Route exact path="/" component={HomePage} /> */}
```
pnavarrc, line 65: Consider removing the comment here, once the experiment is over we will probably just rename the component

src/components/RegionItem/RegionItem.tsx
```diff
@@ -26,12 +27,20 @@ const RegionItem: React.FC<{ region: Region }> = ({ region }) => {
   return (
     <StyledLink to={region.relativeUrl}>
       <Wrapper>
-        <CircleIcon $iconColor={iconColor} />
+        <IconContainer>
```
pnavarrc, line 30: I forgot to add tracking for clicks on the home page items, can you add some tracking?

src/screens/HomePage/New/NewHomepage.tsx
```diff
@@ -0,0 +1,143 @@
...
+    <>
+      <EnsureSharingIdInUrl />
+      <AppMetaTags
+        canonicalUrl="/"
+        pageTitle="Realtime U.S. COVID Map & Vaccine Tracker"
+        pageDescription={getPageDescription()}
+      />
+      <HomepageStructuredData />
+      <VaccinationsBanner />
+      <NewHomePageHeader />
```
pnavarrc, line 91: We can do this post-launch, but it might be better to only create versions for the part of the homepage that is different to avoid code duplication (if we had to change something in the home page, we will have to do it twice until the experiment is over)

pnavarrc (approved):
Looks good, added some comments!

PR #2594, opened 2021-01-29
Homepage A/B test setup

chasulin:


pnavarrc (approved):
Looks good! Maybe we can remove the elements in common from both home pages so we don't have to duplicate changes in the coming weeks, but we can punt that after shipping!

PR #2601, opened 2021-01-29
Fix lint error

chasulin:


ghop02 (approved):

src/common/hooks/useGeolocation.ts
```diff
@@ -59,31 +59,30 @@ function fetchGeolocationData() {
...
-    isp: 'Google LLC',
-    org: 'Google Public DNS',
-    as: 'AS15169 Google LLC',
-    asname: 'GOOGLE',
-    mobile: false,
-    proxy: false,
-    hosting: true,
-  });
-}
+// function mockGeolocationData() {
```
ghop02, line 62: would be nice to add a quck comment about why this is commented out when you get the chance!

chasulin: theres a comment at the top of the page: 

> Note: geolocation does not work locally. To test locally, replace fetchGeolocationData() with mockGeolocationData() in the effect.

PR #2613, opened 2021-02-02
Ship new homepage to 100% of users

chasulin:


ghop02 (approved):
Are we committing to this homepage? If so would be good to remove the old code/remove new from the name.  If not, we should probably track and clean up when we get the chance.

Also will this let us load google optimize asynchronously? I know that it's a fairly heavy initial load

pnavarrc (approved):
Looks good!

src/App.js
```diff
@@ -26,8 +26,8 @@ import PageviewTracker, {
   EventCategory,
 } from 'components/Analytics';
 import { COVID_RESPONSE_SIMULATOR_URL } from 'screens/Tools/Tools';
+import NewHomepage from 'screens/HomePage/New/NewHomepage';
```
schamp, line 29: should we rename `NewHomepage` to just `Homepage`, to eliminate potential confusion? (ie., if this is the new one, what is the other one?)

chasulin: yep, this is just a quick PR to end the a/b experiment and ship the new homepage to all traffic. going to clean up + rename in a separate pr

PR #2623, opened 2021-02-02
Fix amount of locations shown in compare modal

chasulin:
[Trello ticket](https://trello.com/c/f5edT7a9/793-4-concerns-that-state-level-compare-charts-only-show-100-counties-when-expanded-even-though-the-copy-says-view-all-counties-in-s)

To test: 

[On prod](https://covidactnow.org/us/missouri-mo/county/jackson_county/?s=1568508), clicking 'View all counties in Missouri' below the compare table only shows the first 100 counties (rather than all 115 in the state)

[With fix](http://localhost:3000/us/missouri-mo/county/jackson_county/?s=1568508)


pnavarrc (approved):
Looks good!

PR #2624, opened 2021-02-02
Ghop02/vaccination alert fields

ghop02:


public/admin/config.yml
```diff
@@ -726,3 +726,59 @@ collections:
...
+                widget: boolean
+                required: false
+              - label: Eligibility Information URL
+                name: eligibilityInfoUrl
+                widget: string
+              - label: Email alert version
+                name: emailAlertVersion
+                default: 0
+                widget: number
+              - label: Phases
```
igorkofman, line 760: estimated population

igorkofman (approved):

public/admin/config.yml
ghop02: Do you mean estimated size of people that would be in a phase?  Is that easy to get?

public/admin/config.yml
```diff
@@ -726,3 +726,59 @@ collections:
...
+                fields:
+                  - label: Group Phase
+                    name: groupPhase
+                    widget: string
+                  - label: Group Tier
+                    name: groupTier
+                    widget: string
+                  - label: Group Description
+                    name: groupDescription
+                    widget: text
```
pnavarrc, line 773: This should probably be Markdown

public/admin/config.yml
```diff
@@ -726,3 +726,59 @@ collections:
...
+                name: emailAlertVersion
+                default: 0
+                widget: number
+              - label: Phases
+                name: eligibilityPhases
+                widget: list
+                collapsed: true
+                fields:
+                  - label: Group Phase
+                    name: groupPhase
```
pnavarrc, line 766: Consider adding descriptions for these fields, or name them in a way that makes it easier to understand what they are (phase title, for example)

public/admin/config.yml
pnavarrc: The designs had an estimated population size for each phase, we could have this as an optional field for states that provide this info

public/admin/config.yml
```diff
@@ -726,3 +726,59 @@ collections:
...
+                  - label: Group Description
+                    name: groupDescription
+                    widget: text
+                  - label: Expanded Definition URL
+                    name: expandedDefinitionUrl
+                    required: false
+                    widget: string
+                  - label: Eligibility Start Date
+                    name: eligibilityStartDate
+                    widget: date
```
pnavarrc, line 780: Most future phases don't have a defined date, or have it in a guesstimate form "May-June" instead of firm dates, consider changing the widget type to text. An alternative could be to keep it as date, and render as "to be determined" if the date is not set, or to always render as Month only if the data is in the future

public/admin/config.yml
```diff
@@ -726,3 +726,59 @@ collections:
...
+                  - label: Updated at
+                    name: updatedAt
```
pnavarrc, line 782: This should probably be date only

public/admin/config.yml
```diff
@@ -726,3 +726,59 @@ collections:
...
+                required: false
+              - label: Eligibility Information URL
+                name: eligibilityInfoUrl
+                widget: string
+              - label: Email alert version
+                name: emailAlertVersion
+                default: 0
+                widget: number
+              - label: Phases
+                name: eligibilityPhases
```
pnavarrc, line 761: I would call this `vaccinationPhases`

src/cms-content/vaccines/state-vaccine-phases.json
```diff
@@ -0,0 +1,20 @@
+{
+  "regions": [
+    {
+      "emailAlertVersion": 0,
+      "eligibilityInfoUrl": "https://covidvaccine.mo.gov/",
+      "fips": "29",
+      "locationName": "Missouri",
+      "eligibilityPhases": [
+        {
+          "eligibilityStartDate": "2020-12-14T18:54:06.354Z",
```
pnavarrc, line 10: nit: I think that `vaccinationPhases` and `startDate` are better names, but feel free to ignore/punt

public/admin/config.yml
ghop02: why? seems like the time could be helpful.  this feels like an internal field

public/admin/config.yml
ghop02: Yeah I was treating this as a day that it actually starts - so if it wasn't a specific date don't use it, but maybe that's right that it should be text.  can always start there

public/admin/config.yml
ghop02: how about just phases?

PR #2627, opened 2021-02-02
implement vaccination alerts workflow (part 1)

pnavarrc:
This PR implements an outline of the vaccination alerts workflow (mostly in the readme) and a script to compare information from the CMS with information from Firebase.


scripts/vaccination_alerts/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    if (
+      !lastVersionInfo ||
+      lastVersionInfo.emailAlertVersion !== currentPhaseInfo.emailAlertVersion
+    ) {
+      updatedInfoByFips[fipsCode] = currentPhaseInfo;
+    }
+  }
+  return updatedInfoByFips;
+}
```
pnavarrc, line 53: I wanted to add unit tests for this function, but CRA only runs tests under `src` 😢

src/cms-content/vaccines/phases.ts
```diff
@@ -0,0 +1,21 @@
...
+export interface RegionVaccinePhaseInfo {
+  emailAlertVersion: number;
+  eligibilityInfoUrl: string;
+  locationName: string;
+  fips: string;
+  phaseGroups: RegionPhaseGroup[];
+}
+
+export const stateVaccinationPhases: RegionVaccinePhaseInfo[] =
+  stateVaccinationInfo.regions;
```
pnavarrc, line 21: This file is not strictly necessary but it's useful to have types for this JSON

scripts/vaccination_alerts/README.md
```diff
@@ -0,0 +1,93 @@
...
+vaccination-alerts:
+  '1500':
+    locations:
+      '11':
+        emails:
+          pablo@covidactnow.org: { sentAt: null }
+          chelsi@covidactnow.org: { sentAt: null }
+      '13':
+        emails:
+          chelsi@covidactnow.org: { sentAt: null }
```
mikelehen, line 75: Hrm...  Keying off of the `snapshotId` made sense for risk alerts, since it uniquely identifies the data that we are alerting off of...  and in particular, this structure allowed us to make the alert sending process resumable / idempotent so that if the alert sending failed halfway through for snapshot 1500, you could just run it again and it would ignore everybody we already emailed (i.e. non-null `sentAt`) and resume sending to everybody else (with null `sentAt`).

To get a similar behavior, I think maybe we'd want a structure something like:
```yaml
vaccination-alerts:
  '11':
    emailAlertVersions:
      '0':
        emails:
          pablo@covidactnow.org: { sentAt: 'yyyy/mm/dd ...' }
          chelsi@covidactnow.org: { sentAt: 'yyyy/mm/dd ...' }
      '3':
        emails:
          pablo@covidactnow.org: { sentAt: 'yyyy/mm/dd ...' }
          chelsi@covidactnow.org: { sentAt: null }
```

This would indicate that we previously successfully sent alerts for location 11 for version 0, but sending alerts for version 3 failed partway through and so you got the email but chelsi didn't.  When we re-ran the script, it should automatically send it to Chelsi but not you.

Feel free to ping me if this doesn't make sense or you want to talk through in more detail.

scripts/vaccination_alerts/README.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+We update the `info` collection with the latest updated date.
```
mikelehen, line 78: Not 100% sure what this refers to.

scripts/vaccination_alerts/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+## 3. email users
+
+We fetch the information for the current snapshot in `vaccination-alerts` to generate pairs of emails and locations
```
mikelehen, line 82: Again, this isn't really snapshot-based.

scripts/vaccination_alerts/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+export interface RegionVaccineVersion {
+  emailAlertVersion: number;
+}
+
+export interface RegionVaccineVersionMap {
+  [fipsCode: string]: RegionVaccineVersion;
+}
+
+export function getCmsVaccinationInfo(): RegionVaccinePhaseInfoMap {
+  return _.chain(stateVaccinationPhases)
```
mikelehen, line 20: Is there a reason this is state-specific and doesn't include counties?

scripts/vaccination_alerts/utils.ts
mikelehen: Oh, hrm.  Maybe we should consider moving the scripts dir or something.

.gitignore
```diff
@@ -46,3 +46,4 @@ package-lock.json
 
 # a scratch file for lil scripts/notes
 scratch
+scripts/vaccination_alerts/vaccination-alerts.json
```
mikelehen, line 49: FWIW- We have a scripts/.gitignore and scripts/alert_emails/.gitignore file.  Not sure what the best practice is for having multiple .gitignore files, but we should maybe consider consolidating into the root .gitignore if we're not a fan (not in this PR). 😁

mikelehen (approved):
Thank you for the amazing README.md file.  That was a really good idea and made it easy to understand where you're going and provide feedback.  In general I think this is very much in the right direction. I suggested a bit of a schema tweak mostly to try to make sure we can support resumability in case we ever have failures in the script.

scripts/vaccination_alerts/utils.ts
ghop02: scripts does feel a bit like the wild wild west - I'm not sure if we run type checks on it automatically and sometimes imports are not found cleanly.  Maybe moving scripts into src would fix taht?

scripts/vaccination_alerts/README.md
```diff
@@ -0,0 +1,93 @@
...
+  content: 'Phase A...'
+'12':
+  emailAlertVersion: 0
+  content: 'Phase 1...'
+'13':
+  emailAlertVersion: 3 # New
+  content: 'Phase 1...'
+```
+
+In this case, for example, we need to send emails for udpates on location "11" and location "13" (location "12" didn't have any updates).
```
ghop02, line 38: *updates

scripts/vaccination_alerts/utils.ts
mikelehen: FWIW- Right now we have a separate tsconfig.json that specifies to generate CommonJS module code, in order for it to work with node.js... https://github.com/covid-projections/covid-projections/blob/3cfc1d8cc2e39c88224da919557f845df04581d1/scripts/tsconfig.json#L4

This will probably be an obstacle to moving the code into `src/` but maybe there's a solution.

I think the existence of the extra tsconfig.json file is also why you have to specify a relative path to imports instead of being able to base it off of `src/` (which I'm guessing is what you meant about clean imports).

.gitignore
pnavarrc: I don't have strong opinions one way or the other, but since we have a `.gitignore` in the scripts directory I'm going to move this there

scripts/vaccination_alerts/README.md
ghop02: alternatively, we could have:

```
vaccination-alerts:
  '11':
        emails:
          pablo@covidactnow.org: { sentAt: 'yyyy/mm/dd ...', lastAlertVersion: '0' }
          chelsi@covidactnow.org: { sentAt: 'yyyy/mm/dd ...', lastAlertVersion: '3' }
```
or

```
vaccination-alerts:
  '11':
        emails:
          pablo@covidactnow.org: 
            alertVersion:
              0: { sentAt: 'yyyy/mm/dd ...' }
              3: { sentAt: 'yyyy/mm/dd ...' }
          chelsi@covidactnow.org: 
            alertVersion:
              0: { sentAt: 'yyyy/mm/dd ...' }
```

scripts/vaccination_alerts/utils.ts
pnavarrc: We only have state information so far, but if we do get state information, we can import it here and export just a map with all the vaccination information

scripts/vaccination_alerts/README.md
pnavarrc: I was thinking of storing the latest email alert version for each location in the `info` collection (see https://github.com/covid-projections/covid-projections/pull/2627/files#diff-582f325175286997a5df8ad7cdca7ff4cac8d9f6a17b3100ac4d4ee4169c2278R15), but maybe we can get the latest version from the `vaccination-alerts` collection you proposed here? https://github.com/covid-projections/covid-projections/pull/2627/files#r569154259

mikelehen: Ah, cool.  That makes sense.  I was mostly thrown off by "latest updated date" but maybe that was supposed to be "data"?

Also, I think I wouldn't update it until after we've actually emailed users, else I think we lose resumability.

scripts/vaccination_alerts/README.md
pnavarrc: I'm going with Michael's recommendation, it feels easier to query if we want stats on the emails

PR #2628, opened 2021-02-02
Get rid of div grouping metro name and state

faihegberg:
This PR adds wrappers to the compare table on the homepage to enable more specific styling of location cells. Within each first column cell of the table, this also enables the location name text to roll up to the first line and be immediately next to the rank element.

### DIFF
--- a/src/components/Compare/Compare.style.tsx
+++ b/src/components/Compare/Compare.style.tsx
@@ -6,6 +6,7 @@ import { COLORS } from 'common';
 import { Metric } from 'common/metric';
 import { Level } from 'common/level';
 import { ChartLocationName } from 'components/LocationPage/ChartsHolder.style';
+import { Link } from 'react-router-dom';
 
 // TODO (Chelsi): consolidate into a theme:
 
@@ -257,6 +258,10 @@ export const TableHeadContainer = styled(TableHead)<{ $isModal?: boolean }>`
   }
 `;
 
+export const StyledLink = styled(Link)`
+  text-decoration: none;
+`;
+
 export const Population = styled.span`
   font-family: Source Code Pro;
   font-size: 0.875rem;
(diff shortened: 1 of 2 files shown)


faihegberg:
[Fix wrapping for metro names on the homepage. Consider truncating](https://trello.com/c/VCMvZSET/770-fix-wrapping-for-metro-names-on-the-homepage-consider-truncating)

pnavarrc:
The styling is not quite what we need (the locations shouldn't be underlined, and we want to keep the spacing around the dot), let's sync tomorrow!

src/components/Compare/CompareTableRow.tsx
```diff
@@ -77,23 +80,32 @@ const CompareTableRow = (props: {
...
-            <br />
-            <LocationInfoWrapper>
-              <Population>
-                {formatEstimate(location.region.population, populationRoundTo)}
-              </Population>
-            </LocationInfoWrapper>
-          </div>
+          <LocationCellWrapper>
+            <LocationRankWrapper>
+              <div>
```
chasulin, line 85: we don't need this inner div anymore now that there's a wrapper

src/components/Compare/CompareTableRow.tsx
```diff
@@ -77,23 +80,32 @@ const CompareTableRow = (props: {
...
+                <Rank>{location.rank}</Rank>
+                <FiberManualRecordIcon />
+              </div>
+            </LocationRankWrapper>
+            <LocationNameWrapper>
+              <div>
```
chasulin, line 91: same as above ^

src/components/Compare/CompareTableRow.tsx
faihegberg: Oh right, forgot to take those out!

src/components/Compare/Compare.style.tsx
```diff
@@ -272,6 +273,25 @@ export const Rank = styled.span`
...
+  flex: 0 0 auto;
+`;
+
+export const LocationNameWrapper = styled.div`
+  flex: 1 1 auto;
+  white-space: normal;
+  text-overflow: initial;
+`;
+
+export const LocationRankLink = styled(Link)`
```
chasulin, line 291: I'd just call this `StyledLink` since its wrapping the whole row and not just the rank

chasulin (approved):
Looks good!

PR #2630, opened 2021-02-03
Change alert emails to workflow_dispatch and auto-detect main snapshot.

mikelehen:
Note I think I can't actually test the manual workflow dispatch until this is merged, but I tested the "auto" snapshot detection...


scripts/alert_emails/utils.ts
```diff
@@ -94,3 +95,17 @@ export function readAlerts(filePath: string): { [fips: string]: Alert } {
...
 }
+
+export async function resolveSnapshot(snapshotId: string): Promise<number> {
+  if (snapshotId === 'auto') {
+    const snap = await fetchMainSnapshotNumber();
+    console.log(`Resolved 'auto' snapshot to ${snap}.`);
+    return snap;
+  } else {
+    const snap = parseInt(snapshotId);
+    if (Number.isNaN(snap)) {
```
pnavarrc, line 106: nit: Maybe using `Number.isFinite` would be better? `Number.isFinite` returns false for `null`, `undefined`, `NaN`, `+Infinite` and - `-Infinite` + any non-numeric value.

Also, consider passing the radix explicitly to `parseInt` since the default radix changes depending on the value of `snapshotId` (it's very unlikely that this would be an issue for us).

pnavarrc (approved):
Looks good! 🙌

PR #2634, opened 2021-02-03
derive count of vaccinated people from metric ratio

TomGoBravo:
change Projection.getVaccinationsInfo to derive count of people from metric ratio and population if not in the API.

## Tested

Changed src/assets/data/data_url.json to snapshot 1571 (or merge `develop`, the latest uses snapshot 1573 which has TN data) and went to /us/tennessee-tn/county/haywood_county/chart/6 and checked that vaccines are now displayed.


TomGoBravo:
[ETL TN County Vaccine Data into API](https://trello.com/c/PgOjusc7/846-etl-tn-county-vaccine-data-into-api)

src/common/models/Projection.ts
```diff
@@ -397,6 +404,15 @@ export class Projection {
       return null;
     }
 
+    const peopleVaccinated = Projection.getActualOrDefault(
```
ghop02, line 407: I don't think you actually need this helper function!  pretty sure you can just do: 
`const peopleVaccinated = actuals.vaccinationsCompleted ?? (percentVaccinated / 100.0) * population`

TomGoBravo: Done in 6973ce2

src/common/models/Projection.ts
```diff
@@ -257,6 +257,7 @@ export class Projection {
       summaryWithTimeseries.actuals,
       metrics,
       metricsTimeseries,
+      summaryWithTimeseries.population,
```
mikelehen, line 260: This is actually stored in the class as `this.population` so you don't need to pass it in.

mikelehen (approved):
Amazing! One small cleanup, but this looks great. Thanks for tackling this!

src/common/models/Projection.ts
TomGoBravo: Thank you for the tip. Done in da48bf914

TomGoBravo:
I did a lousy job of testing this. https://covidactnow.org/us/tennessee-tn/county/shelby_county/chart/6?s=1574514 says "In Shelby County, Tennessee, 332 people (3.5%) have received the first shot". The county has population 940k so that count should be 33326.

mikelehen:
Ooops.  I tested it too and apparently did an equally lousy job! 😞

PR #2647, opened 2021-02-03
vaccination alerts: fetch vaccination info from Firestore

pnavarrc:
This PR updates the generate-vaccination-alerts script to fetch the vaccination info version from Firebase and move the `getFirebase` function to a common location since it will be reused for vaccine alerts.

### DIFF
--- a/scripts/vaccination_alerts/utils.ts
+++ b/scripts/vaccination_alerts/utils.ts
@@ -17,6 +17,8 @@ export interface RegionVaccineVersionMap {
   [fipsCode: string]: RegionVaccineVersion;
 }
 
+const VACCINATION_VERSIONS_COLLECTION = 'info/vaccinationInfoUpdates/locations';
+
 export function getCmsVaccinationInfo(): RegionVaccinePhaseInfoMap {
   return _.chain(stateVaccinationPhases)
     .map(stateInfo => [stateInfo.fips, stateInfo])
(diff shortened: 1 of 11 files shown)


scripts/vaccination_alerts/utils.ts
```diff
@@ -23,14 +24,13 @@ export function getCmsVaccinationInfo(): RegionVaccinePhaseInfoMap {
...
-  return {
-    // '29': {
-    //   emailAlertVersion: 0,
-    // },
-  };
+export async function getFirebaseVaccinationInfo(): Promise<
+  RegionVaccineVersionMap
+> {
+  const db = await getFirestore();
+  const versionDoc = await db.doc('info/vaccinationInfoUpdates').get();
```
mikelehen, line 31: Oh, hrm...  I didn't realize we'd store this as a single document.  I was imagining we'd have a collection with a single document for each fips code...  But storing it as a single document is probably easier and should be fine as long as it doesn't exceed 1MB, which it probably won't unless we start adding additional per-fips data to it...  So we can do this, at least for now!

mikelehen (approved):

scripts/vaccination_alerts/utils.ts
pnavarrc: Let's change it now then so we don't have to worry about that limitation, I will take than into account for the `vaccine-alerts` collection too 👍

pnavarrc: Updated the Firebase structure, let me know if it makes sense!

mikelehen (approved):
Awesome, yeah, this is what I had in mind.

If we wanted to flatten our hierarchy a bit we could collapse the `info/vaccinationInfoUpdates/locations` collection into just a top-level `vaccinationInfoUpdates` collection. I think `info` as a top-level collection isn't super useful.

Oh, also, it looks like we're mixing naming conventions a bit.  It looks like today we use `-` in our collection names (alerts-subscriptions, shared-component-params, etc.) and use camelCase for field names within documents.  Should we stick with that?

That would mean `vaccinationInfoUpdates` would be `vaccination-info-updates` (or maybe `vaccination-location-info` or something?) and also `emailAlertVersions` would be `email-alert-versions` ?

pnavarrc:
The `info` part was just reusing the collection that we have to store the last snapshot that we used for risk alerts, we can flatten the structure a bit.

PR #2650, opened 2021-02-04
Rm Links to Outdated Methodology Whitepaper

BrettBoval:


src/common/metrics/case_density.tsx
```diff
@@ -136,19 +136,7 @@ function renderDisclaimer(): React.ReactElement {
       <ExternalLink href="https://ethics.harvard.edu/files/center-for-ethics/files/key_metrics_and_indicators_v4.pdf">
         “Key Metrics for Covid Suppression”
       </ExternalLink>{' '}
-      by Harvard Global Health Institute and others.
-      <br />
-      When estimating the number of people who will become infected in the
-      course of a year, we rely on the{' '}
```
mikelehen, line 142: We should keep an eye on the inbox report for if we get questions about the estimated infections without this reference anymore.  I'd guess most people don't understand the difference between cases and infections.

src/common/metrics/hospitalizations.tsx
```diff
@@ -158,6 +158,7 @@ function renderDisclaimer(projections: Projections): React.ReactElement {
       <ExternalLink href="/covid-risk-levels-metrics#icu-capacity-used">
         ICU capacity methodology
       </ExternalLink>
+      .
```
mikelehen, line 161: If you were so inclined, we can probably drop the ICU headroom reference above now. 😁

mikelehen (approved):
❤️ ❤️ ❤️

src/common/metrics/case_density.tsx
BrettBoval: 👍

src/common/metrics/hospitalizations.tsx
BrettBoval: 👍

PR #2651, opened 2021-02-04
vaccination alerts: email signup form

pnavarrc:
work in progress

## Notes
- We might not add the vaccination badge on the chips for locations with vaccination information, we need to confirm that
- The form doesn't do anything on submit yet

**Mobile**

**Desktop** (the form expands to fit its parent container)
image


chasulin (approved):
LGTM

PR #2655, opened 2021-02-04
Fix imports to getFirestore().

mikelehen:


pnavarrc (approved):
Just in time for the alert emails!

PR #2657, opened 2021-02-04
Update AxisBottom

chasulin:
[Trello ticket](https://trello.com/c/a7pmgJAZ/791-improve-x-axis-date-labels-on-desktop)


src/components/Charts/Axis.tsx
```diff
@@ -1,32 +1,33 @@
...
 
-export const AxisBottom = (props: SharedAxisProps<Date>) => {
-  const [dateStart, dateStop] = props.scale.domain();
-  const [minPx, maxPx] = props.scale.range() as number[];
-  const isSmallDevice = maxPx - minPx < 600;
-  const dateFormat = isSmallDevice ? 'MMM' : 'MMM DD';
+interface AxisBottomProps {
+  scale: any;
+  innerHeight: number;
+  tickValues: any;
```
chasulin, line 17: replacing these anys!

src/components/Charts/ChartCaseDensity.tsx
```diff
@@ -120,6 +123,9 @@ const ChartCaseDensity: FunctionComponent<{
     />
   );
 
+  const dateScale = getDateScale(minDate, currDate, chartWidth);
```
pnavarrc, line 126: The AxisBottom components needs to use the same scale than the data points for them to be aligned, so using `xScale` here will solve the misalignment problem.

The `xScale` is a bit more complicated to calculate than usual, because we don't know the end date of the scale

image

We know that the startDate should be at `0`, and that `currDate` should go to `innerWidth - labelsWidth` (`520` px), but we don't know the `endDate` of the scale. To calculate that, we create a scale, calculate `endDate` and then adjust the scale

```ts
const xScale = scaleUtc({ 
  domain: ["2020-03-01", "2021-02-04"], 
  range: [0, 520]
});

// xScale("2020-03-01") = 0
// xScale("2021-02-04") = 520
```

now we need to calculate `endDate` such that `xScale(endDate) = 600`, fortunately, `xScale.invert` can make the calculation for us:

```ts
endDate = xScale.invert(600);
```

and then we adjust the domain

```ts
xScale.domain([startDate, endDate]);
```
This logic is in [components/Charts/utils.ts#L144](https://github.com/covid-projections/covid-projections/blob/develop/src/components/Charts/utils.ts#L144)

pnavarrc (changes requested):
The scales need to be fixed, but looking good so far!

src/components/Charts/ChartZones.tsx
```diff
@@ -77,7 +79,8 @@ const ChartZones = ({
   const maxDate = new Date();
   assert(minDate !== undefined, 'Data must not be empty');
 
-  const xScale = getZonesTimeScale(minDate, maxDate, 0, chartWidth);
+  const xScale = getUtcScale(minDate, maxDate, 0, chartWidth);
+  const dateTicks = getTimeAxisTicks(minDate, maxDate);
```
pnavarrc, line 83: We need to use the date range from from `getUtcScale`, because the endDate is of the scale is changed here [src/components/Charts/utils.tsL159](https://github.com/covid-projections/covid-projections/pull/2657/files#diff-94fef4efde3155913fc63fdabe97e9081a90d10b28fd2107de31826e0f239f5fR159)

```ts
const [startDate, endDate] = xScale.domain();
const dateTicks = getTimeAxisTicks(startDate, endDate);
```

pnavarrc (approved):
Looks good!

PR #2664, opened 2021-02-05
Update Snapshot + Update Vaccination Alerts

BrettBoval:
@pnavarrc is the stuff on staging ready to delpoy.

I have one small thing for the daily deploy (need to block Maine Metros) and then I'm ready on my end.


mikelehen:
@pnavarrc's change doesn't really affect the site.  I think this is good to go.

PR #2667, opened 2021-02-05
Fix cypress tests locally and in CI.

mikelehen:
Cypress tests started failing in CI.  I updated the timeout from 60s to 120s and that seems to have fixed it.

They were also failing for me locally, but for a different reason.  Updating the packages and how we initialize the coverage plugin fixed that.

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,5 +1,6 @@
 {
   "baseUrl": "http://localhost:3000",
   "defaultCommandTimeout": 10000,
-  "projectId": "m7sj8t"
+  "projectId": "m7sj8t",
+  "pageLoadTimeout": 120000
 }
--- a/cypress/plugins/index.js
+++ b/cypress/plugins/index.js
@@ -18,5 +18,6 @@
 module.exports = (on, config) => {
   // `on` is used to hook into various events Cypress emits
   // `config` is the resolved Cypress config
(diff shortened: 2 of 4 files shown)


cypress.json
```diff
@@ -1,5 +1,6 @@
 {
   "baseUrl": "http://localhost:3000",
   "defaultCommandTimeout": 10000,
-  "projectId": "m7sj8t"
+  "projectId": "m7sj8t",
+  "pageLoadTimeout": 120000
```
mikelehen, line 5: This seems to make it pass in CI.  The default timeout is 60s and apparently we're going over for some reason (I assume this is basically the time it takes for 'yarn start' to complete but not sure).

cypress/plugins/index.js
```diff
@@ -18,5 +18,6 @@
 module.exports = (on, config) => {
   // `on` is used to hook into various events Cypress emits
   // `config` is the resolved Cypress config
-  on('task', require('@cypress/code-coverage/task'));
+  require('@cypress/code-coverage/task')(on, config)
+  return config
```
mikelehen, line 22: This is following the instructions on https://github.com/cypress-io/code-coverage

Doing this and updating the package versions made this start passing locally for me.

PR #2684, opened 2021-02-08
Info tooltip content in CMS

chasulin:
Uses copy from [figma](https://www.figma.com/file/085lLFwjpkg8JUB05LSUnb/Learn-Product?node-id=121%3A5208) + dummy copy for metrics that aren't in mocks

### DIFF
--- a/src/cms-content/infoTooltips/index.ts
+++ b/src/cms-content/infoTooltips/index.ts
@@ -1,0 +1,8 @@
+import { Markdown } from '../utils';
+
+export interface InfoTooltip {
+  title: string;
+  id: string;
+  content: Markdown;
+  cta: Markdown;
+}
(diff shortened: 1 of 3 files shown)


src/cms-content/infoTooltips/index.ts
```diff
@@ -0,0 +1,8 @@
+import { Markdown } from '../utils';
+
+export interface InfoTooltip {
+  title: string;
+  id: string;
+  body: Markdown;
+  cta: Markdown;
+}
```
pnavarrc, line 8: should we import the JSON and export the content here?

pnavarrc (approved):
Looks good!

src/cms-content/infoTooltips/index.ts
chasulin: yep! gonna in the next pr

PR #2685, opened 2021-02-09
vaccination alerts: generate list of users to email

pnavarrc:
This PR adds the script that generates the list of users and locations to email when we update vaccination information.


scripts/vaccination_alerts/utils.ts
```diff
@@ -58,3 +63,56 @@ export function getUpdatedVaccinationInfo(
...
+ * have updates for ['29', '01'], then we will notify the user for the changes in
+ * vaccination information for '29', even if the user is not subscribed to alerts for
+ * '29' directly.
+ */
+export function getUserLocationsToAlert(
+  userLocations: FipsCode[],
+  locationsToAlert: FipsCode[],
+): FipsCode[] {
+  return _.chain(userLocations)
+    .map(fipsCode => fipsCode.substr(0, 2))
```
mikelehen, line 114: I think this will do the wrong thing for metros since metros (apparently?) have random fips codes...  New York City metro has fips 35620, so we would end up emailing them when New Mexico (fips 35) gets new vaccine info. 😬 

So I assume we need to look up the region in regions_db and check if it's a county, and then check for `region.state.fipsCode` or whatever.

mikelehen: Alternatively, we could just drop this logic for now, since 88% of people that sign up for a county also sign up for the state, and that'll probably increase with the new vaccination signup UI.

mikelehen (changes requested):
Nice! Couple issues but this is looking good.

scripts/vaccination_alerts/utils.ts
pnavarrc: That's a good point, I tried to simplify but I think that the correct way is to fetch the state of the region using our regions API.

scripts/vaccination_alerts/utils.ts
```diff
@@ -64,23 +68,37 @@ export function getUpdatedVaccinationInfo(
...
+ *     ['pablo@can.org', '29'],
+ *     ['chelsi@can.org', '31'],
+ *     ['chris@can.org', '31']
+ *   ];
+ *
+ *   groupByFips(emailFipsList)
+ *   {
+ *     '29': ['pablo@can.org'],
+ *     '31': ['chelsi@can.org', 'chris@can.org']
+ *   }
```
mikelehen, line 92: Wow, great comment. thank you!

mikelehen (approved):

PR #2687, opened 2021-02-09
Add vaccine dose distribution copy.

mikelehen:
image

### DIFF
--- a/src/common/metrics/vaccinations.tsx
+++ b/src/common/metrics/vaccinations.tsx
@@ -69,8 +69,8 @@ function renderStatus(projections: Projections): React.ReactElement {
   if (info.dosesDistributed && info.ratioDosesAdministered) {
     distributedText = (
       <Fragment>
-        This this means that {locationName} has administered{' '}
-        {formatPercent(info.ratioDosesAdministered)} of their supply of{' '}
+        {locationName} has administered{' '}
+        {formatPercent(info.ratioDosesAdministered)} of their current supply of{' '}
         {formatInteger(info.dosesDistributed)} vaccine doses.
       </Fragment>
     );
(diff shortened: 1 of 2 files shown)


src/common/metrics/vaccinations.tsx
```diff
@@ -65,13 +65,24 @@ function renderStatus(projections: Projections): React.ReactElement {
   const peopleVaccinated = formatInteger(info.peopleVaccinated);
   const percentVaccinated = formatPercent(info.ratioVaccinated, 1);
 
+  let distributedText = <Fragment />;
+  if (info.dosesDistributed && info.ratioDosesAdministered) {
+    distributedText = (
+      <Fragment>
+        This this means that {locationName} has administered{' '}
```
ghop02, line 72: Two this's here.  Also, "this means that" feels a bit unnecessary - being pedantic, it doesn't really imply the % usage of total supply, it only really implies the total doses administered.  Could this just be `{locationName} has administered`?

src/common/metrics/vaccinations.tsx
```diff
@@ -65,13 +65,24 @@ function renderStatus(projections: Projections): React.ReactElement {
...
+        {formatPercent(info.ratioDosesAdministered)} of their supply of{' '}
```
ghop02, line 73: current supply?

ghop02 (commented):
couple small copy notes!

src/common/metrics/vaccinations.tsx
mikelehen: Gah, good catch on "This this". 😞 

"This means that" was bugging me too, so I removed it. But it is the copy that Debbie had approved.  But I think it's probably an improvement.

src/common/metrics/vaccinations.tsx
mikelehen: Yeah, I was thinking about that too.  Done.

ghop02 (approved):
LGTM!

PR #2691, opened 2021-02-09
Metric enum into its own file

chasulin:
I was getting a circular dependency error in a working branch. It seems like the cause might be declaring `Metric` in `metric.ts`--where we also import everything from `common/metric/*each-individual-metric-file*`--each of which import+use `Metric`. This PR pulls the enum into its own file

Screen Shot 2021-02-09 at 3 52 02 PM

### DIFF
--- a/src/common/metric.tsx
+++ b/src/common/metric.tsx
@@ -9,20 +9,9 @@ import { fail } from 'common/utils';
 import { MetricDefinition } from './metrics/interfaces';
 import { formatDecimal, formatPercent } from 'common/utils';
 import { isNumber } from 'lodash';
+import { Metric } from './metricEnum';
 
-export enum Metric {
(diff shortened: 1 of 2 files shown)


src/common/metric.tsx
```diff
@@ -1,3 +1,4 @@
+import { Metric } from 'common/metricEnum';
```
pnavarrc, line 1: We could re-export `Metric` from this file, so we don't have to change the imports in all the other files changed in this branch

pnavarrc (approved):
Looks good, I think it would be better to re-export `Metric` from `common/metrics` to avoid having to remember to import from 2 places (which seems to be very common)

pnavarrc (approved):
Looks good!

PR #2692, opened 2021-02-09
Add vaccination phases internal page

ghop02:
available at `/internal/vaccine-eligibility`.

I wrote this relying less on styled components and more on the material ui system (just because I'm more familiar).  Had to override margins for typography header (which would be nice to look into in general) but otherwise fairly straightforward!

image

### DIFF
--- a/src/screens/internal/VaccinationPhases/VaccinationPhases.tsx
+++ b/src/screens/internal/VaccinationPhases/VaccinationPhases.tsx
@@ -14,7 +14,7 @@ interface StateVaccineDataProps {
   regionData: RegionVaccinePhaseInfo;
 }
 
-const PipeDelimatedRow = ({
+const PipeDelimitedRow = ({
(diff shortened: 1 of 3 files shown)


src/screens/internal/VaccinationPhases/VaccinationPhases.tsx
```diff
@@ -0,0 +1,139 @@
...
+import ExternalLink from 'components/ExternalLink';
+import { MarkdownContent } from 'components/Markdown';
+
+import { sortBy } from 'lodash';
+
+interface StateVaccineDataProps {
+  regionData: RegionVaccinePhaseInfo;
+}
+
+const PipeDelimatedRow = ({
```
pnavarrc, line 17: Delimated → Delimited?

pnavarrc (approved):
Looks good!

PR #2696, opened 2021-02-10
Adding script to parse vaccination information

ghop02:


BrettBoval (approved):
Thanks for running the updater script. LGTM.

BrettBoval:
We used this script to initially port over the data from the collab spreadsheet to the cms. Since then, we've refactored the cms with a new schema. I think this has served its purpose and doesn't need to be updated to the new schema.

PR #2706, opened 2021-02-10
vaccination alerts: send test email

pnavarrc:
This PR implements a script to send a test vaccination alert email and the logic to render the email, based on the vaccination information on the CMS.

**Usage**

```bash
$ yarn vaccinations-send-test-email pablo@covidactnow.org 12  # Florida
```

**Note**:
- Follow the instructions on https://github.com/covid-projections/covid-projections/blob/develop/scripts/alert_emails/readme.md#signing-up-for-the-correct-things to get the necessary keys to send the email.


scripts/vaccination_alerts/utils.ts
```diff
@@ -141,3 +144,55 @@ export function getUserLocationsToAlert(
...
+  const fipsCode = vaccinationInfo.fips;
+  const region = regions.findByFipsCodeStrict(fipsCode);
+
+  const currentPhases = vaccinationInfo.phaseGroups
+    .filter(phaseGroup => phaseGroup.currentlyEligible)
+    .map((phaseGroup, index, allCurrentPhases) => {
+      const isMostRecentlyEligible = index === allCurrentPhases.length - 1;
+
+      return {
+        title: phaseGroup.tier
```
ghop02, line 168: random, but It might be a good idea to make the PhaseGroup a class and add things like `phaseGroup.title` or something

scripts/vaccination_alerts/send-test-email.ts
```diff
@@ -0,0 +1,149 @@
...
+    ],
+  },
+};
+
+const outputPath = path.join(__dirname, 'vaccination-alert.html');
+
+async function main(emailAddress: string) {
+  const emailService = new EmailService();
+
+  const fipsCode = '08';
```
ghop02, line 94: I know this is just a test but would be good to pass this in!

pnavarrc: We are using mock data, we just need to make sure that we have the FIPS code in the mock data. An alternative could be to just try to load the content directly from the CMS

scripts/vaccination_alerts/utils.ts
pnavarrc: Maybe on a follow up PR, but I like the idea!

scripts/vaccination_alerts/utils.ts
```diff
@@ -141,3 +144,70 @@ export function getUserLocationsToAlert(
...
+
+  const mostRecentPhase = _.last(currentPhases);
+  const title = mostRecentPhase?.title
+    ? `${region.fullName} is now in ${mostRecentPhase.title} of vaccination`
+    : 'FALLBACK TITLE';
+
+  const data = {
+    title,
+    subtitle: 'People who are eligible to be vaccinated now include:',
+    sourceName: `${region.fullName} Health Department`,
```
pnavarrc, line 187: TODO: We need to add the name of the source to the CMS

scripts/vaccination_alerts/utils.ts
```diff
@@ -141,3 +151,88 @@ export function getUserLocationsToAlert(
...
+    sourceUrl: vaccinationInfo.eligibilityInfoUrl,
+    eligibilityUrl: vaccinationLinks?.eligibilityInfoUrl,
+    locationName: region.fullName,
+    currentPhases,
+    colors,
+  };
```
pnavarrc, line 207: note to self: I think we should probably move the function that generates the data to send in the emails closer to the CMS, since all the information is taken from there

scripts/vaccination_alerts/utils.ts
```diff
@@ -141,3 +156,30 @@ export function getUserLocationsToAlert(
...
+  // Writing the file locally is useful while styling
+  // fs.writeFileSync('alert.html', emailHtmlContent);
+
+  return {
+    Subject: emailData.emailSubjectLine,
+    To: [emailAddress],
+    Html: emailHtmlContent,
+    From: 'Covid Act Now Alerts <noreply@covidactnow.org>',
+    ReplyTo: 'noreply@covidactnow.org',
+    Group: 'VACCINATION_ALERTS',
```
mikelehen, line 183: The alert emails include the date here which lets us calculate open/click rate for each individual alerts push. We should probably do the same to be consistent? 

https://github.com/covid-projections/covid-projections/blob/b969ebec3def3b48abf00d8262ddcdcaae3ea82a/scripts/alert_emails/utils.ts#L89

scripts/vaccination_alerts/vaccination-alert-template-source.html
```diff
@@ -0,0 +1,247 @@
+<html>
```
mikelehen, line 1: Hrm. Can you add a note to the readme to explain what's up with this -source.html file vs the main .html file?  Seems this isn't directly used anywhere?

src/cms-content/vaccines/email-alerts.ts
```diff
@@ -0,0 +1,94 @@
...
+      const isLastCurrentPhase = index === allCurrentPhases.length - 1;
+      return formatVaccinationPhaseInfo(phaseInfo, isLastCurrentPhase);
+    });
+
+  const mostRecentPhase = currentPhases.find(
+    phaseInfo => phaseInfo.isMostRecentPhase,
+  );
+
+  const title = mostRecentPhase?.title
+    ? `${region.fullName} is now in ${mostRecentPhase.title} of vaccination`
```
mikelehen, line 78: Does this mean if the eligibility groups for an existing phase changed, we should set `isMostRecentPhase` to `false` so that the email says "has updated vaccination information" instead of sending a redundant email with "is now in <existing phase> of vaccination" ?

Or do we not care?

I don't have strong feelings, but wondering if we should have a conversation or document expectations in a "How to update the CMS vaccine alert info" playbook somewhere or something.

Feel free to punt if I'm confused / over-thinking this.

scripts/vaccination_alerts/vaccination-alert-template.html
```diff
@@ -0,0 +1,111 @@
...
+              <strong>See where and how to get vaccinated</strong>
+            </div>
+            <div>on the {{sourceName}} website</div>
+          </div>
+        </a>
+      </div>    
+      {{/if}}
+
+      <div class="section description center-text" style="color: #4f4f4f; margin-top: 24px; text-align: center;">
+        Because eligibility varies throughout {{locationName}}, you may also want to check your county or city’s health department website.
```
mikelehen, line 55: I'm not sure we're plumbing `locationName` in properly?  It was blank in my test email.

scripts/vaccination_alerts/vaccination-alert-template.html
```diff
@@ -0,0 +1,111 @@
...
+          <p style="margin: 5px auto;">Anyone can subscribe using this link:</p>
+          <a class="footer__link" style="color: #4f4f4f;" href="https://covidactnow.org/alert_signup">
+            https://covidactnow.org/alert_signup
+          </a>
+        </div>
+
+        <div class="footer__section" style="margin-bottom: 50px;">
+          <h2 style="color: #000; font-size: 18px; margin-bottom: 5px; text-align: center;">Change your alert settings</h2>
+          <p style="margin: 5px auto;">
+            <a class="footer__link" style="color: #4f4f4f;" href="https://covidactnow.org/alert_signup">Edit your locations</a>
```
mikelehen, line 100: I think this should go to the /alert_unsubscribe page (same as the actual "unsubscribe from all alerts" link below).  And it needs to encode the email address into the link. https://github.com/covid-projections/covid-projections/blob/11c10502451ef5ee08fc1256875e8958c9dd0529/scripts/alert_emails/utils.ts#L56

mikelehen (approved):
Nice!  This generally looks good.  I think there are a couple minor thinks we need to fix with the template data (at least before shipping), but I was able to run it and it generally seems to work! 😁

src/cms-content/vaccines/email-alerts.ts
pnavarrc: We are manually marking updates on the CMS vaccination information, but as a whole for the state, not on individual phases. Here, I'm assuming that we only do that when a vaccination phase starts or a new group is included in the "most recent" active phase, and that we are not emailing people if phases that started before update their definitions (I'm not sure if that happens).

The logic here is more of a fallback in case the `currentPhases` array is empty, but I guess in that case we should fail the job.

PR #2709, opened 2021-02-10
Make vaccine chart more forgiving of missing 2nd dose data.

mikelehen:
If we really want to support this, well need to make additional updates to the copy, but this gets us closer and resolves the symptom reported in https://trello.com/c/gmpQgViy/859-oh-vaccine-timeseries-rendering-mixup

Before:
image

After:
image


pnavarrc (approved):
Looks good!

PR #2710, opened 2021-02-10
Refactoring generate users to email

ghop02:
As I was going through the code, I restructured the existing functions a bit in a way that made slightly more sense to me - lmk if you like this approach!

### DIFF
--- a/scripts/vaccination_alerts/generate-users-to-email.ts
+++ b/scripts/vaccination_alerts/generate-users-to-email.ts
@@ -5,9 +5,68 @@ import {
   EmailFips,
   getLocationsToAlert,
   readVaccinationAlerts,
-  getUserLocationsToAlert,
   groupByFips,
+  DEFAULT_ALERTS_FILE_PATH,
+  getStateFipsCodesSet,
 } from './utils';
+
+import * as yargs from 'yargs';
+import { FipsCode } from '../../src/common/regions';
+import { RegionVaccinePhaseInfo } from '../../src/cms-content/vaccines/phases';
+
+const buildEmailsToAlertByFips = (
+  fipsCodesToAlert: FipsCode[],
+  alertSubscriptions: FirebaseFirestore.QuerySnapshot<
+    FirebaseFirestore.DocumentData
+  >,
+) => {
+  const emailFipsList: EmailFips[] = [];
+
+  alertSubscriptions.forEach(emailLocationsDoc => {
+    const email = emailLocationsDoc.id;
+    const { locations } = emailLocationsDoc.data();
+
+    // Currently only alerting on states, so pull state fips codes from subscribed locations.
+    const subscribedStateFips = getStateFipsCodesSet(locations);
+
+    const userFipsToAlert = subscribedStateFips
+      .intersection(fipsCodesToAlert)
+      .value();
+
+    emailFipsList.push(
+      ...userFipsToAlert.map(fips => [email, fips] as EmailFips),
+    );
+  });
+
+  // Once we have a list of all the [email, fipsCode] pairs to email, we group them
+  // by fipsCode and update the collection with vaccination alerts.
+  return groupByFips(emailFipsList);
+};
+
+// Adds blank sent records for each email for given alert to Firebase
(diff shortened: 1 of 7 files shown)


scripts/vaccination_alerts/generate-users-to-email.ts
```diff
@@ -5,10 +5,69 @@ import {
...
+  alertSubscriptions.forEach(emailLocationsDoc => {
+    const email = emailLocationsDoc.id;
+    const { locations } = emailLocationsDoc.data();
+
+    // Currently only alerting on states, so pull state fips codes from subscribed locations.
+    const subscribedStateFips = getStateFipsCodesSet(locations);
+
+    const userFipsToAlert = subscribedStateFips
+      .intersection(fipsCodesToAlert)
+      .value();
```
pnavarrc, line 34: It feels a bit weird for me to break a `_.chain` sequence, consider returning an array from `getStateFipsCodesSet` so we can just use `_.intersection` here.

scripts/vaccination_alerts/generate-users-to-email.ts
```diff
@@ -84,7 +108,14 @@ async function main() {
...
 if (require.main === module) {
-  main()
+  const argv = yargs.options({
+    input: {
+      default: DEFAULT_ALERTS_FILE_PATH,
+      description: 'Updated alerts input path.',
+    },
+  }).argv;
+
+  main(argv.input)
```
pnavarrc, line 118: 💯  - I'm going to update my script to send test emails after we merge this!

scripts/vaccination_alerts/generate-vaccine-alerts-info.ts
```diff
@@ -4,33 +4,58 @@ import {
...
   const firebaseInfo = await getFirebaseVaccinationInfo();
 
+  console.log('Getting snapshot of alerts to send');
   const vaccinationAlertUpdates = getUpdatedVaccinationInfo(
     cmsInfo,
     firebaseInfo,
   );
 
-  fs.writeFileSync(alertsFilePath, JSON.stringify(vaccinationAlertUpdates));
+  for (const fips of Object.keys(vaccinationAlertUpdates)) {
```
pnavarrc, line 32: nit: consider using `const [fips, current] of Object.entries(vaccinationAlertUpdates)`

scripts/vaccination_alerts/generate-vaccine-alerts-info.ts
```diff
@@ -4,33 +4,58 @@ import {
...
+    path: {
+      default: DEFAULT_ALERTS_FILE_PATH,
+      description: 'Output path.',
+    },
+  }).argv;
+
+  main(argv.path)
     .then(() => {
-      console.log(`Done. Generated ${alertsFilePath}`);
+      console.log(`Done. Generated `);
```
pnavarrc, line 58: Maybe log the path of the file here?

pnavarrc (approved):
Added some comments, looks good!

scripts/vaccination_alerts/generate-users-to-email.ts
```diff
@@ -84,7 +108,14 @@ async function main() {
 }
 
 if (require.main === module) {
-  main()
+  const argv = yargs.options({
```
pnavarrc, line 111: I think we need to add yargs to the dependencies of the project

scripts/vaccination_alerts/generate-users-to-email.ts
pnavarrc: missing semicolon (not that I care, but maybe Prettier is not running for you?)

scripts/vaccination_alerts/generate-users-to-email.ts
ghop02: yeah I think it's pulled in as a dependency but definitely should be explicitly defined - will do

scripts/vaccination_alerts/generate-users-to-email.ts
ghop02: I don't think there's a semicolon there because it's a chained operation

PR #2711, opened 2021-02-10
Add script to send email

ghop02:
TODO: 
 * Generate the template
 * Update/pull from firebase probably rather than reading the alert files


scripts/vaccination_alerts/send-emails.ts
```diff
@@ -0,0 +1,206 @@
...
+    dryRun: {
+      default: true,
+    },
+    email: {
+      type: 'string',
+    },
+    fips: {
+      default: 'string',
+    },
+  }).argv;
```
pnavarrc, line 195: Consider removing the `email` parameter here since we have a dedicated script for that. I would make `batchSize` optional too

scripts/vaccination_alerts/send-emails.ts
```diff
@@ -0,0 +1,206 @@
...
+        const emails = await alertEmailService.fetchEmailsForAlert(alert);
+        return (
+          emails
+            // if an email is specified, filter out all other emails. Used for testing
+            .filter(email => (singleEmail ? email === singleEmail : true))
+            .map(email => ({ email, alert }))
+        );
+      }),
+    ),
+  );
```
pnavarrc, line 141: The function to render and send the emails only needs the FIPS code (I just finished that), so we can probably simplify here. I think that we should remove the `singleEmail` logic to make this function easier to read

scripts/vaccination_alerts/send-emails.ts
```diff
@@ -0,0 +1,206 @@
...
+  console.info(`Total emails to be sent: ${emailsSent}.`);
+  console.info(
+    `Unique Email addresses: ${Object.keys(uniqueEmailAddresses).length}.`,
+  );
+  console.info(`Invalid emails removed: ${invalidEmailAddresses}.`);
+
+  if (!dryRun) {
+    // setLastSnapshotNumber(db, `${currentSnapshot}`);
+    // If we aren't in a dry run but there's clearly some non trivial errors exit with error
+    if (emailsSent < 1 || errorCount > 1) {
```
pnavarrc, line 167: I think that we need to make this script a bit more forgiving, there are some invalid emails in our Firebase collection (you can check the logs of the risk level alerts to see some of them).

pnavarrc (commented):
We need to integrate the email rendering logic from my PR, let's re-review and test once that's ready!

scripts/vaccination_alerts/send-emails.ts
ghop02: ah yeah i wasn't 100% sure what to do here

scripts/vaccination_alerts/send-emails.ts
```diff
@@ -0,0 +1,206 @@
...
+} from '../alert_emails/email-service';
+import * as yargs from 'yargs';
+import { DEFAULT_ALERTS_FILE_PATH, readVaccinationAlerts } from './utils';
+import { RegionVaccinePhaseInfo } from '../../src/cms-content/vaccines/phases';
+
+const BATCH_SIZE = 20;
+
+class SendVaccinationAlertsService {
+  db: FirebaseFirestore.Firestore;
+  emailService: EmailService;
```
mikelehen, line 20: `private readonly` on these? (though see other comment on constructor)

scripts/vaccination_alerts/send-emails.ts
```diff
@@ -0,0 +1,206 @@
...
+
+  invalidEmailCount: number = 0;
+  errorCount: number = 0;
+  emailSent: number = 0;
+
+  uniqueEmailAddress: { [email: string]: number } = {};
+  locationsWithEmails: { [fips: string]: number } = {};
```
mikelehen, line 27: This is written but never read, I think?  Should use it or drop it?

scripts/vaccination_alerts/send-emails.ts
```diff
@@ -0,0 +1,206 @@
...
+
+  constructor(db: FirebaseFirestore.Firestore, emailService: EmailService) {
+    this.db = db;
+    this.emailService = emailService;
+  }
+
+  private async onEmailSent(email: string, alert: RegionVaccinePhaseInfo) {
+    this.updateSentEmailCounters(email, alert);
+    const docPath = `${alert.fips}/emailVersions/${alert.emailAlertVersion}/emails/${email}`;
```
mikelehen, line 36: Is it too late to make `emailVersions` into `email-versions` so we're using a consistent naming scheme for collection names?

scripts/vaccination_alerts/send-emails.ts
```diff
@@ -0,0 +1,206 @@
...
+    return this.db
+      .collection('vaccination-alerts')
```
mikelehen, line 38: Ideally we'd have constants for all of these collection names, but can punt for now.

Also it's a bit arbitrary that we do `collection('vaccination-alerts').doc(docPath)` instead of just including `vaccination-alerts/` at the beginning of `docPath` or doing `.collection('vaccination-alerts').doc(alert.fips).collection('emailVersions')...`

scripts/vaccination_alerts/send-emails.ts
```diff
@@ -0,0 +1,206 @@
...
+    }
+  }
+
+  private updateSentEmailCounters(
+    email: string,
+    alert: RegionVaccinePhaseInfo,
+  ) {
+    this.emailSent += 1;
+    this.uniqueEmailAddress[email] = (this.uniqueEmailAddress[email] || 0) + 1;
+    // Todo don't make just fips
```
mikelehen, line 65: Can you expand this TODO or fix it before merging?

scripts/vaccination_alerts/send-emails.ts
```diff
@@ -0,0 +1,206 @@
...
+    const key = (alert.fips, alert.emailAlertVersion);
```
mikelehen, line 66: Uhh...  This looks like it's trying to make a tuple, but that's not a JavaScript feature, so I think it's equivalent to `const key = alert.emailAlertVersion` which is almost certainly not what we want? 😁 

easiest bet is probably `const key = `${alert.fips}_${alert.emailAlertVersion}` ?

scripts/vaccination_alerts/send-emails.ts
```diff
@@ -0,0 +1,206 @@
...
+    this.locationsWithEmails[key] = (this.locationsWithEmails[key] || 0) + 1;
+  }
+
+  async fetchEmailsForAlert(alert: RegionVaccinePhaseInfo) {
+    const docPath = `${alert.fips}/emailVersions/${alert.emailAlertVersion}/emails/`;
```
mikelehen, line 71: This ends up being a collection path, not a doc path.

scripts/vaccination_alerts/send-emails.ts
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * Given a list of users to email and a list of locations, email those
+ * users with alert information about those locations.
+ *
+ * Takes in a file with {[fips: string]: {...data}} and file like {[fips: string]: [emails]}
+ *
+ * Also takes in a parameter to actual send the email. Otherwise will just generate the files
+ *
+ * You can run via `yarn send-emails fipsToAlertFilename currentSnapshot [send]`
```
mikelehen, line 123: I think this whole comment is out-of-date.  Remove or improve? 😁

scripts/vaccination_alerts/send-emails.ts
```diff
@@ -0,0 +1,206 @@
...
+  singleEmail?: string,
+  fips?: string,
+) {
+  const alertsByFips = readVaccinationAlerts(alertPath);
+  const alerts = _.values(alertsByFips);
+
+  const emailService = new EmailService();
+  const db = getFirestore();
+
+  const alertEmailService = new SendVaccinationAlertsService(db, emailService);
```
mikelehen, line 132: Why pass in `db` and `emailService` instead of having the constructor create them internally?

scripts/vaccination_alerts/send-emails.ts
```diff
@@ -0,0 +1,206 @@
...
+  emailService: EmailService;
+
+  invalidEmailCount: number = 0;
+  errorCount: number = 0;
+  emailSent: number = 0;
+
+  uniqueEmailAddress: { [email: string]: number } = {};
+  locationsWithEmails: { [fips: string]: number } = {};
+
+  constructor(db: FirebaseFirestore.Firestore, emailService: EmailService) {
```
mikelehen, line 29: The abbreviated way to do this would be to drop the `db` / `emailService` field definitions above and do:

`constructor(private readonly db: FirebaseFirestore.Firestore, private readonly emailService: EmailService) { }`

But also see my other comment about maybe removing these as constructor arguments, since we always pass the same values.

scripts/vaccination_alerts/send-emails.ts
```diff
@@ -0,0 +1,206 @@
...
+        alertEmailService.sendAlertEmail(data.email, data.alert, dryRun),
+      ),
+    );
+  }
+
+  alertEmailService.logEmailsToBeSentStats();
+
+  if (dryRun) {
+    return;
+  }
```
mikelehen, line 165: It looks like you removed the code that checks `alertEmailService.errorCount`, but I think we need that.  If there are any errors, we should make sure the process fails with a non-zero exit code so that the workflow fails and we are notified that there was a problem.  Else it'll end up just silently failing.

Note that we explicitly handle invalid emails separately from `errorCount` so those won't cause a problem. https://github.com/covid-projections/covid-projections/pull/2711/files#diff-f89fbdcacb69b99011a18b990ed557c1c1481810027222a5e6ae761de4075e25R103

mikelehen (approved):
Looks good, except I think we lost some important error handling.  I also left a bunch of nits, but feel free to punt and focus on testing the functionality which is probably more important. 😁

scripts/vaccination_alerts/send-emails.ts
pnavarrc: Now or never!

scripts/vaccination_alerts/send-emails.ts
pnavarrc: this is actually okay, because we don't throw on invalid emails :)

scripts/vaccination_alerts/send-emails.ts
pnavarrc: Adding this back!

scripts/vaccination_alerts/firestore-subscriptions.ts
```diff
@@ -0,0 +1,112 @@
...
+    return this.db.collection(VACCINATION_INFO_UPDATES).doc(fipsCode).set({
+      'email-alert-version': emailAlertVersion,
+    });
+  }
+
+  public async getAlertSubscriptions() {
+    return this.db.collection(ALERT_SUBSCRIPTIONS).get();
+  }
+
+  private async getEmailsForVersion(
```
mikelehen, line 44: I'd add `Collection` to this name and maybe add a return type to the function.

scripts/vaccination_alerts/firestore-subscriptions.ts
```diff
@@ -0,0 +1,112 @@
...
+    fipsCode: string,
+    emailAlertVersion: number,
+  ) {
+    return this.db
+      .collection(VACCINATION_ALERTS)
+      .doc(fipsCode)
+      .collection('email-versions')
+      .doc(`${emailAlertVersion}`)
```
mikelehen, line 52: Could just be `.doc(emailAlertVersion)`

scripts/vaccination_alerts/firestore-subscriptions.ts
```diff
@@ -0,0 +1,112 @@
...
+    emailAlertVersion: number,
+    emailAddress: string,
+  ) {
+    const emailsCollection = await this.getEmailsForVersion(
+      fipsCode,
+      emailAlertVersion,
+    );
+    return emailsCollection
+      .doc(emailAddress)
+      .set({ sentAt: admin.firestore.FieldValue.serverTimestamp() });
```
mikelehen, line 93: can you make this an `update()` instead of `.set()`?  Shouldn't make a big difference but it's slightly safer / more correct (won't delete any other fields in the document).

scripts/vaccination_alerts/firestore-subscriptions.ts
```diff
@@ -0,0 +1,112 @@
...
+      fipsCode,
+      emailAlertVersion,
+    );
+    const querySnapshot = await emailsCollection
+      .where('sentAt', '==', null)
+      .get();
+    return querySnapshot.docs.map(emailDoc => emailDoc.id);
+  }
+
+  public async createEmailToSend(
```
mikelehen, line 70: Rename to `markEmailToSend()` since it's not creating an email and so it's symmetric with `markEmailAsSent()` ?

scripts/vaccination_alerts/generate-users-to-email.ts
```diff
@@ -46,18 +44,19 @@ const buildEmailsToAlertByFips = (
 
 // Adds blank sent records for each email for given alert to Firebase
 const commitUpdatedEmailAlertToSend = async (
```
mikelehen, line 46: Maybe rename to `markEmailAlertsToSend` to be consistent with other naming?

scripts/vaccination_alerts/send-emails.ts
```diff
@@ -0,0 +1,196 @@
...
+    console.info(`Total emails to be sent: ${this.emailSent}.`);
+    console.info(`Unique Email addresses: ${uniqueEmails.length}.`);
+    console.info(`Invalid emails removed: ${this.invalidEmailCount}.`);
+  }
+
+  public logErrors() {
+    const { emailSent, invalidEmailCount } = this;
+    console.log('Errors sending vaccination alert emails');
+    console.log(`Emails sent: ${emailSent}`);
+    console.log(`Invalid emails removed: ${invalidEmailCount}`);
```
mikelehen, line 69: Maybe log `errorCount` too?

scripts/vaccination_alerts/send-emails.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  alertEmailService.logEmailsToBeSentStats();
+
+  if (dryRun) {
+    return;
+  }
+
+  // Exit with error if there is any errors sending the emails
+  if (alertEmailService.errorCount >= 1) {
+    alertEmailService.logErrors();
```
mikelehen, line 162: Hrm.  As written, this is only going to log the "Invalid emails removed" message if there were _other_ errors as well.  I think maybe we should just call logErrors() always and have it log both the fatal and non-fatal errors?  Not sure.

mikelehen (approved):

scripts/vaccination_alerts/firestore-subscriptions.ts
pnavarrc: It can't, TS complains that the argument for `.doc` should be a string (`emailAlertVersion` is a number)

mikelehen: Oh... right! Makes sense. 😁

mikelehen (approved):

PR #2712, opened 2021-02-10
Edit default signup regions

chasulin:
Adds metro's states to default alert signup locations
[Trello ticket](https://trello.com/c/07bDUrRi/886-add-metro-states-to-the-alerts-default-signup-when-youre-on-a-metro-area)


pnavarrc (approved):
Looks good!

PR #2714, opened 2021-02-11
handle failure to write to localStorage more gracefully

pnavarrc:
We write a flag to `localStorage` to determine whether or not to enable FullStory recordings. Some users block applications from writing to `localStorage` and in this case, writing to it would fail (also if the storage is full).

This PR handles that failure more gracefully, catching the error if we can't write to `localStorage`

### DIFF
--- a/src/common/fullstory.ts
+++ b/src/common/fullstory.ts
@@ -22,6 +22,7 @@ enum RecordSession {
 export function initFullStory() {
   initStorage();
 
+  // If we can't find the value in localStorage we won't enable recordings in FullStory
   if (localStorage.getItem(FULLSTORY_RECORD_KEY) === RecordSession.TRUE) {
     FullStory.init({ orgId: 'XEVN9' });
   }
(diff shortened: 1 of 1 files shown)


src/common/fullstory.ts
```diff
@@ -32,8 +33,14 @@ function initStorage() {
     return;
```
mikelehen, line 33: I think just referencing `localStorage` can throw an exception.  In fact, it's the above line that is showing up in sentry.  https://sentry.io/organizations/covidactnow/issues/2188103312/?project=5444052&referrer=slack

The recommended way to detect if storage is available is here: https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#testing_for_availability

mikelehen (changes requested):
I think we need to do a bit more to ensure localStorage is really available.

src/common/fullstory.ts
```diff
@@ -32,8 +37,43 @@ function initStorage() {
...
+        // Firefox
+        e.code === 1014 ||
+        // test name field too, because code might not be present
+        // everything except Firefox
+        e.name === 'QuotaExceededError' ||
+        // Firefox
+        e.name === 'NS_ERROR_DOM_QUOTA_REACHED') &&
+      // acknowledge QuotaExceededError only if there's something already stored
+      storage &&
+      storage.length !== 0
```
mikelehen, line 76: FWIW- We could probably treat "quota exceeded" as if storage isn't available and just replace this whole catch block with `return false;`.  But probably (hopefully) won't come up in practice.

mikelehen (approved):

pnavarrc:
[Render Website with All Cookies Turned Off](https://trello.com/c/Ww0n90tS/907-render-website-with-all-cookies-turned-off)

PR #2739, opened 2021-02-15
Features/razzle static gen

schamp:
This is a big mess, but you can have a look and see what I got stuck on.  I suspect there's something wrong in the webpack or the loadable configuration, but I can't figure it out.

Right now, the initial load is nice and fast, but then things fall apart when trying to hydrate.

Feel free to clone and `yarn install` and then `yarn start-razzle` for dev or `yarn export-and-serve` to see the SSR in action.

There are a few misc cleanup commits in here I may pull out as separate PRs.

Some notes here: https://www.dropbox.com/scl/fi/u2omix3pfxj7cnvziiuy7/Static-Site-Generation-prototyping.paper?dl=0&rlkey=u61dl2oodvklxua38k8kyd2l2


mikelehen:
Thanks! Will poke around. FYI- I pulled in (at least 3 of) your cleanup commits and merged them as https://github.com/covid-projections/covid-projections/pull/2742.

src/server.tsx
```diff
@@ -0,0 +1,265 @@
...
+        ${
+          assets.client.css
+            ? `<link rel="stylesheet" href="${assets.client.css}">`
+            : ''
+        }
+        ${
+          process.env.NODE_ENV === 'production'
+            ? `<script src="${assets.client.js}" defer></script>`
+            : `<script src="${assets.client.js}" defer crossorigin></script>`
+        }
```
mikelehen, line 233: So if you look at the index.html getting served by razzle, it contains this near the top:
```
  <!-- razzle assets -->
        <script src="http://localhost:3001/static/js/bundle.js" defer crossorigin></script>
  <!-- end razzle assets -->
```

and then near the bottom:
```
  <!-- loadable-components script tags -->
  <script id="__LOADABLE_REQUIRED_CHUNKS__" type="application/json">[]</script><script id="__LOADABLE_REQUIRED_CHUNKS___ext" type="application/json">{"namedChunks":[]}</script>
<script async data-chunk="client" src="http://localhost:3001/static/js/bundle.js"></script>
  <!-- end loadable-components script tags -->
  <!-- loadable-components link tags -->
  <link data-chunk="client" rel="preload" as="script" href="http://localhost:3001/static/js/bundle.js">
  <!-- end loadable-components link tags -->
```

So bundle.js is getting loaded twice which loads everything twice and causes the concurrent rendering errors, etc.

I experimentally just removed this `<script>` include here and it seems like things are a lot closer to working, though I didn't test very deeply, and I don't know for sure what the "right" fix is.

mikelehen (commented):
I think I found at least one problem... 😁 

Also, FYI- I merged the 3 commits that had console warning fixes, so we're a tiny bit cleaner now.  Thanks!

src/server.tsx
schamp: ah, nice, I bet you're right.  I'll poke around the docs and see if I can find an explanation.

schamp:
definitely not going with this this time around : )

PR #2742, opened 2021-02-16
Fix some console warnings.

mikelehen:
Pulling in some console warnings that @schamp fixed in https://github.com/covid-projections/covid-projections/pull/2739.


schamp:
awesome, thanks : )

PR #2750, opened 2021-02-16
enable vaccination alerts workflow

pnavarrc:
This PR enables the vaccination alerts workflow. Since we want to be able to send alerts only for specific states, I added a check to verify that the version in the CMS is 1 or more to actually send the alerts, which is a bit of an undocumented feature (I'm open to better ideas on how to do this!)

### DIFF
--- a/.github/workflows/vaccination-alerts.yml
+++ b/.github/workflows/vaccination-alerts.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Generate the list of users and locations to email
         run: yarn vaccinations-generate-users-to-email
 
-      - name: Senv vaccination alert emails
+      - name: Send vaccination alert emails
         run: yarn vaccinations-send-alert-emails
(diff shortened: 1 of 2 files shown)


.github/workflows/vaccination-alerts.yml
```diff
@@ -16,10 +25,46 @@ jobs:
...
+          SERVICE_ACCOUNT: ${{ secrets.FIREBASE_SERVICE_ACCOUNT_PROD }}
+
       - name: Generate the vaccination info update
         run: yarn vaccinations-generate-alerts
 
       - name: Generate the list of users and locations to email
         run: yarn vaccinations-generate-users-to-email
-      # Send the email
+
+      - name: Senv vaccination alert emails
```
mikelehen, line 51: Send

mikelehen (approved):

PR #2752, opened 2021-02-17
increase target size for HI and PR

pnavarrc:
Increase the target size for Hawaii and Puerto Rico in the US map. I added an huge invisible border to the paths with opacity set to `0` to make the tap size bigger (new tap areas highlighted below). 

image

I also set `tabIndex=-1` on the link and paths to prevent the map from becoming a keyboard focus trap. A user trying to use `tab` to navigate the page will have to go through all the states before reaching the compare table. We provide a good way to navigate to states already, so I think it's better to skip the map completely from keyboard-navigation for now.

### DIFF
--- a/src/components/Map/Map.style.js
+++ b/src/components/Map/Map.style.js
@@ -2,10 +2,7 @@ import styled from 'styled-components';
 import Typography from '@material-ui/core/Typography';
 import { COLOR_MAP } from 'common/colors';
 
-// TODO(pablo): We don't need these wrapping divs anymore
 export const USMapWrapper = styled.div``;
-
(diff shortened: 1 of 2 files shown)


pnavarrc:
[Expand tap target area on map for states: PR and HI](https://trello.com/c/BsuNRW0B/387-expand-tap-target-area-on-map-for-states-pr-and-hi)

pnavarrc:
[Accessibility · Remove focus trap on US Map](https://trello.com/c/MM70usqT/798-accessibility-remove-focus-trap-on-us-map)

mikelehen (approved):
Looks good!

PR #2753, opened 2021-02-17
remove the blog link from the footer

pnavarrc:
I also changed `div` → `footer` to make it more semantic/accessible.

Note: I initially added `role="contentinfo"` to the footer, but since the footer is directly under `body` that's not necessary (see [WAI-ARIA Practices - ContentInfo](https://www.w3.org/TR/wai-aria-practices-1.2/#aria_lh_contentinfo))

### DIFF
--- a/src/components/Footer/Footer.style.tsx
+++ b/src/components/Footer/Footer.style.tsx
@@ -1,7 +1,7 @@
 import styled from 'styled-components';
 import palette from 'assets/theme/palette';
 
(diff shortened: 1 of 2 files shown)


pnavarrc:
[Remove "Blog" from the footer so it's not discoverable](https://trello.com/c/ahUqaQjX/921-remove-blog-from-the-footer-so-its-not-discoverabe)

chasulin (approved):

PR #2762, opened 2021-02-17
CCVI data

chasulin:
Parses CCVI data and adds `ccvi-data.json`


scripts/ccvi/parse_ccvi_counties.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  const statecCsvText = fs.readFileSync(
+    path.join(__dirname, 'ccvi_states.csv'),
+  );
+  const [, ...stateRows] = await csvParse(statecCsvText);
+
+  const countyCsvText = fs.readFileSync(
+    path.join(__dirname, 'ccvi_counties.csv'),
+  );
+  const [, ...countyRows] = await csvParse(countyCsvText);
```
pnavarrc, line 35: nit: Maybe create a function to avoid repetition here? Also, there is a typo in `statecCsvText` → `stateCsvText`

```ts
const stateRows = readCsv(stateCsvPath);
const countyRows = readCsv(countyCsvPath);
```

src/common/data/index.ts
```diff
@@ -0,0 +1,24 @@
...
+  theme2: number;
+  theme3: number;
+  theme4: number;
+  theme5: number;
+  theme6: number;
+  theme7: number;
+}
+
+export interface FipsToCcviMap {
+  [key: string]: RegionCcviItem;
```
pnavarrc, line 21: Maybe rename `key` → `fipsCode`

src/common/regions/regions_data.ts
```diff
@@ -69,12 +70,14 @@ export const getStateFips = (region: Region): string | null => {
...
   return chain(state_dataset)
     .map(stateInfo => {
+      const ccviData = fipsToCcviData[stateInfo.state_fips_code];
       return new State(
         stateInfo.state,
         stateInfo.state_url_name,
         stateInfo.state_fips_code,
         stateInfo.population,
         stateInfo.state_code,
+        ccviData || {},
```
pnavarrc, line 80: It might be better to default to `null` so we can tell if we don't have CCVI data for a location before we try to access the themes.

pnavarrc (approved):
Looks good!

ghop02:
It looks like that json file is about 700 kb (not sure what it is compressed).   Do we want to be defaulting to async methods so that we're not adding a pretty significant bump in bundle size?

PR #2765, opened 2021-02-17
Fix compare bug + alignment

chasulin:
A previous bug fix to the compare modal introduced a new one. When the pinned row is ranked #1, no other locations are populating the modal (image below). This fixes.

Screen Shot 2021-02-17 at 5 54 09 PM

Also, alignment of the compare filters looked a bit off. This addresses that.

Before:
Screen Shot 2021-02-17 at 5 47 42 PM

After:
Screen Shot 2021-02-17 at 5 47 46 PM

### DIFF
--- a/src/components/Compare/Filters.style.tsx
+++ b/src/components/Compare/Filters.style.tsx
@@ -21,7 +21,7 @@ export const Container = styled.div<{
 }>`
   display: flex;
   margin: ${({ $isModal }) => $isModal && '1rem auto 0'};
-  padding: ${({ $isModal }) => ($isModal ? '0 0 0 1.75rem' : '0.75rem 1rem')};
+  padding: ${({ $isModal }) => ($isModal ? '0 0 0 1.75rem' : '0.75rem 0')};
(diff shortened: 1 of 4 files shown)


pnavarrc:
# Testing Notes

### 1. The pinned locations shows up twice, once pinned and once in the list below.

- Go to https://covid-projections-git-casulin-fix-compare-modal-bug.covidactnow.vercel.app/us/georgia-ga/county/taliaferro_county/?s=1602544
- Sort by population ascending
- Open the modal

image

chasulin:
this rocks! very thorough 🙏🏼
to each point:
1- this is the intended behavior. the location should only be removed from the list if it ranks #1
2- this is a bug. going to fix in a separate PR.
3- also intended behavior (same as point #1)
4- hm yeah. the rank is `0` because we are filtering for metro counties, and the current/pinned county isn't a metro county. instead of `0`, it should probably read `--`. will fix in a separate pr.

pnavarrc (approved):

PR #2766, opened 2021-02-18
Adding script to pull vaccine phases in from spreadsheet

ghop02:
Adds in a script to refresh the CMS data with data from the spreadsheet.  In the future I'd love to automate this a bit more but it works as a cli command for now

### DIFF
--- a/scripts/vaccination_alerts/parse-vaccination-phases.ts
+++ b/scripts/vaccination_alerts/parse-vaccination-phases.ts
@@ -1,0 +1,168 @@
+import fetch from 'node-fetch';
+import { groupBy, sortBy } from 'lodash';
+import * as yargs from 'yargs';
+import path from 'path';
+import fs from 'fs-extra';
+import {
+  stateVaccinationPhases,
+  RegionVaccinePhaseInfo,
+} from '../../src/cms-content/vaccines/phases';
+
+import regions from '../../src/common/regions/region_db';
+
+const parse = require('csv-parse/lib/sync');
+
+const PHASES_PATH = path.join(
+  __dirname,
+  '..',
+  '..',
+  'src',
+  'cms-content',
+  'vaccines',
+  'state-vaccine-phases.json',
+);
+
+const GOOGLE_SHEETS_CSV_URL =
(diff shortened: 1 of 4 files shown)


scripts/vaccination_alerts/parse-vaccination-phases.ts
```diff
@@ -0,0 +1,168 @@
...
+  console.log(
+    `In CMS but not spreadsheet: ${inCmsNotSpreadsheet.map(
+      state => state.locationName,
+    )}`,
+  );
+
+  if (dryRun) {
+    console.log('Dry run -- exiting');
+    return;
+  }
```
pnavarrc, line 140: I'm not sure if we need the dry run mode, the user can always discard changes to the CMS locally before committing them

scripts/vaccination_alerts/parse-vaccination-phases.ts
```diff
@@ -0,0 +1,168 @@
...
+      {
+        regions: sortBy(
+          [...updatedMatchingRecords, ...newStates, ...inCmsNotSpreadsheet],
+          state => state.locationName,
+        ),
+      },
+      null,
+      2,
+    ),
+  );
```
pnavarrc, line 154: Consider creating a variable for the data item so we can check the type before writing to the file.

scripts/vaccination_alerts/parse-vaccination-phases.ts
```diff
@@ -0,0 +1,168 @@
...
+    state: row['State'],
+    eligibilityUrl: row['Eligibility Information URL'],
+    currentlyEligible: row['Currently Eligible'].trimRight() !== '',
+    startDate: row['Estimated Start Date'],
+    phase: row['Phase'],
+    tier: row['Tier'],
+    groupDetails: row['Detailed Group'],
+    generalGroup: row['General Group'],
+    descriptionUrl: row['Expanded description url'],
+    notes: row['Notes for Internal QA'],
```
pnavarrc, line 52: Can we use indices instead to make this more resilient?

pnavarrc (approved):
Looks good!

scripts/vaccination_alerts/parse-vaccination-phases.ts
ghop02: hmm I'm not sure that will really make it more resilient - very possible that a column is added and breaks it.  with indices I'm more concerned the failure will be silent

PR #2770, opened 2021-02-18
Add support for vaccine alert dry runs.

mikelehen:
So @BrettBoval hopefully doesn't have to go poking in Firebase to figure out what alerts are going to be sent. 😁


pnavarrc (approved):
Looks good!

PR #2771, opened 2021-02-18
Use dynamic import to load Firebase to reduce vendor bundle size by ~300k

mikelehen:
I noticed Firebase was eating a pretty big chunk of our vendor bundle size (300k before gzip, 85k after) and we only need it if somebody clicks the share button or signs up for alerts, so I moved it to a dynamic import.

I also deleted the `Newsletter` component which we no longer seem to use.


pnavarrc (approved):
The code is much cleaner and leaner 🙌 🚀

src/screens/AlertUnsubscribe/AlertUnsubscribe.tsx
```diff
@@ -81,13 +85,15 @@ const AlertUnsubscribe = () => {
...
           <BodyCopy>To update preferences for {email}:</BodyCopy>
           <UpdatePreferencesFormWrapper>
-            <AutocompleteRegions
-              regions={regions.all()}
-              selectedRegions={selectedLocations}
-              onChangeRegions={(e, newRegions) =>
-                handleSelectChange(newRegions)
-              }
-            />
+            <div style={{ width: '100%' }}>
```
mikelehen, line 88: dunno if this is best, but it works.

PR #2781, opened 2021-02-18
Vaccination disclaimer/tooltip edit

chasulin:
Under vaccinations chart on county+metro pages, there is currently no tooltip with data source info. This PR removes the "where our data comes from" piece of disclaimer copy when there is no tooltip linked.

Before:
Screen Shot 2021-02-18 at 3 58 44 PM

After:
Screen Shot 2021-02-18 at 3 59 29 PM


pnavarrc (approved):
Looks good!

PR #2783, opened 2021-02-18
Vaccination eligibility module button block

chasulin:


src/components/VaccinationEligibilityBlock/ButtonBlock.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  return (
+    <Wrapper>
+      <StyledLinkButton
+        to="#shareContainer"
+        {...sharedTrackingProps}
+        trackingLabel="Vaccination alerts"
+        startIcon={<EmailAlertIcon />}
+        aria-label="Vaccination alerts signup"
+      >
```
pnavarrc, line 23: Actually I forgot to implement the HashLink version for the `LinkButton` component, this might not work. 😢  - I will test on the preview link

src/components/VaccinationEligibilityBlock/ButtonBlock.tsx
```diff
@@ -0,0 +1,41 @@
...
+        Get notified when eligibility changes
+      </StyledLinkButton>
+      {signupLink && (
+        <StyledLinkButton
+          href={signupLink}
+          {...sharedTrackingProps}
+          trackingLabel="Where to get vaccinated"
+          endIcon={<OpenInNewIcon />}
+          aria-label="Vaccination information"
```
pnavarrc, line 32: We don't need an aria label when the link has text in any of the children elements, the screen reader will read:

> button: See where and how to get vaccinated

we do need `aria-label` for buttons without text (icon-only buttons) though!

pnavarrc (approved):
Looks good!

PR #2785, opened 2021-02-19
Switch createFilterOptions with matchSorter experiment

faihegberg:
Experimenting with using `matchSorter` instead of `createFilterOptions`. Not ready for review.


faihegberg:
After some experimentation and more research into the `matchSorter`, it proves to be far too slow in showing the results from the search bar to the user. Unfortunately, `matchSorter` also does not have a `filterLimit` property or similar which we can utilize in hopes of speeding things up.

PR #2786, opened 2021-02-19
store the vaccination-alerts file in the workflow

pnavarrc:


pnavarrc:
I don't think we need this, closing!

PR #2790, opened 2021-02-19
preview: vaccination info blocks

pnavarrc:


src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -148,9 +154,16 @@ const ChartsHolder = ({ projections, region, chartId }: ChartsHolderProps) => {
           isMobile={isMobile}
           region={region}
         />
-        <MainContentInner>
-          <RegionVaccinationBlock region={region} />
-        </MainContentInner>
+        {!showVaccinationBlock && (
```
mikelehen, line 157: We seem to be showing _a_ vaccination block regardless of this boolean.  Perhaps rename it and add a comment to make it clearer what this is actually toggling.

Also, could/should we move the `<MainContentInner>` element outside of this conditional since it seems to be used in both cases?

src/components/VaccinationEligibilityBlock/VaccinationEligibilityBlock.tsx
```diff
@@ -1,34 +1,43 @@
-import React from 'react';
-import { Heading2, Paragraph } from 'components/Markdown';
-import TabsPanel, { TabInfo } from 'components/TabsPanel';
+import React, { useState } from 'react';
+import { HashLink } from 'react-router-hash-link';
 import { Region, getStateName } from 'common/regions';
```
mikelehen, line 3: HashLink is unused...

src/components/VaccinationEligibilityBlock/VaccinationEligibilityBlock.tsx
```diff
@@ -1,34 +1,43 @@
...
 const VaccinationEligibilityBlock: React.FC<{ region: Region }> = ({
   region,
 }) => {
+  const state = getRegionState(region);
+  assert(
+    state,
+    `VaccinationEligibilityBlock cannot be used for locations with more than one state`,
+  );
+
   // TODO: Assert that the region is a county, state or a single-state metro
```
mikelehen, line 32: obsolete?

src/components/VaccinationEligibilityBlock/VaccinationEligibilityBlock.tsx
```diff
@@ -1,34 +1,43 @@
...
-  const stateName = getStateName(region);
-  const eligibilityData = getEligibilityInfo(region);
+  const stateName = getStateName(state);
```
mikelehen, line 33: Can you do `state.fullname` and then `stateName` won't be `null`-able and you can drop the `Your state` fallbacks below?

mikelehen (approved):
Looks great! Some minor style nits, but nothing major.

PR #2792, opened 2021-02-19
Remove metric export from common/metric to fix circular dependency

ghop02:
This directly fixes https://trello.com/c/5fjRbVaH - I believe there is a circular dependency introduced with importing region.  I think the other option is to move all of the `getMetric.*` functions out into another module but don't have strong opinions and this worked for now


pnavarrc (approved):
Thanks Chris!

PR #2802, opened 2021-02-20
Add Texas Alert

BrettBoval:
Texas counties are starting to report cases again, but are way lower. I think we should keep them for now, add this alert, and not send out email alerts for Texas on Monday.

https://trello.com/c/TdspuIeM/952-texas-reporting-dip-winter-weather-feb-2020

### DIFF
--- a/src/components/LocationPage/Notifications.tsx
+++ b/src/components/LocationPage/Notifications.tsx
@@ -12,7 +12,13 @@ import InfoIcon from '@material-ui/icons/Info';
 import HospitalizationsAlert, {
   isHospitalizationsPeak,
 } from './HospitalizationsAlert';
-import { State, County, Region, MetroArea } from 'common/regions';
+import {
+  State,
+  County,
+  Region,
+  MetroArea,
+  getFormattedStateCode,
+} from 'common/regions';
 import { trackEvent, EventCategory, EventAction } from 'components/Analytics';
 
 const EXPOSURE_NOTIFICATIONS_STATE_FIPS = [
@@ -58,7 +64,7 @@ const NotificationArea: React.FC<{ projections: Projections }> = ({
   } else if (
     // TODO(2021/2/20): TX Winter Reporting
     // https://trello.com/c/TdspuIeM/952-texas-reporting-dip-winter-weather-feb-2020
-    region.fipsCode.startsWith('48')
+    getFormattedStateCode(region)?.includes('TX')
   ) {
     notification = Notification.TXFeb2021Winter;
   } else if (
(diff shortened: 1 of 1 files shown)


src/components/LocationPage/Notifications.tsx
```diff
@@ -98,11 +105,29 @@ const NotificationArea: React.FC<{ projections: Projections }> = ({
...
     </React.Fragment>
   );
 };
 
+const TXFeb2021WinterCopy: React.FC<{ locationName: string }> = ({
+  locationName,
+}) => {
+  return (
+    <Copy>
+      In February 2021, {locationName} experienced extreme winter weather that
```
mikelehen, line 122: How sure are we that we want to say this for all counties in Texas?  It's a big state. Was all of Texas incapacitated, or just some regions?

At minimum perhaps we should say `Texas` instead of `{locationName}` to imply less precision.

src/components/LocationPage/Notifications.tsx
```diff
@@ -48,12 +48,19 @@ const NotificationArea: React.FC<{ projections: Projections }> = ({
...
     None,
   }
   let notification: Notification;
 
   if (showExposureNotifications(region)) {
     notification = Notification.ExposureNotifications;
+  } else if (
+    // TODO(2021/2/20): TX Winter Reporting
+    // https://trello.com/c/TdspuIeM/952-texas-reporting-dip-winter-weather-feb-2020
+    region.fipsCode.startsWith('48')
```
mikelehen, line 61: This breaks in the face of MSA's, e.g. Watertown-Fort Drum, NY has code 48060.  I'll make a fix and push it to this branch.

src/components/LocationPage/Notifications.tsx
BrettBoval: There will be variations in impact, but I wouldn't trust any Texas data for the rest of the month. I'm fine with changing it to say `Texas`.

src/components/LocationPage/Notifications.tsx
mikelehen: Fixed.

src/components/LocationPage/Notifications.tsx
mikelehen: Changed.

mikelehen (approved):

src/components/LocationPage/Notifications.tsx
BrettBoval: We should be worried about giving people over confidence that cases are continuing to decrease. Given delays in reports, decreased test seeking, and closed testing centers, it'll be hard to get a positivity PCR test regardless of the true infection rate.

PR #2809, opened 2021-02-20
a11y: fix the double-focus bug on banner buttons

pnavarrc:
The banner buttons were implemented as a button inside a link, which caused both elements to have a tab stop when navigating using the keyboard. Now that the `LinkButton` component supports `HashLink`, we can replace the implementation, which avoids the double <kbd>tab</kbd> problem.

### DIFF
--- a/src/components/Banner/VaccinationsBanner.style.tsx
+++ b/src/components/Banner/VaccinationsBanner.style.tsx
@@ -2,6 +2,7 @@ import styled from 'styled-components';
 import MuiButton from '@material-ui/core/Button';
 import { materialSMBreakpoint } from 'assets/theme/sizes';
 import { COLOR_MAP } from 'common/colors';
+import LinkButton from 'components/LinkButton';
 
 export const Wrapper = styled.div`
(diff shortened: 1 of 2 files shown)


src/components/Banner/VaccinationsBanner.style.tsx
```diff
@@ -2,6 +2,7 @@ import styled from 'styled-components';
 import MuiButton from '@material-ui/core/Button';
 import { materialSMBreakpoint } from 'assets/theme/sizes';
```
mikelehen, line 3: MuiButton is unused now.

src/components/Banner/VaccinationsBanner.tsx
```diff
@@ -1,5 +1,4 @@
 import React, { Fragment } from 'react';
-import { HashLink } from 'react-router-hash-link';
 import { ButtonContainer } from './Banner.style';
 import { trackEvent, EventAction, EventCategory } from 'components/Analytics';
```
mikelehen, line 3: trackEvent is unused now.

mikelehen (approved):
Cleaner too!

PR #2816, opened 2021-02-22
Typescript conversion

schamp:
Convert a few lingering files to typescript and clean up a few type errors.  This will smooth the way for more static generation stuff, and is a good idea anyway (found a few small bugs).

It does touch a lot of stuff so if you feel like bashing on it a bit to make sure I didn't miss anything I'd appreciate that.


pnavarrc:
Thanks for tackling this @schamp! I think this PR is too big to be reviewed and tested as a single unit, is there an obvious way to split it? Otherwise I might cherry-pick changes from here into smaller PRs so we can merge with confidence.

schamp:
sure, I could try splitting it up.  let me take a stab at that tonight.

schamp:
if I can isolate pure type-only changes (vs. ones that affect behavior) I'm sure that'd make it easier.

schamp:
split into a number of other smaller PRs, so abandoning this one.

PR #2821, opened 2021-02-22
add eligibility link for PR + unit tests

pnavarrc:
This PR adds an eligibility URL for Puerto Rico (using the link from the [CDC: How Do I Get a Vaccine? section](https://www.cdc.gov/vaccines/covid-19/index.html). The target link shows the vaccination phases for Puerto Rico (in Spanish).

I also added unit tests to check that we have the eligibility URL for all the states, and removed the hardcoded check that was hiding that section specifically for Puerto Rico. The tests will show up per state, that way if one ever fails, we will know the state that needs to be fixed:

```
 PASS  src/cms-content/vaccines/index.test.
  vaccination information
    eligibility links
      ✓ AK has eligibility url (2ms)
      ✓ AL has eligibility url
      ✓ AR has eligibility url
      ...
```

We should add tests for the vaccination phases as well, but we might restructure the CMS so all the vaccination information is in one place, so I think it makes more sense to add tests then.

### DIFF
--- a/src/components/VaccinationBlock/VaccinationBlock.tsx
+++ b/src/components/VaccinationBlock/VaccinationBlock.tsx
@@ -1,0 +1,29 @@
+import React from 'react';
+import { Region, MetroArea } from 'common/regions';
+import { getVaccineInfoByFips } from 'cms-content/vaccines/phases';
+import RegionVaccinationBlock from 'components/RegionVaccinationBlock';
+import VaccinationEligibilityBlock from 'components/VaccinationEligibilityBlock';
+import ErrorBoundary from 'components/ErrorBoundary';
+
+const VaccinationBlock: React.FC<{ region: Region }> = ({ region }) => {
+  const vaccineInfo = getVaccineInfoByFips(region.fipsCode);
+
+  const isMultiStateMetro =
+    region instanceof MetroArea && region.states.length > 1;
+
+  // We only have vaccination phases information for states, so we show only the vaccination
+  // links for multi-state metro areas or when we don't have vaccination info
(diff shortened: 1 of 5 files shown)


pnavarrc:
[CMS - Add eligibility link for PR](https://trello.com/c/KODH7JZ0/972-cms-add-eligibility-link-for-pr)

mikelehen (approved):
Code changes LGTM, but the vaccine block isn't actually showing up yet. 😁

src/components/VaccinationBlock/VaccinationBlock.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { Region, MetroArea } from 'common/regions';
+import { getVaccineInfoByFips } from 'cms-content/vaccines/phases';
+import RegionVaccinationBlock from 'components/RegionVaccinationBlock';
+import VaccinationEligibilityBlock from 'components/VaccinationEligibilityBlock';
+import ErrorBoundary from 'components/ErrorBoundary';
+
+const VaccinationBlock: React.FC<{ region: Region }> = ({ region }) => {
+  const vaccineInfo = getVaccineInfoByFips(region.fipsCode);
```
mikelehen, line 9: I think this isn't working for counties.

Perhaps we can make `getVaccineInfo()` automatically map county=>state and return the appropriate info?  And then it can probably do the same for metro=>state and do the "multi state metro" logic to return `null` in that case?

mikelehen (changes requested):

pnavarrc:
Closing in favour of https://github.com/covid-projections/covid-projections/pull/2825

PR #2826, opened 2021-02-23
convert screens/internal to typescript

schamp:


src/screens/internal/AllStates/AllStates.tsx
```diff
@@ -3,22 +3,26 @@ import { LoadingScreen } from './AllStates.style';
...
-function State({ state }) {
+function State({ state }: { state: StateType }) {
   const projections = useProjectionsFromRegion(state);
 
   // Projections haven't loaded yet
   if (!projections) {
     return <LoadingScreen></LoadingScreen>;
   }
-  const stateName = projections.stateName;
+  const stateName = state.name;
```
schamp, line 25: projections didn't have stateName, but we're passing a state in, so it seems reasonable to use the state name from that, if we have found a projectino.

src/screens/internal/AllStates/AllStates.tsx
```diff
@@ -3,22 +3,26 @@ import { LoadingScreen } from './AllStates.style';
...
 import { MetricChart } from 'components/Charts';
-import regions from 'common/regions';
+import regions, { State as StateType } from 'common/regions';
 
 function AllStates() {
-  return regions.states.map(state => {
-    return <State key={state.stateCode} state={state} />;
-  });
+  return (
+    <>
```
schamp, line 10: the context in which AllStates() was used required a single child, not a list, so I wrapped in a <>

src/screens/internal/AllStates/AllStates.tsx
schamp: this actually fixes a bug

pnavarrc (approved):
Looks good!

PR #2828, opened 2021-02-23
convert components/Map|MiniMap to typescript

schamp:


src/components/Map/Map.tsx
```diff
@@ -72,7 +79,7 @@ function Map({
         />
       </div>
       {!hideInstructions && (
-        <MapInstructions $isMiniMap={isMiniMap}>
```
schamp, line 75: $isMiniMap didn't do anything on MapInstructions, so I removed it.

src/components/Map/geo-albers-usa-territories.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'geo-albers-usa-territories';
```
schamp, line 1: this module doesn't have typings, so declaring this got me past some errors, but I'd be happy to find a better approach.

src/components/Map/Map.tsx
```diff
@@ -39,27 +46,27 @@ function Map({
         <Legend hideLegendTitle={hideLegendTitle}>
           <LegendItem
             key={'legend-5'}
-            title={LOCATION_SUMMARY_LEVELS[Level.SUPER_CRITICAL].summary}
+            title={LOCATION_SUMMARY_LEVELS[Level.SUPER_CRITICAL].summary || ''}
```
mikelehen, line 49: FWIW- I'm not sure why `summary` is optional on `LevelInfoMap`.  I think it should always be defined. 🤔   You could explore fixing that if you felt inclined, but also okay to skip that rabbit hole for now. 😁

src/components/Map/USACountyMap.tsx
```diff
@@ -0,0 +1,204 @@
...
+                    .map(geo => {
+                      const fipsCode = geo.id;
+                      const state = regions.findByFipsCode(fipsCode);
+
+                      if (
+                        state === null ||
+                        !Boolean(state) ||
+                        !(state instanceof StateType)
+                      ) {
+                        return null;
```
mikelehen, line 130: Rather than trying to have all of this be conditional on something that shouldn't happen, I think you can use `findByFipsCodeStrict` and we'll just crash if the state is missing.

The old code would have already crashed in this case, since it was using `state.stateCode` before checking if `state` was defined.

src/components/Map/Map.tsx
schamp: It *is* always defined, but because of how the types are set up, typescript can't deduce that (I think).  This would be nice to follow up with better indexing or an enum or something, but I didn't want this work to get stuck on that.

src/components/Map/USACountyMap.tsx
```diff
@@ -0,0 +1,204 @@
...
+ *
+ * This is special cased from the normal map display. The mariana islands are
+ * small enough that simply showing the islands is not a UX that works.
+ */
+const MarianaIslands = ({
+  fill,
+  onMouseEnter,
+  onMouseLeave,
+  onClick,
+  tabIndex,
```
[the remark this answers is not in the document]
schamp: typescript FTW

src/components/Map/USACountyMap.tsx
schamp: agree, I'll just coerce to a state and eliminate the if check

PR #2829, opened 2021-02-23
convert src/assets|common to typescript

schamp:


src/common/colors.ts
```diff
@@ -80,19 +81,19 @@ export const LEVEL_COLOR = {
 };
 
 export function colorFromLocationSummary(
-  summary,
-  defaultColor = COLOR_MAP.GRAY_LIGHT,
```
schamp, line 84: GRAY_LIGHT doesn't exist on COLOR_MAP

pnavarrc (approved):
Looks good! The file `src/common/us_state_centers.ts` is not used anywhere, we could get rid of it.

schamp:
> Looks good! The file `src/common/us_state_centers.ts` is not used anywhere, we could get rid of it.

nice, removed, thanks

PR #2831, opened 2021-02-23
Trim unnecessary precision off of metric values in summaries file.

mikelehen:
We had a lot of unnecessary precision in the metric values in the summaries file.  I'm now rounding to the exact precision we need, which saves ~80kb off of our gzipped main bundle.


pnavarrc (approved):
Looks good! I think that maybe later we could move the responsibility of rendering values to each individual metric file (under the metric definition interface), but this works for now.

PR #2837, opened 2021-02-23
convert some components to typescript

schamp:
these are ones that were more independent and easier to include in a smaller group


pnavarrc (approved):
✅ thanks for updating!

PR #2840, opened 2021-02-23
convert final screens and components to typescript

schamp:
this is the last of the screens and components, one more after this with some top-level files.


src/components/LocationPage/SocialButtons.tsx
```diff
@@ -52,7 +58,7 @@ const SocialButtons = props => {
         </TwitterShareButton>
       </SocialShareButton>
       <SocialShareButton {...buttonProps} color="#007fb1">
-        <LinkedinShareButton url={url} hashtags={[hashtag]} title={quote}>
+        <LinkedinShareButton url={url} title={quote}>
```
schamp, line 61: hashtags is not valid LinkedinSharedButton interface, probably copied from the twitter one.

src/screens/Embed/ProjectionsTab.tsx
```diff
@@ -33,7 +39,7 @@ export default function DataPage({
           <Grid item xs={12}>
             <Spacer />
           </Grid>
-          <Grid container item xs={12} align="center" justify="center">
+          <Grid container item xs={12}>
```
schamp, line 42: also not valid props for Grid

PR #2842, opened 2021-02-23
Start to add Data API specific content - remove embed :'(

ghop02:
This updates the Data API page content.  There will be many more changes to come, but the content here comes from Tom and I.  Planning on adding more sections other than just license.  

Would like to get this in so that the renamed tools page makes more sense.


pnavarrc (approved):
Looks good!

PR #2845, opened 2021-02-23
Add state abbreviation when checking user input

faihegberg:
This PR enables the user to enter state abbreviation by implementing a functionality where as soon as the user enters just 2 letters (not case-sensitive), check for whether these 2 letters represent a state. If so, surface the state to the top of the dropdown menu.

Some notes:
- This PR adds an additional hook, `checkForStateMatch`, which determines whether `Autocomplete` should search from the beginning of the user input string or at any point in the user input string. This serves the exact same purpose as the `checkForZipcodeMatch` hook. But I kept it separate to keep the state check functionality more explicit and in case we need to add additional functionality to the state check in the future. If this is not necessary or is overkill, we can probably combine `checkForStateMatch` and `checkForZipcodeMatch` into one hook because they technically achieve the same thing.
- Adding the state abbreviation check logic results in multiple `if-else` in `onInputChange` and `stringifyOption`. So I tried to clean it up a little with some shorthand. Please let me know if you think this sacrifices readability or makes the code unintuitive, I can revert it back 😄

### DIFF
--- a/src/components/Search/Homepage/HomepageSearchAutocomplete.tsx
+++ b/src/components/Search/Homepage/HomepageSearchAutocomplete.tsx
@@ -1,7 +1,7 @@
 import React, { useState } from 'react';
 import { Autocomplete } from '@material-ui/lab';
 import { createFilterOptions } from '@material-ui/lab/useAutocomplete';
-import { Region, MetroArea } from 'common/regions';
+import { Region, MetroArea, State } from 'common/regions';
 import {
   Wrapper,
   StyledTextField,
@@ -31,6 +31,7 @@ const HomepageSearchAutocomplete: React.FC<{
   const [input, setInput] = useState('');
   /* We only check for a zipcode match when the input is all numbers and has a length of 5: */
   const [checkForZipcodeMatch, setCheckForZipcodeMatch] = useState(false);
+  const [checkForStateMatch, setCheckForStateMatch] = useState(false);
   const [noOptionsCopy, setNoOptionsCopy] = useState('No location found');
   const [isOpen, setIsOpen] = useState(false);
   const { countyToZipMap } = useCountyToZipMap();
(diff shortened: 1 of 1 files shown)


src/components/Search/Homepage/HomepageSearchAutocomplete.tsx
```diff
@@ -43,30 +44,39 @@ const HomepageSearchAutocomplete: React.FC<{
...
     const isStringOfDigits = /^\d+$/.test(value);
+    const isStringOfTwoLetters =
+      value.length === 2 && /^[a-zA-Z]+$/.test(value);
     if (isStringOfDigits) {
       setNoOptionsCopy('Enter a valid 5-digit zip code');
-      if (value.length === 5) setCheckForZipcodeMatch(true);
-      else setCheckForZipcodeMatch(false);
+      value.length === 5
+        ? setCheckForZipcodeMatch(true)
+        : setCheckForZipcodeMatch(false);
```
chasulin, line 53: want to keep with the if/else syntax for readability?

src/components/Search/Homepage/HomepageSearchAutocomplete.tsx
```diff
@@ -118,7 +128,8 @@ const HomepageSearchAutocomplete: React.FC<{
           onChange={onSelect}
           getOptionSelected={getOptionSelected}
           filterOptions={createFilterOptions({
-            matchFrom: checkForZipcodeMatch ? 'any' : 'start',
+            matchFrom:
+              checkForZipcodeMatch || checkForStateMatch ? 'any' : 'start',
```
chasulin, line 132: if we set `matchFrom` to 'any' whenever the input is 2 letters, it starts checking for matches against any part of a location name, which isnt exactly what we want. for example, an input of 'ct' returns 'Distri**ct** of Columbia' and 'Grand Jun**ct**ion metro, CO':

Screen Shot 2021-02-23 at 4 29 34 PM

faihegberg:
After some experimentation and test implementation, we concluded that implementing this right now will introduce some unwanted side effects. Summary of actions taken for when this is revisited in the future:
- Tried implementing `MatchSorter` in the `FilterOptions` property of `Autocomplete` component. This ended up taking far too slow to load results for users.
- Implemented a hook to check whether the user enters exactly 2 letters. When this happens, check for state and surface it if there's a match. But this has an unwanted side effect of returning all locations where the 2 letters are included anywhere in the name of those locations (e.g. "FL" will return "Florida" first but "Cape Coral metro, FL" immediately after instead of "Flint metro, MI"), which is not the use case that we're looking for.

PR #2847, opened 2021-02-23
Visually hidden div for accessibility 

chasulin:
- Extracts `VisuallyHidden` div into something more generic + reusable for accessibility
- Uses uuid to give each tooltip a unique ID for `aria-describedby` attribute


src/components/InfoTooltip/DisclaimerTooltip.tsx
```diff
@@ -1,14 +1,17 @@
...
+import VisuallyHiddenDiv from 'components/VisuallyHidden/VisuallyHidden';
 import { useBreakpoint } from 'common/hooks';
 import { tooltipAnchorOnClick } from 'components/InfoTooltip';
+import { v4 as uuidv4 } from 'uuid';
 
 const DisclaimerTooltip: React.FC<StyledTooltipProps> = props => {
   const [isOpen, setIsOpen] = useState(false);
   const isMobile = useBreakpoint(600);
 
+  const idForAccessability = `${uuidv4()}`;
```
pnavarrc, line 13: You don't need the string interpolation, can just use

```ts
const idForAccessability = uuidv4();
```

src/components/InfoTooltip/InfoTooltip.tsx
```diff
@@ -19,6 +20,8 @@ const InfoTooltip: React.FC<StyledTooltipProps> = props => {
     props.trackCloseTooltip();
   };
 
+  const idForAccessability = `${uuidv4()}`;
```
pnavarrc, line 23: Also here

src/components/VisuallyHidden/VisuallyHidden.tsx
```diff
@@ -0,0 +1,32 @@
...
+  overflow: hidden;
+  position: absolute;
+  white-space: nowrap;
+  width: 1px;
+`;
+
+const VisuallyHiddenDiv: React.FC<{
+  content: React.ReactNode;
+  elemId: string;
+}> = ({ content, elemId }) => {
```
pnavarrc, line 24: I don't think we need the `VisuallyHiddenDiv` component, we can just pass the id, children and role to the `HiddenDiv` component directly:

```tsx
<HiddenDiv id={idForAccessability} role="tooltip">
  {props.title}
</HiddenDiv>
```

pnavarrc (approved):
Looks good!

src/components/VisuallyHidden/VisuallyHidden.tsx
chasulin: per our chat, keeping it as is

PR #2859, opened 2021-02-24
Cms content entry

Ryan-Fuentes:
site changes for Connecticut (no new vaccine alerts), adds join us + contact us sections to about page, adds logos for api page (but not yet pulled into the ui)


BrettBoval (approved):
LGTM

PR #2861, opened 2021-02-24
Add register button

ghop02:
Adds register button to Data API
image


src/screens/DataApi/DataApi.style.tsx
```diff
@@ -1,7 +1,28 @@
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
 import { MarkdownContent } from 'components/Markdown';
 import { COLOR_MAP } from 'common/colors';
 import theme from 'assets/theme';
+import LinkButton from 'components/LinkButton';
+
+const SharedButtonStyles = css`
```
chasulin, line 7: you can import this in from SectionButton.style instead of rewriting

src/screens/DataApi/DataApi.style.tsx
```diff
@@ -1,7 +1,28 @@
...
+  font-size: 1rem;
+  text-transform: none;
+  text-decoration: none;
+  color: inherit;
+  padding: ${0.5 * theme.spacing(1)}px ${theme.spacing(2)};
+`;
+
+export const GreenLinkButton = styled(LinkButton)`
```
chasulin, line 15: oh! this too

src/screens/DataApi/DataApi.style.tsx
ghop02: Do you think the `SectionButton` stuff should eventually be moved to a more common place? Feels more uncomfy to import directly from learn code (though I'm not sure I 100% follow that philosophy)

chasulin: buttons are soon to be redone anyway, so i guess its just fine to leave this all as is

chasulin (approved):

PR #2862, opened 2021-02-24
Sort location search results by name (asc) then population (desc)

faihegberg:
Currently the search bar on the homepage and location page returns location results in the following order: state, metro area, county. Each of these groups are sorted alphabetically.

This PR sorts search results within each group alphabetically, but if two locations have the same name, then it will sort these locations by population in descending order.

### DIFF
--- a/src/common/regions/utils.ts
+++ b/src/common/regions/utils.ts
@@ -22,6 +22,20 @@ export function belongsToState(county: County, stateFips: string | null) {
 const sortByPopulation = (regions: Region[]): Region[] =>
   sortBy(regions, region => -region.population);
 
+/* Sort regions by name in ascending order.
+If two regions have the same name, sort them by population in descending order. */
+function sortByNameThenPopulation(locations: Region[]) {
+  const sortedLocations = [...locations];
+  sortedLocations.sort(compareRegionsByNameAndPopulation);
+  return sortedLocations;
+}
+
+function compareRegionsByNameAndPopulation(region1: Region, region2: Region) {
+  if (region1.name === region2.name)
+    return region1.population < region2.population ? 1 : -1;
+  else return region1.name < region2.name ? -1 : 1;
+}
+
 /**
  * Returns a list of regions in the order that is most relevant given the
  * region passed as argument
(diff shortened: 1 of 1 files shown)


src/common/regions/utils.ts
```diff
@@ -32,6 +46,7 @@ const sortByPopulation = (regions: Region[]): Region[] =>
  */
 export function getAutocompleteRegions(region?: Region): Region[] {
   const { states, metroAreas, counties } = regions;
+  [states, metroAreas, counties].forEach(sortByNameThenPopulation);
```
pnavarrc, line 49: This implementation works, but it mutates the `states`, `metroAreas` and `counties` [in place](https://en.wikipedia.org/wiki/In-place_algorithm) instead or returning a new array of sorted results.

In this particular case, that doesn't cause any problems, but in general, mutating function parameters can make code difficult to debug. Consider creating a copy and sorting the copy. The code might be easier to read if we separate the comparison function too

```ts
function compareRegionsByNameAndPopulation(regionA: Region, regionB: Region) {
  /* ... */
}

function sortByNameAndPopulation = (locations: Region[]) => {
  const sortedLocations = [...locations]; // this creates a copy
  sortedLocations.sort(compareRegionsByNameAndPopulation);
  return sortedLocations;
}

// ...

return concat<Region>(
  sortByNameAndPopulation(states),
  sortByNameAndPopulation(metroAreas),
  sortByNameAndPopulation(counties),
)
```

As I mentioned, this change is not strictly necessary and is more of a coding styling thing, I always try to avoid mutating function arguments when I can!

src/common/regions/utils.ts
```diff
@@ -22,6 +22,20 @@ export function belongsToState(county: County, stateFips: string | null) {
...
+const sortByNameThenPopulation = (locations: Region[]) => {
+  locations.sort((region1, region2) => {
+    if (region1.name < region2.name) return -1;
+    else if (region1.name > region2.name) return 1;
+    else if (region1.name === region2.name) {
+      if (region1.population < region2.population) return 1;
+      else if (region1.population > region2.population) return -1;
+    }
+    return 0;
+  });
```
pnavarrc, line 36: Is there a way to simplify this a bit?

```ts
if (region1.name === region2.name) {
  return region1.population > region2.population ? -1 : 1;
} else {
  return region1.name < region2.name ? -1 : 1
}
```

pnavarrc (commented):
Added some notes!

pnavarrc (approved):
Looks good!

pnavarrc:
[King WA ranked behind King TX](https://trello.com/c/DE5Xu9T9/780-king-wa-ranked-behind-king-tx)

PR #2863, opened 2021-02-24
remove location name on vaccination events

pnavarrc:
[Trello](https://trello.com/c/kJpd5apB/908-google-analytics-add-a-custom-dimension-for-the-type-of-page-state-county-metro-home-etc-to-filter-search-events) - Remove location names from vaccination event labels.

The original intent of adding the location name on the events tracking interactions with the vaccination block was to be able to tell in which locations user were interacting with the links more often. That makes difficult to aggregate the events by type of link (Eligibility or Options) across all locations.

Also, there we have [Content Groups](https://support.google.com/analytics/answer/2853423?hl=en) set up for state, metro and county pages, so we can use that as secondary dimension when looking at vaccination events.


pnavarrc:
[Analytics - remove location names from event labels](https://trello.com/c/kJpd5apB/908-analytics-remove-location-names-from-event-labels)

mikelehen (approved):
Makes sense, thanks!

PR #2864, opened 2021-02-24
analytics: remove unused tracking scripts and fix page title tracking

pnavarrc:
[Trello](https://trello.com/c/KQdb7uEx/740-update-the-page-title-in-ga-every-time-we-navigate) - update the page title when tracking page views.

The page title should be tracked automatically by `ReactGA.pageview`, however, the page view event is fired before the title has time to update. This PR adds a small delay (it seems to be enough) between the change in the URL and the moment when we track the page view so it has the correct title.

I also commented out optimize and removed the linkid plugin (both unused).

### DIFF
--- a/src/components/Analytics/PageviewTracker.tsx
+++ b/src/components/Analytics/PageviewTracker.tsx
@@ -44,8 +44,11 @@ function usePageTracking() {
   useEffect(() => {
     if (initialized) {
       // We only add the legacy tracker explicitly here, because by default, GA
-      // will send always send hits to the default tracker.
-      ReactGA.pageview(pathname, [legacyTracker.name]);
+      // will send always send hits to the default tracker. We add a delay to
+      // give the app time to update the page title.
+      setTimeout(() => {
+        ReactGA.pageview(pathname, [legacyTracker.name], document.title);
+      }, 200);
     }
   }, [initialized, pathname]);
 }
--- a/src/components/Analytics/PageviewTracker.tsx
+++ b/src/components/Analytics/PageviewTracker.tsx
@@ -17,7 +17,6 @@ function initializeGA() {
       ...options,
     },
   ]);
-  ReactGA.plugin.require('linkid');
 }
 
 /**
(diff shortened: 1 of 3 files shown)


pnavarrc:
[Update the page title in GA every time we navigate](https://trello.com/c/KQdb7uEx/740-update-the-page-title-in-ga-every-time-we-navigate)

src/components/Analytics/PageviewTracker.tsx
```diff
@@ -45,8 +44,11 @@ function usePageTracking() {
   useEffect(() => {
     if (initialized) {
       // We only add the legacy tracker explicitly here, because by default, GA
-      // will send always send hits to the default tracker.
-      ReactGA.pageview(pathname, [legacyTracker.name]);
+      // will send always send hits to the default tracker. We add a delay to
+      // give the app time to update the page title.
+      setTimeout(() => {
+        ReactGA.pageview(pathname, [legacyTracker.name], document.title);
+      }, 200);
```
mikelehen, line 51: FWIW-  This makes me a tiny bit nervous... e.g. people that bounce off the page might be less likely to be counted as a pageview now...  If you think of it, it would be good to sanity check the page views in GA after a few days to make sure it looks stable.

src/components/Experiment/Experiment.tsx
```diff
@@ -24,6 +24,9 @@ import {
  *
  * The code always defaults to version A if the experiment is not running
  * (or ended) and we still have both versions in the code.
+ *
+ * Note: The optimize script needs to be uncommented from index.html before
+ * starting experiments.
```
mikelehen, line 29: FWIW I wonder if there's an npm module for google-optimize that we could dynamically import here.  That'd be ideal, since then we wouldn't load the script until you actually get to a page that has the experiment.  Something to maybe poke at briefly next time we do an experiment.

mikelehen (approved):

src/components/Analytics/PageviewTracker.tsx
pnavarrc: I tested closing the tab immediately after navigating to a page and we don't capture the page view - I'm going to try something else, good catch!

PR #2865, opened 2021-02-24
remove vaccination icon from email alerts form

pnavarrc:
[Trello](https://trello.com/c/6cbTzuj9/984-get-rid-of-the-sweeet-vaccine-icon-on-alert-signups) - Remove the vaccine icon on alert signups. It also removes the info block that explains the icon, and the modal that was included with it.

I removed the location name from the source link tracking because there is a different way to get the locations in the events view in GA.


pnavarrc:
[Get rid of the sweeet vaccine icon on alert signups](https://trello.com/c/6cbTzuj9/984-get-rid-of-the-sweeet-vaccine-icon-on-alert-signups)

chasulin (approved):
lgtm

PR #2869, opened 2021-02-25
CCVI tooltips in cms

chasulin:
Uses stand-in copy

### DIFF
--- a/src/cms-content/tooltips/index.ts
+++ b/src/cms-content/tooltips/index.ts
@@ -1,8 +1,11 @@
 import { Markdown } from '../utils';
 import infoToolTipContent from './info-tooltips.json';
 import metricCaculationTooltipContent from './metric-calculation-tooltips.json';
+import ccviTooltipContent from './ccvi-tooltips.json';
 import { Metric } from 'common/metricEnum';
 import { assert } from 'common/utils';
+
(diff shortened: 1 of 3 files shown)


src/cms-content/tooltips/index.ts
```diff
@@ -48,3 +51,34 @@ export const locationPageHeaderTooltipContent: Tooltip = getTooltipContentById(
...
+function getCcviTooltipById(id: string) {
+  const tooltipContent = ccviTooltipContent.subscoreTooltips.find(
+    item => item.id === id,
+  );
+  assert(
+    tooltipContent,
+    `Ccvi tooltip content unexpectedly not found for id: ${id}`,
+  );
+  return tooltipContent;
+}
```
pnavarrc, line 72: It might be easier to create a map instead of finding through the array each time:

```
const tooltipContentById = _.keyBy(ccviTooltipContent, item => item.id);
function getCcviTooltipById(id: string) {
  const tooltipContent = tooltipContentById[id];
  assert(
    tooltipContent,
    `Ccvi tooltip content unexpectedly not found for id: ${id}`,
  );
  return tooltipContent;
}
```

pnavarrc (approved):
Looks good! Let's add unit tests against each ID later to make sure we don't miss any before launching

PR #2871, opened 2021-02-25
analytics: setup amplitude

pnavarrc:
[Trello](https://trello.com/c/SdOBDCcd/1004-dual-write-our-data-into-amplitude-in-addition-to-google-analytics) - Setup Amplitude and wire up events

**Note**: Amplitude records events with a name (event type) and an object of properties. I'm using the event category as event type and passing all the other GA event properties as an object: 

| **Google Analytics** | **Amplitude** |
| --- | --- |
| Event category | Event type |
| Event category | eventCategory |
| Event action | eventAction |
| Event label | eventLabel |
| Event value | eventValue |

The Chrome extension [Amplitude Instrumentation Explorer](https://chrome.google.com/webstore/detail/amplitude-instrumentation/acehfjhnmhbmgkedjmjlobpgdicnhkbp/related?hl=en-US) shows the events in a popup, we are registering the same number of events than GA.

I created a separated project for non-production environments (Covid Act Now Dev)


pnavarrc:
[Dual write our data into Amplitude in addition to Google Analytics.](https://trello.com/c/SdOBDCcd/1004-dual-write-our-data-into-amplitude-in-addition-to-google-analytics)

src/components/Analytics/PageviewTracker.tsx
```diff
@@ -3,6 +3,18 @@ import { useEffect, useState } from 'react';
...
 import ReactGA from 'react-ga';
 import { defaultTracker, legacyTracker } from './utils';
+import amplitude from 'amplitude-js';
+
+const AMPLITUDE_KEY_DEV = '9273bc15ce71641291d471c9f17895a5';
+const AMPLITUDE_KEY_PROD = 'c92804b9b1f5323200e94002a76a86a9';
+
+function initializeAmplitude() {
+  const amplitudeKey =
+    process.env.NODE_ENV === 'production'
```
mikelehen, line 13: Hrm...  we have an existing `getEnvironment()` function that uses `process.env.REACT_APP_ENVIRONMENT`... Not 100% sure if they behave the same, but let's reuse that if we can.

https://github.com/covid-projections/covid-projections/blob/6add9e54681a519c44af9b1c72a4b98595a81a9c/src/common/utils/environment.ts#L8

mikelehen (approved):
Adds 26k to our bundle, but otherwise looks great! 😛

PR #2889, opened 2021-02-26
proof of concept: next.js

pnavarrc:
# Proof of Concept: Location Pages with Next.js

- Performance
- Loading data for each location only
- Static generation of content: 
  - SEO & faster rendering

## Routing

Routing in Next.js is defined by the folder structure of the `pages` directory. In this demo, the `pages/index.tsx` file represents the home page, and `pages/us/[state]/county/[county].tsx` represents a county page. The files or folders in square brackets `[state]` and `[county].tsx` represent dynamic parts of the URL

```sh
pages
├── index.tsx
└── us
    └── [state]
        └── county
            └── [county].tsx
```

## Page

The Next.JS page is just a React component that takes props, with two additional exported functions, `getStaticProps` and `getStaticPaths`

```tsx
function CountyPage({ region, locationSummary }) {
  // render the component
  return <div>{/* ... */}</div>;
}

export async function getStaticProps() {}
export async function getStaticPaths() {}
export default CountyPage;
```

### `getStaticPaths`

This function returns all the possible paths for the `CountyPage` component and additional context information for the component that can be used to generate static props. In this case, it just uses the regions to generate the relative path of the county page. Note that `common/regions` is imported dynamically to make sure that is not a dependency of the `CountyPage` component (not included in the bundle)

```ts
export async function getStaticPaths() {
  const { default: regions } = await import('../../../../src/common/regions');

  const paths = regions.counties.map(
    county => `/us/${county.state.urlSegment}/county/${county.urlSegment}/`,
  );
  return { paths, fallback: false };
}
```

### `getStaticProps`

This function uses the URL params and potentially more contextual information to generate static props for the component. In this case, we are dynamically importing the locationSummary for each county and the main region props. Note that the returned `props` need to be JSON serializable

```ts
export async function getStaticProps({ params }) {
  const { state, county } = params;
  const { default: regions } = await import('../../../../src/common/regions');

  const region = regions.findCountyByUrlParams(state, county);
  // get projections

  const locationSummaries = await import(
    '../../../../src/common/location_summaries'
  );
  const locationSummary = locationSummaries.getSummaryFromFips(region.fipsCode);

  return {
    props: {
      region: {
        fullName: region.fullName,
        name: region.name,
        fipsCode: region.fipsCode,
        population: region.population,
        state: {
          fipsCode: region.state.fipsCode,
          stateCode: region.state.stateCode,
        },
      },
      locationSummary,
    },
  };
}

```

## Static Build

Next.Js uses the `generateStaticPaths` function to generate the tree of files for the county pages, and uses `getStaticProps` to pass the props which will end up being hardcoded in the resulting HTML file. The generated HTML file will have the values hardcoded + the JSON props in the HTML itself, along with the JS code to render the page

```html
  <body>
    <div id="__next">
      <div>
        <h1>Adams County, Washington</h1>
        <span>risk level: 2</span>
        <h2>
          Overall risk:
          <!-- -->2
        </h2>
        <ul>
          <li>level: 0, value: 0.62</li>
          <li>level: 0, value: 0.028</li>
          <li>level: 4, value: null</li>
          <li>level: 2, value: 14.3</li>
          <li>level: 4, value: 0.105</li>
        </ul>
      </div>
    </div>
    <script id="__NEXT_DATA__" type="application/json">
      {
        "props": {
          "pageProps": {
            "region": {
              "fullName": "Adams County, Washington",
              "name": "Adams County",
              "fipsCode": "53001",
              "population": 19983,
              "state": { "fipsCode": "53", "stateCode": "WA" }
            },
            "locationSummary": {
              "level": 2,
              "metrics": {
                "0": { "value": 0.62, "level": 0 },
                "1": { "value": 0.028, "level": 0 },
                "2": { "value": null, "level": 4 },
                "5": { "value": 14.3, "level": 2 },
                "6": { "value": 0.105, "level": 4 }
              },
              "ccvi": 0.76854505
            }
          },
          "__N_SSG": true
        },
        "page": "/us/[state]/county/[county]",
        "query": { "state": "washington-wa", "county": "adams_county" },
        "buildId": "Ozy0I5bUkjgK25utMmGhE",
        "nextExport": false,
        "isFallback": false,
        "gsp": true
      }
    </script>
</body>
```

The files in the `out` directory can be statically served. The generated HTML code render really fast (100 performance score), however, the demo is not realistic at all, since the demo page doesn't have any styling.

# Testing

- Run on dev `yarn next:dev` (open localhost:3000)
- To create a static export: `yarn next:build && yarn next:export`
- To serve the static export: `yarn next:serve` (served in port 5000 by default)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
 # See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
 
+.next
+out
+
 # storybook
 storybook-static
 
--- a/next-env.d.ts
+++ b/next-env.d.ts
@@ -1,0 +1,2 @@
+/// <reference types="next" />
+/// <reference types="next/types/global" />
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "js-yaml": "^4.0.0",
     "lodash": "^4.17.19",
     "moment": "^2.24.0",
+    "next": "^10.0.7",
     "postinstall-postinstall": "^2.1.0",
     "query-string": "^6.11.1",
     "react": "^16.13.0",
@@ -128,7 +129,12 @@
     "storybook": "start-storybook -p 6006 -s public --no-dll",
     "build-storybook": "build-storybook -s public --no-dll",
     "analyze": "cross-env REACT_APP_ENVIRONMENT=prod yarn build && source-map-explorer build/static/js/*.js",
-    "generate-ccvi": "ts-node-script -r tsconfig-paths/register scripts/ccvi/parse_ccvi.ts"
+    "generate-ccvi": "ts-node-script -r tsconfig-paths/register scripts/ccvi/parse_ccvi.ts",
+    "next:dev": "next dev",
+    "next:build": "next build",
+    "next:start": "next start",
+    "next:export": "next export",
+    "next:serve": "serve -s out"
   },
   "lint-staged": {
     "{src,scripts,firebase/functions/src,}/**/*.{css,scss,js,ts,tsx}": [
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -1,0 +1,24 @@
+import Link from 'next/link';
+
+function HomePage() {
+  return (
+    <>
+      <div>
+        <ul>
+          <li>
+            <Link href="/us/washington-wa/county/king_county">
+              King County, WA
+            </Link>
+          </li>
+        </ul>
+      </div>
+      <main>{/* Content */}</main>
+      <footer>
+        {/* Embed */}
+        {/* Footer */}
+      </footer>
+    </>
+  );
+}
+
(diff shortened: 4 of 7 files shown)


schamp:
a few notes:
1. per the docs (https://nextjs.org/docs/basic-features/data-fetching) the page definition files can statically `import` anything you want, it won't affect load time, as they are just used on the server and to statically generate the client.  so no need to `await import` on those if you don't want to.

actually, that's the only note.  I think the performance is going to be good but I worry about the impact / amount of rework to deal with the router changes.  I'm also curious to see how much longer the build (which statically generates all of these) takes.

pnavarrc:
Closing to clean up our PRs, labelling it with performance to make it easier to find!

PR #2891, opened 2021-02-27
Vulnerabilities footer links

chasulin:
Adds footer link block:

Screen Shot 2021-02-26 at 11 47 46 PM

To test:
`yarn storybook`
'VulnerabilitiesFooterLinks' under 'Shared Components'

Also adds utils (`renderStateDescription`, `renderCountyDescription`) to get region score description text (per 'location page copy' section of [the copy doc](https://docs.google.com/document/d/1svP2I63M6JUGuNAM1QoAa6IOC8Vd7uYluenRi8wT_Zk/edit#))

### DIFF
--- a/src/common/ccvi/renderRegionDescription.tsx
+++ b/src/common/ccvi/renderRegionDescription.tsx
@@ -12,7 +12,7 @@ export function renderStateDescription(
     return (
       <>
         is one of the most vulnerable states, with{' '}
-        <strong>{overallScore}</strong>% of the population in a high
+        <strong>{scoreAsPercent}</strong> of the population in a high
         vulnerability area.{' '}
       </>
     );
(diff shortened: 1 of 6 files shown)


src/common/ccvi/renderRegionDescription.tsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react';
+import { CcviLevel, getCcviLevel } from './index';
+import { formatPercent } from 'common/utils';
+
+export function renderStateDescription(
```
chasulin, line 5: should these go in `CCVI_LEVEL_INFO_MAP` 🤔 ? or cleaner to keep them here

pnavarrc: Maybe moving it to the map would be better so all the level information is in the same place, under a `renderDescription` function property in case we need to pass parameters such as the location name.

pnavarrc (approved):
Looks good!

src/common/ccvi/renderRegionDescription.tsx
chasulin: adding a todo + going to merge in as is for now

PR #2901, opened 2021-02-28
remove unused code

schamp:
found some code that wasn't used, deleting it to help clarify true dependencies on large data imports


mikelehen (approved):
Nice!

PR #2904, opened 2021-02-28
clean up compare style

schamp:
one more of these

### DIFF
--- a/src/components/Compare/Compare.style.tsx
+++ b/src/components/Compare/Compare.style.tsx
@@ -309,7 +309,7 @@ const SharedCellStyles = css`
 `;
 
 export const LocationNameCell = styled.td<{
-  iconColor?: Level;
+  $iconColor?: Level;
   sortByPopulation: boolean;
 }>`
   ${SharedCellStyles}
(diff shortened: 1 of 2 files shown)


src/components/Compare/Compare.style.tsx
```diff
@@ -382,7 +382,7 @@ export const Row = styled(TableRow)<{
   background-color: ${({ index, $headerRowBackground }) =>
     !isNumber(index)
       ? `${$headerRowBackground}`
-      : index % 2 === 0
+      : (index ?? 0) % 2 === 0
```
mikelehen, line 385: Hrm... I'm not sure why this change was necessary.  `isNumber` is typed as:

```
        isNumber(value?: any): value is number;
```

So typescript's type narrowing should know that at this point in the code it is a number and you shouldn't need a `??` fallback.

Were you seeing something different? Am I confused?

mikelehen (approved):
LGTM with one non-blocking confusion on my part.

src/components/Compare/Compare.style.tsx
schamp: you are right, and I was confused by it as well, but next's typescript compiler was complaining, and this took care of it.  I can take it out for the time being, as it's unnecessary, but since it is still correct and works, I'm inclined to leave it as one less thing to cleanup later.

PR #2905, opened 2021-03-01
fix style props

schamp:
more of these I overlooked the first time around.

### DIFF
--- a/src/components/LocationPage/LocationPageHeader.style.tsx
+++ b/src/components/LocationPage/LocationPageHeader.style.tsx
@@ -20,12 +20,12 @@ export const ColoredHeaderBanner = styled(Box)`
 `;
 
 export const Wrapper = styled(Box)<{
-  condensed?: Boolean;
-  headerTopMargin: number;
-  headerBottomMargin: number;
+  $condensed?: Boolean;
+  $headerTopMargin: number;
+  $headerBottomMargin: number;
 }>`
   ${props =>
-    props.condensed
+    props.$condensed
       ? `
     display: block;
   `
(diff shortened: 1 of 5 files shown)


src/components/LocationPage/LocationPageHeader.style.tsx
```diff
@@ -20,12 +20,12 @@ export const ColoredHeaderBanner = styled(Box)`
 `;
 
 export const Wrapper = styled(Box)<{
-  condensed?: Boolean;
-  headerTopMargin: number;
-  headerBottomMargin: number;
+  $condensed?: Boolean;
```
pnavarrc, line 23: I think this is supposed to be `boolean` instead of `Boolean`, maybe we can fix that as well in this PR?

src/components/SummaryStats/SummaryStats.style.tsx
```diff
@@ -4,19 +4,19 @@ import palette from 'assets/theme/palette';
 import { COLOR_MAP } from 'common/colors';
 
 export const SummaryStatsWrapper = styled(Box)<{
-  condensed?: Boolean;
-  isHeader?: Boolean;
+  $condensed?: Boolean;
+  $isHeader?: Boolean;
```
pnavarrc, line 8: Also here

pnavarrc (approved):
Thanks for cleaning up - all the instances of `Boolean` as a type should be `boolean`, feel free to update or leave that for later

src/components/LocationPage/LocationPageHeader.style.tsx
schamp: surprised the typechecker didn't freakout about this, hm.

PR #2906, opened 2021-03-01
Remove invalid override from MuiSwitch

schamp:
MUI was complaining about this.

### DIFF
--- a/src/assets/theme/overrides/MuiSwitch.tsx
+++ b/src/assets/theme/overrides/MuiSwitch.tsx
@@ -17,5 +17,4 @@ export default {
       'outline-offset': '2px',
     },
   },
-  focusVisible: {},
 };


schamp:
this might not be quite right

pnavarrc:
It looks like the switch component works with this change, however, `MuiSwitch` inherits from `IconButton` which in turns inherits from `ButtonBase`, and [`focusVisible` is a customizable property for `ButtonBase`](https://material-ui.com/api/button-base/#css), so I don't think is necessarily incorrect.

chasulin (approved):
focus state still works, so i think this is fine

Screen Shot 2021-03-02 at 10 39 56 AM

schamp:
thanks for having a look.  there were actually additional warnings related to this change, and to eliminate them I removed some additional things that I haven't incorporated into this PR yet.   and I was digging into something else and haven't had time to test, so I want to revisit this later.

schamp:
I've lost context on this, so will abandon.  I may try again later.

PR #2907, opened 2021-03-01
Remove unused REVERSED_STATES

schamp:
part of earlier cleanup I overlooked, no longer used anywhere in the source tree


pnavarrc (approved):
Thanks for cleaning up!

PR #2910, opened 2021-03-01
Add vulnerability to compare table.

mikelehen:
1. Adds "Vulnerability Level" to compare.
2. Tweaks paddings / margins / font size so it fits.

image

### DIFF
--- a/src/components/Compare/columns.ts
+++ b/src/components/Compare/columns.ts
@@ -13,6 +13,10 @@ const CCVI_COLUMN_ID = 100;
 /**
  * Represents a column in the compare table (e.g. for the case density metric or CCVI).
  * Implementations must define the name, getValue() function, etc. as appropriate.
+ *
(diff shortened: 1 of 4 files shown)


src/components/Compare/columns.ts
```diff
@@ -13,6 +13,10 @@ const CCVI_COLUMN_ID = 100;
 /**
  * Represents a column in the compare table (e.g. for the case density metric or CCVI).
  * Implementations must define the name, getValue() function, etc. as appropriate.
+ *
+ * TODO(michael): Refactor this to have a `cellContent(row: SummaryForCompare)` method that
+ * returns the rendered React component and then we can probably drop textAlign, fontSize,
+ * getFormattedValue(), getIconColor().
```
mikelehen, line 19: I might take a stab at this refactor later today, since I think it'd be cleaner, but I wanted to get this PR sent out so we can get it into the preview link.

chasulin (approved):

PR #2924, opened 2021-03-02
DEI: add percent population vulnerable for states

chasulin:


src/common/ccvi/renderRegionDescription.tsx
```diff
@@ -18,46 +19,50 @@ export function renderRegionDescription(
...
 ): React.ReactElement {
   const level = getCcviLevel(overallScore);
 
   if (level === CcviLevel.VERY_HIGH) {
     return (
       <>
-        is one of the <strong>most vulnerable</strong> states, with{' '}
-        {vulnerablePopulation} of the population in a high vulnerability area.
+        is one of the <strong>most vulnerable</strong> states
+        {percentPopulationVulnerable
```
pnavarrc, line 30: I think it's better to check if `percentPopulationVulnerable` is a finite number instead, else it will not render when the percentage is 0

```
Number.isFinite(percentPopulationVulnerable) && (
  `, with ...`
)
```

pnavarrc (approved):
Looks good!

PR #2933, opened 2021-03-03
Vulnerability Notifications Above the Fold

BrettBoval:
https://trello.com/c/tpbwti4V/1039-vulnerability-callout-in-the-alert-section-on-a-location-page.

Also removed stale notification code for NYC and TX.


src/components/LocationPage/Notifications.tsx
```diff
@@ -50,37 +47,27 @@ const NotificationArea: React.FC<{ projections: Projections }> = ({
...
-    // https://trello.com/c/TdspuIeM/952-texas-reporting-dip-winter-weather-feb-2020
-    getFormattedStateCode(region)?.includes('TX')
-  ) {
-    notification = Notification.TXFeb2021Winter;
-  } else if (
-    // TODO(2020/12/22): Remove NYC notice after it's been up for a week or so.
-    ['36047', '36061', '36005', '36081', '36085'].includes(region.fipsCode)
-  ) {
-    notification = Notification.NYCCounty;
   } else if (isHospitalizationsPeak(projections.primary)) {
```
BrettBoval, line 62: If we are cleaning house should we remove this too.

By the time this could trigger again, we would be solidly into a 4th wave and likely already want different messaging. And if we need this type of logic we could also revert the changes later.

pnavarrc (changes requested):
The vulnerability block doesn't render for PR and Mariana Islands, we need to handle that case.

BrettBoval:
> The vulnerability block doesn't render for PR and Mariana Islands, we need to handle that case.

Pesky edge cases strike again. I guess our showVulnerability function doesn't always get to return true.

pnavarrc (approved):

PR #2943, opened 2021-03-04
Vulnerabilities scroll bug

chasulin:
Vulnerabilities hashlink scroll doesn't account for the header/searchbar height. This adds a scroll offset


ghop02 (approved):
Might be nice to make that a constant like `const SEARCH_BAR_OFFSET = 180` or something to communicate what the 180 is

PR #2946, opened 2021-03-04
analytics: track events in FullStory

pnavarrc:
- Cleanup `fullstory.ts` a bit (dynamic import and made the code follow a similar structure than `amplitude.ts`)
- Add tracking of events in FullStory when FS tracking is enabled
- Fixed the "open tooltip" event, it was being tracked multiple times (hover triggers multiple times when the user moves the mouse)

**Testing**
- In the Chrome console (in local dev), set the tracking flag in `localStorage` to `"true"` `localStorage.setItem('FULLSTORY_RECORD', 'true')` and reload the page
- Click around the app with the Network tab open, you should see some requests to FullStory
- Change the flag to `false` to disable recording `localStorage.setItem('FULLSTORY_RECORD', 'false')` and reload the page
- Click around the app, there shouldn't be more requests to FullStory

### DIFF
--- a/src/common/fullstory.ts
+++ b/src/common/fullstory.ts
@@ -1,5 +1,3 @@
-import * as FullStory from '@fullstory/browser';
-
 /**
  * We want to sample FullStory recordings so we don't hit the limit so quickly. We do this
  * by storing a flag in localStorage and only initializing FullStory for the percentage of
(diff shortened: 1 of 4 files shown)


pnavarrc:
[Fullstory - Enable data layer to automatically integrate GA events](https://trello.com/c/n5gWa38a/834-fullstory-enable-data-layer-to-automatically-integrate-ga-events)

src/common/fullstory.ts
```diff
@@ -9,27 +7,36 @@ import * as FullStory from '@fullstory/browser';
...
-  if (localStorage.getItem(FULLSTORY_RECORD_KEY) === RecordSession.TRUE) {
-    FullStory.init({ orgId: 'XEVN9' });
-  }
+async function getFullStory() {
+  return isRecordingEnabled() ? import('@fullstory/browser') : null;
+}
+
+export async function initFullStory() {
+  return getFullStory().then(fs => (fs ? fs.init({ orgId: 'XEVN9' }) : null));
+}
```
mikelehen, line 34: nit: I think these functions would be more readable if they returned `void` and used `await` and `if` like so:

```
export async function initFullStory(): void {
  const fs = await getFullStory();
  if (fs) {
    fs.init({ orgId: 'XEVN9' });
  }
}
```

If you wanted to avoid boilerplate and make a helper, you could do:

```
// TODO: Use proper type instead of any.
export async function withFullStory(fn: (fs: any) => void): void {
  const fs = await getFullStory();
  if (fs) {
    fn(fs);
  }
}
...
withFullStory(fs => fs.init({ordId: 'KEVN9'}));
```

src/common/fullstory.ts
```diff
@@ -9,27 +7,36 @@ import * as FullStory from '@fullstory/browser';
...
+  return localStorage.getItem(FULLSTORY_RECORD_KEY) === RecordSession.TRUE;
+}
 
-  // If we can't find the value in localStorage we won't enable recordings in FullStory
-  if (localStorage.getItem(FULLSTORY_RECORD_KEY) === RecordSession.TRUE) {
-    FullStory.init({ orgId: 'XEVN9' });
-  }
+async function getFullStory() {
+  return isRecordingEnabled() ? import('@fullstory/browser') : null;
+}
```
mikelehen, line 30: FWIW- reading from localStorage is more expensive than reading from a local variable (localStorage is synced across tabs, etc.) and I don't know if `import` has any overhead to worry about, but it's tempting to cache the result of `getFullStory()`, especially since I assume this gets called several times on page load (to track the pageview, perf metrics, etc.)...

Probably over-optimization though, so feel free to punt. 😁

mikelehen (approved):
LGTM with usual nits. 😁

src/common/fullstory.ts
pnavarrc: will punt for now, I want to make the analytics tracking logic a bit more unified

PR #2951, opened 2021-03-04
update the menu items on location search to match the homepage search

pnavarrc:
- Reduces the bundle size by ~ 74kB (the state SVGs are gone)


pnavarrc:
will implement a different solution, closing!

PR #2964, opened 2021-03-06
We should be showing the state level link in counties

igorkofman:


igorkofman:
Btw, I don't really understand what's happening with this comment:
  // If we don't have vaccination information for a given state, we fallback to
  // show the eligibility links.

PR #2969, opened 2021-03-06
Add disclaimers / blocked metrics to the CMS.

mikelehen:
* Introduces "Region Overrides" in the CMS for blocking metrics or adding disclaimers for one or more regions.
  * I set it up so you can apply an override to a region, a region + subregions, or just subregions of the region, since we sometimes want to block a state, sometimes counties/metros in a state, or sometimes both.
  * Right now overrides can only be targeted at metrics, but I think we could expand it to allow other per-region overrides down the road (e.g. adding an alert at the top of the location page, etc.).
* Moves our currently-blocked metrics to the CMS.
* Adds disclaimers for CA, CT, NY, and NH.

https://trello.com/c/UWWTbT8T/1017-data-disclaimers-in-cms
https://trello.com/c/rWS7ewoZ/1016-add-disclaimer-for-california-adjusted-case-rate
https://trello.com/c/3OVL7zDA/956-add-note-about-ny-test-positivity-discrepancy
https://trello.com/c/snz4arOC/957-re-add-connecticut-test-positivity-discrepancy-notice


src/cms-content/region-overrides/index.ts
```diff
@@ -0,0 +1,67 @@
+import RegionOverrides from './region-overrides.json';
+import { Markdown } from 'cms-content/utils';
+import { getMetricDefinition } from 'common/metric';
+import { Metric } from 'common/metricEnum';
+import regions, { Region } from 'common/regions';
+
+type Include = 'region' | 'subregions' | 'region-and-subregions';
```
pnavarrc, line 7: No enum?

src/common/regions/regions_data.ts
```diff
@@ -152,6 +152,10 @@ export const statesByFips = fromPairs(
   states.map(state => [state.fipsCode, state]),
 );
 
+export const statesByStateCode = fromPairs(
+  states.map(state => [state.stateCode, state]),
+);
```
pnavarrc, line 157: I recently found about `_.keyBy`, it feels a bit more readable to me to create maps by a specific property

```ts
const statesByStateCode = keyBy(states, state => state.stateCode);
```

src/components/LocationPage/ChartBlock.tsx
```diff
@@ -86,4 +84,26 @@ function ChartBlock(props: {
...
+  // Preface with any region override disclaimer text.
+  const overrideDisclaimer = getRegionMetricOverride(region, metric)
+    ?.disclaimer;
+  return (
+    <>
+      {overrideDisclaimer && <MarkdownContent source={overrideDisclaimer} />}
+      {standardDisclaimer}
+    </>
+  );
+}
```
pnavarrc, line 107: Should we move this to `metricDefinition` and call `metricDefinition.renderDisclaimer` instead?

public/admin/config.yml
```diff
@@ -770,3 +776,63 @@ collections:
...
+                label: Include
+                widget: select
+                default: region
+                options:
+                  - label: Region
+                    value: region
+                  - label: Subregions (counties / metros)
+                    value: subregions
+                  - label: Region and subregions (counties / metros)
+                    value: region-and-subregions
```
pnavarrc, line 808: Isn't this last option more like: `state and all its counties and metros` / `metro and all its counties`?

src/cms-content/region-overrides/index.ts
```diff
@@ -0,0 +1,67 @@
...
+  metric: Metric,
+): RegionOverride | undefined {
+  const metricDefinition = getMetricDefinition(metric);
+  for (const override of getOverridesForRegion(region)) {
+    if (override.metric === metricDefinition.metricId) {
+      return override;
+    }
+  }
+  return undefined;
+}
```
pnavarrc, line 30: Consider creating a map between CMS metric IDs and the values in the enum instead - I feel like this might make it easier to end up with a circular dependency (which is close to happen in metrics, metric definitions, etc).

```ts
const metricByOverrideMetricId = {
  'metric.caseDensity': Metric.CASE_DENSITY,
  // ...
}

function getRegionMetricOverride(region: Region, metric: Metric) {

}
```

I think this is better because each individual metric will need to call this file to render elements, for example, in `common/metrics/case_density.tsx`:

```tsx
// part of the metric definition...
function renderDisclaimer(region: Region) {
  if (getRegionMetricOverride(Metric.CASE_DENSITY)) {
    return "...";
  } else {
    return "...";
  }
}
```

src/cms-content/region-overrides/index.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/** Parses the string "region" out of the specified override and returns it as a Region class. */
+function getRegionFromOverride(override: RegionOverride): Region {
+  if (/^[A-Z][A-Z]$/.test(override.region)) {
+    return regions.findByStateCodeStrict(override.region);
+  } else {
+    return regions.findByFipsCodeStrict(override.region);
+  }
+}
```
pnavarrc, line 67: Consider adding a unit test to check that all the overrides correspond to real regions so we can catch mistakes (invalid state codes, etc)

pnavarrc (commented):
Added some notes!

src/cms-content/region-overrides/index.ts
mikelehen: Thanks.  In my mind it seemed useful to give metrics a persistent string identifier that we could use, but we don't actually need it and this ends up cleaner.

src/cms-content/region-overrides/index.ts
mikelehen: Added!

src/common/regions/regions_data.ts
mikelehen: SGTM!

src/components/LocationPage/ChartBlock.tsx
mikelehen: Since the code is the same for every metric, I'd rather not duplicate it across each one.  I could move it to `metric.tsx`, but I'm not sure that's much better.  Feels better to keep it close to where it's being used.

mikelehen (commented):
Thanks! Made suggested tweaks except for moving to `metricDefinition.renderDisclaimer`.  Let me know what you think!

src/cms-content/region-overrides/index.ts
pnavarrc: oh yeah, I think having a string for each metric might be useful, I'm a bit concerned about the dependencies between files

pnavarrc (approved):
Looks good!

PR #2981, opened 2021-03-08
Experiment with using Luxon in projection.ts

faihegberg:
This PR experiments with replacing moment with Luxon in `projection.ts`.


pnavarrc:
[Migrate time utils from moment to luxon.](https://trello.com/c/PKsdIfBC/1064-migrate-time-utils-from-moment-to-luxon)

PR #2982, opened 2021-03-08
Removed Source Deck Link in Explore

BrettBoval:
We've deprecated the google slides source deck doc. Eventually we would like a similar "find where this data comes from" as we do for the metrics, but for now I'm removing the old links and just keeping it generic.

### DIFF
--- a/src/components/Explore/Explore.tsx
+++ b/src/components/Explore/Explore.tsx
@@ -13,7 +13,6 @@ import { useTheme } from '@material-ui/core/styles';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import { ParentSize } from '@vx/responsive';
 import { useModelLastUpdatedDate } from 'common/utils/model';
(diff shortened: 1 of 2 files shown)


src/components/Explore/Explore.tsx
```diff
@@ -64,8 +63,7 @@ function trackShare(label: string, value?: number) {
 function getNoDataCopy(metricName: string, locationNames: string) {
   return (
     <p>
-      We don't have {metricName} data for {locationNames}. Learn more about{' '}
-      <ExternalLink href={DATA_SOURCES_URL}>our data sources</ExternalLink>.
+      We don't have {metricName} data for {locationNames}.
```
pnavarrc, line 66: Linting error above (ExternalLink is not used) ☝️

pnavarrc (approved):
Added a note, looks good otherwise!

PR #3002, opened 2021-03-09
X-axis updates (explore + metric charts)

chasulin:
- Removes 'today' marker from explore
- All charts now show abbreviated months, abbreviated year for Dec '20 and Jan '20, and every-other date tick on screens narrower than 960px
- Consolidates utils


pnavarrc (approved):
Looks great!

PR #3011, opened 2021-03-10
add error boundaries around location page blocks

pnavarrc:
[Error Boundaries](https://reactjs.org/docs/error-boundaries.html) catch JavaScript errors in children components and prevents the error from breaking the entire app. Currently, we have an error boundary in the Router, so if a feature anywhere in a page throws an error, we show "Something went wrong" on the entire page.

location page showing "something went wrong"

For example, this happened when we forgot to include vaccination data for the West Virginia ([see error in Sentry](https://sentry.io/organizations/covidactnow/issues/2227716089/events/bef2ca4c507f4a749d450597e2799401/)). This PR implements a component that includes the error boundary to wrap location page blocks. 

```tsx
<LocationPageBlock>
  <VulnerabilitiesBlock scores={ccviScores} region={region} />
</LocationPageBlock>
```

In this case, if the `VulnerabilitiesBlock` component crashes, it will show an error message and log the error to Sentry, but all the other blocks will render correctly.

the error message shows only for the feature block

## Testing

This will only be shown in production builds, in development React renders a different component with the stack trace. To test:

- Simulate a feature crashing, add the following line to any feature
  ```tsx
  new URL('~~~');
  ```
- Build and serve the app in production mode by running `yarn serve`. 
- Open a location page. You should see the error message on the feature only, instead of the full-page "something went wrong" messsage.


src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -23,13 +19,18 @@ import {
...
+import Recommend from 'components/Recommend';
+import ShareModelBlock from 'components/ShareBlock/ShareModelBlock';
+import VaccinationEligibilityBlock from 'components/VaccinationEligibilityBlock';
 import VulnerabilitiesBlock from 'components/VulnerabilitiesBlock';
-import { useCcviForFips } from 'common/hooks';
-import { useScrollToElement } from 'common/hooks';
+import ChartBlock from './ChartBlock';
+import LocationPageBlock from './LocationPageBlock';
+import LocationPageHeader from './LocationPageHeader';
+import { ChartContentWrapper } from './ChartsHolder.style';
```
pnavarrc, line 33: I sorted the dependencies a bit and merged two imports from the same file into one.

pnavarrc:
[Put error boundary around individual features not whole location page](https://trello.com/c/tNqWfQYA/970-put-error-boundary-around-individual-features-not-whole-location-page)

chasulin (commented):
nice! a few thoughts-
- want to add to the homepage as well? ie. currently if compare crashes on the homepage, the original page-wide error boundary pops up
- maybe we should change the language of the error message. "An error occurred while rendering this component" is very eng-y
- possible to style it? the error message is currently full bleed and the top/bottom spacing is a little awkward:

Screen Shot 2021-03-10 at 3 46 56 PM

chasulin (approved):

PR #3023, opened 2021-03-11
cms vaccines: update the emailAlertVersion to be a string field

pnavarrc:
[Trello](https://trello.com/c/Aus9dlOE/1027-update-the-data-type-for-emailalertversion-on-the-cms) - The UI to change numbers is error-prone, so it's easier for the team to enter values as strings. In this PR, I change the type in the CMS config and the JSON file, but leave the external interface of the vaccination phases info unchanged (the emailAlertVersion is exposed as a number)

- Updated the config to take strings
- Updated the vaccine information (phases) to be strings
- Added unit tests to ensure that `emailAlertVersion` is a number
- Add unit tests to ensure that there are no repeated states

### DIFF
--- a/public/admin-vaccines/config.yml
+++ b/public/admin-vaccines/config.yml
@@ -102,8 +102,8 @@ collections:
                 required: false
               - label: Email alert version
                 name: emailAlertVersion
-                default: 0
-                widget: number
+                default: '0'
+                widget: string
               - label: Notes
                 name: notes
                 required: false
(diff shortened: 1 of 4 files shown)


pnavarrc:
[Update the data type for emailAlertVersion on the CMS](https://trello.com/c/Aus9dlOE/1027-update-the-data-type-for-emailalertversion-on-the-cms)

chasulin (approved):

PR #3026, opened 2021-03-11
Remove "outbreak" copy from website and stick to risk terminology.

mikelehen:
https://docs.google.com/document/d/1MfsA5X6jCaWg9NsVVJdd7j6IbA17daa5aDZPOObNEn8/edit

### DIFF
--- a/src/screens/LocationPage/utils.ts
+++ b/src/screens/LocationPage/utils.ts
@@ -1,8 +1,9 @@
 import { LOCATION_SUMMARY_LEVELS } from 'common/metrics/location_summary';
-import { formatMetatagDate } from 'common/utils';
+import { formatMetatagDate, formatPercent } from 'common/utils';
 import { Region, State, County, MetroArea } from 'common/regions';
 import { replace } from 'lodash';
 import { LocationSummariesByFIPS } from 'common/location_summaries';
+import { Metric } from 'common/metricEnum';
 
 function locationName(region: Region) {
   if (region instanceof State) {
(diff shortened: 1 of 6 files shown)


src/screens/LocationPage/utils.ts
```diff
@@ -22,7 +23,18 @@ export function getPageTitle(region: Region): string {
 
 export function getPageDescription(region: Region): string {
   const date = formatMetatagDate();
-  const { level: alarmLevel } = LocationSummariesByFIPS[region.fipsCode];
+  const { level: alarmLevel, metrics } = LocationSummariesByFIPS[
+    region.fipsCode
+  ];
   const levelInfo = LOCATION_SUMMARY_LEVELS[alarmLevel];
-  return `${date} COVID RISK LEVEL: ${levelInfo.detail(locationName(region))}`;
```
mikelehen, line 27: This broke because `detail()` now returns a `ReactElement` containing a link but the description needs to be plain text.  But also, this description (which google shows in search results) ended up pretty brief, so I added our vaccination metric to it, which we're trying to highlight more anyway.

src/screens/LocationPage/utils.ts
```diff
@@ -22,7 +23,18 @@ export function getPageTitle(region: Region): string {
...
+  const vaccinationRatio = metrics[Metric.VACCINATIONS]?.value;
+  const vaccinationText = vaccinationRatio
+    ? ` and ${formatPercent(
+        vaccinationRatio,
+        1,
+      )} of the population has received at least one vaccine dose`
+    : '';
+  return `${date}: ${locationName(region)} is at ${
+    levelInfo.name
```
chasulin, line 38: maybe lowercase the level? "xxx is at Medium COVID risk level" --> "xxx is at medium COVID risk level"

chasulin (approved):

src/screens/LocationPage/utils.ts
mikelehen: oops, good catch, thanks!

PR #3032, opened 2021-03-11
Fonts file

chasulin:
Starts to centralize our font styles

Adds `common/theme` folder, currently with `fonts.ts` (later going to add `colors.ts` into it as well)

In `fonts.ts`, we are exporting each set of font styles (ie. roboto + bold) as a set of css properties, that can be called as `${fonts.regularBookBold}`


pnavarrc (approved):
🙌

PR #3036, opened 2021-03-11
Adding data coverage table

ghop02:
Adds a data coverage table.  This is good for a pass, but still a couple small details to be worked out:

- [ ] finalize metrics to be included
- [ ] verify tooltip content

image


src/components/Explore/utils.ts
```diff
@@ -329,7 +329,7 @@ export function getSocialQuote(regions: Region[], metric: ExploreMetric) {
   return '';
 }
 
-const pluralize = (num: number, singular: string, plural: string) =>
+export const pluralize = (num: number, singular: string, plural: string) =>
```
pnavarrc, line 332: Maybe we should move this to `common/utils` somewhere?

src/screens/DataApi/DataApi.style.tsx
```diff
@@ -25,6 +25,11 @@ export const GreenLinkButton = styled(LinkButton)`
   }
 `;
 
+export const BlueLinkButton = styled(LinkButton)`
+  color: ${COLOR_MAP.BLUE};
+  text-transform: none;
+`;
```
pnavarrc, line 31: Consider using the `ExternalLink` component instead. This is mostly okay as the LinkButton is actually a link in the DOM, but the hover state makes it look like a button.

Ideally, we want to use buttons for actions and links for navigations, but we don't always follow that.

src/screens/DataApi/DataApi.tsx
```diff
@@ -36,6 +39,25 @@ export const sidebarSections: TocItem[] = [
...
+    <Grid container spacing={1}>
+      <Grid item xs={12}>
+        <DataCoverageTable rows={FIELDS_PRETTY_NAMES} />
+      </Grid>
+      <Grid item xs={12}>
+        <BlueLinkButton
+          trackingCategory={EventCategory.API}
+          trackingLabel="Data API: View all metrics"
+          href="https://apidocs.covidactnow.org/data-definitions"
+        >
```
pnavarrc, line 53: If you use `ExternalLink`, remember to call the tracking function in the `onClick` prop

```tsx
<ExternalLink href="..." onClick={() => trackViewAllMetrics()}>
  View more metrics
</ExternalLink>

```

src/screens/DataApi/DataCoverageTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+  );
+};
+
+const tooltipText = (coverage: CoverageDetails, regionType: RegionType) => {
+  switch (regionType) {
+    case RegionType.COUNTY:
+      return (
+        `Supported in ${coverage.regionsAvailable} counties (of ${coverage.totalRegions}), ` +
+        `or ${coverage.statesAvailable} states and territories (of ${coverage.totalStates}).`
+      );
```
pnavarrc, line 90: nit: I think you don't need to concatenate the strings when using template strings, you can just have a long line

src/screens/DataApi/DataCoverageTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+  if (status.regionsAvailable === 0) {
+    return (
+      <TextWithTooltip text="No data" tooltipContent="No data available." />
+    );
+  }
+
+  const availableRatio = status.regionsAvailable / status.totalRegions;
+
+  if (availableRatio >= 0.95) {
+    return <CheckIcon color="secondary" />;
```
pnavarrc, line 111: Consider adding the `aria-label` attribute to the icons to ensure that the information is available for users that need assistive technologies

scripts/update_data_coverage_summary.ts
```diff
@@ -94,7 +84,8 @@ async function main(outputFolder: string) {
   allSummaries.forEach(regionSummary => {
     _.forEach(['riskLevels', 'metrics', 'actuals'], firstLevel => {
       // summaryDataSection is a sub object of a `RegionSummary`.  Accessing it
-      // directly generates a typescript error, but we know it will always have a value.
+      // directly generates a typescript error, but we know it will always have a value
+      // as it's part of the
```
pnavarrc, line 88: I think there is more to that phrase 😄

src/screens/DataApi/DataApi.tsx
```diff
@@ -36,6 +39,25 @@ export const sidebarSections: TocItem[] = [
   },
 ];
 
+const DataCoverageSection: React.FC<{}> = () => {
+  return (
+    <Grid container spacing={1}>
+      <Grid item xs={12}>
+        <DataCoverageTable rows={FIELDS_PRETTY_NAMES} />
+      </Grid>
+      <Grid item xs={12}>
```
pnavarrc, line 48: nit: We can probably move the link inside the other `<Grid item xs={12}>` container (unless you just want the spacing)

src/screens/DataApi/DataCoverageTable.stories.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import CoverageTable, {
+  COVERAGE_SUMMARY,
+} from 'screens/DataApi/DataCoverageTable';
+
+export default {
```
pnavarrc, line 6: nit: I would probably move the `DataCoverageTable` component to the `components` folder, but here is okay too

src/screens/DataApi/DataCoverageTable.style.tsx
```diff
@@ -0,0 +1,77 @@
...
+  font-size: 14px;
+`;
+
+export const StyledTableHead = styled(TableHead)`
+  font-family: Roboto;
+  font-weight: 500;
+  font-size: 13px;
+  text-transform: uppercase;
+  background-color: white;
+`;
```
pnavarrc, line 30: We are trying to keep the minimum text size in the site to be 16px for accessibility - The text doesn't look super small to me, but maybe check how it looks at 16px?

src/screens/DataApi/DataCoverageTable.style.tsx
```diff
@@ -0,0 +1,77 @@
...
+  }
+`;
+
+export const UnderlinedText = styled(Typography)`
+  text-decoration-line: underline;
+  text-decoration-style: dotted;
+  text-underline-position: under;
+  vertical-align: middle;
+  font-size: 14px;
+`;
```
pnavarrc, line 22: Design feedback: The dotted underline makes me think that I would see the tooltip by hovering on the text, but when I do nothing happens. Maybe we should make the tooltip show up when hovering the text or the icon?

image

src/screens/DataApi/DataCoverageTable.style.tsx
```diff
@@ -0,0 +1,77 @@
...
+  scope: 'row',
+}))`
+  width: 30%;
+
+  @media (max-width: ${materialSMBreakpoint}) {
+    padding: 8px;
+    padding-left: 0px;
+    width: 20%;
+  }
+`;
```
pnavarrc, line 44: nit: No https://material-ui.com/api/table-head/ ? It's probably okay since you are telling it to use `th`

src/screens/DataApi/DataCoverageTable.style.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export const MetricTableCell = styled(TableCell).attrs(props => ({
+  component: 'th',
+  align: 'left',
+  scope: 'row',
+}))`
```
pnavarrc, line 50: The component should be `td` right?

src/screens/DataApi/DataCoverageTable.style.tsx
```diff
@@ -0,0 +1,77 @@
...
+  width: 23%;
+
+  @media (max-width: ${materialSMBreakpoint}) {
+    padding: 4px;
+  }
+`;
+
+export const MetricTableRow = styled(TableRow)`
+  &:nth-of-type(even) {
+    background-color: #fafafa;
```
pnavarrc, line 75: Consider updating it to take variables from `common/colors` (that would be `GREY_0`)

src/screens/DataApi/DataCoverageTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+  { displayName: 'Deaths', field: 'actuals.deaths' },
+  { displayName: 'Total testing volume', field: 'actuals.positiveTests' },
+];
+
+export const FIELDS_PRETTY_NAMES = FIELDS.map(field => {
+  const record = COVERAGE_SUMMARY.find(record => field.field === record.name);
+  if (record) {
+    return {
+      ...record,
+      ...{ name: field.displayName },
```
pnavarrc, line 49: nit: no need to create an object here

```ts
return {
  ...record,
  name: field.displayName
}
```

src/screens/DataApi/DataCoverageTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+  { displayName: 'Positive test rate', field: 'metrics.testPositivityRatio' },
+  { displayName: '% Vaccinated', field: 'metrics.vaccinationsInitiatedRatio' },
+  { displayName: 'ICU capacity used', field: 'metrics.icuCapacityRatio' },
+  { displayName: 'Hospitalization usage', field: 'metrics.icuCapacityRatio' },
+  { displayName: 'Deaths', field: 'actuals.deaths' },
+  { displayName: 'Total testing volume', field: 'actuals.positiveTests' },
+];
+
+export const FIELDS_PRETTY_NAMES = FIELDS.map(field => {
+  const record = COVERAGE_SUMMARY.find(record => field.field === record.name);
```
pnavarrc, line 45: nit: consider filtering the fields that have records and then mapping to make the sequence easier to read

```ts
// I'm not sure if creating the functions below is worth the trouble...
FIELDS
  .filter(hasRecord)
  .map(fieldToRecord)
  .filter((record): record is MetricCoverage);


```

src/screens/DataApi/DataCoverageTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+      tooltipContent={tooltipContent}
+    />
+  );
+};
+
+const CoverageRow: React.FC<{ row: MetricCoverage }> = ({ row }) => {
+  return (
+    <MetricTableRow key={row.name}>
+      <MetricTableCell>{row.name}</MetricTableCell>
+      <AvailabilityTableCell>
```
pnavarrc, line 137: Remember to add keys on children items for the same type

```tsx
<AvailabilityTableCell key="state-cell">
  {/* ... */}
</AvailabilityTableCell />
```

src/screens/DataApi/DataCoverageTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+const CoverageTable: React.FC<{ rows: MetricCoverage[] }> = ({ rows }) => {
+  return (
+    <TableContainer>
+      <Table aria-label="Data Coverage Table">
+        <StyledTableHead>
+          <TableRow>
+            <MetricTableCellHeader>Metric</MetricTableCellHeader>
+            <AvailabilityTableCell>States</AvailabilityTableCell>
+            <AvailabilityTableCell>Counties</AvailabilityTableCell>
+            <AvailabilityTableCell>Metros</AvailabilityTableCell>
```
pnavarrc, line 159: nit: we need keys here as well

src/screens/DataApi/DataCoverageTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+          </TableRow>
+        </StyledTableHead>
+        <TableBody>
+          {rows.map(row => (
+            <CoverageRow row={row} />
+          ))}
```
pnavarrc, line 165: and here :)

pnavarrc (changes requested):
Added some notes on accessibility + some other suggestions, let me know when it's ready for a second pass!

src/screens/DataApi/DataCoverageTable.tsx
ghop02: ah yeah I split them somewhat intentionally because I think it's easier to read. Do we have a pattern of not doing so?

src/screens/DataApi/DataCoverageTable.tsx
ghop02: hmm I didn't think you need for this case - I've only been able to find that this is true for lists/iterators.

src/screens/DataApi/DataCoverageTable.style.tsx
ghop02: So actually [it looks like](https://material-ui.com/api/table-cell/) any `TableCell`s that have a parent context of a `TableHead` (which this does) automatically turn into `th`

src/screens/DataApi/DataCoverageTable.style.tsx
ghop02: Yeah I think so

pnavarrc (approved):
Looks good!

PR #3038, opened 2021-03-12
fix arrow buttons for compare

pnavarrc:
[Trello](https://trello.com/c/EDnr5bGw/1095-up-down-sort-buttons-are-invisible-in-compare-modal-until-hover) - Fix the compare table arrow buttons

**Modal**
image

**Inline**
image


chasulin (approved):
sweeet, thank you!

PR #3041, opened 2021-03-12
Move STATES_MAP closer to where it's used

schamp:
STATES_MAP was only used in one place, so move it there so that not everyone who imports `locations.ts` gets all that data, too.


src/common/locations.ts
```diff
@@ -3,7 +3,6 @@ import US_STATE_DATASET from 'components/MapSelectors/datasets/us_states_dataset
 import { each, has } from 'lodash';
```
mikelehen, line 3: some unused imports

mikelehen (approved):
A couple unused imports left, but otherwise this looks fine.

That said:
1. FWIW- We've actually migrated the bulk of our codebase to use our RegionDb class (https://github.com/covid-projections/covid-projections/blob/221444b06c2d74ec62cf2573b742be836b9a6f49/src/common/regions/region_db.ts#L12).  We should be able to drop STATE_MAP entirely and switch over to it.  We might need to add a `findByStateCode()` method akin to `findByStateCodeStrict()` for this particular use case though.
2. AFAIK, with tree shaking the bundler shouldn't pull in STATES_MAP unless it's actually needed by the calling code.  So I'm surprised this code change makes any difference.

src/common/locations.ts
schamp: thx, will get.

schamp:
> A couple unused imports left, but otherwise this looks fine.
> 
> That said:
> 
> 1. FWIW- We've actually migrated the bulk of our codebase to use our RegionDb class (https://github.com/covid-projections/covid-projections/blob/221444b06c2d74ec62cf2573b742be836b9a6f49/src/common/regions/region_db.ts#L12
>    ).  We should be able to drop STATE_MAP entirely and switch over to it.  We might need to add a `findByStateCode()` method akin to `findByStateCodeStrict()` for this particular use case though.

I'll revisit this and see if there's an easy way to do that, if not, I may just land this and follow up with that later.

> 2. AFAIK, with tree shaking the bundler shouldn't pull in STATES_MAP unless it's actually needed by the calling code.  So I'm surprised this code change makes any difference.

From what I can tell, this allows the next.js builder/bundler to break STATES_MAP out into a separate chunk and not load it on all the pages.  I may look deeper though based on what you say.

schamp:
@mikelehen it was pretty straightforward to remove STATES_MAP entirely, so I just did that.  Thanks for the suggestion : )

src/common/utils/model.ts
```diff
@@ -81,9 +80,10 @@ export function fetchAllStateProjections(snapshotUrl: string | null = null) {
       APIRegionSubPath.STATES,
     );
     return all
-      .filter(
-        summaryWithTimeseries =>
-          getStateName(summaryWithTimeseries.state || '') !== undefined,
+      .filter(summaryWithTimeseries =>
+        summaryWithTimeseries?.state
+          ? Boolean(regions.findByStateCode(summaryWithTimeseries.state))
```
mikelehen, line 85: Minor (feel free to ignore if you just want to merge as-is), but using `Boolean()` as a way to check for null seems a bit confusing / indirect.

I'd probably use a `!== null` check and/or create an `isKnownStateCode()` helper.

mikelehen (approved):
Nice! This looks great.  Thanks for the cleanup.  Saves a bit of bundle size and just makes the code more manageable. 😁

src/common/utils/model.ts
schamp: Ah, thanks for the callout.  I'm accustomed to prefer `Boolean` to indicate we care in a boolean context whether the value is truthy or not, and avoid issues of `null` vs `undefined`, etc.  If we were using `strictNullChecks`, maybe it would matter less to me.  Or, if I hadn't been burned by `undefined`s failing a `!== null` check too many times to count : )

I'll merge as is, but agree that an `isKnownStateCode` would be a worthy improvement.

PR #3044, opened 2021-03-12
Lazy load state geometry

schamp:
not sure about this yet.


pnavarrc:
@schamp we are removing the state SVGs (I have a branch with that in progress), we are going to replace them with dots like in the home page search.

schamp:
Oh nice, I'll just cancel this then

PR #3046, opened 2021-03-12
demo: time range selector for daily new cases chart

pnavarrc:
image


mikelehen:
😮

schamp:
I've been looking at the massive (1.2M) statically rendered HTML for each location page with next.js, and ~700K of it looks like chart SVG data, and 200K supporting JSON init data.  If we could cut the graph and init data to 7 days worth by default and load the rest lazily / on demand that could really reduce the default amount of data we render and shove down to the client on first load.

PR #3051, opened 2021-03-13
ForwardRef for Mui complaints

schamp:
This lets these things take a forwarded ref, they don't do anything with it, but it avoids some very long log messages from Mui components that clutter the error and warning stream.

### DIFF
--- a/src/components/Compare/ModalCompare.tsx
+++ b/src/components/Compare/ModalCompare.tsx
@@ -12,7 +12,7 @@ import Filters from 'components/Compare/Filters';
 import { LockBodyScroll } from 'components/Dialog';
 import { Region, MetroArea } from 'common/regions';
 
-const ModalCompare = (props: {
+interface ModalCompareProps {
   stateName?: string;
   county: any | null;
   setShowModal: any;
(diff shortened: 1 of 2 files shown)


src/components/Compare/ModalCompare.tsx
```diff
@@ -12,117 +12,124 @@ import Filters from 'components/Compare/Filters';
...
+          homepageSliderValue={props.homepageSliderValue}
+          setHomepageSliderValue={props.setHomepageSliderValue}
+          region={props.region}
+          vaccinesFirst={props.vaccinesFirst}
+          vulnerabilityFirst={props.vulnerabilityFirst}
+        />
+      </Fragment>
+    );
+  },
+);
```
pnavarrc, line 133: Consider wrapping the component before exporting it instead to reduce nesting and the size of the diff

```tsx
export default React.forwardRef((props, ref) => 
  <ModalCompare {...props} />
);
```

pnavarrc (approved):
Thanks for the fix? Added a suggestion

src/components/Compare/ModalCompare.tsx
schamp: great idea, thanks.  why didn't I think of that?

PR #3055, opened 2021-03-14
Make EmailService automatically fall back to half speed on throttling errors.

mikelehen:
This will let us gracefully handle if we're sending risk alerts and vaccine alerts at the same time.

https://trello.com/c/X9Yc96oq/1065-allow-concurrent-sending-of-vaccine-and-alert-emails


mikelehen:
cc/ @BrettBoval

pnavarrc (approved):
Looks good!

PR #3057, opened 2021-03-15
Split ShareCardImage into Home and Location versions

schamp:
This resolves a todo by @mikelehen and also simplifies migration to next.js

I thought about making a component that takes {children} instead of copying the wrappers, but wasn't sure it was worth the hassle.  Feel free to press for that though, if you think it is.

It also became evident that the homepage does *not* include the `<ScreenshotReady>` component like the individual location-based share cards do.  Is that intentional, or a bug?  Let me know, I can fix it while I'm at it.

### DIFF
--- a/src/screens/internal/ShareImage/HomeShareCardImage.tsx
+++ b/src/screens/internal/ShareImage/HomeShareCardImage.tsx
@@ -10,7 +10,6 @@ import { Header } from './LocationShareCardImage';
  * screenshot that we then use as our OpenGraph image.
  */
 const HomeShareCardImage = () => {
-  // FIXME: do we need <ScreenshotReady> on this one? it was missing before
   return (
     <DarkScreenshotWrapper className={SCREENSHOT_CLASS}>
       <Header isHomePage />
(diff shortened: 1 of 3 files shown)


src/screens/internal/ShareImage/HomeShareCardImage.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { DarkScreenshotWrapper } from './ShareImage.style';
+import { SCREENSHOT_CLASS } from 'components/Screenshot';
+import { Header } from './LocationShareCardImage';
+
+/**
+ * Screen that just shows the appropriate share card so that we can take a
+ * screenshot that we then use as our OpenGraph image.
+ */
+const HomeShareCardImage = () => {
+  // FIXME: do we need <ScreenshotReady> on this one? it was missing before
```
mikelehen, line 13: It gets added by a nested component here once we've loaded the necessary data and rendered the map: https://github.com/covid-projections/covid-projections/blob/ead6033a754e31b5573b16884227f05dc298c532/src/components/Map/USACountyMap.tsx#L191

mikelehen (approved):
I think you can remove the FIXME, but otherwise looks good!

src/screens/internal/ShareImage/HomeShareCardImage.tsx
schamp: ah, thanks : )

PR #3064, opened 2021-03-15
use new menu items for location page search

pnavarrc:
- Remove the state icon for the menu item results on the location page (replace them with the 'dot' icon)
- Remove state SVG shapes and other unused files
- Remove some extra containers

Will rename the `NewMenuItem` component to `MenuItem` on a separate PR (also fix the width bug, which is unrelated to this change)

image


src/components/Header/SearchHeader.tsx
```diff
@@ -41,37 +39,31 @@ const SearchHeader = ({
...
-                <>Counties&nbsp;&nbsp;</>
-                <MapIcon />
-              </MapToggle>
+      <MenuBarWrapper>
+        <SelectorWrapper
+          onClick={() => setMobileMenuOpen(false)}
+          $isNarrowMobile={isNarrowMobile}
+        >
+          <SearchAutocomplete
+            filterLimit={getFilterLimit(region)}
```
chasulin, line 48: unrelated, so feel free to punt, but i think we can remove this dynamic filter limit and just set it to a static number. now that we've taken out the location-specific ordering (ie. state's counties first, etc.), we don't need the location-specific filter limit. we always just rank alphabetized states first, and so currently each state is showing a different amount of states, which is strange

chasulin (approved):

pnavarrc:
[Performance: Remove state SVGs and use dots on the location page search](https://trello.com/c/zbUrph7A/1066-performance-remove-state-svgs-and-use-dots-on-the-location-page-search)

PR #3072, opened 2021-03-16
Split Map view from SocialLocationPreview

schamp:
Split out the map-specific view, so that the location-specific version can have non-optional props, simplifies and smooths the way for next.js migration.

Depends on the ShareCardImage pull request (https://github.com/covid-projections/covid-projections/pull/3057) being landed first.  I'll have to clean up and rebase after that lands.


mikelehen (approved):
Looks great!  Thanks for tackling all this cleanup.

PR #3078, opened 2021-03-16
upgrade caniuse-lite

pnavarrc:


ghop02 (approved):

chasulin (approved):
im not sure exactly what the use case of this is, but it doesnt look like its used anywhere in the code. if thats the case, maybe we delete? but if im missing something, go ahead and merge !

ghop02:
> im not sure exactly what the use case of this is, but it doesnt look like its used anywhere in the code. if thats the case, maybe we delete? but if im missing something, go ahead and merge !

I think it's used by browsersync when running `yarn start` but i could be wrong (I remember seeing a line about that)

pnavarrc:
This is used by `babel/preset-env` to decide how to transform our ES6 code into browser-compatible code. For example, if the syntax `const { a, b } = data` is not available in some browsers, `babel/preset-env` will convert our code to `const a = data.a; const b = data.b` at build time.

Keeping the compatibility list updated removes unnecessary transformations as more browsers support new syntax. TL;DR: This is not a super critical update, it makes our build process a bit better.

PR #3080, opened 2021-03-16
Migrate UTC conversion instances to time utils

faihegberg:
This PR migrates all direct usages of `moment` for UTC conversion to using time utils instead.


pnavarrc (approved):
Looks good! Consider renaming the enum to start with uppercase and maybe keep it singular (`TimeFormat` instead of `timeFormats`), that reads a bit better when using it

```ts
TimeFormat.YYYY_MM_DD
```

PR #3085, opened 2021-03-16
Implement hooks to load geographies async

pnavarrc:
- Copy `counties-10m.json` and `states-10m.json` to `common/data` and implement the corresponding import functions
- Implement the hooks to import state and county geographies.
- Update the `RegionMap` component to load geographies using the hooks


src/components/RegionMap/RegionMap.tsx
```diff
@@ -143,36 +148,22 @@ function getProjectionConfig(
   return { scale, center };
 }
 
-function buildCountyGeometries(countyFipsList: string[]) {
-  const objects = COUNTIES_JSON.objects;
-  const geometries = objects.counties.geometries.filter(geoCounty =>
-    countyFipsList.includes(geoCounty.id),
+function buildCountyGeometries(
+  countiesTopoJson: any,
```
ghop02, line 152: Can this be a type?

src/components/RegionMap/RegionMap.tsx
```diff
@@ -18,18 +17,27 @@ const RegionMap: React.FC<{
...
-    countyFipsList,
-  ]);
+  const projectionConfig = useMemo(() => {
+    return countiesTopoJson
+      ? getProjectionConfig(countiesTopoJson, width, height)
+      : null;
+  }, [countiesTopoJson, width, height]);
+
+  if (!statesTopoJson || !countiesTopoJson || !projectionConfig) {
+    return null;
```
ghop02, line 39: What is the effect of returning null here is there just empty content? Potentially could use something like suspense suspense.  When this is null will we want to have filler?

src/components/RegionMap/RegionMap.tsx
```diff
@@ -18,18 +17,27 @@ const RegionMap: React.FC<{
...
+      allCountiesTopoJson
+        ? buildCountyGeometries(allCountiesTopoJson, countyFipsList)
+        : null,
+    [countyFipsList, allCountiesTopoJson],
   );
 
-  const statesTopoJson = useMemo(() => getStateGeometries(countyFipsList), [
-    countyFipsList,
-  ]);
+  const projectionConfig = useMemo(() => {
```
ghop02, line 32: This isn't fetching anything is it? I would place this after the if statement
```
if (!statesTopoJson || !countiesTopoJson) { return null; }
const projectionConfig = useMemo(() => {
  return getProjectionConfig(...);
}, [...])
```

src/components/RegionMap/RegionMap.tsx
pnavarrc: Returning `null` doesn't render anything. In this case, I think it's okay because the MiniMap only renders open by default when the screen width is over 1350px, and even in that case, they only see the map as empty while the location page loads.

src/components/RegionMap/RegionMap.tsx
pnavarrc: `useMemo` cannot be called conditionally. I'm memoizing that call because there is a bunch of math going on in calculating the centroid and scale for a given set of county TopoJSON objects, so we probably don't want to run that too often.

ghop02: cool! Could you add that as a comment about why useMemo is used there?

src/components/RegionMap/RegionMap.tsx
pnavarrc: That somehow breaks the calls to `geoBounds` and `geoCentroid`, I'm not sure how to make those compatible 🤷‍♂️

mikelehen: I think TypeScript's `typeof import()` syntax helps here.  See https://github.com/covid-projections/covid-projections/pull/3088

We could probably use that to remove some of our existing hand-defined types too...

mikelehen (commented):
Just to be clear, this isn't going to have any net improvement, right? Because we're still importing `counties-10m.json` synchronously in USACountyMap, right, so it'll still be bundled into the main bundle?

I assume that the bundling is smart enough to not *additionally* load it asynchronously from this new bundle (given it's already in the main bundle), but if it's _not_ that smart, this would actually have a net performance decrease, right?

pnavarrc:
> Just to be clear, this isn't going to have any net improvement, right?

correct, I'm working on making the homepage map async as well, but I want to do that separately because the loading state might need more attention.

mikelehen (approved):
Suggest merging the PR mentioned in https://github.com/covid-projections/covid-projections/pull/3085#discussion_r595593041 and it'd be great to get the homepage map to take advantage of this soon, but otherwise, looks good to me!

pnavarrc:

PR #3087, opened 2021-03-16
Add share widgets to FAQ

faihegberg:
This PR experiments with adding share widgets to each FAQ. All props of `SmallShareButtons` are not implemented yet as this is simply an exploration.

### DIFF
--- a/src/screens/Learn/Faq/Section.tsx
+++ b/src/screens/Learn/Faq/Section.tsx
@@ -7,7 +7,7 @@ import { trackEvent, EventAction, EventCategory } from 'components/Analytics';
 
 const Section = (props: { section: FaqSection }) => {
   const { section } = props;
-  const url = 'https://covidactnow.org/faq#';
+  const url = 'https://covidactnow.org/faq';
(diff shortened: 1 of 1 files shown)


src/screens/Learn/Faq/Section.tsx
```diff
@@ -13,6 +22,13 @@ const Section = (props: { section: FaqSection }) => {
           <Fragment key={`faq-question-${i}`}>
             <ItemName id={question.questionId}>{question.question}</ItemName>
             <MarkdownContent source={question.answer} />
+            <SmallShareButtons
+              shareUrl={url + question.questionId}
```
pnavarrc, line 26: it's better to use `urlJoin` here instead of concatenating

chasulin: echoing this ^

chasulin (approved):

src/screens/Learn/Faq/Section.tsx
faihegberg: 👍

PR #3088, opened 2021-03-16
Use types for topo json files.

mikelehen:
I noticed TypeScript's `typeof import()` syntax could help here. 😁   Only minimally tested, but seems to work.


mikelehen:
Feel free to merge this into your branch or just take it as inspiration / whatever.

pnavarrc:
Nice!

pnavarrc:

PR #3089, opened 2021-03-17
load counties async, states sync

pnavarrc:
[Trello](https://trello.com/c/kX8BKtQF/1061-performance-load-map-async-to-decrease-main-bundle-size) - Load geographies asynchronously

Note that I'm not loading the state geographies async yet because we will need a better loading state for that to look nice. When selecting the counties map, the map briefly disappears before rendering the map with the counties. Subsequent toggles are much smoother 

**Changes**
- Simplifies the USACountyMap
- Move the map tooltip logic to the USACountyMap component
- Implement the MapCounties component, which uses the async version of the counties

This reduces the main bundle size by ~209KB.

```
384.43 KB (-209.19 KB)  build/static/js/main.a3e677ea.chunk.js
349.8 KB (+207 B)       build/static/js/6.d2ee4c54.chunk.js
244.53 KB (+118.92 KB)  build/static/js/25.3b206c5d.chunk.js
...
```


pnavarrc:

src/components/Map/USACountyMap.tsx
```diff
@@ -1,87 +1,49 @@
...
+import React, { useState } from 'react';
 import { Link } from 'react-router-dom';
 import { ComposableMap, Geographies, Geography } from 'react-simple-maps';
 import { geoAlbersUsaTerritories } from 'geo-albers-usa-territories';
+import ReactTooltip from 'react-tooltip';
 import { colorFromLocationSummary } from 'common/colors';
 import { useSummaries } from 'common/location_summaries';
 import { ScreenshotReady } from 'components/Screenshot';
 import regions, { State as StateType } from 'common/regions';
+import stateGeographies from 'common/data/states-10m.json';
```
mikelehen, line 10: Out of curiosity, why not make this async too?  Would save another 35k of gzipped JSON that has to be downloaded/parsed/rendered-as-svg before FCP / LCP can happen, right?

mikelehen (approved):
Nice! Looks like this saves 209KB from our gzipped main bundle (594K => 385K).  🥳 

I'd be curious about loading the states async too (basically treat the map as an async resource, like it would be if it was an image)...  but maybe the UX ends up undesirable?

src/components/Map/USACountyMap.tsx
pnavarrc: The rendering of the map is not ideal while the data loads - I want to try having a good loading state to minimize the impact of loading the states async first. I'm going to try that tomorrow :)

PR #3091, opened 2021-03-17
Use region instead of projections

schamp:
We don't need projections in the notification block, so use the region, which is already available.  This will help unblock deferred-loading of projection data on location pages, which should help FCP.


mikelehen (approved):
LGTM.

PR #3092, opened 2021-03-17
Pull `projections` out of LocationPageHeader

schamp:
We don't actually need the whole projections in the location page header, so pull it out and pass in or compute what we need (alarmLevel / isCounty).

This will also help defer loading of projection details for metrics charts below while still rendering a helpful header above the fold.

### DIFF
--- a/src/components/LocationPage/ChartsHolder.tsx
+++ b/src/components/LocationPage/ChartsHolder.tsx
@@ -177,7 +177,7 @@ const ChartsHolder = ({ projections, region, chartId }: ChartsHolderProps) => {
     <>
       <ChartContentWrapper>
         <LocationPageHeader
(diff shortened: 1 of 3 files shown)


mikelehen (approved):
Nice! 

FYI- I think where we ultimately want to end up is rendering the location page header via data from the location summary (a big JSON file that is in our main bundle and used to render the compare table). It should have everything we need.  If you go that direction and have questions, holler!

https://github.com/covid-projections/covid-projections/blob/31fe51e2dbe089df3431efc8f2211ad2a45c80d4/src/common/location_summaries.ts#L25

schamp:
Nice, if I can get what I need from that and defer loading the actual projections it will be really nice.  One small obvious step at a time : )

PR #3097, opened 2021-03-17
map: minor performance improvements

pnavarrc:
I'm not 100% sure that this will make a difference, but it might.

- Applying effects on hover to all the path children of the Map would potentially impact the county paths, so I removed it from the USAMapWrapper and moved it to the state geographies instead.


pnavarrc:
This needs more research, closing!

PR #3098, opened 2021-03-18
Migrate time manipulation utils to Luxon

faihegberg:
This PR migrates time additions and subtractions to Luxon and adds a time util to parse time from Unix timestamp (in milliseconds).

### DIFF
--- a/src/common/utils/time-utils.ts
+++ b/src/common/utils/time-utils.ts
@@ -1,5 +1,4 @@
 import { DateTime } from 'luxon';
-import moment from 'moment';
 
 export enum TimeFormat {
   YYYY_MM_DD = 'yyyy-MM-dd', // 2020-03-01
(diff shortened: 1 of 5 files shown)


src/common/utils/time-utils.ts
```diff
@@ -1,31 +1,41 @@
...
 
 export enum TimeUnit {
-  HOURS = 'hours',
-  DAYS = 'days',
-  WEEKS = 'weeks',
-  MONTHS = 'months',
+  HOURS,
+  DAYS,
+  WEEKS,
+  MONTHS,
```
pnavarrc, line 18: It's better to add a string values to enums to make it easier to debug (unless the enum represents numeric amounts)

```ts
export enum TimeUnit {
  HOURS = 'hours',
  DAYS = 'days',
  // ...
}
```

if we do that, we can use their string value in other places, for example

```ts
function subtractTime(date: Date, amount: number, unit: TimeUnit): Date {
  // If unit = TimeUnit.HOURS, then it becomes "hours" when
  // building and then we get amountOptions = { "hours": amount }; 
  const amountOption = { [unit]: amount };
  return DateTime.fromJSDate(date).minus(amountOption).toJSDate();
}
```

we can do the same for the `addTime` function

pnavarrc (approved):
Added some suggestions to simplify the code, but looks good otherwise!

PR #3104, opened 2021-03-18
fix tooltip dates for charts

pnavarrc:


src/components/Charts/utils.ts
```diff
@@ -199,3 +199,11 @@ export function getTimeAxisTicks(from: Date, to: Date) {
     addTime(dateFrom, i, TimeUnit.MONTHS),
   );
 }
+
+export function getColumnDate({ x }: { x: number }): Date {
```
ghop02, line 203: I think referring to this as "column date" is a bit unclear - i'm not entirely sure what "column date" means without more investigation.  I think this number is a unixtime milli representation right? Maybe a util in time utils `formatUnixtimeMilliDate(x: number): string` or similar could be more clear?

pnavarrc: This takes a `Column` from `common/models/Projection` and returns a date, in this particular case we want to interpret the date as UTC (since unix milliseconds don't have timezone info). I think it's too specific to move to non-chart contexts, but if it comes up in other places we should move it to `time-utils`

pnavarrc: Updated the types to clarify a bit

pnavarrc:
[Bug: Chart tooltips show the wrong date (timezone issue)](https://trello.com/c/3pss1Cmo/1126-bug-chart-tooltips-show-the-wrong-date-timezone-issue)

mikelehen (approved):

PR #3110, opened 2021-03-19
add missing aria-labels on the new footer

pnavarrc:
- The `ExternalLink` component was not passing additional link props to the underlying `a` element
- The link to the logo didn't have an alt-text (set to "Covid Act Now")

### DIFF
--- a/src/components/NewFooter/MenuContent.tsx
+++ b/src/components/NewFooter/MenuContent.tsx
@@ -46,7 +46,7 @@ const MenuContent: React.FC = () => {
         ))}
       </Section>
       <Section>
-        <LogoWrapper to="/">
+        <LogoWrapper to="/" aria-label="Covid Act Now">
           <LogoNonUrl />
         </LogoWrapper>
         <AboutCopy>{aboutUs}</AboutCopy>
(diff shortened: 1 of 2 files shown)


src/components/NewFooter/MenuContent.tsx
```diff
@@ -46,7 +46,7 @@ const MenuContent: React.FC = () => {
         ))}
       </Section>
       <Section>
-        <LogoWrapper to="/">
+        <LogoWrapper to="/" aria-label="Covid Act Now">
           <LogoNonUrl />
```
mikelehen, line 50: FWIW- An alternative fix might be to add a `title` to the LogoNonUrl `<svg>` tag, which I think is the svg-equivalent of `alt` text on an image.  But the aria-label on the link seems fine too!

mikelehen (approved):
Thanks for catching / fixing!

src/components/NewFooter/MenuContent.tsx
pnavarrc: `title` is not recommended in most cases because it's inconsistently handled by screen readers

mikelehen: Ah, excellent to know. Thanks!

PR #3112, opened 2021-03-19
Use summary for location page header

schamp:
This allows deferred loading of the projections needed for the other graphs below the fold.

### DIFF
--- a/src/common/models/Projections.ts
+++ b/src/common/models/Projections.ts
@@ -8,6 +8,12 @@ import { LocationSummary, MetricSummary } from 'common/location_summaries';
 import { RegionSummaryWithTimeseries } from 'api/schema/RegionSummaryWithTimeseries';
 import { County, Region } from 'common/regions';
 
+export type MetricValue = number | null;
+
+export type MetricValues = {
+  [metric in Metric]: MetricValue;
+};
+
 /**
  * The complete set of data / metrics and related information for a given
  * location (state or county).
@@ -77,12 +83,12 @@ export class Projections {
     return this.getMetricValue(metric) !== null;
   }
 
-  getMetricValue(metric: Metric): number | null {
+  getMetricValue(metric: Metric): MetricValue {
     return this.primary.getMetricValue(metric);
   }
 
-  getMetricValues(): { [metric in Metric]: number | null } {
-    const result = {} as { [metric in Metric]: number | null };
+  getMetricValues(): MetricValues {
+    const result = {} as MetricValues;
     for (const metric of ALL_METRICS) {
       result[metric] = this.getMetricValue(metric);
     }
--- a/src/common/models/Projections.ts
+++ b/src/common/models/Projections.ts
@@ -8,10 +8,8 @@ import { LocationSummary, MetricSummary } from 'common/location_summaries';
 import { RegionSummaryWithTimeseries } from 'api/schema/RegionSummaryWithTimeseries';
 import { County, Region } from 'common/regions';
 
-export type MetricValue = number | null;
-
 export type MetricValues = {
-  [metric in Metric]: MetricValue;
+  [metric in Metric]: number | null;
 };
 
 /**
@@ -83,7 +81,7 @@ export class Projections {
     return this.getMetricValue(metric) !== null;
   }
 
-  getMetricValue(metric: Metric): MetricValue {
+  getMetricValue(metric: Metric): number | null {
     return this.primary.getMetricValue(metric);
   }
 
--- a/src/components/LocationPage/ChartsHolder.tsx
+++ b/src/components/LocationPage/ChartsHolder.tsx
@@ -181,9 +181,7 @@ const ChartsHolder = ({ region, chartId }: ChartsHolderProps) => {
     scrollToRecommendations();
   }, [chartId, metricRefs, isRecommendationsShareUrl]);
 
-  const initialFipsList = useMemo(() => {
-    return projections?.primary?.fips ? [projections.primary.fips] : [];
-  }, [projections]);
+  const initialFipsList = [region.fipsCode];
 
   const ccviScores = useCcviForFips(region.fipsCode);
 
--- a/src/components/LocationPage/ChartsHolder.tsx
+++ b/src/components/LocationPage/ChartsHolder.tsx
@@ -8,7 +8,6 @@ import {
 } from 'common/hooks';
 import { ALL_METRICS } from 'common/metric';
 import { Metric } from 'common/metricEnum';
-import { Projections } from 'common/models/Projections';
 import { Region, State, getStateName } from 'common/regions';
 import { getRecommendationsShareUrl } from 'common/urls';
 import {
@@ -33,6 +32,10 @@ import ChartBlock from './ChartBlock';
 import LocationPageBlock from './LocationPageBlock';
 import LocationPageHeader from './LocationPageHeader';
 import { ChartContentWrapper } from './ChartsHolder.style';
+import { useProjectionsFromRegion } from 'common/utils/model';
+import { MetricValues, Projections } from 'common/models/Projections';
+import { LoadingScreen } from 'screens/LocationPage/LocationPage.style';
+import { LocationSummary, useSummaries } from 'common/location_summaries';
 
 // TODO: 180 is rough accounting for the navbar and searchbar;
 // could make these constants so we don't have to manually update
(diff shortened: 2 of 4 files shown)


src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -44,13 +47,89 @@ const scrollTo = (div: null | HTMLDivElement, offset: number = 180) =>
     behavior: 'smooth',
   });
 
-interface ChartsHolderProps {
+interface RecommendationsProps {
   projections: Projections;
+  recommendationsRef: React.RefObject<HTMLDivElement>;
+}
+
+const Recommendations = ({
```
schamp, line 55: I broke Recommendations out as a wrapper for the recommend thing because 1) it was almost entirely independent and 2) its props depended largely on projection data.   By splitting it, it's easier to conditionally load based on projection data being available.

schamp: Is it important to break it out into a separate component file?  As it wasn't shared or used by anything else, it seemed okay to leave it here, but please let me know what you prefer.

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -44,13 +47,89 @@ const scrollTo = (div: null | HTMLDivElement, offset: number = 180) =>
...
+
+const summaryToStats = (summary: LocationSummary): MetricValues => {
+  const stats = {} as MetricValues;
+  for (const metric of ALL_METRICS) {
+    stats[metric] = summary.metrics[metric]?.value ?? null;
+  }
+  return stats;
+};
+
+const ChartsHolder = ({ region, chartId }: ChartsHolderProps) => {
```
schamp, line 128: now ChartsHolder doesn't need projections as a prop, so we can render it before projections are loaded.

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -102,46 +181,18 @@ const ChartsHolder = ({ projections, region, chartId }: ChartsHolderProps) => {
...
-  );
-
-  const recommendsShareUrl = getRecommendationsShareUrl(region);
-
-  const alarmLevel = projections.getAlarmLevel();
-  const recommendsShareQuote = getShareQuote(
-    projections.locationName,
-    alarmLevel,
-  );
+    return projections?.primary?.fips ? [projections.primary.fips] : [];
```
schamp, line 185: I'm not thrilled about how this works with projections possibly undefined, but I didn't see a better way, open to suggestions if you have any.

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -102,46 +181,18 @@ const ChartsHolder = ({ projections, region, chartId }: ChartsHolderProps) => {
...
-  }`;
+  const ccviScores = useCcviForFips(region.fipsCode);
 
-  // TODO(Chelsi): make these 2 functions less redundant?
-  const recommendationsFedModalCopy = getModalCopyWithFedLevel(
-    projection,
-    projections.locationName,
-    projections.getMetricValues(),
-  );
+  if (!locationSummary) {
```
schamp, line 190: finally after all the hooks, if there's no location summary we can't render at all.

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -170,15 +221,13 @@ const ChartsHolder = ({ projections, region, chartId }: ChartsHolderProps) => {
...
-  const ccviScores = useCcviForFips(region.fipsCode);
-
   // TODO(pablo): Create separate refs for signup and share
   return (
     <>
       <ChartContentWrapper>
         <LocationPageHeader
-          alarmLevel={projections.getAlarmLevel()}
-          stats={projections.getMetricValues()}
+          alarmLevel={alarmLevel}
```
schamp, line 229: now LocationPageHeader has no dependencies on Projections \o/ and can get everything from the locationSummary.

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -200,30 +249,29 @@ const ChartsHolder = ({ projections, region, chartId }: ChartsHolderProps) => {
...
-            shareUrl={recommendsShareUrl}
-            shareQuote={recommendsShareQuote}
-            recommendationsRef={recommendationsRef}
-            feedbackFormUrl={recommendationsFeedbackForm}
-            fedLevel={getFedLevel(projections.primary)}
-            harvardLevel={getHarvardLevel(projections.primary)}
-            harvardModalLocationCopy={recommendationsHarvardModalCopy}
-            fedModalLocationCopy={recommendationsFedModalCopy}
-          />
+          {!projections ? (
```
schamp, line 252: and now each other thing displaying is conditional on projections being available.

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -200,30 +249,29 @@ const ChartsHolder = ({ projections, region, chartId }: ChartsHolderProps) => {
...
-              <ChartBlock
-                key={metric}
-                metric={metric}
-                projections={projections}
-                chartRef={metricRefs[metric]}
-                isMobile={isMobile}
-                region={region}
-                stats={stats}
-              />
+              {!projections ? (
```
schamp, line 262: It seemed better to have a separate loading thing for each chart rather than one loading thing for the whole bunch of them but I'm happy to  hange this if preferred.

src/screens/LocationPage/LocationPage.tsx
```diff
@@ -42,17 +39,7 @@ function LocationPage({ region }: LocationPageProps) {
           mobileMenuOpen={mobileMenuOpen}
           setMobileMenuOpen={setMobileMenuOpen}
         />
```
schamp, line 41: now ChartsHolder is responsible for getting its own projections and loading them, so we can render it unconditionally.

src/components/LocationPage/ChartsHolder.tsx
pnavarrc: Thanks for adding all these useful comments! Separating the parts that depend on projections makes sense and makes the above the fold part load much faster compared to the rest (it feels almost instant in this branch), that's a huge improvement!

We are starting a redesign of the location page in a week or two, so we will have to implement/modify these blocks anyways, so don't worry about moving Recommendations to its own file, we can do that as part of the redesign. We will keep the structure of rendering what we can without projections for sure though!

src/components/LocationPage/ChartsHolder.tsx
pnavarrc: This can be probably reworked by using `region.fipsCode` instead. The Explore component takes a list of FIPS codes and grabs the projections itself, and since here we have the page region already we can use that instead.

src/components/LocationPage/ChartsHolder.tsx
pnavarrc: yeah, hooks cannot be called conditionally, I think we should move the hooks inside the components they belong to, but we can take care of that during the redesign (some of them are needed by more than one block I think)

src/components/LocationPage/ChartsHolder.tsx
schamp: ah, nice!  I was originally thinking to do that, but then I saw "no, wait, it's a list, and regions only have one FIPS code" and didn't think it through to see we're building a list of one here.

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -102,46 +181,16 @@ const ChartsHolder = ({ projections, region, chartId }: ChartsHolderProps) => {
...
-  );
-
-  const recommendsShareUrl = getRecommendationsShareUrl(region);
-
-  const alarmLevel = projections.getAlarmLevel();
-  const recommendsShareQuote = getShareQuote(
-    projections.locationName,
-    alarmLevel,
-  );
+  const initialFipsList = [region.fipsCode];
```
schamp, line 184: no reason to `useMemo` any more.

src/common/models/Projections.ts
```diff
@@ -8,6 +8,12 @@ import { LocationSummary, MetricSummary } from 'common/location_summaries';
 import { RegionSummaryWithTimeseries } from 'api/schema/RegionSummaryWithTimeseries';
 import { County, Region } from 'common/regions';
 
+export type MetricValue = number | null;
```
mikelehen, line 11: Hrm, not sure this is a helpful abstraction.  `MetricValue` doesn't immediately imply to me that it's a nullable value and so I think the code may actually be clearer just sticking with `number | null` rather than introducing extra indirection here.  So I'd prefer drop this type.

mikelehen (commented):
🎉 

One minor nit but this looks great!

mikelehen (approved):

src/common/models/Projections.ts
schamp: fair point.  I was trying to simplify the typing, because of having to pass it for props in a couple places.  But I can revert this bit here.

PR #3114, opened 2021-03-19
Fix typo in alert emails utils

faihegberg:


faihegberg:
[Regression: Send Vaccine Email Alerts Fails with new time handling](https://trello.com/c/z2GXD6m3/1129-regression-send-vaccine-email-alerts-fails-with-new-time-handling)

pnavarrc (approved):
Thanks for the fix! I tested using `yarn send-test-email`, works fine now :)

PR #3132, opened 2021-03-22
Create start of time util and get time diff util

faihegberg:
This PR creates two utils, get start of time util which gets the start of a timeframe based on a specified time unit, and get time difference util which calculates the difference between two points of time in the specified time unit.

### DIFF
--- a/src/common/utils/time-utils.test.ts
+++ b/src/common/utils/time-utils.test.ts
@@ -1,56 +1,90 @@
 import {
-  TimeFormat,
+  DateFormat,
   TimeUnit,
   formatDateTime,
+  formatUTCDateTime,
   parseDateString,
   parseDateUnix,
-  dateToUTC,
   addTime,
   subtractTime,
+  getStartOf,
+  getTimeDiff,
 } from 'common/utils/time-utils';
 
 // Date object constructor takes a zero-indexed month number.
 
-const testDate = new Date(2020, 2, 1);
-
 describe('time formatting', () => {
+  const testDate = new Date(2020, 2, 1);
   test('Date in ISO format', () => {
-    expect(formatDateTime(testDate, TimeFormat.YYYY_MM_DD)).toBe('2020-03-01');
+    expect(formatDateTime(testDate, DateFormat.YYYY_MM_DD)).toBe('2020-03-01');
   });
-  test('Dash seperated date', () => {
-    expect(formatDateTime(testDate, TimeFormat.MM_DD_YYYY)).toBe('03/01/2020');
+  test('Dash separated date', () => {
+    expect(formatDateTime(testDate, DateFormat.MM_DD_YYYY)).toBe('03/01/2020');
   });
   test('Date in full month, day, year', () => {
-    expect(formatDateTime(testDate, TimeFormat.MMMM_D_YYYY)).toBe(
+    expect(formatDateTime(testDate, DateFormat.MMMM_D_YYYY)).toBe(
       'March 1, 2020',
     );
   });
   test('Date in shorthand month, day, year', () => {
-    expect(formatDateTime(testDate, TimeFormat.MMM_DD_YYYY)).toBe(
+    expect(formatDateTime(testDate, DateFormat.MMM_DD_YYYY)).toBe(
       'Mar 01, 2020',
     );
   });
   test('Date in full month and unpadded day', () => {
-    expect(formatDateTime(testDate, TimeFormat.MMMM_D)).toBe('March 1');
+    expect(formatDateTime(testDate, DateFormat.MMMM_D)).toBe('March 1');
   });
   test('Past date', () => {
-    expect(formatDateTime(new Date(2015, 1, 1), TimeFormat.YYYY_MM_DD)).toBe(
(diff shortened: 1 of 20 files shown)


src/common/utils/time-utils.test.ts
```diff
@@ -115,3 +117,50 @@ describe('subtract time', () => {
...
+describe('get start of time', () => {
+  test('Start of hour', () => {
+    expect(getStartOf(new Date(2020, 2, 1, 0, 30), TimeUnit.HOURS)).toEqual(
+      testDate,
+    );
+  });
+  test('Start of day', () => {
+    expect(getStartOf(new Date(2020, 2, 1, 8, 30), TimeUnit.DAYS)).toEqual(
+      testDate,
+    );
```
pnavarrc, line 130: I think it's better to not use variables in test cases with custom input to make it easier for us to compare the expected input and output.

```ts
expect(getStartOf(new Date(2020, 2, 1, 8, 30), TimeUnit.DAYS))
  .toEqual(new Date(2020, 2, 1, 0, 0));
```

If a group of functions share the input value (the formatting functions, for example), it might make sense to use a variable for the input. Consider moving `testDate` to be inside `describe('time formatting', () => { ... })` block

src/common/utils/time-utils.test.ts
```diff
@@ -115,3 +117,50 @@ describe('subtract time', () => {
...
+  });
+  test('Difference in weeks', () => {
+    expect(
+      getTimeDiff(testDate, new Date(2020, 2, 15), TimeUnit.WEEKS),
+    ).toEqual(-2);
+  });
+  test('Difference in months', () => {
+    expect(
+      getTimeDiff(testDate, new Date(2020, 5, 1), TimeUnit.MONTHS),
+    ).toEqual(-3);
```
pnavarrc, line 164: Consider adding a test case for positive diffs? Just one or two examples where the first date passed to `getTimeDiff` is after the date passed as second argument

src/common/utils/time-utils.ts
```diff
@@ -69,3 +72,15 @@ function getTimeUnitOption(amount: number, unit: TimeUnit): DurationObject {
...
 }
+
+export function getStartOf(date: Date, unit: TimeUnit): Date {
+  return moment(date).startOf(unit).toDate();
+  // TODO (Fai): Migrate to below implementation.
+  // return DateTime.fromJSDate(date).startOf(unit).toJSDate();
+}
+
+export function getTimeDiff(date1: Date, date2: Date, unit: TimeUnit) {
+  return moment(date1).diff(date2, unit);
```
pnavarrc, line 83: Consider adding a comment to say what the function does, and maybe an example. I think it's easy to not remember which date are you supposed to pass first, including maybe an example?

```ts
/**
 * Calculate the difference between 2 dates in the given 
 * time unit. Example:
 * 
 *   getTimeDiff(dateTomorrow, dateToday, TimeUnit.DAYS));  // 1
 */
```

pnavarrc (changes requested):
Added some notes!

pnavarrc (approved):
Looks great!

src/common/utils/time-utils.test.ts
```diff
@@ -7,13 +7,14 @@ import {
...
 } from 'common/utils/time-utils';
 
 // Date object constructor takes a zero-indexed month number.
 
-const testDate = new Date(2020, 2, 1);
-
 describe('time formatting', () => {
+  const testDate = new Date(2020, 2, 1);
   test('Date in ISO format', () => {
     expect(formatDateTime(testDate, TimeFormat.YYYY_MM_DD)).toBe('2020-03-01');
```
chasulin, line 19: just looked at the `TimeFormat` enum--maybe naming it `DateFormat` would make more sense?

src/common/utils/time-utils.test.ts
```diff
@@ -1,56 +1,90 @@
...
-
 describe('time formatting', () => {
+  const testDate = new Date(2020, 2, 1);
   test('Date in ISO format', () => {
-    expect(formatDateTime(testDate, TimeFormat.YYYY_MM_DD)).toBe('2020-03-01');
+    expect(formatDateTime(testDate, DateFormat.YYYY_MM_DD)).toBe('2020-03-01');
   });
   test('Dash seperated date', () => {
-    expect(formatDateTime(testDate, TimeFormat.MM_DD_YYYY)).toBe('03/01/2020');
+    expect(formatDateTime(testDate, DateFormat.MM_DD_YYYY)).toBe('03/01/2020');
```
pnavarrc, line 22: typo, sep**e**rated → **separated**

src/common/utils/time-utils.test.ts
```diff
@@ -1,56 +1,90 @@
...
 
+describe('utc time formatting', () => {
+  // Pacific timezone is set as the default in package.json for testing consistency.
+  const testDate = new Date(2020, 2, 1, 20); // March 1st 2020 8 pm. Pacific.
+  test('Date in ISO format', () => {
+    expect(formatUTCDateTime(testDate, DateFormat.YYYY_MM_DD)).toBe(
+      '2020-03-02',
+    );
+  });
+  test('Dash seperated date', () => {
```
pnavarrc, line 57: same

PR #3135, opened 2021-03-23
performance: preload fonts

pnavarrc:
- Preload the fonts. Lighthouse keeps saying that they are in the critical rendering path. I tested this branch and couldn't see an improvement, but no harm either, so probably worth trying.
- Remove extra semicolon in a JSX statement.

**Note**
The linter re-indented the `index.html` file, but the only changes are 

```html
<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
<link
  rel="preload"
  as="style"
  href="https://fonts.googleapis.com/css2?family=Source+Code+Pro:wght@400;700&family=Roboto:wght@400;500;700&display=swap"/>
```

### DIFF
--- a/src/screens/internal/ShareImage/HomeShareCardImage.tsx
+++ b/src/screens/internal/ShareImage/HomeShareCardImage.tsx
@@ -14,7 +14,7 @@ const HomeShareCardImage = () => {
     <DarkScreenshotWrapper className={SCREENSHOT_CLASS}>
       <Header isHomePage />
       <ShareCardWrapper isHomePage>
-        <SocialLocationPreviewMap />;
+        <SocialLocationPreviewMap />
       </ShareCardWrapper>
     </DarkScreenshotWrapper>
   );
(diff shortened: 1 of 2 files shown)


src/screens/internal/ShareImage/HomeShareCardImage.tsx
```diff
@@ -14,7 +14,7 @@ const HomeShareCardImage = () => {
     <DarkScreenshotWrapper className={SCREENSHOT_CLASS}>
       <Header isHomePage />
       <ShareCardWrapper isHomePage>
-        <SocialLocationPreviewMap />;
+        <SocialLocationPreviewMap />
```
mikelehen, line 17: 😬

mikelehen (approved):
Looks good.  I agree I don't perceive any difference and I'm not too surprised.

It seems like Chrome is queuing the download of the fonts at the same time as the bundles, but the bundles take much longer to download and nothing can actually be rendered on the page until the bundle completes...  so even though the font isn't technically blocking, it ends up happening before the page renders anything so Chrome / Light House considers it blocking, I guess.

image

PR #3142, opened 2021-03-23
add aria-label to nav bar home and back links

pnavarrc:
- Add aria-labels for the Logo and back links on the NavBar component.
- For logos and links that go to the home page, it's considered better to spell out the name of the website rather than having something more contextual like "Home" or "Back"

### DIFF
--- a/src/components/AppBar/NavBar.tsx
+++ b/src/components/AppBar/NavBar.tsx
@@ -112,7 +112,11 @@ const NavBar: React.FC = () => {
     <Style.AppBar position="sticky" color="transparent" elevation={0}>
       <Style.Toolbar>
         {isLocationPage(pathname) && (
-          <Style.BackLink to="/" onClick={() => trackNavigation('Back Home')}>
+          <Style.BackLink
+            to="/"
+            onClick={() => trackNavigation('Back Home')}
+            aria-label="Covid Act Now"
+          >
             <ArrowBack />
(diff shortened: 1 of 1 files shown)


src/components/AppBar/NavBar.tsx
```diff
@@ -112,14 +112,19 @@ const NavBar: React.FC = () => {
     <Style.AppBar position="sticky" color="transparent" elevation={0}>
       <Style.Toolbar>
         {isLocationPage(pathname) && (
-          <Style.BackLink to="/" onClick={() => trackNavigation('Back Home')}>
+          <Style.BackLink
+            to="/"
+            onClick={() => trackNavigation('Back Home')}
+            aria-label="Covid Act Now"
```
chasulin, line 118: should this be more descriptive than "covid act now"? something like "back to covid act now homepage" ?

pnavarrc: Let's do that 👍

pnavarrc: (back to Covid Act Now). I read somewhere that in footer and top navs just having the name of the website is better because is less verbose, and it's a well known pattern for most users.

chasulin: oh solid, sounds good!

chasulin (approved):

PR #3143, opened 2021-03-23
Compare filters clean up

chasulin:
Adds `useCompareSliderValueMap` hook to update the MUI Slider component's numerical value whenever the filter selection changes on homepage or location page. We were previously passing `setHomepageSliderValue` and `setSliderValue` down as props and doing this update in multiple places in the compare code. This unifies into a single hook that is called once from the top (in `CompareMain`)


src/common/utils/compare.ts
```diff
@@ -365,3 +365,34 @@ export const getCompareSubheader = (region: Region): string => {
     return 'Compare counties';
   }
 };
+
+// Value maps for filters' slider components (0, 50, and 99 are numerical values required by MUI Slider).
```
chasulin, line 369: These maps were previously in Filter.tsx and CompareMain.tsx. Moved them all here

src/components/Compare/CompareMain.tsx
```diff
@@ -175,15 +163,16 @@ const CompareMain = (props: {
...
-  // Location page slider:
-  const defaultSliderValue = scopeValueMap[geoScope];
-  const [sliderValue, setSliderValue] = useState(defaultSliderValue);
-
-  // Homepage slider:
-  const defaultHomepageSliderValue = homepageScopeValueMap[homepageScope];
-  const [homepageSliderValue, setHomepageSliderValue] = useState(
-    defaultHomepageSliderValue,
-  );
+  // Since the route isn't changing when navigating between county pages within the same state, state variables weren't resetting. This forces a reset:
```
chasulin, line 166: This useEffect is already in develop, but **this current** PR's branch is based off of the branch where the effect was added, so the change is showing up here as well

src/common/hooks/useHomepageCompareSliderMap.ts
```diff
@@ -0,0 +1,25 @@
...
+  homepageScope: HomepageLocationScope,
+) {
+  const defaultHomepageSliderValue = homepageScopeValueMap[homepageScope];
+  const [homepageSliderValue, setHomepageSliderValue] = useState(
+    defaultHomepageSliderValue,
+  );
+
+  useEffect(() => {
+    setHomepageSliderValue(homepageScopeValueMap[homepageScope]);
+  }, [homepageScope]);
```
pnavarrc, line 22: I think `setHomepageSliderValue` needs to be a dependency as well

src/common/hooks/useLocationPageCompareSliderMap.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export default function useLocationPageCompareSliderMap(
+  geoScope: GeoScopeFilter,
+) {
+  const defaultSliderValue = scopeValueMap[geoScope];
+  const [sliderValue, setSliderValue] = useState(defaultSliderValue);
+
+  useEffect(() => {
+    setSliderValue(scopeValueMap[geoScope]);
+  }, [geoScope]);
```
pnavarrc, line 17: `setSliderValue` should be a dependency here too

pnavarrc (approved):
Looks good!

src/common/hooks/useHomepageCompareSliderMap.ts
chasulin: hm the linter doesnt complain when its omitted. just double checked [in the docs](https://reactjs.org/docs/hooks-reference.html) - ("React guarantees that setState function identity is stable and won’t change on re-renders. This is why it’s safe to omit from the useEffect or useCallback dependency list.")

src/common/hooks/useLocationPageCompareSliderMap.ts
chasulin: same as above^

src/common/hooks/useHomepageCompareSliderMap.ts
pnavarrc: nice!

PR #3144, opened 2021-03-23
Update blocked metrics / disclaimers.

mikelehen:
* NH no longer needs its disclaimer.
* Dallas County test positivity is fixed.
* Oklahoma counties are reporting zero new cases. Disabled. https://trello.com/c/HdAKfp49/1139-all-ok-counties-except-2-no-longer-reporting-cases
* Added disclaimer for IL counties where 1+ dose number may be an underestimate. https://trello.com/c/buXoYmKy/1140-il-county-first-doses-underreporting


src/cms-content/region-overrides/region-overrides.json
```diff
@@ -75,11 +53,19 @@
       "context": "https://trello.com/c/7tqJyFna/1099-disabled-white-pine-county-nv-vaccine-data-due-to-anomoly"
     },
     {
-      "include": "region",
+      "include": "subregions",
       "metric": "metrics.caseDensity",
       "blocked": true,
-      "region": "47157",
-      "context": "https://trello.com/c/aB0EunUd/1108-disabled-shelby-county-tn-due-to-trailing-zeros-bringing-dnc-down"
+      "region": "OK",
```
ghop02, line 59: Worth adding a disclaimer here? Not sure if it shows up if it's fully blocked

ghop02 (approved):

src/cms-content/region-overrides/region-overrides.json
mikelehen: Yeah, it won't show up right now so there's no benefit, but that's a future improvement I've been thinking about...

PR #3161, opened 2021-03-25
Remove duplicate state + county topology data

schamp:
These files were in the components/Map folder, but only the ones in common/data were being used.  Out they go!


pnavarrc (approved):
Thanks for cleaning up! 🧹🧹🧹🧹

PR #3163, opened 2021-03-25
Change update location summaries to query by state rather than pulling counties in all at once

ghop02:
We finally hit the limit of string size for parsing a json file for the version of node we're on.  This splits up the querying by states - we should be good for a while (and maybe can even upgrade our version of node now!)

### DIFF
--- a/scripts/county_history_data/generate.ts
+++ b/scripts/county_history_data/generate.ts
@@ -8,11 +8,11 @@
 import fs from 'fs-extra';
 import path from 'path';
 import _ from 'lodash';
-import { fetchAllCountyProjections } from '../../src/common/utils/model';
+import { fetchAllCountyProjectionsByState } from '../../src/common/utils/model';
 import { Projection } from '../../src/common/models/Projection';
 
 async function main() {
-  const allProjections = await fetchAllCountyProjections();
+  const allProjections = await fetchAllCountyProjectionsByState();
   const result = {} as {
     [date: string]: {
       [fips: string]: { population: number; caseDensity: number };
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -66,6 +66,19 @@ export class Api {
   }
 
   /**
+   * Fetches the summary+timeseries for every region in the specified state
+   */
+  async fetchCountySummariesWithTimeseriesForState(
+    state: State,
+  ): Promise<RegionSummaryWithTimeseries[]> {
+    const all = await this.fetchApiJson<AggregateRegionSummaryWithTimeseries>(
+      `county/${state.stateCode}.timeseries.json`,
+    );
+    assert(all != null, 'Failed to load timeseries');
+    return all;
+  }
+
+  /**
    * Fetches the summary+timeseries for a region. Returns null if not found.
    */
   async fetchSummaryWithTimeseries(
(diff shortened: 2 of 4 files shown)


scripts/county_history_data/generate.ts
```diff
@@ -8,11 +8,11 @@
 import fs from 'fs-extra';
 import path from 'path';
 import _ from 'lodash';
-import { fetchAllCountyProjections } from '../../src/common/utils/model';
+import { fetchAllCountyProjectionsByState } from '../../src/common/utils/model';
 import { Projection } from '../../src/common/models/Projection';
 
 async function main() {
-  const allProjections = await fetchAllCountyProjections();
+  const allProjections = await fetchAllCountyProjectionsByState();
```
mikelehen, line 15: I think you can undo this change?

src/common/utils/model.ts
```diff
@@ -100,7 +100,14 @@ export function fetchAllStateProjections(snapshotUrl: string | null = null) {
 
 /** Returns an array of `Projections` instances for all counties. */
 const cachedCountiesProjections: { [key: string]: Promise<Projections[]> } = {};
-export function fetchAllCountyProjections(snapshotUrl: string | null = null) {
+export function fetchAllCountyProjections(
+  snapshotUrl: string | null = null,
+  queryByState: boolean = true,
```
mikelehen, line 105: I don't think we really need this flag.  We can just merge these methods and always query by state.

src/common/utils/model.ts
```diff
@@ -123,6 +130,48 @@ export function fetchAllCountyProjections(snapshotUrl: string | null = null) {
...
+  state: State,
+  snapshotUrl: string | null = null,
+) {
+  snapshotUrl = snapshotUrl || getSnapshotUrlOverride();
+  async function fetch() {
+    const all = await new Api(
+      snapshotUrl,
+    ).fetchCountySummariesWithTimeseriesForState(state);
+    return all
+      .filter(summaryWithTimeseries => {
```
mikelehen, line 158: I think we can drop this filter because this function should never be called for states not in our regions instance (this filtering originally showed up when e.g. we started adding territories to the backend but the frontend didn't know about them).

src/api/index.ts
```diff
@@ -65,6 +65,19 @@ export class Api {
     return all;
   }
 
+  /**
+   * Fetches the summary+timeseries for every region in the specified state
```
mikelehen, line 69: region=>county ?

mikelehen (approved):
Thanks for knocking this out.  Would like a few small cleanups, but overall looks good.

src/common/utils/model.ts
ghop02: This is actually filtering counties not states - I'd lean towards keeping it just to make sure we don't include counties that  aren't matching 1:1 (or we can continue with the strict call).  but can remove if you feel strongly

mikelehen: Oh, I see what you mean.  Yeah, I guess I don't feel super strongly but I also can't immediately think of a case where the backend would have extra counties for a state and we would want to silently ignore that, so I probably lean towards removing this code, which it looks like you did in the follow-up PR.  Thanks!

PR #3165, opened 2021-03-25
Switching implementation of fetch all counties to just use query by states

ghop02:
+ a little bit of other misc feedback

### DIFF
--- a/src/common/utils/model.ts
+++ b/src/common/utils/model.ts
@@ -99,9 +99,7 @@ export function fetchAllStateProjections(snapshotUrl: string | null = null) {
 }
 
 const cachedCountiesProjections: { [key: string]: Promise<Projections[]> } = {};
-export async function fetchAllCountyProjections(
-  snapshotUrl: string | null = null,
-) {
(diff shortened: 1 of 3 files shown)


src/common/utils/model.ts
```diff
@@ -98,49 +98,24 @@ export function fetchAllStateProjections(snapshotUrl: string | null = null) {
...
-        // We don't want to return county projections for counties that we
-        // don't have in the regions instance
-        const region = regions.findByFipsCode(summaryWithTimeseries.fips);
-        return region instanceof County && Boolean(region.state);
-      })
-      .map(summaryWithTimeseries => {
-        const region = regions.findByFipsCodeStrict(summaryWithTimeseries.fips);
-        return new Projections(summaryWithTimeseries, region);
-      });
+    return allProjections.flatMap(obj => obj);
```
mikelehen, line 111: Looks like you can just do `.flat()` ?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

mikelehen (approved):

PR #3168, opened 2021-03-25
Trims FAQ ids

chasulin:
Noticed while [qa'ing this PR](https://github.com/covid-projections/covid-projections/pull/3087) that some share URLs weren't working properly because their IDs were entered into the CMS with trailing white space. This fixes


pnavarrc (approved):
🙏

PR #3174, opened 2021-03-25
Fix date precision on explore

faihegberg:
This PR gets rid of floating point date values on Explore charts.

### DIFF
--- a/src/components/Explore/utils.ts
+++ b/src/components/Explore/utils.ts
@@ -343,9 +343,9 @@ const pluralizeDays = (num: number) => pluralize(num, 'day', 'days');
  * today, 1 day ago, 5 days ago, 3 weeks and 2 days ago, 5 weeks ago, etc.
  */
 export function weeksAgo(dateFrom: Date, dateTo: Date) {
-  const totalDays = getTimeDiff(dateTo, dateFrom, TimeUnit.DAYS);
+  const totalDays = Math.floor(getTimeDiff(dateTo, dateFrom, TimeUnit.DAYS));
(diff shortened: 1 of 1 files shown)


src/components/Explore/utils.ts
```diff
@@ -343,9 +343,9 @@ const pluralizeDays = (num: number) => pluralize(num, 'day', 'days');
  * today, 1 day ago, 5 days ago, 3 weeks and 2 days ago, 5 weeks ago, etc.
  */
 export function weeksAgo(dateFrom: Date, dateTo: Date) {
-  const totalDays = getTimeDiff(dateTo, dateFrom, TimeUnit.DAYS);
+  const totalDays = Math.floor(getTimeDiff(dateTo, dateFrom, TimeUnit.DAYS));
   const totalWeeks = Math.floor(totalDays / 7);
-  const numDays = totalDays % 7;
+  const numDays = Math.floor(totalDays % 7);
```
chasulin, line 348: nice! 2 little nits-

maybe rename `numDays` to `numDaysWithinPastWeek` or something like that? to make the distinction between this and `totalDays`

another option for the rounding is to use `formatDecimal` in `common/utils/index`

pnavarrc: There is `formatInteger` too (which adds thousands separators, but in this case that would never happen because the number would always be between 0 and 6)

pnavarrc:
[Weird Explore Tooltip Sigfigs](https://trello.com/c/5m4kyd0n/1168-weird-explore-tooltip-sigfigs)

pnavarrc (approved):
Looks good! 👏

PR #3179, opened 2021-03-26
Text tooltips update

chasulin:
Implements new styling for tooltip anchor text ([figma](https://www.figma.com/file/l4Lra5SwXH2NX6XtZdDVWd/Buttons-and-Links?node-id=66%3A607)), applies to data api table tooltips and chart disclaimer tooltips. Renames `DisclaimerTooltip` to `TextTooltip`

[Trello ticket](https://trello.com/c/jVsSBn7E/1068-implement-new-design-pattern-for-text-that-triggers-tooltips)

### DIFF
--- a/src/components/InfoTooltip/TextTooltip.tsx
+++ b/src/components/InfoTooltip/TextTooltip.tsx
@@ -2,7 +2,7 @@ import React, { useState } from 'react';
 import {
   StyledTooltip,
   StyledCloseIcon,
-  SmallTooltipAnchorText,
+  TooltipAnchorText,
 } from './Tooltip.style';
 import { StyledTooltipProps } from 'components/InfoTooltip';
 import VisuallyHiddenDiv from 'components/VisuallyHidden/VisuallyHidden';
@@ -15,10 +15,6 @@ const TextTooltip: React.FC<StyledTooltipProps> = props => {
   const isMobile = useBreakpoint(600);
 
   const idForAccessability = uuidv4();
-
-  const AnchorTextComponent = props.altAnchorComponent
-    ? props.altAnchorComponent
-    : SmallTooltipAnchorText;
 
   return (
     <>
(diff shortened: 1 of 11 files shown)


chasulin:
[Implement new design pattern for text that triggers tooltips](https://trello.com/c/jVsSBn7E/1068-implement-new-design-pattern-for-text-that-triggers-tooltips)

src/components/InfoTooltip/TextTooltip.tsx
```diff
@@ -1,17 +1,25 @@
...
-const DisclaimerTooltip: React.FC<StyledTooltipProps> = props => {
+const TextTooltip: React.FC<StyledTooltipProps> = props => {
   const [isOpen, setIsOpen] = useState(false);
   const isMobile = useBreakpoint(600);
 
   const idForAccessability = uuidv4();
 
+  const AnchorTextComponent = props.altAnchorComponent
+    ? props.altAnchorComponent
+    : SmallTooltipAnchorText;
```
chasulin, line 21: is there a better way? prob. but if/when we start implementing more text tooltips with varying styling, we can pass an anchor text component in every time. this feels good enough for now though

src/components/InfoTooltip/Tooltip.style.tsx
```diff
@@ -84,9 +84,20 @@ export const StyledMarkdown = styled(MarkdownBody)`
...
+`;
+
+export const SmallTooltipAnchorText = styled.span`
+  ${tooltipAnchorText};
   font-size: 0.875rem;
 `;
+
+export const LargeTooltipAnchorText = styled.span`
+  ${tooltipAnchorText};
+  font-size: 1rem;
```
mikelehen, line 102: Hrm...  It looks like the only thing `SmallTooltipAnchorText` and `LargeTooltipAnchorText` override is the font-size...   Can we just rely on the component using the tooltip to set the correct font size?  E.g. if I want a big font I could do:

```
<span style={{fontSize: '30px'}}><TextToolTip ... /></span>
```

In fact, I think maybe that already happens?  If I just remove the `font-size` setting here and on line 97 so that `SmallTooltipAnchorText` and `LargeTooltipAnchorText` are identical and don't set the font size, I think everything still renders exactly the same?  So maybe we don't need these and the `altAnchorComponent` thing?

mikelehen (approved):
This looks good with one potential cleanup, but I think Igor also may have wanted a bit more changes in terms of the hover/click interaction.  Right now when you hover, you get a "hand" mouse pointer indicating that it is clickable, like a link, but when you click you just get a blue focus rectangle:

image

I think Igor suggested either:
1. On hover the mouse pointer doesn't change to a hand pointer and clicking does nothing (no blue focus rectangle).
2. You get the hand pointer and clicking correctly navigates you to the underlying source link within the tooltip.

I think 1 is easier and works in all cases (including the data coverage table) so that makes more sense to me, but feel free to check with Josh if we want more input.  And feel free to submit this as-is and either update the trello ticket for the remaining work or open a new one.  Thanks!

src/components/InfoTooltip/Tooltip.style.tsx
chasulin: ahh yes, if i dont pass a font size, it grabs it from the parent. updating!

chasulin:
implementing option #1 and will revisit with josh on monday

PR #3197, opened 2021-03-28
remove unused functions and imports

schamp:
I found these to be unused when looking at performance stuff, may as well remove.

### DIFF
--- a/src/common/locations.ts
+++ b/src/common/locations.ts
@@ -1,5 +1,12 @@
 /** Helpers for dealing with the State / Counties dataset. */
-import US_STATE_DATASET from 'components/MapSelectors/datasets/us_states_dataset_01_02_2020.json';
+
+export interface Location {
+  state_fips_code: string;
+  full_fips_code?: string;
+  population: number;
+  state_code: string;
+  state: string;
+}
 
 /** Aggregations supported by the Explore chart. */
 export const AGGREGATED_LOCATIONS: Location[] = [
@@ -18,29 +25,3 @@ export const AGGREGATED_LOCATIONS: Location[] = [
     state_code: 'NAMC',
   },
 ];
-
-export interface Location {
-  county?: string;
-  county_url_name?: string;
-  county_fips_code?: string;
-  state_fips_code: string;
-  full_fips_code?: string;
-  population: number;
-  state_code: string;
-  state: string;
-  state_url_name?: string;
-}
-
-// TODO(michael): Make this return County type and fix up references.
-export function findCountyByFips(fips: string) {
-  const statesData = US_STATE_DATASET.state_county_map_dataset as any;
-  for (const state in statesData) {
-    const countiesData = statesData[state].county_dataset;
-    for (const county of countiesData) {
-      if (String(county.full_fips_code) === String(fips)) {
-        return county;
-      }
-    }
-  }
-  return undefined;
-}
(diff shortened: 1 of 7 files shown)


schamp:
the adjacency data file is still used by `regions_data.ts` so we can't delete that file yet.

src/common/utils/compare.ts
```diff
@@ -71,30 +62,6 @@ export function getAllCountiesOfState(stateCode: string): SummaryForCompare[] {
     .map(getLocationObj);
 }
 
-export function getAllMetroCounties(): SummaryForCompare[] {
```
schamp, line 74: by code search and TSC, none of these were used, easy removal!
and since this was the only use of `getCountyMsaCode` now, we can get rid of that too.

src/components/Compare/Compare.style.tsx
```diff
@@ -412,7 +412,7 @@ export const Row = styled(TableRow)<{
   }
 `;
 
-export const Footer = styled.div<{ isCounty: any }>`
+export const Footer = styled.div`
```
schamp, line 415: unused prop.

src/components/Compare/CompareMain.tsx
```diff
@@ -249,14 +248,9 @@ const CompareMain = (props: {
     return null;
   }
 
-  // TODO(chris): pusing down use of county as far as possible, but underlying compare
-  // table code not yet ready to stop using county as an input, so querying the legacy Location
-  // county API here.
-  const locationCounty = findCountyByFips(currentCounty?.region.fipsCode || '');
```
schamp, line 255: this analysis was a little trickier.  Ultimately, everywhere this was used was just boolean, which could be ascertained by just passing currentCounty in a boolean context, saving this expensive lookup (which scanned looped through all states and counties!) and in fact, eliminating the only use of this function.

src/components/Compare/CompareMain.tsx
```diff
@@ -249,14 +248,9 @@ const CompareMain = (props: {
...
   const sharedProps = {
     stateName: props.stateName,
     stateId,
-    county: locationCounty,
```
schamp, line 259: we can use `currentCounty` in this object in every case where we need `county`

src/components/Compare/CompareTable.tsx
```diff
@@ -279,7 +277,7 @@ const CompareTable = (props: {
         region={region}
       />
       {!props.isModal && (
-        <Footer isCounty={props.county}>
```
schamp, line 282: removed this unused prop.

src/components/Compare/Filters.tsx
```diff
@@ -17,7 +17,7 @@ import HomepageSlider from './HomepageSlider';
 const Filters = (props: {
   isHomepage?: boolean;
   stateId?: string;
-  county?: any | null;
+  isCounty: boolean;
```
schamp, line 20: change this to a boolean and fix the name

src/common/locations.ts
```diff
@@ -1,17 +1,13 @@
...
-import has from 'lodash/has';
-import US_STATE_DATASET from 'components/MapSelectors/datasets/us_states_dataset_01_02_2020.json';
-import countyAdjacencyMsa from './data/county_adjacency_msa.json';
 
-interface AdjacencyData {
-  [fips: string]: {
-    adjacent_counties: string[];
-    msa_code?: string;
-  };
+export interface Location {
```
schamp, line 3: not all these props are needed anymore.

chasulin (approved):
looks good to me! thanks for the clean up!

PR #3200, opened 2021-03-29
New button system

chasulin:
Implements new uniform button system ([figma](https://www.figma.com/file/l4Lra5SwXH2NX6XtZdDVWd/Buttons-and-Links?node-id=309%3A1375)) using theming. Adds to storybook, doesn't yet pull into the frontend.

We will have 5 main buttons: `OutlinedButton`, `FilledButton`, `LargeOutlinedButton`, `LargeFilledButton`, `TextButton`. Each is built off of `BaseButton`, which receives props and returns the correct type of button without styles (ie. `to` prop outputs a react router `Link`, `to` prop with `#` returns a HashLink, etc etc). Styles for each of the 5 buttons are applied in `MainButtons.style.tsx`.

In terms of use, we should use these 5 buttons **directly** in the component (rather adding any further styles with styled()) in order to keep uniformity. They are unaware of their context, so margins should be implemented in a parent/wrapping component.


chasulin:
[Implement new button and link styles](https://trello.com/c/D6A7AFNf/1046-implement-new-button-and-link-styles)

src/assets/theme/buttons.tsx
```diff
@@ -0,0 +1,71 @@
...
+  disabledText: string;
+  icon: string;
+  disabledIcon: string;
+  fontFamily: any; // fix
+}
+
+const newBlue = '#3567FD';
+const newBlueDark = '#002CB4';
+const newBlueLight = 'rgba(53, 103, 253, .15)';
+const lightGrayFigma = '#e0e0e0';
```
chasulin, line 27: will move all of these into colors or palette in a separate pr, probably when implementing the color map josh linked in figma

src/components/ButtonSystem/MainButtons.style.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export const LargeFilledButton = styled(BaseButton)`
+  ${largeButtonSizing};
+  ${props => button(props.theme.palette.buttons[ButtonType.FILL])};
+`;
+
+export const TextButton = styled(BaseButton)`
+  ${textButtonSizing};
+  ${props => button(props.theme.palette.buttons[ButtonType.TEXT])};
+`;
```
pnavarrc, line 92: nit: Maybe you can create a function that given a theme and `ButtonType` it will return the styles?

```ts
function getButtonStyles(theme: Theme, buttonType: ButtonType) {
  return theme.buttons[buttonType];
}


export const TextButton = styled(BaseButton)`
  ${textButtonSizing};
  ${props => getButtonStyle(props.theme, ButtonType.TEXT)};
`;
```

src/components/ButtonSystem/Storybook/ButtonSystem.stories.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import ButtonSystem from './ButtonSystem';
+
+export default {
+  title: 'Button System/All',
```
pnavarrc, line 5: I would probably just call it "Buttons", and add more stories for some additional cases, such as the button group and icon-only button

src/components/ButtonSystem/index.ts
```diff
@@ -0,0 +1,15 @@
...
+  TextButton,
+} from './MainButtons.style';
+
+export {
+  OutlinedButton,
+  FilledButton,
+  LargeOutlinedButton,
+  LargeFilledButton,
+  TextButton,
+};
```
pnavarrc, line 15: I think you can do 

```ts
export * from './MainButtons.style';
```

src/components/ButtonSystem/BaseButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+interface TrackingProps {
+  trackingCategory: EventCategory;
+  trackingAction?: EventAction;
+  trackingLabel: string;
+}
+
+export type LinkButtonProps = LinkProps & TrackingProps & StyledButtonProps;
+
+const BaseButton: React.FC<LinkButtonProps> = props => {
```
pnavarrc, line 36: Just writing here so we don't forget, let's make sure that the `trackingCategory` and `trackingLabel` props are mandatory for both buttons and links. We can make `trackingAction` non-mandatory since it's kind of implied by the type of element ( `NAVIGATION` for links and `CLICK` for regular buttons)

pnavarrc (commented):
Added some notes!

src/assets/theme/buttons.tsx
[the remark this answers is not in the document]
chasulin: [comment above](https://github.com/covid-projections/covid-projections/pull/3200#discussion_r603687457) addresses this !

src/components/ButtonSystem/MainButtons.style.tsx
chasulin: per our convo, nixing this

src/components/ButtonSystem/BaseButton.tsx
chasulin: all set

src/components/ButtonSystem/BaseButton.style.ts
```diff
@@ -0,0 +1,12 @@
+import MuiButton from '@material-ui/core/Button';
+import styled from 'styled-components';
+
+export const BaseMuiButton = styled(MuiButton).attrs(props => ({
+  disableRipple: true,
+  disableFocusRipple: true,
+}))`
+  &.Mui-focusVisible {
+    outline: blue auto 1px;
+    outline: -webkit-focus-ring-color auto 1px;
```
chasulin, line 10: lets discuss focus state in today's meeting with josh. as you pointed out, the blue focus outline isnt visible on the filled button

src/components/ButtonSystem/MainButtons.style.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  &:hover {
+    text-decoration: underline;
+  }
+`;
+
+/**
+ * Main button variations:
+ */
+export const OutlinedButton = styled(BaseButton)`
```
ghop02, line 69: This is a small drive by comment, but I wonder if it would be worth having these different variants accessible by either a prop (similar to material ui variant field `variant="filled"`, `variant="outlined"`, etc.) or something like: 
```
export {
  Outlined: OutlinedButton,
  Filled: FilledButton
}
```
so you can call something similar to
```
Buttons.Outlined
```
would be useful.

I think having all the different variants makes the API feel more streamlined but may just be a personal preference!

chasulin: in terms of using a prop-- i thought about this, but i landed on the current setup so that we dont need to import 2 things (a button + an enum for the prop) when implementing, and we could just import 1 thing (an intuitively-named button component)

src/components/ButtonSystem/BaseButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+};
+
+const Button: React.FC<StyledButtonProps> = props => {
+  const hasOnClick = props.onClick;
+  assert(hasOnClick, 'Button needs an onClick');
+
+  const { onClick, ...otherProps } = props;
+
+  return <BaseMuiButton {...otherProps} onClick={onClick} />;
+};
```
pnavarrc, line 67: This can be probably simplified a bit

```tsx
const Button: React.FC<StyledButtonProps> = ({ onClick, ...otherProps }) => {
  assert(onClick, 'Button needs an onClick');
  return <BaseMuiButton {...otherProps} onClick={onClick} />;
}
```

src/assets/theme/buttons.tsx
```diff
@@ -0,0 +1,76 @@
...
+    border: 'transparent',
+    text: newBlue,
+    backgroundHover: 'transparent',
+    borderHover: 'transparent',
+    textHover: newBlueDark,
+    icon: COLOR_MAP.GRAY_BODY_COPY,
+    fontFamily: fonts.regularBook,
+    disabledBackground: COLOR_MAP.GREY_1, // ?
+    disabledText: COLOR_MAP.GRAY_BODY_COPY, // ?
+    disabledIcon: COLOR_MAP.GREY_3, // ?
```
pnavarrc, line 72: Let's confirm with Josh, review accessibility and remove the `// ?` post-launch

src/components/ButtonSystem/MainButtons.style.tsx
ghop02: Ooo the double import thing makes sense.  i think material ui handles that a bit differently by exporting a type, something like `export type FormControlClassKey = 'root' | 'marginNormal' | 'marginDense' | 'fullWidth';`

Then the prop is typed as that type, but can define it easily as a string.  (also i went down a rabbit hole looking into this because i've kinda always been curious how they do it, but they just [make a map of variant -> component](https://github.com/mui-org/material-ui/blob/17fe2c49d1cd081036c549ada51f49e0061b2923/packages/material-ui/src/TextField/TextField.js#L129) and return that - kinda cool!)

src/components/ButtonSystem/BaseButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+type BaseButtonProps = TrackingProps & (LinkButtonProps | StyledButtonProps);
+
+const BaseButton: React.FC<BaseButtonProps> = props => {
+  const isLink = props.href || props.to;
+  const isButton = props.onClick && !isLink;
+  assert(
+    isLink || isButton,
+    "Button needs either a redirect ('href' or 'to' if triggering navigation) or an onClick (if triggering a non-redirect action)",
+  );
```
pnavarrc, line 83: I don't think we need this assertion, we already already have assertions for both the `LinkButton` and the `Button` components ... or alternatively, remove those assertions and keep the assertion here.

pnavarrc (approved):
Looks good!

src/assets/theme/buttons.tsx
chasulin: ah yep. will add more explicit todo

src/components/ButtonSystem/MainButtons.style.tsx
chasulin: ooo handy + good to know!! gonna keep this pr as is

PR #3202, opened 2021-03-29
Try finding email textbox element and focusing on it

faihegberg:
This PR simply experiments with focusing on the email textbox once the user clicks the "Receive alerts" button.


chasulin (approved):
nice!!

PR #3211, opened 2021-03-30
Update Banner

BrettBoval:
See https://trello.com/c/P5xQ8cLZ/1190-update-banner-to-highlight-states-with-rising-case-numbers

- Add 4 more states that have increasing hospitalizations on the recommendation of Igor.
- Add text into the banner highlighting that cases are also rising.

### DIFF
--- a/src/components/Banner/SpringSurgeBanner.tsx
+++ b/src/components/Banner/SpringSurgeBanner.tsx
@@ -25,7 +25,8 @@ const BannerInner: React.FC = () => {
   return (
     <InnerContainer>
       <Body>
-        Hospitalizations are rising in {SeveralStatesLink}. Help prevent another
+        Daily new cases are flat or rising in two-thirds of states and
+        hospitalizations are rising in {SeveralStatesLink}. Help prevent another
         COVID wave by continuing to take precautions and getting vaccinated when
         you are eligible.
       </Body>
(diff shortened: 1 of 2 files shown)


src/components/Banner/SpringSurgeBanner.tsx
```diff
@@ -25,7 +25,8 @@ const BannerInner: React.FC = () => {
   return (
     <InnerContainer>
       <Body>
-        Hospitalizations are rising in {SeveralStatesLink}. Help prevent another
+        Daily new cases are flat or rising in two-thirds of states and
```
mikelehen, line 28: FWIW- 32 states have r(t) of 1.01 or higher, so you could probably just say "are rising". I'm fine either way though.

mikelehen (approved):
This is great, thank you!

BrettBoval:
I decided to keep the states to highlight in the explore hospitalization hard-coded because I want a manual curation step. Same with keeping the flat or rising vague to two-thirds.

We will have to update this, but I'm not worried because this is high visibility for the team at the moment.

src/components/Banner/SpringSurgeBanner.tsx
BrettBoval: 1.01 is within the noise for increasing in my mind, especially given the vagaries in reporting recently. You could make the case that 0.98 should be included in "essentially flat" too, which is why I kept it at two-thirds instead of 32 states. It also makes it more evergreen without losing much messaging.

PR #3222, opened 2021-03-30
Switch from using moment to time utils in projection.ts

faihegberg:
This PR migrates moment usages in projection.ts to time utils.

### DIFF
--- a/src/common/utils/time-utils.test.ts
+++ b/src/common/utils/time-utils.test.ts
@@ -75,14 +75,14 @@ describe('utc time formatting', () => {
 });
 
 describe('date string parsing', () => {
-  test('ISO format to date object', () => {
-    expect(parseDateString('2020-03-01T00:00:00.000')).toEqual(
-      new Date(2020, 1, 29, 16),
-    );
-  });
   test('ISO format in UTC to date object', () => {
     expect(parseDateString('2020-03-01T00:00:00.000Z')).toEqual(
       new Date(2020, 1, 29, 16),
+    );
+  });
+  test('ISO format to UTC date object with time', () => {
+    expect(parseDateString('2020-12-31T23:59:59.999Z')).toEqual(
+      new Date(2020, 11, 31, 15, 59, 59, 999),
     );
   });
   test('Date string to date object', () => {
--- a/src/common/utils/time-utils.test.ts
+++ b/src/common/utils/time-utils.test.ts
@@ -77,11 +77,16 @@ describe('utc time formatting', () => {
 describe('date string parsing', () => {
   test('ISO format to date object', () => {
     expect(parseDateString('2020-03-01T00:00:00.000')).toEqual(
-      new Date(2020, 2, 1),
+      new Date(2020, 1, 29, 16),
+    );
+  });
+  test('ISO format in UTC to date object', () => {
+    expect(parseDateString('2020-03-01T00:00:00.000Z')).toEqual(
+      new Date(2020, 1, 29, 16),
     );
   });
   test('Date string to date object', () => {
-    expect(parseDateString('2020-03-01')).toEqual(new Date(2020, 2, 1));
+    expect(parseDateString('2020-03-01')).toEqual(new Date(2020, 1, 29, 16));
   });
   test('Invalid date string', () => {
     expect(() => parseDateString('Hello')).toThrow();
--- a/src/common/models/Projection.ts
+++ b/src/common/models/Projection.ts
@@ -2,7 +2,6 @@ import first from 'lodash/first';
 import last from 'lodash/last';
 import findIndex from 'lodash/findIndex';
 import findLastIndex from 'lodash/findLastIndex';
-import moment from 'moment';
 import { ActualsTimeseries } from 'api';
 import {
   ActualsTimeseriesRow,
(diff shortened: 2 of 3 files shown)


mikelehen (changes requested):
I'm not 100% sure why, but this seems to tweak the graphs slightly.  the chart is offset just slightly (a pixel or so) and we lose the "Mar" indicator in the axis.  I'm guessing we're ending up with a start date of like "2020 Mar 1 XX:00:00" instead of "2020 Mar 1 00:00:00" based on timezone or something?  Curiously, the first data point still has "Mar 1, 2020" in the tooltip so we didn't lose an entire day.

**Before:**
image

**After:**
image

In general, it might be nice if there was a class we could use to represent dates (without a time or timezone) as opposed to DateTime objects, to avoid any accidental timezone issues and so we can be really explicit about where we are dealing with absolute dates and where we're dealing with an actual date+time.  

I guess we could create our own wrapper class around a `DateTime` that ensures that it has a standardized time (`00:00` presumably) and timezone, but that might be too tedious and not sure how much it would solve.  Barring that, we probably just need to establish really clear patterns and documentation to try to avoid mistakes.  Also, I could be totally wrong about what's broken here and this isn't really the issue. 🤷

schamp:
> I'm not 100% sure why, but this seems to tweak the graphs slightly. the chart is offset just slightly (a pixel or so) and we lose the "Mar" indicator in the axis. I'm guessing we're ending up with a start date of like "2020 Mar 1 XX:00:00" instead of "2020 Mar 1 00:00:00" based on timezone or something? Curiously, the first data point still has "Mar 1, 2020" in the tooltip so we didn't lose an entire day.
> 
> **Before:**
> image
> 
> **After:**
> image
> 
> In general, it might be nice if there was a class we could use to represent dates (without a time or timezone) as opposed to DateTime objects, to avoid any accidental timezone issues and so we can be really explicit about where we are dealing with absolute dates and where we're dealing with an actual date+time.
> 
> I guess we could create our own wrapper class around a `DateTime` that ensures that it has a standardized time (`00:00` presumably) and timezone, but that might be too tedious and not sure how much it would solve. Barring that, we probably just need to establish really clear patterns and documentation to try to avoid mistakes. Also, I could be totally wrong about what's broken here and this isn't really the issue. 🤷

Wondering whether something like:
`DateTime.fromISO(date, {
        zone: 'utc',
})
`
in `time_utils.parseDateString` might take care of that...

faihegberg:
> > I'm not 100% sure why, but this seems to tweak the graphs slightly. the chart is offset just slightly (a pixel or so) and we lose the "Mar" indicator in the axis. I'm guessing we're ending up with a start date of like "2020 Mar 1 XX:00:00" instead of "2020 Mar 1 00:00:00" based on timezone or something? Curiously, the first data point still has "Mar 1, 2020" in the tooltip so we didn't lose an entire day.
> > **Before:**
> > image
> > **After:**
> > image
> > In general, it might be nice if there was a class we could use to represent dates (without a time or timezone) as opposed to DateTime objects, to avoid any accidental timezone issues and so we can be really explicit about where we are dealing with absolute dates and where we're dealing with an actual date+time.
> > I guess we could create our own wrapper class around a `DateTime` that ensures that it has a standardized time (`00:00` presumably) and timezone, but that might be too tedious and not sure how much it would solve. Barring that, we probably just need to establish really clear patterns and documentation to try to avoid mistakes. Also, I could be totally wrong about what's broken here and this isn't really the issue. 🤷
> 
> Wondering whether something like:
> `DateTime.fromISO(date, { zone: 'utc', }) `
> in `time_utils.parseDateString` might take care of that...

@mikelehen thanks for raising this and @schamp thanks for suggesting the fix. It looks like whether "Mar" is displayed or not depends on the start date of the data set. The part that appeared mysterious is that the current implementation (using `moment`) would return Mar 1st 2020 as the earliest date, and the new implementation (using `Luxon`) would also return Mar 1st 2020 as the earliest date. That said, I did notice a very slight difference of a few hours between the two. `moment.utc()` will return the given time in UTC. Luxon, it turns out, will return the given time in the local system's timezone! So Andrew's suggestion is correct and should fix the issue 😄. I tried implementing this and "Mar" is back on the chart, and I checked out the few other usages of this method too to be sure it's working as intended.

faihegberg:
Yikes! Looks like I'll need to adjust two of the unit tests too since I didn't know about Luxon instantiating the object not necessarily in UTC.

src/common/utils/time-utils.test.ts
```diff
@@ -77,11 +77,16 @@ describe('utc time formatting', () => {
 describe('date string parsing', () => {
   test('ISO format to date object', () => {
     expect(parseDateString('2020-03-01T00:00:00.000')).toEqual(
-      new Date(2020, 2, 1),
+      new Date(2020, 1, 29, 16),
```
mikelehen, line 80: I was expecting this would break for some timezones, but I see we are overriding the `yarn jest` command to use `America/Los_Angeles` timezone, so this works as expected as long as you run the tests via that command...  But it'd be nice if our tests were not timezone-dependent for the most part.  Could we do `new Date('2020-03-01T00:00:00.000Z')` here instead?

mikelehen (approved):

src/common/utils/time-utils.test.ts
faihegberg: Yea, @pnavarrc and I decided to implement `America/Los_Angeles` timezone when we run tests to ensure tests are consistent timezone wise. I'm not sure exactly how our tests would interpret timezone if we do not specify a timezone at all (dependent on server?). I can definitely adjust the date to include `Z` though!

mikelehen: Yeah, to be clear, I mean that instead of `new Date(2020, 1, 29, 16)` you could do `new Date('2020-03-01T00:00:00.000Z')`.  Since we're explicitly specifying the timezone (via `Z`), you don't need to know that the tests run in PDT which is +8 from UTC or whatever it is. 😁

PR #3254, opened 2021-04-03
restructure regions data

schamp:
Add a script and preprocess the regions data to reduce JS eval time for regions data.

Move data we don't need any more out of the data folder.

Still testing whether this improves perf.


schamp:
looks like it eliminates ~1MB of loaded and processed JSON data

schamp:
according to `yarn analyze` reduces main bundle size by ~300K (17%): 
old: 
image
new: 
image

and a not very scientific sample via page speed insights indicates it shaves main bundle evaluation time by a few seconds:
old:
image

new:
image


still looking at whether this unlocks a simple way to defer loading the whole thing as part of the main bundle.

mikelehen:
Nice!  Will take a look later today or tomorrow but sounds like this may turn out to be a nice win! 🎉

src/common/regions/preprocessed_regions_data.ts
```diff
@@ -0,0 +1,33 @@
...
+import values from 'lodash/values';
+import mapValues from 'lodash/mapValues';
+
+import statesByFipsJson from 'common/data/states_by_fips.json';
+import countiesByFipsJson from 'common/data/counties_by_fips.json';
+import metroAreasByFipsJson from 'common/data/metro_areas_by_fips.json';
+
+import { State, County, MetroArea } from './types';
+
+export const statesByFips = mapValues(statesByFipsJson, (v: any) => {
```
mikelehen, line 11: It'd be better to just drop the `: any` type and do `v => {` so TypeScript infers the type rather than specifying it as `any` which means TypeScript won't do any type-checking on it.

src/common/locations.ts
```diff
@@ -1,17 +1,13 @@
...
+  full_fips_code?: string;
+  population: number;
+  state_code: string;
+  state: string;
 }
 
-const ADJACENT_COUNTIES: AdjacencyData = countyAdjacencyMsa.counties;
-
 /** Aggregations supported by the Explore chart. */
 export const AGGREGATED_LOCATIONS: Location[] = [
```
mikelehen, line 12: FWIW- We should be able to delete this in favor of https://github.com/covid-projections/covid-projections/blob/398c97144382962d4dc97827d65dbc118b46cb8b/src/common/regions/preprocessed_regions_data.ts#L33 and this whole file could go away!  (but feel free to punt. you've done enough great cleanup in this PR. 😁 )

src/common/regions/utils.ts
```diff
@@ -6,13 +6,65 @@ import values from 'lodash/values';
...
+// some of the existing state based logic over.  Ideally we will be able
+// to remove these at some point, but they are helpful in the meantime.
+
+export function getStateName(region: Region): string | null {
+  if (region.regionType === RegionType.COUNTY) {
+    return (region as County).state.fullName;
+  }
+  if (region.regionType === RegionType.STATE) {
+    return (region as State).fullName;
+  }
```
mikelehen, line 25: FYI- I guess this is existing code, so feel free to leave it as-is, but you could clean it up (take advantage of TypeScript type narrowing) with:
```
if (region instanceof County) {
  return region.state.fullName;
}
if (region instanceof State) {
  return region.fullName;
}
```

and similar below...

mikelehen (changes requested):
Wow!  This is great.  There's a lot of nice cleanup here and this change wasn't as disruptive as I was worried it might be!

I left a bunch of feedback, mostly minor.  I'd like to take another pass before giving it a full 👍  but in general I'm a fan.

A couple higher-level things:
1. We should decide if we think "minifying" JSON names is worthwhile.  I pinged @pnavarrc in case he wants to weigh in.
2. There's a lot of `any` use that I'd like to remove.  Not sure if that was just for expediency or you ran into issues, but feel free to ping me if you want help defining types.

src/common/regions/types.ts
```diff
@@ -22,6 +22,15 @@ const fipsToPrincipalCityRenames: FipsToPrincipalCityName = {
   [NY_METRO_FIPS]: 'New York City',
 };
 
+// PODT Region object to aid serialization/deserialization w/ JSON
+export interface RegionObject {
+  n: string;
+  u: string;
+  f: FipsCode;
+  p: number;
+  r: RegionType;
```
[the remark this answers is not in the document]
schamp: for interface types I prefer to be explicit, because for humans it can be harder to infer types than it is for typescript : )

src/common/regions/types.ts
```diff
@@ -22,6 +22,15 @@ const fipsToPrincipalCityRenames: FipsToPrincipalCityName = {
   [NY_METRO_FIPS]: 'New York City',
 };
 
+// PODT Region object to aid serialization/deserialization w/ JSON
+export interface RegionObject {
+  n: string;
```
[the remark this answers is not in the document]
schamp: I thought about gzip, and my reasoning went something along these lines:
1. probably wouldn't hurt
2. we control both the input and output formats, and they're defined right next to each other, so verbosity is less important (and I don't expect these objects to be used apart from their "hydrated" classes).
3. I thought making this file smaller would both make it less of a red herring when targeting for cleanup by looking at file size.

But yeah if either of you feel strongly about it I can put them back, no prob.

scripts/prepare_regions_data/regions_data_do_not_use.ts
```diff
@@ -1,74 +1,24 @@
+/*
+Should only be used by script for pre-processing regions data
```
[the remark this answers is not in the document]
schamp: good catch, this name was from when the file still lived in the source tree, before I realized I could move it here to `scripts`.

src/common/regions/types.ts
```diff
@@ -87,6 +102,21 @@ export class State extends Region {
...
+      n: this.name,
+      u: this.urlSegment,
+      f: this.fipsCode,
+      p: this.population,
+      r: this.regionType,
+      s: this.stateCode,
+    };
+  }
+
+  public static fromObject(obj: StateObject) {
```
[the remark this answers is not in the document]
schamp: TIL stringify would take advantage of this.  But because `County` and `MetroArea` still require the lookups to find the referenced regions during parsing, I want to leave it like this to make it consistent between those three types.

In the other diff (https://github.com/covid-projections/covid-projections/pull/3261) I found an approach that I like a little better, which brings the `statesByFips` lookup into the main bundle separately from `countiesByFips` and `metroAreasByFips`, because it's so much smaller and useful as a reference.  To make that work (I think) we need to separate the class definitions into separate files to avoid a circular dependency (e.g, `MetroArea` -> `statesByFips` -> `State`).  Or I suppose we could just bring `statesByFips` definition into that types file.  

There are some tradeoffs with that (loading the lookup data in a `types` file, splitting up class definitions into separate files when having them all together is nice).  There are also some dependencies still I think, getting direct County references out of MetroArea and using CountyFips instead (and doing the lookup where used).

So I see a couple of options:
1. fix the other things in the diff as-is, then make any other changes necessary to eliminate the lookup requirement, then do it by lookup so `toObject` in `County` and `MetroArea` doesn't require a lookup to be passed in in a second PR
2. bring the minimum number of changes necessary to unlock this capability into this PR.

After thinking about the choices, I think at this moment I prefer something like this:
a. bring the "statesByFips" loading and hydration right into `types.ts` so we don't have to split `types.ts` to avoid the circular dependency.  any place that needs `statesByFips` is probably going to need those other types anyway.
b. roll those changes into this PR and land it all as a unit

I'll be working on this stuff tonight, maybe I'll make an example commit with those changes so you can see what I mean, it would at least help me identify any remaining dependencies of that change.

src/common/locations.ts
schamp: good catch, I'll look into it.

schamp:
Thanks for the review!  No issues with your comments, I can do the types, I was just being lazy to see if the approach was workable before putting more energy into it (in my day job I have a linter that yells at me about `any` and I've been trained to let that catch things that slip through the cracks).

src/common/regions/preprocessed_regions_data.ts
schamp: it's a little tricky with the types inferred from the loaded JSON files, but I'll figure something out.

schamp: I'm getting a lot of: 
```
$ tsc
src/common/regions/preprocessed_regions_data.ts(18,14): error TS2322: Type 'Dictionary<boolean>' is not assignable to type '{ [fips: string]: State; }'.
  Index signatures are incompatible.
    Type 'boolean' is not assignable to type 'State'.
src/common/regions/preprocessed_regions_data.ts(25,14): error TS2322: Type 'Dictionary<boolean>' is not assignable to type '{ [fips: string]: County; }'.
  Index signatures are incompatible.
    Type 'boolean' is not assignable to type 'County'.
src/common/regions/preprocessed_regions_data.ts(32,14): error TS2322: Type 'Dictionary<boolean>' is not assignable to type '{ [fips: string]: MetroArea; }'.
  Index signatures are incompatible.
    Type 'boolean' is not assignable to type 'MetroArea'.

```

with an attempt that currently looks like this, if you can help me sort it out : (
```export const statesByFips: {[fips: string]: State} = mapValues(
  statesByFipsJson, State.fromObject);
  /*(v: StateObject): State => {
    return State.fromObject(v);
  },
);
*/
export const countiesByFips: {[fips: string]: County} = mapValues(
  countiesByFipsJson,
  (v: CountyObject): County => {
    return County.fromObject(v, statesByFips);
  },
);

export const metroAreasByFips: {[fips: string]: MetroArea} = mapValues(
  metroAreasByFipsJson,
  (v: MetroAreaObject): MetroArea => {
    return MetroArea.fromObject(v, statesByFips, countiesByFips);
  },
);
```

schamp: I think this is probably why I went with 'any' the first time around :lolsob:

src/common/regions/utils.ts
schamp: I'll leave it for now.

scripts/prepare_regions_data/prepare-regions-data.ts
```diff
@@ -0,0 +1,52 @@
...
+  // convert the constructed object map to one suitable for serializing
+  const toObjMap = (m: { [index: string]: any }): { [index: string]: any } => {
+    const result: any = {};
+    Object.keys(m).forEach((k: string) => {
+      result[k] = m[k].toObject();
+    });
+    return result;
+  };
+
+  await Object.keys(files).forEach(async (file: any) => {
```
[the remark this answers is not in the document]
schamp: nice improvement, thanks : )

src/common/regions/preprocessed_regions_data.ts
schamp: oh I got it figured out, I needed to cast the imported JSON in to the expected type explicitly, then typescript could do its thing.

src/common/locations.ts
schamp: yep, this was pretty straightforward.  it turns out, since customareas are 'state' regions, we can just use their fullname directly and don't need to even bring in customAreasByFips in the place it's used.

mikelehen:
As meta feedback, force pushing is fine up until code reviews start, but now I'm unable to see your new changes (in the github UI) without re-reviewing the entire PR which is 😢 .  Github also has trouble tracking comments when the commit they were made against disappears.  So it's generally best to just do normal commits after reviews have started.

schamp:
Oh no, sorry.  I didn't realize github did this : (

src/common/regions/types.ts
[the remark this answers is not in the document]
mikelehen: Sorry, what I meant to say is that we don't actually need to write the RegionType field to the JSON file since all of the regions in a given file will be of the same type (state, county, metro).  In fact, your code already doesn't use this field so I think it's just a simple matter of removing it from `RegionObject` interface and the `toObject()` methods.  Not sure if it'll save anything post-gzip, but it is probably cleaner not to write unnecessary data regardless.  😁

src/common/regions/types.ts
```diff
@@ -22,6 +22,15 @@ const fipsToPrincipalCityRenames: FipsToPrincipalCityName = {
   [NY_METRO_FIPS]: 'New York City',
 };
 
+// PODT Region object to aid serialization/deserialization w/ JSON
+export interface RegionObject {
+  n: string;
+  u: string;
+  f: FipsCode;
+  p: number;
+  t: RegionType;
```
mikelehen, line 31: Replied to the previous comment thread, but adding it again here just in case that gets lost.  I think we don't need to include `t` on the `RegionType` or write it in `toObject()`. We're not using it anywhere.

src/common/regions/types.ts
```diff
@@ -22,6 +22,15 @@ const fipsToPrincipalCityRenames: FipsToPrincipalCityName = {
   [NY_METRO_FIPS]: 'New York City',
 };
 
+// PODT Region object to aid serialization/deserialization w/ JSON
```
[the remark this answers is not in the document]
mikelehen: How about drop PODT and just say "JSON representation of a Region object"?

side note: [Fun memo from Elon Musk about acronyms.](https://gist.github.com/klaaspieter/12cd68f54bb71a3940eae5cdd4ea1764)

tsconfig.json
```diff
@@ -16,5 +16,5 @@
     "noEmit": true,
     "jsx": "react"
   },
-  "include": ["src"]
+  "include": ["src", "scripts/prepare_regions_data"]
```
mikelehen, line 19: Heh.  Guessing this was temporary for development, until your other PR lands?

mikelehen (approved):
Looks great.  Thanks for getting this done and cleaning up all the code along the way.

src/common/regions/types.ts
schamp: True.  I had thought of this, but was leaving it in, because if we end up breaking out individual files for location pages we're going to need to be able to inspect the object itself to figure out what kind of region it is.  But, perhaps IAGNI, and also, we can write region type to those files specifically and just inspect it to hydrate the right one, so we don't need it in the files with everything and so I'll remove it and regenerate the data,

tsconfig.json
schamp: I think my editor did that actually, but yeah, I'll remove it.

schamp:
I spent the first 10 years of my career in the defense industry, which changes a person.  I think our record was either the 6-letter helmet display acronym, or the 5-letter 2-layer weapon sensor acronym.  But TBH I thought PODT was like JSON, well known industry wide 🤷

PR #3257, opened 2021-04-04
Make EmailAlertsFooter use Fips for defaultRegions

schamp:
This is part of a larger plan to eliminate nested regions in region DB, to allow simpler partial loading of region data.

The email alerts footer component already had a dependency on the regions database, so we can look things up by FIPS ode and just pass FIPS code in, instead of the whole region.


schamp:
Looks like it depends more strictly on https://github.com/covid-projections/covid-projections/pull/3254 , so will either clean up or wait until that lands.

schamp:
Turns out this is not useful yet, so closing, may try again if it looks worth it.

PR #3265, opened 2021-04-05
Vaccination component UI update

chasulin:
(Storybook only) Adds alternate vaccination eligibility component for states with all adults eligible


pnavarrc (approved):
Looks good!

PR #3282, opened 2021-04-06
Recommendations section updates

chasulin:
- Removes sources and methodology modal
- Updates context text to reference CDC only
- Removes "share with" instructions


pnavarrc (approved):
Looks good!

PR #3283, opened 2021-04-06
Mega menu to all users

chasulin:
- Ships mega menu to all users, with a new experiment (API link first in list vs. last in list)
- Deletes code from old menu variation
- Reorgs files for mega menu+footer+menu content


src/cms-content/footer/index.ts
```diff
@@ -1,4 +1,5 @@
 import { Markdown } from '../utils';
+import { shuffle } from 'lodash';
```
pnavarrc, line 2: It's better to import only the function that we need performance-wise (we can install a plugin to do that automatically, but that's a bit of a hassle)

```ts
import shuffle from 'lodash/shuffle';
```

src/components/AppBar/DonateButton/index.ts
```diff
@@ -0,0 +1,3 @@
+import { DonateButton, DonateButtonHeart } from './DonateButton';
+
+export { DonateButton, DonateButtonHeart };
```
pnavarrc, line 3: we can simplify this by doing

```ts
export * from './DonateButton';
```

src/components/MenuContent/MenuContent.tsx
```diff
@@ -0,0 +1,85 @@
...
+}> = ({ featuredSections, onClick }) => {
+  const apiLink = featuredSections.filter(
+    section => section.iconId === SectionId.API,
+  );
+  const dailyDownloadLink = featuredSections.filter(
+    section => section.iconId === SectionId.DAILY_DOWNLOAD,
+  );
+  const alertsLink = featuredSections.filter(
+    section => section.iconId === SectionId.ALERTS,
+  );
```
pnavarrc, line 28: Maybe using `find` would be better, that way you can just add the items to the array without the `...`

```ts
const apiLink = featuredSections.find(
  ({ iconId }) => iconId === SectionId.API,
);

const orderedSectionsA = [apiLink, alertsLink, dailyDownloadLink];
```

creating a map by Id might be more compact too

```ts
const sectionById = keyBy(featuredSections, section => section.iconId);

const orderedSectionsA = [
  sectionsById[SectionId.API],
  // ...
]
```

pnavarrc (approved):
Added some notes, looks good!

src/components/MenuContent/MenuContent.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const orderedSectionsA = [...apiLink, ...alertsLink, ...dailyDownloadLink];
+  const orderedSectionsB = [...alertsLink, ...dailyDownloadLink, ...apiLink];
+
+  return (
```
pnavarrc, line 33: These 2 arrays (and the code above) can be moved outside the render method, since they don't need to recalculated each time the component renders

chasulin: as in into the cms index? they use the `featuredSections` prop

chasulin: its a pretty small calculation- im tempted to just keep as is and 🚢

chasulin:
[Ship mega menu to 100% of users + new experiment that moves the API link to the top of the list for 50%](https://trello.com/c/1iE4kHH7/1237-ship-mega-menu-to-100-of-users-new-experiment-that-moves-the-api-link-to-the-top-of-the-list-for-50)

PR #3287, opened 2021-04-07
Enforce typescript on scripts folder

schamp:
Saw this on trello and it looked simple, and indeed it was.


schamp:
huh, `yarn tsc` runs fine locally for me, but *crashes* on CI?  any ideas here?

tsconfig.json
```diff
@@ -16,5 +16,5 @@
     "noEmit": true,
     "jsx": "react"
   },
-  "include": ["src"]
```
mikelehen, line 19: FWIW- I think this means that website code under src/ would now be able to import code from scripts/ which we probably don't really want. 

Can we just tweak CI so that it runs `tsc` inside the `scripts` directory (which has its own [tsconfig.json](https://github.com/covid-projections/covid-projections/blob/develop/scripts/tsconfig.json) already)?

scripts/vaccination_alerts/parse-vaccination-phases.ts
```diff
@@ -63,6 +63,7 @@ const buildRegionGroup = (
     locationName: state,
     fips: regions.findByFullName(state.trimRight())!.fipsCode,
     eligibilityInfoUrl: firstRow.eligibilityUrl,
+    vaccinationSignupUrl: '', // this does not appear to be available in the spreadsheet
```
mikelehen, line 66: FWIW- Brett says we can delete this script, so feel free to do that if it's easier.

mikelehen (changes requested):

scripts/vaccination_alerts/parse-vaccination-phases.ts
schamp: nice.

tsconfig.json
schamp: Hm, you're right, what if we updated `"tsc"` in `packages.json` to run both `--project tsconfig.json` and `--project scripts/tsconfig.json`?

Or do we want to avoid running it on `scripts` outside of CI?

mikelehen: Ah, that sounds perfect.

I suspect nobody runs `yarn tsc` manually (I didn't even realize it existed until you made this PR) and so in practice, updating it there will only affect CI.  But in general I don't think it hurts to always run against both.  I'd only hesitate if we ended up making `yarn start` slower or something, but I assume they're unrelated?

schamp: ha ha, I run it all the time, it's my background sanity check 

I don't believe `yarn start` is related.  I'll go ahead and proceed with this change. 

... so I did, turns out the tsconfig in scripts was broken, so I fixed that.

This should work for everybody, and doesn't add much time.

schamp:
Looks like the typechecker is still failing, with an OOM.  Can we allocate more resources to that job, or is there some other fix?

mikelehen (approved):
Bleh.  Code looks good but not sure why CI is crashing...  I think github runners have a fixed-size of 7GB (https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) and we probably don't want to have to spin up our own hardware.  I'm surprised it's failing.  Our codebase isn't that big or anything.

Thoughts:
1. Google around for tsc out-of-memory errors and see if anything turns up (might be a known issue or maybe there's a node.js option to give node.js heap more memory or something)
2. Could try splitting it into two separate tsc commands, e.g. `tsc --project tsconfig.json && tsc --project scripts/tsconfig.json`
3. Could try running on an env similar to the GitHub one (linux, same version of node, etc.) to see if we can repro it.
4. Could try experimentally deleting scripts/ code until the problem goes away to see if there's a particular script that's triggering it or something.

3/4 are definitely not very fun so hopefully 1 or 2 would pan out. 😁

scripts/tsconfig.json
```diff
@@ -4,5 +4,6 @@
     "module": "commonjs",
     "noImplicitAny": true,
     "strictNullChecks": true
-  }
+  },
+  "include": ["../src", "."]
```
mikelehen, line 8: As an aside, it's curious that you had to add this.

This tsconfig.json is used whenever we run one of our scripts with e.g.:
```
ts-node-script -r tsconfig-paths/register scripts/vaccination_alerts/parse-vaccination-phases.ts
```

and it works in that context. 🤷

schamp: yeah, that is weird.  I can't explain this.  though it stands to reason that a type-script script-running engine would compile typescript according to the tsconfig.json it finds.

PR #3295, opened 2021-04-07
Close mega menu when clicking logo, donate button, or back arrow

chasulin:


src/components/AppBar/DonateButton/DonateButton.tsx
```diff
@@ -21,11 +21,14 @@ export const DonateButton = () => (
   </DonateButtonWrapper>
 );
 
-export const DonateButtonHeart = () => (
+export const DonateButtonHeart: React.FC<{ closeMenu: () => void }> = ({
+  closeMenu,
+}) => (
```
pnavarrc, line 26: It's probably better to not add a new prop just use the `onClick` handler like you did for the other top nav links, but we will need to set the type of the `DonateButtonHeart` to be `LinkButton`

```tsx
// StyledDonateButtonHeart is a LinkButton
import { LinkButtonProps } from 'components/LinkButton';

export const DonateButtonHeart: React.FC<LinkButtonProps> = ({ onClick }) => (
  <StyledDonateButtonHeart
    to="/donate"
    {/* ... */}
    onClick={onClick}
  >
    Donate
  </StyledDonateButtonHeart>
);
```

Then we can pass `onClick`, that way we keep the separation of concerns between the `DonateButton` and the mega menu intact

```tsx
<DonateButtonHeart onClick={closeMenu} />
```

pnavarrc (approved):
Looks good, other than the new prop for the `DonateButtonHeart`, added a suggestion for it

src/components/AppBar/DonateButton/DonateButton.tsx
```diff
@@ -21,11 +22,12 @@ export const DonateButton = () => (
   </DonateButtonWrapper>
 );
 
-export const DonateButtonHeart = () => (
+export const DonateButtonHeart: React.FC<LinkProps> = ({ onClick }) => (
```
chasulin, line 25: tracking props are declared in `DonateButton.tsx` (and passed to both donate button variations), so i took `LinkProps` instead of `LinkButtonProps`

PR #3297, opened 2021-04-08
Update homepage banner.

mikelehen:
image

### DIFF
--- a/src/components/Banner/SpringSurgeBanner.tsx
+++ b/src/components/Banner/SpringSurgeBanner.tsx
@@ -18,21 +18,21 @@ const BannerInner: React.FC = () => {
   const { totalVaccinationsInitiated } = usaAggregation;
   const millionVaccinations = Math.round(totalVaccinationsInitiated / 1000000);
 
-  const ManyStatesLink = (
+  const HospitalizationsAreRisingTooLink = (
     <HashLink
       to="#explore-hospitalizations"
       onClick={trackClick}
       scroll={(element: HTMLElement) => scrollWithOffset(element, -80)}
     >
-      many states
+      hospitalizations are rising too
     </HashLink>
   );
 
   return (
     <InnerContainer>
       <Body>
-        Daily new cases are rising in half of states and hospitalizations are
-        rising in {ManyStatesLink}. Help prevent another COVID wave by
+        Daily new cases are rising in about half of the states and{' '}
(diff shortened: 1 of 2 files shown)


src/components/Banner/SpringSurgeBanner.tsx
```diff
@@ -13,22 +14,28 @@ const trackClick = () => {
...
+  const millionVaccinations = Math.round(totalVaccinationsInitiated / 1000000);
+
+  const ManyStatesLink = (
     <HashLink
       to="#explore-hospitalizations"
       onClick={trackClick}
       scroll={(element: HTMLElement) => scrollWithOffset(element, -80)}
     >
-      several states
+      many states
```
mikelehen, line 27: I can revert if you're not a fan, but about half of the states that have rising cases also have rising hospitalizations so "many" feels appropriate.

src/screens/HomePage/HomePage.tsx
```diff
@@ -67,15 +67,15 @@ export default function HomePage() {
...
     'MN',
     'NJ',
-    'WV',
-    'CT',
-    'IL',
-    'MA',
-  ]; // Updated on 29 March 2021
+    'PA',
+    'PR',
+  ]; // Updated on 7 April 2021 -- this is an illustrative list but not exhaustive
```
mikelehen, line 78: We could add several more states, but the graph just gets noisy, hurts performance (we're downloading the API endpoint for each state), and also some of the "low but rising" states just don't really pop when on a normalized graph with the other ones, so I don't think we should try to make this be an exhaustive list...

mikelehen:
[Update Banner to "Daily new cases are rising in half of the states, ..."](https://trello.com/c/KnVsTuI6/1238-update-banner-to-daily-new-cases-are-rising-in-half-of-the-states)

src/components/Banner/SpringSurgeBanner.tsx
BrettBoval: I like it. "Many" does feel appropriate.

src/screens/HomePage/HomePage.tsx
BrettBoval: Agreed that we can treat this as a representative, not exhaustive, list.

BrettBoval (approved):

PR #3299, opened 2021-04-08
Enable attribution tracking in amplitude.

mikelehen:
Note that the utm params get set as user properties on the session and so only happens at the beginning of a session.  To test it I had to clear some stuff in localStorage / cookies to force a new session (which also accidentally logged me out of gmail, so be careful I guess).

image


pnavarrc (approved):
Tested on a clean profile with the Amplitude extension installed, work as expected!

pnavarrc:
[Amplitude: Configure web attribution](https://trello.com/c/aXvYXjzt/1154-amplitude-configure-web-attribution)

PR #3300, opened 2021-04-08
Add utm source data to risk alert emails.

mikelehen:
I'm not bothering doing the vaccination alert emails because they aren't going to be around much longer (most states already have open eligibility and all will by April 19).


pnavarrc (approved):
Looks good!

PR #3304, opened 2021-04-08
clean up region classes some

schamp:
This embeds the `statesByFips` lookup so that we can construct counties and metroareas that reference states directly from the FIPS that we store, making toJSON and fromJSON possible with no additional parameters.  To make this work, `statesByFips` must be constructed in the `regions/types.ts` package, to void a circular dependency.  I was worried this would smell too bad to actually use, but after trying it, it wasn't as bad as I thought.

To unlock this, metroAreas must not refer directly to county objects, so this diff stores countyFips directly instead.  In all cases but one this actually simplified things, since we were mapping across counties to get the FIPS code.  In the one case it didn't, we already imported the regions DB so a lookup was straightforward.

While doing this I thought of another optimization, which might be a bridge too far or not worth it, but I may explore it a bit if it turns out to be easy -- the `urlsSegments` for counties and states derive directly from their names.  So if I can construct a simple function that matches the one used to generate them, we can regenerate them when constructing the region rather than storing the data.  I guess step one in that exploration would be exporting w/o urlsegments to see how much space would actually base saved.

Anyway, this PR isn't essential, but it illustrates what I was trying to describe to @mikelehen and figured I'd put it up.  If we decide to merge it, then great, if not, NBD.

### DIFF
--- a/src/common/regions/types.ts
+++ b/src/common/regions/types.ts
@@ -1,4 +1,7 @@
 import urlJoin from 'url-join';
+import mapValues from 'lodash/mapValues';
+
+import statesByFipsJson from 'common/data/states_by_fips.json';
 
 export type FipsCode = string;
 export type ZipCode = string;
@@ -21,6 +24,11 @@ const fipsToPrincipalCityRenames: FipsToPrincipalCityName = {
   [DC_METRO_FIPS]: 'Washington DC',
   [NY_METRO_FIPS]: 'New York City',
 };
+
+export const statesByFips = mapValues(
+  statesByFipsJson as { [fips: string]: StateObject },
+  v => State.fromJSON(v),
+);
 
 // JSON-serializable representation of a Region object
 export interface RegionObject {
@@ -61,7 +69,7 @@ export abstract class Region {
     return `${this.name} (fips=${this.fipsCode})`;
   }
 
-  abstract toObject(): RegionObject;
+  abstract toJSON(): RegionObject;
 }
 
 export interface StateObject extends RegionObject {
@@ -102,7 +110,7 @@ export class State extends Region {
     );
   }
 
-  public toObject(): StateObject {
+  public toJSON(): StateObject {
     return {
       n: this.name,
       u: this.urlSegment,
@@ -112,9 +120,8 @@ export class State extends Region {
     };
   }
 
-  public static fromObject(obj: StateObject): State {
-    const s = new State(obj.n, obj.u, obj.f, obj.p, obj.s);
-    return s;
+  public static fromJSON(obj: StateObject): State {
+    return new State(obj.n, obj.u, obj.f, obj.p, obj.s);
   }
 }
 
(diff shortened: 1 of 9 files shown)


schamp:
there's something not quite right about the build which I'll look into, but this PR is the general idea.

src/common/regions/types.ts
```diff
@@ -206,15 +218,18 @@ export interface MetroAreaObject extends RegionObject {
...
+
   constructor(
     name: string,
     urlSegment: string,
     fipsCode: FipsCode,
     population: number,
-    public counties: County[],
-    public states: State[],
+    statesFips: FipsCode[],
+    public readonly countiesFips: FipsCode[],
```
mikelehen, line 229: Can you add a comment explaining that we intentionally only store FIPS codes instead of `County` objects so that we can construct metro objects without loading all the county objects (or whatever makes sense to you to explain this)?

mikelehen (approved):
TBH I'm on the fence on this one.

On the one hand:
1) If I squint I can see how this might make it easier to remove counties / metros from the main bundle down the road (though I don't know what the exact path will be so we may never benefit from that).
2) The code changes aren't too bad.

On the other hand, it feels like a mild wart that `MetroArea` now only contains FIPS codes of counties instead of the richer `County` objects (which turns out not to be much of a sacrifice in practice but seems like a less-useful API in general, regardless).

But there's also something appealing about County / Metro objects being directly serializable to/from JSON as long as we have the 50 states in the main bundle. So I'm fine with merging this if you're a fan.

mikelehen:
FWIW- I just noticed that we have `County.adjacentCountiesFips` (a list of FIPS rather than rich `County` objects) already, so I guess we aren't really consistent about using the rich types.  So I'm more fine with changing `MetroArea.counties` to `MetroArea.countiesFips` 🤷

schamp:
TBH I'm also not sure how keen I am on this approach.  I think mostly what I like is that it unlocks loading location page-specific regions as their own individual data files, which could allow better display of data above the fold on location pages while lazy loading the rest for the charts down lower.

IOW, if I have a single JSON metro area object, I can reconstitute it into the actual MetroArea Region object, looking up its state in the (small) table, and deferring loading all counties until that data is needed later.  I *think* that simplifies the path for what I described above, if (as mentioned in the other PR) we also have a separate urlSegment -> FIPS mapping we can use.

So I'm trying to explore these ideas in discrete chunks that can be individually reviewed and merged if they provide marginal benefit, while clarifying the problem space and unlocking new ways of thinking about how to approach deferred loading most of the data.  Lol that's a lot of buzzwords but I hope you get the idea.

schamp:
I will explore a few more of those things later today (I have the day off) and this weekend, and then merge if it looks promising enough.

schamp:
So this unblocks a quite a few of the dependencies on the regions database, so I am going to merge it.

PR #3308, opened 2021-04-08
compute url segments at load

schamp:
This depends on https://github.com/covid-projections/covid-projections/pull/3304

Compute urlSegments at load time, to eliminate them from the downloaded and parsed JSON.
Ungzipped size reduction:
counties_by_fips.json: 439488 -> 371328 (~15% reduction: ~70kb)
metro_areas_by_fips.json: 47620 -> 38099 (~20% reduction ~9kb)
states_by_fips.json: 3820 -> 2812 (~26% reduction ~1kb)

Not so useful for states and metro areas but may be meaningful for counties.

Mostly just seeing if this might be worth it, so no great shakes if we don't think it's compelling.


scripts/prepare_regions_data/prepare-regions-data.ts
```diff
@@ -39,6 +45,48 @@ async function main() {
   }
 }
 
+/**
+ * We optimize storing the URL segments, which can be computed from names.
```
mikelehen, line 49: This first sentence doesn't quite make sense to me without more context.  Maybe something like:
```
/** 
 * Our location URLs must be consistent over time to avoid broken URLs, but to
 * save space we don't want to persist them in the frontend code.  So this
 * function validates that the computation functions used in the frontend code
 * generate url segments that exactly match the url segments stored in the raw
 * region data files.
 */
```

mikelehen (approved):
Looks like this saves 16k after gzip which isn't huge but is probably worth the changes. Thanks!

That said, it did give me slight pause in that it kind of moves us towards a world where we can't remove the county regions from the main bundle, since we rely on it more strongly for routing...  whereas if we instead had a url_segment=>FIPS mapping that lived in the main bundle, we could theoretically figure out how to lazy load the full county region objects lazily as needed or something.

But that said, at this point the county objects are pretty small (the biggest part is the name which is comparable to if we stored url segments in the main bundle). So that's probably fine.

But that did make me realize that the `adjacentCountiesFips` data inside of `County` objects is kinda' nontrivial in size and is also only used in only one [tiny place](https://github.com/covid-projections/covid-projections/blob/8e05802d23bfbfc0005de1efedb568d482c8cc9b/src/common/utils/compare.ts#L71).  So we might want to extract that back out to its own file and lazy-load it (which I recognize is sort of a step backwards; sorry).  Opened a [trello item](https://trello.com/c/kIRBVErp/1246-perf-lazy-load-county-adjacency-info).

schamp:
Thanks for the review, I'm not entirely sure how keen I am on this one, myself.  I was doing some other explorations (will work on them this weekend) on what breaking urlSegment -> fips mappings into their own files (to support the routing) would do.

Mostly I think the implications for this change are that we'd still want it around, because I believe we use urlSegment in a bunch of places (besides the router) to build up links.

But I think I'll hold off merging it until I have more clarity about that approach, something better might present itself.  If it does look like this is worth it, I'll go ahead and merge.

schamp:
refactored to compute the urlSegment on demand, to avoid generating and storing at import time, and also because it's obvious and cleaner.

schamp:
I've rebased and fixed these conflicts, will force-push that and land it

scripts/prepare_regions_data/regions_data.ts
```diff
@@ -15,18 +15,21 @@ import {
...
-      ),
-  );
+  return map(state_dataset, stateInfo => {
+    stateFipsToUrlSegment[stateInfo.state_fips_code] = stateInfo.state_url_name;
+    return new State(
+      stateInfo.state,
+      stateInfo.state_fips_code,
+      stateInfo.population,
+      stateInfo.state_code,
+    );
```
mikelehen, line 31: FWIW- If you wanted you could drop the separate validation code and the `stateFipsToUrlSegment` maps and just do the validation here:

```
const s = new State(...)
assert(s.urlSegment === stateInfo.state_url_name, "Computed URL segment must match persisted one.");
return s;
```

and same for counties / metros...  I don't feel strongly though.

mikelehen (approved):

scripts/prepare_regions_data/regions_data.ts
schamp: ah, good idea.  but, yeah, maybe I'll save that if I have a reason to touch it later.

PR #3312, opened 2021-04-08
Remove finally() usage that breaks Firefox 59.2

mikelehen:
Alternative option would be to install a polyfill like https://www.npmjs.com/package/promise.prototype.finally which is more future-proof but so far we don't have a precedent of using polyfills.


pnavarrc (approved):
Thanks for fixing this!

PR #3313, opened 2021-04-08
More theme cleanup

chasulin:
- Removes mostly unused `ThemeColors`
- Moves mega menu theme into main theme (instead of accessing via palette)

### DIFF
--- a/src/components/Button/Button.style.ts
+++ b/src/components/Button/Button.style.ts
@@ -46,11 +46,11 @@ const Outlined = styled(Button).attrs(props => ({
 `;
 
 export const ButtonPrimary = styled(Outlined)`
-  color: ${theme.colors.green};
+  color: ${COLOR_MAP.GREEN.BASE};
 `;
 
 export const ButtonSecondary = styled(Outlined)`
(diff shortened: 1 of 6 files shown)


src/components/Button/Button.style.ts
```diff
@@ -46,11 +46,11 @@ const Outlined = styled(Button).attrs(props => ({
 `;
 
 export const ButtonPrimary = styled(Outlined)`
-  color: ${theme.colors.green};
+  color: ${COLOR_MAP.GREEN.BASE};
 `;
 
 export const ButtonSecondary = styled(Outlined)`
-  color: ${theme.colors.lightBlue};
```
pnavarrc, line 53: I think we don't use these buttons anywhere, we can probably remove them. Also, should it be `${props => props.theme.colors.green}`?

pnavarrc (approved):
Looks good!

src/components/Button/Button.style.ts
pnavarrc: ^ please ignore 😅

PR #3316, opened 2021-04-09
cleanup getUrlAndShareQuote

schamp:
The Region-derived classes do exactly what we need for this, so we can simplify this function and avoid reaching into the data members of those classes and duplicating the behavior.


mikelehen (approved):
Nice!

PR #3341, opened 2021-04-13
add getRegionsDB and useRegionsDB utilities

schamp:
These unblock a bunch other changes to defer loading the regions DB at build time.  This diff itself should not affect performance, as the functions aren't called anywhere yet.


schamp:
Once this is merged, I'll create misc. other PRs from the feature/decouple_region_db branch, for easier testing and perf regression testing.  That will let us focus our attention on the more potentially significant ones without getting them mixed up with the trivial ones.

mikelehen (approved):

PR #3349, opened 2021-04-14
precompute largest metro areas

schamp:
This is expensive to compute on each pageload and requires the regions DB, so just precompute when generating the other files and include it in the main bundle.


src/screens/HomePage/utils.ts
```diff
@@ -1,10 +1,26 @@
...
+const largestMetroFipsForExplore = largestMetroFipsForExploreJson as FipsCode[];
+
+// This is used to compute data at build time, because it governs a fixed list
+// we can precompute.  So if you need to change this value,
+// re-run `yarn prepare-regions-data` and check in the updated files.
+export const DEFAULT_EXPLORE_GEOLOCATION_COUNT = 5;
 
 /* Gets fips of 5 largest metro areas to set as homepage explore presets */
 export function getLargestMetroFipsForExplore(numRegions: number) {
+  return take(largestMetroFipsForExplore, numRegions);
```
mikelehen, line 17: Ehh...  I don't think it's worth all of this machinery to generate a list of 5 FIPS that never changes.  It creates some indirection that's going to be hard for somebody to unravel when they come across this code, so I would just remove the `numRegions` parameter (the function comment already implies that this will always be 5) and return `["35620","31080","16980","19100","26420"]` here.  And then undo the rest of this PR.  It's easy enough to write a one-off dummy script to generate different FIPS codes if we want to rework this in the future.  Sorry!

mikelehen (changes requested):

src/screens/HomePage/utils.ts
schamp: no ha ha, this makes sense, thanks for the suggestion : )

scripts/prepare_regions_data/prepare-regions-data.ts
```diff
@@ -21,7 +21,12 @@ import {
   generateStateUrlSegment,
   generateCountyUrlSegment,
   generateMetroAreaUrlSegment,
+  getRegionsDB,
 } from '../../src/common/regions';
+import {
+  computeLargestMetroFipsForExplore,
+  DEFAULT_EXPLORE_GEOLOCATION_COUNT,
+} from '../../src/screens/HomePage/utils';
```
mikelehen, line 29: I think these 3 imports are all unused now?

src/screens/HomePage/utils.ts
```diff
@@ -1,10 +1,27 @@
...
 import take from 'lodash/take';
-import regions from 'common/regions';
+import { RegionDB } from 'common/regions';
 
-/* Gets fips of 5 largest metro areas to set as homepage explore presets */
-export function getLargestMetroFipsForExplore(numRegions: number) {
+// This is used to compute data at build time, because it governs a fixed list
+// we can precompute.  So if you need to change this value,
+// re-run `yarn prepare-regions-data` and check in the updated files.
+export const DEFAULT_EXPLORE_GEOLOCATION_COUNT = 5;
```
mikelehen, line 9: No longer used.

src/screens/HomePage/utils.ts
```diff
@@ -1,10 +1,27 @@
...
+
+/* Gets fips of 5 largest metro areas to set as homepage explore presets 
+  These are precomputed by hand using the function below.
+*/
+export function getLargestMetroFipsForExplore() {
+  return ['35620', '31080', '16980', '19100', '26420'];
+}
+
+/**
+ * Use this function to re-compute the values for the function above, if needed
```
mikelehen, line 19: This code is trivial enough to re-write, I'd just drop this rather than keep around an unused function. 🤷

mikelehen (approved):

src/screens/HomePage/utils.ts
schamp: I'm inclined to keep it 🤷, while I agree it is trivial, it provides context and details at almost no cost.

PR #3350, opened 2021-04-14
Search in navbar of non-location-pages

chasulin:
Adds search to navbar for all non-location-pages

[Figma](https://www.figma.com/file/JGqemgZwlG9e5DQbZpdKgy/Redesign?node-id=1560%3A110)

On homepage: renders once user has scrolled 560px. on mobile, the donate button disappears at that scroll point.
On all other non-location-pages: always visible on desktop+mobile. no donate button on mobile.


src/components/AppBar/NavBar.tsx
```diff
@@ -11,8 +11,11 @@ import { useBreakpoint } from 'common/hooks';
 
 const isLocationPage = (pathname: string) => pathname.includes('/us');
 
-const NavBar: React.FC = () => {
-  const [isMenuOpen, setMenuOpen] = useState(false);
+const NavBar: React.FC<{
+  renderSearch?: () => React.ReactElement;
+  renderSecondaryElement?: () => React.ReactElement;
+}> = ({ renderSearch, renderSecondaryElement }) => {
+  const [menuOpen, setMenuOpen] = useState(false);
```
chasulin, line 18: just renamed from `isMenuOpen` to `menuOpen`

src/components/AppBar/NavBar.tsx
```diff
@@ -77,22 +80,27 @@ const NavBar: React.FC = () => {
         )}
         <Link
           to="/"
-          style={{ display: 'inline-flex' }}
+          style={{ display: 'inline-flex', zIndex: 10000 }}
```
chasulin, line 83: manually moving the logo to be on top of the absolutely-positioned div that wraps the search bar so the logo is clickable/not covered

src/components/AppBar/NavBar.style.ts
```diff
@@ -10,7 +10,7 @@ import palette from 'assets/theme/palette';
 import { COLOR_MAP } from 'common/colors';
 import { materialSMBreakpoint } from 'assets/theme/sizes';
 
-const desktopNavHeight = 64;
+export const desktopNavHeight = 84;
```
chasulin, line 13: addition of search calls for 20 more px

src/components/AppBar/NavBar.tsx
chasulin, line 83: manually moving the logo to be on top of the absolutely-positioned div that wraps the search bar so the logo is clickable/not covered

pnavarrc: An alternative would be to wrap each section in flex containers and put the absolutely positioned div inside the middle container

```html
<flex-container>
  <flex-item>
    {/* Logo */}
  </flex-item>
  <flex-item>
    { renderSearch && renderSearch() }
  </flex-item>
  <flex-item>
    { renderMenuItems && renderMenuItems() }
  </flex-item>
</flex-container>
```

in this case, `renderSearch` should be responsible of wrapping the absolutely positioned div in a positioned one, that way the `NavBar` doesn't "need to know" that some of the render props are using position absolute

```tsx
const renderSearch = () => (
  <ContainerPositionRelative>
    <Search {...props} />
  </ContainerPositionRelative>
);
```

For `renderSecondaryElement`, we could just make it so it renders whatever is after the search, that way we don't need to special-case the donate button

```tsx
const renderMenuItems: () => (
  <>
    <DonateButton />
    <HamburgerMenu />
  </>
)
```

or 

```tsx
const renderMenuItems: () => (
    <HamburgerMenu />
)
```

chasulin: removed special casing of the donate button. re: the first part (wrapping each in flex item)--search needs to be in an absolutely positioned div that the navbar items are aware of, so that it sits in the middle of the navbar and not in the middle of its flex container (would be off-center)

mikelehen (approved):
Code looks great (nice use of theme variables) and seems to behave as expected!

chasulin:
[Add search field to top nav on non-location pages](https://trello.com/c/rA4QCJlQ/1167-add-search-field-to-top-nav-on-non-location-pages)

PR #3351, opened 2021-04-14
Make location router use new lookups

schamp:
This generates and uses urlSegment -> FIPS mappings and uses them in the location router.  It still depends on the regions DB, but breaks one of the dependencies and unlocks simpler decoupling later.

It does put urlSegment back into the main bundle, but the net effect shouldn't be significant (since we removed it earlier)


schamp:
we decided this wasn't the way to go

PR #3352, opened 2021-04-14
remove region DB load from vaccines phases test

schamp:
We have a `stateByFips` lookup we can use now instead.

### DIFF
--- a/src/cms-content/vaccines/phases.test.ts
+++ b/src/cms-content/vaccines/phases.test.ts
@@ -1,9 +1,11 @@
 import sortBy from 'lodash/sortBy';
-import regions from 'common/regions';
+import values from 'lodash/values';
+
+import { statesByFips } from 'common/regions';
 import { getVaccineInfoByFips, verifyOneItemPerState } from './phases';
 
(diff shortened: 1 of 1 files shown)


src/cms-content/vaccines/phases.test.ts
```diff
@@ -1,9 +1,11 @@
 import sortBy from 'lodash/sortBy';
-import regions from 'common/regions';
+import values from 'lodash/values';
+
+import { statesByFips } from 'common/regions';
 import { getVaccineInfoByFips, verifyOneItemPerState } from './phases';
 
-describe('vaccination info is complete for each state', () => {
-  const states = sortBy(regions.states, state => state.fullName);
+describe('vaccination info is complete for each state', async () => {
```
mikelehen, line 7: Why `async`?

mikelehen (approved):

src/cms-content/vaccines/phases.test.ts
schamp: ah, you know, I don't need that, will remove.

PR #3360, opened 2021-04-15
Recommendations cleanup

chasulin:
We no longer reference HGHI or the federal task force. This removes all related code (+ removes leftover code from the old sources and methodology modal)

TODO:
- [x] confirm copy


src/common/utils/recommend.ts
```diff
@@ -304,10 +184,11 @@ export const summaryByLevel = {
 export function getShareQuote(locationName: string, alarmLevel: Level): string {
   const locationNameWithAbbreviation = getAbbreviatedCounty(locationName);
 
-  const unknownShareQuote = `These are the White House Coronavirus Task Force’s official COVID recommendations for ${locationNameWithAbbreviation}:`;
+  // This needs updating (still mentions WH):
+  const unknownShareQuote = `These are the COVID guidelines set by the CDC:`;
```
chasulin, line 188: gonna double check about this copy

PR #3374, opened 2021-04-17
Reduce CCVI precision to ~2 digits to reduce import JSON size

schamp:
exploratory PR To see how much this might improve perf (vs. also splitting by region). 

ccvi.json before = ~843KB unzipped
ccvi.json after = ~632KB unzipped

Mapping scores to levels is at e.g. 0.2, 0.4., etc. so no need to load and parse CCVI scores like 0.2532445, right?

But I'm still checking for other uses so if we actually need that precision, we should keep it.


schamp:
Execution time of loading and parsing CCVI data (its own chunk!):

Before: 
image

After:
image

So seems significant enough to land on its own if we're okay with it while investigating other ways of breaking it up.

schamp:
Also appears to meaningfully reduce gzipped chunk size:
image

scripts/ccvi/parse_ccvi.ts
```diff
@@ -38,6 +38,9 @@ async function main() {
   const metroRows = await readCsv('ccvi_metros.csv');
 
   const rows = [...stateRows, ...countyRows, ...metroRows];
+  const numberize = (x: any) => {
+    return Number(toNumber(x).toString().substring(0, 4));
```
schamp, line 42: Cuts stringified float number to 4 characters, e.g., "0.21341451345" => "0.21"

We can do a more rigorous rounding / precision reduction if it's warranted, this was a quick-and-dirty way to see if it would be an improvement.

At the very least would probably want to add a comment. explaining what we're doing.

mikelehen: How about `Number(x.toFixed(2))` ?

mikelehen (approved):
Nice!

I'm planning to move this data into the API projections which will allow us to only load the data for a single region instead of all regions, but this is a nice improvement in the interim!

schamp:
Ah, good to know, I was just in the middle of exploring separating out the ccvi data.  But I like your approach including it with the projections in the API, that seems like a good place to put it.  So I'll stop going down that path.

PR #3375, opened 2021-04-18
Make snapshot useEffect dependency to avoid continual fetching

schamp:
This re-fetches the snapshot every time.  We didn't notice when it was using the version provided by the bundle, but it's evident when using an override.

Setting the dependency to snapshot version should prevent re-fetching and object identity-based state update.

### DIFF
--- a/src/common/location_summaries.ts
+++ b/src/common/location_summaries.ts
@@ -47,9 +47,9 @@ export async function fetchSummaries(snapshotNumber: number) {
 export function useSummaries(): SummariesMap | null {
   const [summaries, setSummaries] = useState<SummariesMap | null>(null);
 
-  const snapshot = getSnapshotOverride() || currentSnapshot();
   useEffect(() => {
     async function fetch() {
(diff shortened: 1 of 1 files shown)


src/common/location_summaries.ts
```diff
@@ -62,7 +62,7 @@ export function useSummaries(): SummariesMap | null {
       setSummaries(summaries);
     }
     fetch();
-  });
```
mikelehen, line 65: This seems fine, but rather than factoring `snapshot` out and then making it a dependency, could you just add `[]` as the dependency list here?

mikelehen (approved):
Thanks for chasing this down!

src/common/location_summaries.ts
schamp: I could.  It would mean it wouldn't be updated if we ever make snapshot update outside of a navigation, e.g., something like a built-in that auto-cycles through data like a gif-maker without navigation.

But maybe WAGNI, so sure I can make that change.

PR #3388, opened 2021-04-19
Location page: vaccinations progress bar, shared components, spark lines wip

chasulin:
Pr includes:
- Vaccination progress bar
- Spark line component (still work to be done)
- A container div to use as parent of each location page section
- A subtitle +chevron icon to reuse throughout the location page


src/components/NewLocationPage/Shared/SectionContainer.tsx
```diff
@@ -0,0 +1,12 @@
...
+ */
+
+import React from 'react';
+import { Container } from './Shared.style';
+
+const SectionContainer: React.FC = ({ children }) => {
+  return <Container>{children}</Container>;
+};
+
+export default SectionContainer;
```
pnavarrc, line 12: This component is not needed, the Container already received `children` + all the other props. Consider just export it with a name from `NewLocationPage/Shared/index.ts`


```tsx
export { Container as SectionContainer } from './Shared.style';
```

src/components/NewLocationPage/SparkLines/SingleSparkLine.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { scaleLinear } from '@vx/scale';
+import { LinePath } from '@vx/shape';
+import { curveNatural } from '@vx/curve';
+import { COLOR_MAP } from 'common/colors';
+import { Point } from './utils';
+
+const SingleSparkLine: React.FC<{
+  data: Point[];
+  chartWidth: number;
+  chartHeight: number;
```
pnavarrc, line 11: Maybe just call them `width` and `height`?

src/components/NewLocationPage/SparkLines/SingleSparkLine.tsx
```diff
@@ -0,0 +1,64 @@
...
+  capY: number;
+}> = ({ data, chartWidth, chartHeight, capY }) => {
+  const yScale = scaleLinear({
+    domain: [0, capY],
+    range: [chartHeight, 0],
+  });
```
pnavarrc, line 17: I don't think we need `capY`, we should probably use

```tsx
import { max as d3Max } from 'd3-array';

const yScale = scaleLinear({
  domain: [0, d3max(data, d => d.y)],
  range[height, 0]
}
```

src/components/NewLocationPage/SparkLines/SingleSparkLine.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const xScale = scaleLinear({
+    domain: [0, 29],
+    range: [0, chartWidth],
+  });
+
+  const getXCoord = (p: Point) => xScale(p.x);
+  const getYCoord = (p: Point) => yScale(p.y);
```
pnavarrc, line 25: Maybe we don't need these functions, they make a bit more sense when the calculation is more involved or when we need to reuse them a couple of times.

src/components/NewLocationPage/SparkLines/SingleSparkLine.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const barWidth = 2;
+
+  return (
+    <svg
+      width={chartWidth}
+      height={chartHeight}
+      style={{ border: '1px solid blue' }}
+    >
```
pnavarrc, line 34: maybe remove the styling

src/components/NewLocationPage/SparkLines/SingleSparkLine.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <g>
+        {data.map((p: Point) => {
+          return (
+            <rect
+              fill={COLOR_MAP.GREY_1}
```
pnavarrc, line 39: This will need a key (is inside a `map`)

src/components/NewLocationPage/SparkLines/SingleSparkLine.tsx
```diff
@@ -0,0 +1,64 @@
...
+              x={xScale(p.x) - barWidth / 2}
+              y={yScale(p.y)}
+              width={barWidth}
+              height={chartHeight - getYCoord(p)}
+            />
+          );
+        })}
+      </g>
+      <g stroke="black" strokeWidth={1.5}>
```
pnavarrc, line 48: I think it's better to pass the attributes to `LinePath` directly

src/components/NewLocationPage/VaccinationsProgressBar/VaccinationsProgressBar.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  return (
+    <svg
+      width={barWidth}
+      height={barHeight}
+      role="img"
+      aria-labelledby={titleId}
+    >
+      <title id={titleId}>
+        Progress bar showing {locationName}'s percent vaccinated
```
pnavarrc, line 33: We can add the values too since there is only 2 of them.

pnavarrc (approved):
Added some notes

src/components/NewLocationPage/Shared/SectionContainer.tsx
chasulin: oooo yes this is very true.

PR #3401, opened 2021-04-21
Spark lines with projections

chasulin:
- Plugs projections into spark lines
- Matches [mocks](https://www.figma.com/file/JGqemgZwlG9e5DQbZpdKgy/Redesign?node-id=1903%3A391)

Screen Shot 2021-04-21 at 11 07 02 AM


src/components/NewLocationPage/SparkLineBlock/SparkLine.tsx
```diff
@@ -0,0 +1,96 @@
...
+            return (
+              <LinePath
+                stroke="black"
+                strokeWidth={1.5}
+                key={`line-${p.x}`}
+                data={smoothedData}
+                x={getXCoord}
+                y={getYCoord}
+                curve={curveNatural}
+                stroke-linecap="round"
```
pnavarrc, line 66: This should be `strokeLinecap` — alternatively, you can create a style wrapper for the line path element so you can move all the styling there

```tsx
import { LinePath } from '@vx/shape';

export const SparkLinePath = styled(LinePath)`
  stroke: black;
  stroke-width: 1.5px;
  stroke-linecap: round;
`;
```

and then use it

```tsx
<SparkLinePath
  data={smoothedData}
  x={getXCoord}
  y={getYCoord}
  curve={curveNatural}
/>
```

src/components/NewLocationPage/SparkLineBlock/SparkLineBlock.style.tsx
```diff
@@ -0,0 +1,74 @@
...
+  text-overflow: ellipsis;
+`;
+
+export const BlockTitle = styled.h2`
+  margin: 0 0 1.5rem;
+  font-size: 1rem;
+  line-height: 1.4;
+  color: ${COLOR_MAP.GRAY_BODY_COPY};
+  text-transform: uppercase;
+`;
```
pnavarrc, line 32: Should this be shared so it can be used in other blocks?

src/components/NewLocationPage/SparkLineBlock/SparkLineBlock.style.tsx
```diff
@@ -0,0 +1,74 @@
...
+  }
+`;
+
+export const StyledLink = styled(Link)`
+  color: black;
+  text-decoration: none;
+
+  &:not(:last-of-type) {
+    margin-right: 1rem;
+    margin-bottom: 1rem;
```
pnavarrc, line 65: The `margin-bottom` shouldn't be conditional right? Also, that's probably something we want to add to the container (`padding-bottom`). Consider moving contextual concerns to the container (like spacing between its children) instead of setting that on the children items themselves

```ts
const SparklineSetContainer = styled.div`
  padding-bottom: 1rem;
  * > {
    &:not(:last-of-type) {
      margin-right: 1rem;
    }
  }
`;

const StyledLink = styled(Link)`
  {/* only style what's inside the link */}
`;
```

```tsx
<SparklineSetContainer>
  <StyledLink to="/">
    <ChartTitle title={title} />
    <SparkLine rawData={rawData} smoothedData={smoothedData} />
  </StyledLink>
</SparklineSetContainer>
```

src/components/NewLocationPage/SparkLineBlock/utils.ts
```diff
@@ -0,0 +1,118 @@
...
+ * Returns both the raw and smoothed series for the given metric.
+ * Raw series used by bar chart, smoothed series used by line chart.
+ */
+export function getAllSeriesForMetric(
+  seriesList: Series[],
+  projection: Projection,
+): SeriesWithData[] {
+  return seriesList.map((item: Series) => ({
+    ...item,
+    data: cleanSeries(projection.getDataset(item.datasetId)),
```
pnavarrc, line 84: Discussed on Slack, we need to filter data from the last 30 days here

src/components/NewLocationPage/SparkLineBlock/SparkLineSet.tsx
```diff
@@ -0,0 +1,38 @@
...
+      {SPARK_LINE_METRICS.map((metric: SparkLineMetric) => {
+        const { title, seriesList } = sparkLinesMetricData[metric];
+        const metricSeries = getSparkLineSeriesFromProjection(
+          seriesList,
+          projection,
+        );
+        const rawData = getDataFromSeries(metricSeries[0]);
+        const smoothedData = getDataFromSeries(metricSeries[1]);
+        return (
+          <StyledLink to="/">
```
pnavarrc, line 28: This needs a key

pnavarrc (commented):
Added some notes, I can review again after the data filter update!

src/components/NewLocationPage/SparkLineBlock/SparkLineBlock.style.tsx
chasulin: im not sure i agree? `StyledLink` is the wrapper--i think it makes sense to put the margins on the children's wrappers, no? re: the margin bottom in particular, its a bit of a cheat way to make sure that there is spacing if the spark lines wrap

Screen Shot 2021-04-21 at 2 23 37 PM

pnavarrc (approved):
Looks good!

src/components/NewLocationPage/SparkLineBlock/SparkLine.tsx
chasulin: updated from `stroke-linecap` to `strokeLinecap`. as for making a style wrapper, i was getting a typescript error and gave up on investigating it for now 😅

PR #3403, opened 2021-04-21
Add basic notes block appearance features

faihegberg:
This PR creates a notes block for the new location page. This PR focuses only on establishing the appearance of the component so it is a WIP. Other functionalities (four props including icon, title, body, redirect URL) have not been built and passed into this component yet.


src/components/NewLocationPage/NotesBlock/NotesBlock.style.tsx
```diff
@@ -0,0 +1,18 @@
...
+  display: inline-block;
+  flex-direction: column;
+  font-weight: bold;
+`;
+
+export const TextComponent = styled.div`
+  margin-top: 10px;
+  margin-bottom: 10px;
+  font-weight: normal;
+  color: #4f4f4f;
```
chasulin, line 13: this color is in `COLOR_MAP`. rather than hardcoding the hex code, you can import `COLOR_MAP` and use `COLOR_MAP.GREY_4`

src/components/NewLocationPage/Shared/Shared.style.tsx
```diff
@@ -5,7 +5,7 @@ import { COLOR_MAP } from 'common/colors';
 
 export const SectionContainer = styled.div`
   display: flex;
-  flex-direction: column;
+  flex-direction: row;
```
chasulin, line 8: lets avoid editing the styles of reusable parent div. instead, you can add a wrapping div with `display:flex` inside of `SectionContainer`

src/components/NewLocationPage/NotesBlock/NotesBlock.style.tsx
```diff
@@ -0,0 +1,18 @@
+import styled from 'styled-components';
+
+export const TextContainer = styled.div`
```
chasulin, line 3: i think an unstyled div would render things properly. so just:
`export const TextContainer = styled.div```

src/components/MenuContent/Menu.style.tsx
```diff
@@ -225,5 +225,4 @@ export const LogoWrapper = styled(Link)`
 
 export const NonWrappingSpan = styled.span`
   white-space: nowrap;
-  ${props => props.theme.fonts.regularBookMidWeight};
```
chasulin, line 228: ooo actually- separate of this pr, i think i have to fix the font weight of the `TextAndIconWithSpecialWrapping`

src/components/NewLocationPage/NotesBlock/NotesBlock.style.tsx
```diff
@@ -0,0 +1,18 @@
...
+  display: inline-block;
+  flex-direction: column;
+  font-weight: bold;
```
chasulin, line 6: re: this comment, once i make that fix, we wont need `font-weight: bold` here

src/components/NewLocationPage/NotesBlock/NotesBlock.style.tsx
```diff
@@ -0,0 +1,18 @@
...
+`;
+
+export const TextComponent = styled.div`
```
chasulin, line 9: we want text to render in the dom as text. making this a `p` element instead of a `div` would do the trick

src/components/NewLocationPage/NotesBlock/NotesBlock.style.tsx
```diff
@@ -0,0 +1,18 @@
...
+  margin-top: 10px;
+  margin-bottom: 10px;
```
chasulin, line 11: `SectionContainer`'s padding should properly take care of the spacing below this text block, so we can nix this `margin-bottom`

src/components/NewLocationPage/NotesBlock/NotesBlock.style.tsx
```diff
@@ -0,0 +1,18 @@
+import styled from 'styled-components';
+
+export const TextContainer = styled.div`
+  display: inline-block;
+  flex-direction: column;
+  font-weight: bold;
+`;
+
+export const TextComponent = styled.div`
+  margin-top: 10px;
```
chasulin, line 10: its best to use `rem` instead of `px`.[ this article summarizes](https://engageinteractive.co.uk/blog/em-vs-rem-vs-px) this concept pretty well. implementation-wise, 1rem = 16px, so this^ would be `margin-top: .625rem`

src/assets/images/VulnerabilityIcon.svg
```diff
@@ -0,0 +1,9 @@
...
+</svg>
```
chasulin, line 9: this works! you can also place this svg into a react component. a good example to reference is `AlertsIcon` in `AlertsIcon.tsx`, and you can see it put to use in `GetAlertsButton.tsx` (imported on line 2, implemented on line 17)

src/components/NewLocationPage/NotesBlock/NotesBlock.style.tsx
faihegberg: Since I am adjusting it to just `export const TextContainer = styled.p`, would I still need to specify the margins? 🤔

chasulin: the default margins that come with a `p` are probably rendering, yeah? id override those and match the margin-top to what is in the mocks

faihegberg: Yep! Ignore me 🙈 Turns out I was looking at the wrong component! This makes sense.

src/components/NewLocationPage/NotesBlock/NotesBlock.style.tsx
```diff
@@ -0,0 +1,19 @@
...
+  font-size: initial;
+  text-transform: none;
+`;
+
+export const TextContainer = styled.div``;
+
+export const TextComponent = styled.p`
+  color: ${COLOR_MAP.GREY_4};
+  font-weight: initial;
+`;
```
faihegberg, line 15: When I don't specify margin, the space between the paragraph and the title seems bigger than in the mock. Is it worthwhile to explicitly specify the top margin here or setting it to `initial`? (Although when I tried setting it to initial, it makes the space just a tiny bit too narrow).

chasulin: yeah, paragraph elements come with default margins. i would just override them and set the margin top to whatever the amount of space there is between title+body in the mocks. so something like `margin: .5rem 0 0;` . this sets all margins to be 0 except for the top-margin which is .5rem

src/components/NewLocationPage/NotesBlock/NotesBlock.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+const body =
+  'New York City metro is more likely to experience severe physical and economic suffering from COVID, and to face harder, longer recovery.';
+const redirectUrl = 'http://www.google.com';
+
+export const Example = () => {
+  if (!highVulnerability) {
+    return null;
+  }
+  return (
+    <BaseButton
```
chasulin, line 24: it looks like `BaseButton` is putting some unwanted styling on the children/NotesBlock. i think we need to make a styled BaseButton that explicitly doesn't style its children. something like:

```
export const StyledBaseButton = styled(BaseButton)`
font-weight: inherit;
font-size: inherit;
text-transform: none;
letter-spacing: 0;
`
```

And then import it into the file and swap `BaseButton` for `StyledBaseButton`

src/components/NewLocationPage/NotesBlock/NotesBlock.style.tsx
```diff
@@ -0,0 +1,19 @@
+import styled from 'styled-components';
+import { COLOR_MAP } from 'common/colors';
+
+export const SectionContentContainer = styled.div`
+  display: flex;
+  font-size: initial;
+  text-transform: none;
```
chasulin, line 7: [once we implement StyledBaseButton](https://github.com/covid-projections/covid-projections/pull/3403/files#r618804311), we can remove `font-size: initial;` and `text-transform: none;`

src/components/NewLocationPage/NotesBlock/NotesBlock.style.tsx
```diff
@@ -0,0 +1,19 @@
...
+`;
+
+export const TextContainer = styled.div``;
+
+export const TextComponent = styled.p`
+  color: ${COLOR_MAP.GREY_4};
+  font-weight: initial;
```
chasulin, line 14: [same as above](https://github.com/covid-projections/covid-projections/pull/3403/files#r618805534). we can remove `font-weight: initial;`

chasulin (approved):
looks good! 🚢

PR #3404, opened 2021-04-21
Convert projection.ts file to use native JavaScript dates

faihegberg:
This PR migrates Projection.ts file to use native JavaScript dates.

### DIFF
--- a/src/common/models/Projection.ts
+++ b/src/common/models/Projection.ts
@@ -33,11 +33,6 @@ export function reenableDisabledMetrics(enable: boolean): void {
  */
 export const RT_TRUNCATION_DAYS = 7;
 
-/**
- * We truncate our projections to 4 weeks out.
- */
-export const PROJECTIONS_TRUNCATION_DAYS = 30;
-
 // We require at least 15 ICU beds in order to show ICU Capacity usage.
 // This still covers enough counties to cover 80% of the US population.
 const MIN_ICU_BEDS = 15;
@@ -463,9 +458,9 @@ export class Projection {
    * timeseries) as well as the dates to be consistent (since we keep track of
    * three lists).
    *
-   * In order to this grab the earliest and latest dates from the
-   * timeseries and for every single day in between them fill the each array
-   * (the dates, the actuals and the timeseres(predicted)) with the value at
+   * In order to do this, we grab the earliest and latest dates from the
+   * timeseries and for every single day in between them fill each array
+   * (the dates, the actuals and the timeseries(predicted)) with the value at
    * that date or null.
    */
   private getAlignedTimeseriesAndDates(
@@ -520,17 +515,18 @@ export class Projection {
       // Clone the date since we're about to mutate it below.
       dates.push(new Date(currDate.getTime()));
 
-      // increment the date by one
+      // Increment the date by one.
+      // We specifically use setUTCDate and getUTCDate instead of setDate and getDate,
+      // as the latter two do not take daylight savings into account.
+      // This results in each day being incremented by 24 hours (although two days of the year should be 23 and 25),
+      // leading to days being shifted and displayed incorrectly.
       currDate.setUTCDate(currDate.getUTCDate() + 1);
     }
 
-    const now = new Date();
-    const todayIndex = dates.findIndex(date => date > now);
-    const days = todayIndex >= 0 ? todayIndex : dates.length;
     return {
-      actualTimeseries: actualsTimeseries.slice(0, days),
-      metricsTimeseries: metricsTimeseries.slice(0, days),
-      dates: dates.slice(0, days),
+      actualTimeseries: actualsTimeseries,
+      metricsTimeseries: metricsTimeseries,
+      dates: dates,
     };
   }
 
--- a/src/common/models/Projection.ts
+++ b/src/common/models/Projection.ts
@@ -180,10 +180,7 @@ export class Projection {
       actualTimeseries,
       metricsTimeseries,
       dates,
-    } = this.getAlignedTimeseriesAndDates(
-      summaryWithTimeseries,
-      PROJECTIONS_TRUNCATION_DAYS,
-    );
+    } = this.getAlignedTimeseriesAndDates(summaryWithTimeseries);
     const metrics = summaryWithTimeseries.metrics;
 
     this.metrics = metrics || null;
@@ -473,7 +470,6 @@ export class Projection {
    */
   private getAlignedTimeseriesAndDates(
     summaryWithTimeseries: RegionSummaryWithTimeseries,
-    futureDaysToInclude: number,
   ) {
     const actualsTimeseriesRaw = summaryWithTimeseries.actualsTimeseries;
     const metricsTimeseriesRaw = summaryWithTimeseries.metricsTimeseries || [];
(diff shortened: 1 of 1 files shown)


src/common/models/Projection.ts
```diff
@@ -485,19 +484,31 @@ export class Projection {
...
     // If we have projections, we use that time range; else we use the actuals.
     // TODO(chris): Is there a reason that this was bound to the projections timeseries first?
     // It cuts off some of the earlier dates
     if (metricsTimeseriesRaw.length > 0) {
-      earliestDate = moment.utc(first(metricsTimeseriesRaw)!.date);
-      latestDate = moment.utc(last(metricsTimeseriesRaw)!.date);
+      [year, month, date] = first(metricsTimeseriesRaw)!
+        .date.split('-')
+        .map(elem => parseInt(elem));
+      earliestDate = new Date(Date.UTC(year, month - 1, date));
```
faihegberg, line 495: `Date.UTC` accepts integers as parameters, not strings. Hence the string manipulation for `[year, month, date]`.

src/common/models/Projection.ts
```diff
@@ -485,19 +484,31 @@ export class Projection {
...
+      [year, month, date] = last(actualsTimeseriesRaw)!
+        .date.split('-')
+        .map(elem => parseInt(elem));
+      latestDate = new Date(Date.UTC(year, month - 1, date));
     }
 
-    earliestDate = moment.utc('2020-03-01');
+    earliestDate = new Date(Date.UTC(2020, 2, 1));
```
faihegberg, line 511: Perhaps I should add a comment here that native JavaScript months are zero-indexed.

src/common/models/Projection.ts
```diff
@@ -510,9 +521,10 @@ export class Projection {
       metricsTimeseriesRaw,
     );
 
-    let currDate = earliestDate.clone();
-    while (currDate.diff(latestDate) <= 0) {
-      const ts = currDate.format('YYYY-MM-DD');
+    let currDate = new Date(earliestDate.getTime());
+
+    while (currDate <= latestDate) {
+      const ts = currDate.toISOString().substring(0, 10);
```
faihegberg, line 527: The most strightforward way I could find to get the date in `YYYY-MM-DD` format in native JavaScript is to first convert it using `toISOString()` then take the first 10 characters 😐 . Probably not the most elegant but other methods seem unnecessarily lengthy.

src/common/models/Projection.ts
mikelehen: Hrm, I think you can just do `new Date(first(metricsTimeseriesRaw)!.date)` since these give me the same result:
image

or am I missing something?

src/common/models/Projection.ts
```diff
@@ -485,19 +484,31 @@ export class Projection {
...
-      earliestDate = moment.utc(first(actualsTimeseriesRaw)!.date);
-      latestDate = moment.utc(last(actualsTimeseriesRaw)!.date);
+      [year, month, date] = first(actualsTimeseriesRaw)!
+        .date.split('-')
+        .map(elem => parseInt(elem));
+      earliestDate = new Date(Date.UTC(year, month - 1, date));
+      [year, month, date] = last(actualsTimeseriesRaw)!
+        .date.split('-')
+        .map(elem => parseInt(elem));
+      latestDate = new Date(Date.UTC(year, month - 1, date));
```
mikelehen, line 508: If we do need to do this complicated string parsing / date creation dance, we should move it to a helper function so it's not duplicated 4 times. 😁

src/common/models/Projection.ts
mikelehen: Again, I think we can just use the string here, but otherwise, yes, a comment would be great. 😁

src/common/models/Projection.ts
```diff
@@ -521,10 +533,10 @@ export class Projection {
       ] as MetricsTimeseriesRow;
       actualsTimeseries.push(actualsTimeseriesrowForDate || null);
       metricsTimeseries.push(metricsTimeseriesRowForDate || null);
-      dates.push(currDate.toDate());
+      dates.push(new Date(currDate.getTime()));
```
mikelehen, line 536: Perhaps add a comment:
```
// Clone the date since we're about to mutate it below.
```

src/common/models/Projection.ts
```diff
@@ -521,10 +533,10 @@ export class Projection {
...
 
       // increment the date by one
-      currDate = currDate.clone().add(1, 'days');
+      currDate.setUTCDate(currDate.getUTCDate() + 1);
     }
 
     // only keep futureDaysToInclude days ahead of today
```
mikelehen, line 542: If you feel like doing some cleanup, we don't actually need `futureDaysToInclude` anymore, so you can remove that parameter and delete this code.

mikelehen (changes requested):
I think we can simplify the date parsing, but otherwise this looks good.  Thanks!

src/common/models/Projection.ts
faihegberg: I decided to use `Date.UTC` to be explicit about instantiating it as a UTC timestamp and to be extra cautious due to the note outlined under "3. Timestamp string" in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date. That said, switching it can potentially make it much more straightforward, and you are correct that parsing _just_ the date value as string (no timestamp attached) should yield the exact same result (but only if we always parse date-only value and in ISO8601 format). I can make the switch to make it more straightforward.

src/common/models/Projection.ts
```diff
@@ -521,17 +517,16 @@ export class Projection {
...
-      currDate = currDate.clone().add(1, 'days');
+      currDate.setUTCDate(currDate.getUTCDate() + 1);
     }
 
-    // only keep futureDaysToInclude days ahead of today
     const now = new Date();
     const todayIndex = dates.findIndex(date => date > now);
-    const days =
-      todayIndex >= 0 ? todayIndex + futureDaysToInclude : dates.length;
+    const days = todayIndex >= 0 ? todayIndex : dates.length;
```
faihegberg, line 529: Not very sure whether we still need `days` variable as is, since `futureDaysToInclude` has been removed?

src/common/models/Projection.ts
```diff
@@ -181,10 +180,7 @@ export class Projection {
       actualTimeseries,
       metricsTimeseries,
       dates,
-    } = this.getAlignedTimeseriesAndDates(
-      summaryWithTimeseries,
-      PROJECTIONS_TRUNCATION_DAYS,
```
faihegberg, line 186: Do we still need `PROJECTIONS_TRUNCATION_DAYS` constant since futureDaysToInclude has been removed? I do not see it referenced elsewhere.

mikelehen: Nope, remove it!

src/common/models/Projection.ts
mikelehen: I think you can just remove `now`, `todayIndex`, `days` and the calls to `slice()`.

src/common/models/Projection.ts
```diff
@@ -521,17 +517,16 @@ export class Projection {
       ] as MetricsTimeseriesRow;
       actualsTimeseries.push(actualsTimeseriesrowForDate || null);
       metricsTimeseries.push(metricsTimeseriesRowForDate || null);
-      dates.push(currDate.toDate());
+      // Clone the date since we're about to mutate it below.
+      dates.push(new Date(currDate.getTime()));
 
       // increment the date by one
-      currDate = currDate.clone().add(1, 'days');
+      currDate.setUTCDate(currDate.getUTCDate() + 1);
```
mikelehen, line 524: Oh, do you mind adding a comment on this explaining that we're using native Date due to the issues you ran into with Luxon regarding daylight saving time?  That way somebody knows to be very careful if they rework this in the future.

faihegberg: I'll add it at the top of the method.

mikelehen (approved):

PR #3405, opened 2021-04-21
Fix bug on TextAndIconWithSpecialWrapping

chasulin:
the correct font weight wasnt being applied to both spans. this fixes + moves the component out of the `MenuContent` folder, since it is now being used outside of the menu


pnavarrc (approved):
Looks good!

PR #3407, opened 2021-04-21
Homepage banner: Update rising hospitalizations states.

mikelehen:
# Old:
image

# New:
image

### DIFF
--- a/src/screens/HomePage/HomePage.tsx
+++ b/src/screens/HomePage/HomePage.tsx
@@ -75,14 +75,14 @@ export default function HomePage() {
     location.hash === '#explore-hospitalizations';
   const risingHospitalizationStates = [
     'DE',
+    'FL',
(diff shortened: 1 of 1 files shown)


src/screens/HomePage/HomePage.tsx
```diff
@@ -75,14 +75,14 @@ export default function HomePage() {
     location.hash === '#explore-hospitalizations';
   const risingHospitalizationStates = [
     'DE',
+    'FL',
     'IL',
     'MD',
     'MI',
-    'MN',
-    'NJ',
+    'NH',
```
mikelehen, line 82: Note MN is still rising but it was crowding out PR which feels more important to highlight so I swapped in NH instead.

BrettBoval (approved):
Looks good to me. Thanks for updating this.

PR #3408, opened 2021-04-22
Move <a> into cells, add onClick to navigate on row

schamp:
To clear the DOM validation warning.  Seems the most straightforward way, which preserves any styling on the links, and allows non-link areas of the row to be clickable to navigate.


mikelehen (approved):
Nice, thanks for grabbing!

PR #3409, opened 2021-04-22
Add forwardRef to prevent warning

schamp:


mikelehen (approved):
FWIW- I think the `ref` that's actually triggering this isn't even used / necessary, so we could just remove it. https://github.com/covid-projections/covid-projections/blob/947bd4274cc459a0646b1fda44af1cde3e4d0afe/src/components/LocationPage/ChartsHolder.tsx#L192

But this is fine too.  It seems reasonable to forward the `ref` through the error boundary. 🤷

PR #3411, opened 2021-04-22
Correct dom validation issue in location page header

schamp:
The tooltip div in the header was inside a `<p>`, which is a no-no.  The fix is to move it outside the `<p>`, and change the style to `inline-block` so that it remains on the same line, instead of automatically being put on the next one.

Before: 
image

After:
image


mikelehen (approved):
Looks good!

PR #3416, opened 2021-04-22
Redirect blog URLs to right places.

mikelehen:
We're going to point blog.covidactnow.org at the main website and turn off the WordPress blog.  These URL redirects were provided by Mikayla.


mikelehen:
[Reroute blog URL to Learm](https://trello.com/c/Q1wnJ05b/1219-reroute-blog-url-to-learm)

PR #3417, opened 2021-04-22
Use USA API endpoint and remove Native American Majority Counties.

mikelehen:
* Adds a new USA region class available via `regions.usa`.
* Replaces aggregations.json with a (much smaller) site-summary.json file that contains arbitrary precomputed data (right now just usa summary data).
* Wires up USA API endpoint and new site-summary.json file to existing components (Explore, NationalText, banners).

This affects the "USA" explore chart, the `NationalText` copy above the homepage explore chart, and the banner on the homepage.  The national numbers are subtly different than our frontend aggregation, which is to be expected because now we're using real USA numbers instead of aggregating the states.

Also, saves 20k in the gzipped main bundle.


src/api/index.ts
```diff
@@ -84,7 +84,11 @@ export class Api {
   async fetchSummaryWithTimeseries(
     region: Region,
   ): Promise<RegionSummaryWithTimeseries | null> {
-    if (region instanceof State) {
+    if (region.fipsCode === '0') {
```
schamp, line 87: since we're using this coded value in more than one place, consider using a named constant (e.g., USA_FIPS_CODE) instead of a string literal so 1) we can trace its usage better and 2) so it's clear what it represents.

mikelehen: 👍  This is a WIP.  Will clean up!

mikelehen:
[Use API based data for the US counts on the homepage](https://trello.com/c/PH0925Wd/1206-use-api-based-data-for-the-us-counts-on-the-homepage)

src/common/models/Projection.ts
```diff
@@ -303,6 +305,25 @@ export class Projection {
     return getRegionMetricOverride(this.region, metric)?.blocked ?? false;
   }
 
+  get twoWeekPercentChangeInCases() {
```
mikelehen, line 308: This "two week percentage" logic moved from NationalText to here so we could store it in the site-summary.json file.

src/common/models/Projection.ts
```diff
@@ -261,9 +261,11 @@ export class Projection {
     return lastValue(this.smoothedDailyCases);
   }
 
-  /** Returns the date when projections end (should be 30 days out). */
+  /** Returns the last date we have (case) data for. */
   get finalDate(): Date {
```
mikelehen, line 265: This was unused so I've repurposed it.

src/components/DataCoverageTable/DataCoverageTable.tsx
```diff
@@ -85,6 +85,8 @@ const tooltipText = (coverage: CoverageDetails, regionType: RegionType) => {
       return `Supported in ${coverage.regionsAvailable} metros (of ${coverage.totalRegions}).`;
     case RegionType.STATE:
       return `Supported in ${coverage.regionsAvailable} states (of ${coverage.totalRegions}).`;
+    case RegionType.NATION:
+      fail('Unexpected RegionType.NATION');
```
mikelehen, line 89: Had to add this to make TypeScript happy.

src/components/NationalText/utils.tsx
```diff
@@ -1,55 +1,31 @@
...
-  );
+    lastDate,
+    twoWeekPercentChangeInCases,
+    twoWeekPercentChangeInDeaths,
+  } = usa;
 
-  const lastDate = last(dates);
   const lastDateFormatted: string = formatUTCDateTime(
-    parseDateUnix(lastDate!),
+    parseDateUnix(lastDate),
```
mimecuvalo, line 28: haha what did `lastDate!` compile to before?? that wasn't a syntax error!?

mimecuvalo (commented):
looks reasonable to me - but i have less context still than most - prbly good to have a 2nd of eyes before landing.

src/components/NationalText/utils.tsx
mikelehen: Hah, `!` is the "[non-null assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator)" operator in TypeScript.  It tells TypeScript that even though the type says the value may be null you pinky-promise-swear that it won't actually be null and so it should let you pretend it won't ever be null.  It's similar to `any` in that whenever you use it, you're overriding TypeScript's type-checking and giving yourself the opportunity to shoot yourself in the foot.

In this case, since `lastDate` is now coming from the JSON file at compile-time, it can't be null (or rather TypeScript will fail the build if we ever screw up and put null in the JSON file), so `!` isn't necessary.

mimecuvalo: ahh gotcha - TIL - thanks! it's been far too long since i've used TypeScript (4 years?) - i don't remember using that back in the day (instead i remember just marking things as `any`)

scripts/update_location_summaries.ts
```diff
@@ -6,6 +6,7 @@ import {
   fetchAllStateProjections,
   fetchAllMetroProjections,
   fetchAllCountyProjections,
+  fetchProjectionsRegion,
```
pnavarrc, line 9: _nit_: It looks like the following imports are not needed anymore

```ts
import { DatasetId } from '../src/common/models/Projection';
import { assert } from '../src/common/utils';
```

also the `INDIGENOUS_FIPS` list is not used anymore

src/common/models/Projection.ts
```diff
@@ -309,6 +311,25 @@ export class Projection {
...
+    return this.getTwoWeekPercentChange(this.smoothedDailyCases);
+  }
+
+  get twoWeekPercentChangeInDeaths() {
+    return this.getTwoWeekPercentChange(this.smoothedDailyDeaths);
+  }
+
+  private getTwoWeekPercentChange(series: any[]): number | null {
+    const lastIndex = indexOfLastValue(series) ?? series.length;
+    const lastTwoWeeks = series.slice(lastIndex - 14, lastIndex + 1);
```
pnavarrc, line 324: Should we use the dates to calculate the value 2 weeks ago instead? For the U.S. series this is probably okay, but for counties with sparse data the date corresponding to 14 days ago could be months ago (for data reported weekly).

We can punt if this is not really a problem, but maybe add a note so we know that this calculation is assuming that we have one data point per day.

src/common/models/Projection.ts
```diff
@@ -309,6 +311,25 @@ export class Projection {
...
+  get twoWeekPercentChangeInCases() {
+    return this.getTwoWeekPercentChange(this.smoothedDailyCases);
+  }
+
+  get twoWeekPercentChangeInDeaths() {
+    return this.getTwoWeekPercentChange(this.smoothedDailyDeaths);
+  }
+
+  private getTwoWeekPercentChange(series: any[]): number | null {
+    const lastIndex = indexOfLastValue(series) ?? series.length;
```
pnavarrc, line 323: _nit_: shouldn't it be `series.length - 1`? I see that we compensate below, feel free to  leave as it for now

```ts
const lastIndex = indexOfLastValue(series) ?? series.length - 1;
```

pnavarrc (approved):
Added some notes, looks good!

src/common/models/Projection.ts
mikelehen: Er, yeah, I think I threw that in to make TypeScript happy since `indexOfLastValue()` can return null, but really the series should never be empty and if it is, this code is going to behave strangely so I just added an assert instead.

The `lastIndex + 1` below does need to stay as-is since .slice() treats the end index as exclusive, not inclusive.

src/common/models/Projection.ts
mikelehen: We actually have a bunch of code that assumes that these timeseries don't have gaps and there is a (possibly null) data point for every day.

This is guaranteed to work because of [`getAlignedTimeseriesAndDates()`](https://github.com/covid-projections/covid-projections/blob/94bd2ae85096794f15ecd77a63ea8cba59b9420d/src/common/models/Projection.ts#L494)

PR #3421, opened 2021-04-22
New summary stats 

chasulin:
Implements the individual desktop+mobile stat. Not yet placed into a block

Location page redesign/Summary stats in storybook


src/components/NewLocationPage/SummaryStatsBlock/SummaryStatsBlock.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
```
chasulin, line 1: ignore this whole component for now. as we chatted about, vaccinations will need to render separately, etc etc

src/components/NewLocationPage/SummaryStatsBlock/utils.ts
```diff
@@ -0,0 +1,45 @@
...
+export interface SummaryStatProps {
+  levelInfo: LevelInfo;
+  formattedValue: string;
+  hasSubLabel: boolean;
+  metricName: string;
+  isMobile: boolean;
+  metric: Metric;
+}
+
+export const orderedStatMetrics = [
```
chasulin, line 16: same as [above](https://github.com/covid-projections/covid-projections/pull/3421/files#r618768534) ^^

src/components/NewLocationPage/SummaryStatsBlock/DesktopSummaryStat.tsx
```diff
@@ -0,0 +1,39 @@
...
+        <MetricLabel>{metricName}</MetricLabel>
+        <StyledChevron />
+      </Row>
+      <Row>
+        <MetricValue value={formattedValue} iconColor={levelInfo.color} />
+        {hasSubLabel && (
+          <MetricSubLabel
+            text={metricSubLabelText[metric]}
+            isMobile={isMobile}
+          />
```
pnavarrc, line 32: I guess in this case `isMobile` will always be false right? If that's the case we don't need the `isMobile` prop in the `DesktopSummaryStat` component

src/components/NewLocationPage/SummaryStatsBlock/DesktopSummaryStat.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const DesktopSummaryStat: React.FC<SummaryStatProps> = ({
+  levelInfo,
+  formattedValue,
+  hasSubLabel,
+  metricName,
+  isMobile,
+  metric,
+}) => {
+  return (
```
pnavarrc, line 20: I think we could make more clear if this component is responsible to know how to render each metric or receive parameters instead. 

- If we pass `metric`, then we could calculate `formattedValue`, `metricName`, `hasSubLabel` inside the component
- If we don't pass the `metric` prop (the props will be computed by the parent component), we could receive `metricSubLabelText` as an optional prop and avoiding having that logic in this component.

src/components/NewLocationPage/SummaryStatsBlock/MetricSubLabel.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+const MetricSubLabel: React.FC<{ text: string[]; isMobile: boolean }> = ({
+  text,
+  isMobile,
+}) => {
+  return (
+    <SubLabelWrapper>
+      <SubLabel>{text[0]}</SubLabel> {!isMobile && <br />}
+      <SubLabel>{text[1]}</SubLabel>
+    </SubLabelWrapper>
```
pnavarrc, line 17: How about making the text wrap and making the wrapper have a maximum fit instead of forcing the break? We can keep it as is if that's too much trouble!

src/components/NewLocationPage/SummaryStatsBlock/MetricValue.tsx
```diff
@@ -0,0 +1,18 @@
...
+  iconColor,
+}) => {
+  return (
+    <>
+      <ValueWrapper>
+        <CircleIcon $iconColor={iconColor} />
+        <Value>{value}</Value>
+      </ValueWrapper>
+    </>
+  );
```
pnavarrc, line 15: Maybe we can remove the wrapping fragment

src/components/NewLocationPage/SummaryStatsBlock/MobileSummaryStat.tsx
```diff
@@ -0,0 +1,39 @@
...
+            isMobile={isMobile}
+          />
+        )}
+      </Row>
+      <Row>
+        <MetricValue value={formattedValue} iconColor={levelInfo.color} />
+        <StyledChevron />
+      </Row>
+    </StatContent>
+  );
```
pnavarrc, line 36: Same here, maybe pass the sub label text as optional parameter from above

src/components/NewLocationPage/SummaryStatsBlock/SummaryStat.tsx
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <>
+      <MobileOnly>
+        <MobileSummaryStat {...statProps} />
+      </MobileOnly>
+      <DesktopOnly>
+        <DesktopSummaryStat {...statProps} />
+      </DesktopOnly>
+    </>
+  );
```
pnavarrc, line 38: an option could be

```tsx
const Component = isMobile 
  ? MobileSummaryStat 
  : DesktopSummaryStat;

return <Component {...statProps} />
```

src/components/NewLocationPage/SummaryStatsBlock/SummaryStatsBlock.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { Metric } from 'common/metricEnum';
+import { MetricValues } from 'common/models/Projections';
+import { orderedStatMetrics } from './utils';
+
+const SummaryStatsBlock: React.FC<{ stats: MetricValues }> = ({ stats }) => {
+  return (
+    <BlockWrapper>
+      {orderedStatMetrics.map((metric: Metric) => {
+        const value = stats[metric] as number;
+        return <SummaryStat metric={metric} value={value} />;
```
pnavarrc, line 13: We need to add a `key` prop here since we are rendering something in the body of a `map`

pnavarrc (approved):
Looks good!

PR #3424, opened 2021-04-23
Fix "days ago" display on chart

faihegberg:
This PR sets midnight as the baseline for date comparison. This is a WIP as I am still testing it further.


chasulin (approved):
👍🏼

PR #3434, opened 2021-04-24
improve getGeolocatedRegions

schamp:
this did an expensive search (scan through all counties to match by zip code) twice, this does it once and reuses the value, and uses a state code lookup table instead of a manual scan

### DIFF
--- a/src/common/regions/utils.ts
+++ b/src/common/regions/utils.ts
@@ -6,7 +6,14 @@ import values from 'lodash/values';
 import isNull from 'lodash/isNull';
 import without from 'lodash/without';
 import regions from './region_db';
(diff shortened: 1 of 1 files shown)


src/common/regions/utils.ts
```diff
@@ -192,10 +183,17 @@ export function getGeolocatedRegions(
   if (geolocation.country !== UNITED_STATES) {
     return null;
   } else {
+    const county = getCountyRegionFromZipCode(
+      geolocation.zipCode,
+      countyToZipMap,
```
schamp, line 188: this is going to help if we try getting zip code mappings by state or county on demand

mikelehen (approved):

PR #3437, opened 2021-04-24
make SearchAutocomplete and HomepageSearchAutocomplete match better

schamp:
These are pretty similar, but are hard to compare because of how they're organized.  put props in the same order, and do some things the same way that seem like they should be consistent.

### DIFF
--- a/src/components/Search/Homepage/HomepageSearchAutocomplete.tsx
+++ b/src/components/Search/Homepage/HomepageSearchAutocomplete.tsx
@@ -46,7 +46,7 @@ const HomepageSearchAutocomplete: React.FC<{
     if (menuOpen) setIsOpen(false);
   }, [menuOpen]);
 
-  const onInputChange = (e: any, value: string) => {
+  const onInputChange = (e: React.ChangeEvent<{}>, value: string) => {
     setInput(value);
     const isStringOfDigits = /^\d+$/.test(value);
     if (isStringOfDigits) {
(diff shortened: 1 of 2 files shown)


mikelehen (commented):
@chasulin can confirm, but I think we're going to be getting rid of `SearchAutocomplete` soon and unifying on `HomepageSearchAutocomplete`.  (This is included in a location page redesign that's in progress.)

Also, FWIW- I noticed that with this change you get some odd "contains the zip code xxx" text even if you search by a non-zipcode:
image

schamp:
> @chasulin can confirm, but I think we're going to be getting rid of `SearchAutocomplete` soon and unifying on `HomepageSearchAutocomplete`. (This is included in a location page redesign that's in progress.)
> 
> Also, FWIW- I noticed that with this change you get some odd "contains the zip code xxx" text even if you search by a non-zipcode:
> image

Ah thanks, didn't know about this plan.  I'll double check, if this is not the case (or is some time out) I'll fix this thing, else close.

schamp:
Looks like this bug is present in the live site: 
image

I think I see a fix though.  I can include it as a part of this, or just make it separate

schamp:
So the bug was that after entering a 5-digit zip code, if you switch the input to "not a string of digits" in one go (e.g., by selecting all and replacing text by typing or pasting) the "entering a zip code" flag wasn't reset to false.  So the fix was "always set that flag to false, unless you're entering a string of 5 digits.", which is in the latest commit here.

schamp:
(but, good find, thanks : )  it didn't turn up in my testing)

chasulin:
yep, thats right-everything will soon use `HomepageSearchAutocomplete`. but yikes, that zip code bug is on prod. thank you @schamp for fixing

chasulin (approved):
looks good + bug fixed! thank you 🙏🏼

schamp:
Thanks for the timely review :)

PR #3438, opened 2021-04-24
Add skip link to Homepage, Location pages, and all PageContent pages

schamp:
Pretty straightforward.  See https://web.dev/headings-and-landmarks/#bypass-repetitive-content-with-skip-links for explanation

### DIFF
--- a/src/components/PageContent/PageContent.tsx
+++ b/src/components/PageContent/PageContent.tsx
@@ -2,6 +2,7 @@ import React, { Fragment, useState } from 'react';
 import NavBar, { NavBarSearch } from 'components/NavBar';
 import SidebarContents from 'components/SidebarContents';
 import ShareModelBlock from 'components/ShareBlock/ShareModelBlock';
+import SkipLink from 'components/SkipLink/SkipLink';
 import { TocItem } from 'cms-content/utils';
 import {
   PageContainer,
@@ -22,6 +23,7 @@ const PageContent: React.FC<{ sidebarItems: TocItem[] }> = ({
 
   return (
     <Fragment>
+      <SkipLink />
       <NavBar
         renderSearch={() => <NavBarSearch menuOpen={menuOpen} />}
         renderSecondaryElement={() => <DesktopOnlyDonateButton />}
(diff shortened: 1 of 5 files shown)


schamp:
[[Accessibility] Implement skip links](https://trello.com/c/zqBLkcow/609-accessibility-implement-skip-links)

mikelehen (changes requested):
Wow, thanks for grabbing this.  I tested this with VoiceOver on mac and I think it's going to need some tweaking to be fully useful.

* On the homepage, right now it scrolls to the map but the map is not currently accessible to non-sighted users (we intentionally remove all the state links from the focus tree by [setting tabIndex to -1]()https://github.com/covid-projections/covid-projections/blob/5a20f15fc70fecc6ef98719bf48a9b57a3b6ea8b/src/components/Map/USACountyMap.tsx#L95.  And so the page scrolls but the skip nav link is still the selected/focused item and so you still have to navigate through the page to get anywhere useful.  I think we would instead want to send them to either the banner or the search bar.  Probably the banner?  I think in general, just skipping the nav bar makes sense.
* On the location page, nothing scrolls and the skip link remains the active/selected item after you click it, so again they'd need to select through the nav bar before they got anywhere useful.  Not 100% sure why that's happening.  Maybe moving the `#main` id closer to the actual content (e.g. the h1 with the location name) would help? 
* On PageContent pages (e.g. https://covid-projections-git-feature-skip-links-covidactnow.vercel.app/about) it seems to successfully navigate to the first header and automatically read it to you ("About Covid Act Now"). So I think that's working as desired.

Not sure if you are on a mac, but if you are and are willing to dig deeper into this, here's a 12min tutorial on VoiceOver: https://www.youtube.com/watch?v=5R-6WvAihms. 😁  Else, maybe @pnavarrc , @chasulin , or I can pick this up this week and play with the homepage / location page experience.  Thanks again for getting a jump on this!

schamp:
Awesome, thanks for the details.  I was testing using the keyboard / tab
key but I'll spend some time getting to know voiceover.

schamp:
So I tried moving things around on the location page, and unfortunately it scrolls underneath the search/nav header : /

So I'll be poking around on this a bit more.  If you think it's worth separating out the `PageContent` bit and landing it separately, we can do that, or just wait until we've figured it out everywhere.

schamp:
bailing on this one for now : (

PR #3446, opened 2021-04-26
Compare highlight fix

chasulin:
[Reorganizing compare row elements in this PR](https://github.com/covid-projections/covid-projections/pull/3408) made it so that the location name was no longer turning blue when hovering the row ([see here](https://covidactnow.org/?s=1783703) on prod when hovering a compare row). This fixes + also reorders some imports


pnavarrc (approved):
Looks good!

PR #3451, opened 2021-04-26
adjust labels of the vaccination chart

pnavarrc:
**Desktop**
image

**Mobile**

### DIFF
--- a/src/components/Charts/Charts.style.ts
+++ b/src/components/Charts/Charts.style.ts
@@ -194,12 +194,28 @@ export const LegendLabel = styled.span`
 `;
 
 // Vaccination Label
-export const VaccinationLabel = styled.text.attrs(props => ({ dx: 8, dy: 7 }))`
-  font-family: Roboto;
-  font-size: 13px;
-  fill: ${props => palette(props).annotation || '#4f4f4f'};
+export const VaccinationLabel = styled.g`
+  rect {
+    fill: white;
+  }
+  text {
+    font-family: Roboto;
+    font-size: 13px;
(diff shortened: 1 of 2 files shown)


src/components/Charts/Charts.style.ts
```diff
@@ -194,12 +194,28 @@ export const LegendLabel = styled.span`
...
-export const VaccinationLabelBold = styled.tspan`
-  font-weight: bold;
+export const VaccinationLabelBold = styled.g`
+  rect {
+    fill: white;
+  }
+  text {
+    font-family: Roboto;
+    font-size: 13px;
+    font-weight: bold;
```
chasulin, line 217: can use ${props => props.theme.fonts.regularBookBold}

src/components/Charts/Charts.style.ts
```diff
@@ -194,12 +194,28 @@ export const LegendLabel = styled.span`
...
-export const VaccinationLabel = styled.text.attrs(props => ({ dx: 8, dy: 7 }))`
-  font-family: Roboto;
-  font-size: 13px;
-  fill: ${props => palette(props).annotation || '#4f4f4f'};
+export const VaccinationLabel = styled.g`
+  rect {
+    fill: white;
+  }
+  text {
+    font-family: Roboto;
```
chasulin, line 202: nit- can use ${props => props.theme.fonts.regularBook}

chasulin (approved):
looks good. mobile might be slightly confusing because the fully vaccinated label sits on the 1st dose line, but this is prob the best solution for now

PR #3456, opened 2021-04-27
Create location name blocks

faihegberg:
This PR creates the location name block for the new location page header.

### DIFF
--- a/src/components/NewLocationPage/LocationName/LocationName.style.tsx
+++ b/src/components/NewLocationPage/LocationName/LocationName.style.tsx
@@ -1,19 +1,21 @@
 import styled from 'styled-components';
 import { materialSMBreakpoint } from 'assets/theme/sizes';
 
-export const RegionContainer = styled.div`
-  @media (max-width: ${materialSMBreakpoint}) {
-    text-align: center;
+export const RegionNameContainer = styled.div`
+  text-align: center;
+  @media (min-width: ${materialSMBreakpoint}) {
+    text-align: left;
   }
 `;
 
-export const RegionNameText = styled.div`
+export const RegionNameText = styled.h1`
+  font-weight: normal;
   font-size: 2rem;
-  line-height: 128%;
+  line-height: 1.3;
+  margin: 0.1rem 0;
 `;
 
 export const MultiStateText = styled.div`
-  margin-top: 0.13rem;
   font-size: 1rem;
   font-weight: 500;
 `;
--- a/src/components/NewLocationPage/LocationName/LocationName.style.tsx
+++ b/src/components/NewLocationPage/LocationName/LocationName.style.tsx
@@ -1,0 +1,19 @@
+import styled from 'styled-components';
+import { materialSMBreakpoint } from 'assets/theme/sizes';
+
+export const RegionContainer = styled.div`
+  @media (max-width: ${materialSMBreakpoint}) {
+    text-align: center;
+  }
+`;
+
+export const RegionNameText = styled.div`
+  font-size: 2rem;
+  line-height: 128%;
+`;
+
+export const MultiStateText = styled.div`
+  margin-top: 0.13rem;
+  font-size: 1rem;
+  font-weight: 500;
+`;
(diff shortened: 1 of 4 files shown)


src/components/NewLocationPage/LocationName/LocationName.style.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export const RegionContainer = styled.div`
+  @media (max-width: ${materialSMBreakpoint}) {
+    text-align: center;
+  }
+`;
+
+export const RegionNameText = styled.div`
+  font-size: 2rem;
+  line-height: 128%;
```
chasulin, line 12: you can round and call this 1.3

src/components/NewLocationPage/LocationName/LocationName.style.tsx
```diff
@@ -0,0 +1,19 @@
+import styled from 'styled-components';
+import { materialSMBreakpoint } from 'assets/theme/sizes';
+
+export const RegionContainer = styled.div`
+  @media (max-width: ${materialSMBreakpoint}) {
```
chasulin, line 5: lets keep with the current format for breakpoints and use `min-width` instead of `max-width`. so:

```
export const RegionContainer = styled.div`
  text-align: center;
  @media (min-width: ${materialSMBreakpoint}) {
    text-align: left;
  }
```
`;

src/components/NewLocationPage/LocationName/LocationName.style.tsx
```diff
@@ -0,0 +1,19 @@
...
+  }
+`;
+
+export const RegionNameText = styled.div`
+  font-size: 2rem;
+  line-height: 128%;
+`;
+
+export const MultiStateText = styled.div`
+  margin-top: 0.13rem;
```
chasulin, line 16: ditto here- can round and call this `.1rem`

src/components/NewLocationPage/LocationName/LocationName.style.tsx
```diff
@@ -0,0 +1,19 @@
+import styled from 'styled-components';
+import { materialSMBreakpoint } from 'assets/theme/sizes';
+
+export const RegionContainer = styled.div`
+  @media (max-width: ${materialSMBreakpoint}) {
+    text-align: center;
+  }
+`;
+
+export const RegionNameText = styled.div`
```
chasulin, line 10: we want to preserve the hierarchy of the page for accessibility purposes (and i think SEO purposes?). in this case, that would mean using the `h1` element for the location name (rather than a `div`). from an accessibility standpoint, this lets assistive technologies know that the location name is the main header of the page

src/components/NewLocationPage/LocationName/LocationName.tsx
```diff
@@ -0,0 +1,57 @@
...
+import {
+  RegionContainer,
+  RegionNameText,
+  MultiStateText,
+} from './LocationName.style';
+
+const LocationName: React.FC<{ region: Region }> = ({ region }) => {
+  if (region instanceof State) {
+    return (
+      <RegionContainer>
```
chasulin, line 13: maybe `RegionNameContainer` instead of `RegionContainer`

src/components/NewLocationPage/LocationName/utils.ts
```diff
@@ -0,0 +1,22 @@
+import { State, County, MetroArea } from 'common/regions';
+
+export function inMultipleStates(region: MetroArea): boolean {
+  return region.stateCodes.includes('-');
+}
+
+export function getRegionName(region: State | County | MetroArea) {
```
chasulin, line 7: this type can just be `Region`, which encompasses all three. you can import it from `common/regions`

src/components/NewLocationPage/LocationName/utils.ts
```diff
@@ -0,0 +1,22 @@
+import { State, County, MetroArea } from 'common/regions';
+
+export function inMultipleStates(region: MetroArea): boolean {
```
chasulin, line 3: maybe something like `isMultiStateMetro` would be more explicit

chasulin (approved):

PR #3463, opened 2021-04-28
Move button inside of NotesBlock container

chasulin:
This moves `StyledBaseButton` inside of `SectionContainer` so `SectionContainer` is always the parent div of each location page module above the fold

(not yet on prod, just storybook- Location page redesign/Notes Block)

### DIFF
--- a/src/components/NewLocationPage/NotesBlock/NotesBlock.tsx
+++ b/src/components/NewLocationPage/NotesBlock/NotesBlock.tsx
@@ -6,22 +6,45 @@ import {
   IconWrapper,
   TextContainer,
   TextComponent,
+  StyledBaseButton,
 } from './NotesBlock.style';
+import { EventAction, EventCategory } from 'components/Analytics';
 
-const NotesBlock: React.FC<{ icon: any; title: string }> = ({
+interface NotesProps {
+  icon: any;
+  title: string;
+  redirectUrl?: string;
+  trackingCategory?: EventCategory;
+}
+
+const NotesBlock: React.FC<NotesProps> = ({
   icon,
   title,
+  redirectUrl,
+  trackingCategory = EventCategory.NONE,
   children,
 }) => {
+  // If there is no href/to/onClick prop passed,
+  // BaseButton knows to render as a span and not a button
+  const buttonProps = redirectUrl ? { href: redirectUrl } : {};
+
   return (
     <SectionContainer>
-      <SectionContentContainer>
-        <IconWrapper>{icon}</IconWrapper>
-        <TextContainer>
-          <LabelWithChevron text={title} />
-          <TextComponent>{children}</TextComponent>
-        </TextContainer>
-      </SectionContentContainer>
+      <StyledBaseButton
+        trackingCategory={trackingCategory}
+        trackingAction={EventAction.CLICK}
+        trackingLabel={`Header notes block: ${title}`}
+        style={{ cursor: redirectUrl ? 'pointer' : 'default' }}
+        {...buttonProps}
+      >
+        <SectionContentContainer>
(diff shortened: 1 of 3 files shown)


src/components/NewLocationPage/NotesBlock/NotesBlock.tsx
```diff
@@ -6,22 +6,45 @@ import {
   IconWrapper,
   TextContainer,
   TextComponent,
+  StyledBaseButton,
 } from './NotesBlock.style';
+import { EventAction, EventCategory } from 'components/Analytics';
 
-const NotesBlock: React.FC<{ icon: any; title: string }> = ({
+interface NotesProps {
+  icon: any;
```
mikelehen, line 14: Can we find the correct type? 😁

mikelehen (approved):

src/components/NewLocationPage/NotesBlock/NotesBlock.tsx
```diff
@@ -6,22 +6,45 @@ import {
...
+  title: string;
+  redirectUrl?: string;
+  trackingCategory?: EventCategory;
+}
+
+const NotesBlock: React.FC<NotesProps> = ({
   icon,
   title,
+  redirectUrl,
+  trackingCategory = EventCategory.NONE,
```
mikelehen, line 24: Also curious why we have a default of NONE here?  Would it be better to require it to always be specified?  or is NONE actually a good default?

src/components/NewLocationPage/NotesBlock/NotesBlock.tsx
mikelehen: `JSX.Element` perhaps?

src/components/NewLocationPage/NotesBlock/NotesBlock.tsx
chasulin: hmm, maybe a generic location page header category as a backup? if the note were clickable, there _should_ be an event category passed in. `NONE` was just a fallback for if the note is ever just text and not clickable (i dont know if this would ever be the case?) 

im gonna leave it as `NONE` for now and revisit when i go over analytics for the new components tomorrow

src/components/NewLocationPage/NotesBlock/NotesBlock.tsx
chasulin: haha typescript is making me unhappy with this one. adding a todo and will be sure to investigate before 🚀

src/components/NewLocationPage/NotesBlock/NotesBlock.tsx
mikelehen: Oh, gotcha.  Another option would be to do something like

```
assert(trackingCategory || !redirectUrl, 'Must supply a trackingCategory if this is a link.')
```

but cool to revisit later. 👍

src/components/NewLocationPage/NotesBlock/NotesBlock.tsx
chasulin: oh whoops--didnt refresh and see your 2nd comment before replying. i think `JSX.Element` does the trick 💫🙏🏼

mikelehen: 👍  

FWIW- To find the type, I just added:

`const foo = (<VulnerabilityIcon />);`

and then hovered over `foo` in VSCode to see what type it said it was.  Not sure if this always works 100% (I suspect it may give you a more complicated type than necessary sometimes), but it can be helpful.

chasulin: oh sweet

PR #3469, opened 2021-04-29
Add updated date to location name header

faihegberg:
This PR adds an "Updated on <DATE>" component below the region name in the new location page header.

We also add a new date-time format to enable usage of the new time util (time formatting as "8 AM PDT", which will be displayed in the tooltip.

All tooltip links site wide are also updated to have a dotted line underneath, not a dashed line underneath, as per new design.

### DIFF
--- a/src/components/NewLocationPage/LocationName/LocationName.style.tsx
+++ b/src/components/NewLocationPage/LocationName/LocationName.style.tsx
@@ -23,7 +23,6 @@ export const MultiStateText = styled.div`
 
 export const UpdatedOnText = styled.h2`
   font-size: 1rem;
-  text-transform: uppercase;
   line-height: 1;
   margin: 0.25rem 0;
   color: ${COLOR_MAP.GREY_4};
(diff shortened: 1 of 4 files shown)


src/components/NewLocationPage/LocationName/LocationName.style.tsx
```diff
@@ -19,3 +20,11 @@ export const MultiStateText = styled.div`
   font-size: 1rem;
   font-weight: 500;
 `;
+
+export const UpdatedOnText = styled.h2`
```
chasulin, line 24: i think this can be a `span` and not an `h2`

src/components/NewLocationPage/LocationName/LocationName.style.tsx
```diff
@@ -19,3 +20,11 @@ export const MultiStateText = styled.div`
...
+  font-size: 1rem;
+  text-transform: uppercase;
+  line-height: 1;
+  margin: 0.25rem 0;
```
chasulin, line 28: i think we can drop the bottom margin (and maybe bump it up a tiny bit), so `margin-top: .5rem`

chasulin (approved):

PR #3496, opened 2021-05-03
Add timeout to correct scrolling position

faihegberg:
This PR adds timeout to correct the scrolling position. Unfortunately, I also tried 700, 800, and 900 milliseconds. They did not give reliable scroll behaviour. Also tried other scrolling methods (e.g. `scrollIntoView`, `scrollTo`, etc.) and the incorrect scrolling persisted. Adjusting the timeout to 1000 milliseconds seems to be the only reliable fix so far.


chasulin:
[Learn links with anchor tags don't reliably scroll to correct position.](https://trello.com/c/yvC6j6Jz/1303-learn-links-with-anchor-tags-dont-reliably-scroll-to-correct-position)

chasulin (approved):
looks like it works!

PR #3499, opened 2021-05-03
Above the fold: move fetching of projections

chasulin:
Makes it so rendering above the fold doesn't depend entirely on projections


pnavarrc (approved):
Looks good!

PR #3501, opened 2021-05-04
Stop input field from toggling open and closed

faihegberg:
This PR adds a condition to check whether the user clicks outside the location search dropdown on the homepage. This enables us to differentiate when to close the dropdown and when to leave it open. The dropdown should only close if the user clicks outside the dropdown.

Not sure whether this is the ideal solution but `Autocomplete` seems to generically recognize a click (whether inside or outside the dropdown) as an `onClose`. Hence the explicit check for a `placeholder` attribute, which is unique to the dropdown.

### DIFF
--- a/src/components/Search/Homepage/HomepageSearchAutocomplete.tsx
+++ b/src/components/Search/Homepage/HomepageSearchAutocomplete.tsx
@@ -147,9 +147,13 @@ const HomepageSearchAutocomplete: React.FC<{
             }
           }}
           onClose={() => {
-            setIsOpen(false);
-            if (setHideMapToggle) {
-              setHideMapToggle(false);
+            // Only close the dropdown if the user clicks outside the dropdown
+            // (i.e. Element in focus will not have a 'placeholder' attribute).
+            if (!document.activeElement?.hasAttribute('placeholder')) {
+              setIsOpen(false);
+              if (setHideMapToggle) {
+                setHideMapToggle(false);
+              }
             }
           }}
           PaperComponent={StyledPaper}


chasulin (approved):
lgtm!

src/components/Search/Homepage/HomepageSearchAutocomplete.tsx
```diff
@@ -147,9 +147,13 @@ const HomepageSearchAutocomplete: React.FC<{
...
-            setIsOpen(false);
-            if (setHideMapToggle) {
-              setHideMapToggle(false);
+            // Only close the dropdown if the user clicks outside the dropdown
+            // (i.e. Element in focus will not have a 'placeholder' attribute).
+            if (!document.activeElement?.hasAttribute('placeholder')) {
+              setIsOpen(false);
+              if (setHideMapToggle) {
+                setHideMapToggle(false);
+              }
```
pnavarrc, line 156: How about giving the input element an ID and use that ID to check if the input element is the active one? I feel like that would be more robust, since the user could click on a different element with a placeholder or the input element might not have one necessarily

```tsx
<div style={{ display: 'flex', alignItems: 'center' }}>
  <StyledTextField
    placeholder="City, county, state, or zip"
    {...params}
    className={searchTextFieldStyles.root}
    $isOpen={isOpen}
    InputProps={{
      ...params.InputProps,
      id={generatedID}
      endAdornment: <SearchBarIcon />,
    }}
  />
  <Hidden smUp>
  <CloseIcon
    onClick={() => setIsOpen(false)}
    $showIcon={isOpen}
  />
  </Hidden>
</div>
// ...

onClose={() => (
  if (document.activeElement?.attribute('id') !== generatedID) {
    // ...
  }
)}

```

faihegberg: Unfortunately, if we generate an ID, it will still not override the generated ID from Material UI. This mismatch will prevent us from comparing whether the focused element is the same element which should be open/closed. After some discussion, we decided to merge this PR as is.

PR #3507, opened 2021-05-04
Spark lines selecting correct explore tab

chasulin:
- Uses ref instead of hash link + maps `SparkLineMetric` to `ExploreMetric`
- Fixes alignment of chevron icons in spark line and summary stat labels
- Makes vax progress bar clickable
- Remove grey circle icon from percent vaccinated summary stat
- Fixes line height of notes block title
- Drops font weight of overall risk level label to 1.4 rem on iphone5

### DIFF
--- a/src/components/LocationPage/ChartsHolder.tsx
+++ b/src/components/LocationPage/ChartsHolder.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useEffect } from 'react';
+import React, { useRef, useEffect, useState } from 'react';
 import { useLocation } from 'react-router-dom';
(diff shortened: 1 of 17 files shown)


pnavarrc:
## Feedback
- The focus outline in the metric charts is visible when clicking on it and persists after clicking
- The past 30 days block doesn't show initially, and the boxes for the spark lines start out narrow and expand when the projection data arrives
- It seems that the spark line charts are re-rendering on scroll, maybe this is caused by the use of `ParentSize` maybe we can memoize the charts to prevent that from happening

PR #3509, opened 2021-05-04
Make sure explore doesn't reset the locations list on every Homepage render.

mikelehen:
Basic issue was that `initialFipsList` property was being recomputed to a different (but identical) Array on every HomePage render and causing the explore chart component to think it had changed (since React only does shallow
comparisons it doesn't matter that the Array was identical).

* Removes no-longer-needed code for showing hospitalizations in the homepage explore chart, since it was causing `initialFipsList` to change on every render.
* Use `useMemo()` to memoize `exploreGeoLocations` to make sure it doesn't change on every render.
* Turns `getLargestMetroFipsForExplore()` function into a `largestMetroFipsForExplore` constant so that the resulting Array is `==` according to React.
* Also added a diagnostic useLogChanges() hook that can be used for future debugging.


pnavarrc (approved):
Looks good!

PR #3511, opened 2021-05-04
Some more explore cleanup which prevents it from re-rendering at all on homepage scroll.

mikelehen:
This builds on #3509 with some non-essential additional cleanup / optimization.

* Remove no-longer-needed `initialChartIndigenousPopulations` prop.
* Change `nationalSummaryText` to a boolean instead of a whole JSX.Element.


pnavarrc (approved):
Thanks for cleaning up!

PR #3512, opened 2021-05-04
Fix bolding of location name edge cases

faihegberg:
This PR addresses issues of location name bolding for the new location page. It implements a check to see whether a region contains a suffix which is likely to be more than one word (e.g. "City and Borough", "Census Area"). If yes, then it will bold the whole name and suffix, for the following reasons:
- There is a high chance the above examples are not the only possible multi-word suffixes out there. Better to bold the whole name and suffix than guess where the name ends and the suffix starts and potentially bold in the wrong place.
- Most multi-word suffixes have the word "and" or "census" as the second last word.

For regions where the last word is not a recognizable suffix (e.g. "District of **Columbia**"), also bold the whole name.

Implementing the above logic after reading "County Variations" in https://en.wikipedia.org/wiki/County_(United_States) 😅

### DIFF
--- a/src/components/NewLocationPage/LocationName/utils.ts
+++ b/src/components/NewLocationPage/LocationName/utils.ts
@@ -1,26 +1,19 @@
 import { Region, State, County, MetroArea } from 'common/regions';
 
-const regionSuffixes = new Set([
-  'metro',
-  'city',
-  'county',
-  'parish',
-  'area',
-  'borough',
-]);
+const regionSuffixes = ['metro', 'city', 'county', 'parish', 'area', 'borough'];
 
 /**
  * Common indicators of regions with multiple-word suffixes
  * include region names with 'and' and 'census'
  * (e.g. 'City and Borough', 'Census Area').
  */
-const multipleWordSuffixIndicator = new Set(['and', 'census']);
+const multipleWordSuffixIndicator = ['and', 'census'];
 
 export function isMultiStateMetro(region: MetroArea): boolean {
   return region.stateCodes.includes('-');
 }
 
-export function getRegionName(region: Region) {
+export function getSplitRegionName(region: Region) {
   if (region instanceof State) {
     return [region.name];
   } else if (region instanceof County) {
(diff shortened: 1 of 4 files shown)


src/components/NewLocationPage/LocationName/utils.ts
```diff
@@ -1,5 +1,21 @@
 import { Region, State, County, MetroArea } from 'common/regions';
 
+const regionSuffixes = new Set([
```
chasulin, line 3: this can just be an array (ditto to `multipleWordSuffixIndicator` below)

const regionSuffixes = [
  'metro',
  'city',
  'county',
  'parish',
  'area',
  'borough',
]

src/components/SharedComponents/StyledRegionName/StyledRegionName.tsx
```diff
@@ -3,21 +3,27 @@
...
   return (
     <Wrapper $truncateText={truncateText}>
-      {regionNameMain} {regionSuffix && <Suffix>{regionSuffix}</Suffix>}
+      {regionNameMain}
+      {regionSuffix && (
+        <Suffix>
+          {regionSuffix !== 'metro' ? ` ${regionSuffix}` : ` ${regionSuffix},`}
+        </Suffix>
+      )}
+      {!regionSuffix && region.regionType !== 'state' && <Suffix>,</Suffix>}
```
chasulin, line 26: when does this return true (not a state but doesnt have a suffix)?

faihegberg: Added comments to clarify :)

chasulin:
we can simplify the logic of `splitRegionName`, since there are only 2 clauses:
1) the suffix is one word, in which case we want to split it out and style it differently
2) every other scenario, in which case the entire name is style the same and we don't need to split it up

```
/**
 * We only split the region name into [main, suffix] and apply unique
 * styles to each if it is a single word suffix (ie. 'County').
 *
 * We don't split the region name for regions with multiple-word suffix (ie. 'Census Area', 'City and Borough'),
 * as it is unknown how many words exactly make up a multiple-word suffix.
 * In this case, we style the whole region name the same.
 */
function splitRegionName(regionName: string) {
  const splitRegion = regionName.split(' ');

  if (
    regionSuffixes.includes(splitRegion[splitRegion.length - 1].toLowerCase())
  ) {
    const suffix = splitRegion.pop();
    const regionNameMain = splitRegion.join(' ');
    return [regionNameMain, suffix];
  } else {
    const regionNameMain = splitRegion.join(' ');
    return [regionNameMain];
  }
}
```

faihegberg:
> /**
>  * We only split the region name into [main, suffix] and apply unique
>  * styles to each if it is a single word suffix (ie. 'County').
>  *
>  * We don't split the region name for regions with multiple-word suffix (ie. 'Census Area', 'City and Borough'),
>  * as it is unknown how many words exactly make up a multiple-word suffix.
>  * In this case, we style the whole region name the same.
>  */

This is so much cleaner! Thank you 😀

PR #3513, opened 2021-05-04
Spark lines maxY update

chasulin:
Uses maxY of the smoothed series only (rather than overall maxY of smoothed and raw)

### DIFF
--- a/src/components/NewLocationPage/SparkLineBlock/utils.ts
+++ b/src/components/NewLocationPage/SparkLineBlock/utils.ts
@@ -101,16 +101,10 @@ export function getSparkLineSeriesFromProjection(
   }));
 }
 
-function getMaxY(series: Column[]) {
-  return d3Max(series, (d: Column) => d.y);
-}
-
-export function getOverallMaxY(seriesA: Column[], seriesB: Column[]) {
-  const maxA = getMaxY(seriesA);
-  const maxB = getMaxY(seriesB);
-  assert(maxA, 'Maximum value unexpectedly not found'); // theres probably a better way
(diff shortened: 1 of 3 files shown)


src/components/NewLocationPage/SparkLineBlock/utils.ts
```diff
@@ -101,16 +101,10 @@ export function getSparkLineSeriesFromProjection(
...
-
-export function getOverallMaxY(seriesA: Column[], seriesB: Column[]) {
-  const maxA = getMaxY(seriesA);
-  const maxB = getMaxY(seriesB);
-  assert(maxA, 'Maximum value unexpectedly not found'); // theres probably a better way
-  assert(maxB, 'Maximum value unexpectedly not found');
-  return Math.max(maxA, maxB);
+export function getMaxY(series: Column[]) {
+  const maxY = d3Max(series, (d: Column) => d.y);
+  assert(maxY, 'Maximum value unexpectedly not found'); // theres probably a better way
```
pnavarrc, line 106: If `maxY` is `0` the assertion will throw, maybe you can use `Number.isFinite` instead?

```ts
assert(Number.isFinite(maxY), '...');
```

pnavarrc (approved):
Added a note, consider adding an error boundary around the SparkLine component so if the assertion throws, only the component will break and not the entire page

src/components/NewLocationPage/SparkLineBlock/utils.ts
chasulin: removed the assertion and added an error boundary

PR #3530, opened 2021-05-06
Explore chart label tweaks.

mikelehen:
* Have overlapping labels move upwards instead of down.
* Ensure we don't show a label for empty series (was showing up in the upper left at `(0, 0)`).

Fixes: 
* https://trello.com/c/Mgz6dgdr/1242-explore-chart-adjust-label-adjustment-function-to-go-up
* https://trello.com/c/ICgj0iIv/1322-imagepng

**Before:**
image

**After:**
image

### DIFF
--- a/src/components/Explore/DateMarker.tsx
+++ b/src/components/Explore/DateMarker.tsx
@@ -15,9 +15,9 @@ const DateMarker: React.FC<{
   seriesList: Series[];
   date: Date;
 }> = ({ left, seriesList, date }) => {
-  const [, seriesSmooth] = seriesList;
-  const pointSmooth = findPointByDate(seriesSmooth.data, date);
-  const dateToCompareTo = pointSmooth ? getColumnDate(pointSmooth) : date;
+  const series = seriesList[0];
+  const point = findPointByDate(series.data, date);
+  const dateToCompareTo = point ? getColumnDate(point) : date;
   const today = new Date();
   today.setHours(0, 0, 0, 0);
   return today < dateToCompareTo ? null : (
--- a/src/components/Explore/MultipleLocationsChart.tsx
+++ b/src/components/Explore/MultipleLocationsChart.tsx
@@ -1,7 +1,7 @@
 import React, { useCallback, Fragment } from 'react';
 import isNumber from 'lodash/isNumber';
 import last from 'lodash/last';
-import max from 'lodash/max';
+import min from 'lodash/min';
 import sortBy from 'lodash/sortBy';
 import { Group } from '@vx/group';
 import { scaleUtc, scaleLinear } from '@vx/scale';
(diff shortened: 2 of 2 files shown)


src/components/Explore/MultipleLocationsChart.tsx
```diff
@@ -133,7 +133,9 @@ const MultipleLocationsChart: React.FC<{
   barOpacity,
   isMobileXs = false,
 }) => {
-  const seriesList = sortSeriesByLast(unsortedSeriesList);
+  const seriesList = sortSeriesByLast(unsortedSeriesList).filter(
+    series => series.data.length > 0,
+  );
```
mikelehen, line 138: This filter ensures we don't include empty series and show a label in the upper left of the chart.

src/components/Explore/DateMarker.tsx
```diff
@@ -15,9 +15,9 @@ const DateMarker: React.FC<{
   seriesList: Series[];
   date: Date;
 }> = ({ left, seriesList, date }) => {
-  const [, seriesSmooth] = seriesList;
-  const pointSmooth = findPointByDate(seriesSmooth.data, date);
-  const dateToCompareTo = pointSmooth ? getColumnDate(pointSmooth) : date;
+  const series = seriesList[0];
```
mikelehen, line 18: I'm not sure why we were using the second (`seriesSmooth`) series before.  All of the series should have their dates aligned, right?  Else the chart is going to be "wrong".  In any case I had to change it to use the first series because when I filtered out empty series, we could end up with a `seriesList` of length 1 after the filtering.

src/components/Explore/MultipleLocationsChart.tsx
```diff
@@ -263,15 +265,19 @@ function formatCurrentValueLabels(
...
     return [
       ...labelInfoList,
       {
-        x: lastPoint?.x ? getXPosition(lastPoint) : 0,
-        y: lastPoint?.y ? Math.max(getYPosition(lastPoint), currMaxY + 14) : 0,
+        x: lastPoint?.x != null ? getXPosition(lastPoint) : 0,
+        y:
+          lastPoint?.y != null
+            ? Math.min(getYPosition(lastPoint), currMinY - 14)
+            : 0,
```
mikelehen, line 280: Had to add `!= null` because having a y value of `0` is expected and we want to position it normally.

mikelehen:
[Explore: The location label shows on the left where the location doesn't have data for the selected variable](https://trello.com/c/ICgj0iIv/1322-explore-the-location-label-shows-on-the-left-where-the-location-doesnt-have-data-for-the-selected-variable)

mikelehen:
[Explore chart: Adjust label adjustment function to go up](https://trello.com/c/Mgz6dgdr/1242-explore-chart-adjust-label-adjustment-function-to-go-up)

src/components/Explore/DateMarker.tsx
pnavarrc: that works, I think I took the smoothed version because in my mind it was less likely to have missing data (but in reality, both series should have the dates we need)

pnavarrc (approved):
Looks good! Thanks for implementing the fixes! 🙌

PR #3537, opened 2021-05-06
Get location summaries synchronously from the main bundle if possible.

mikelehen:
Fixes https://trello.com/c/Mu3twLii/1306-fix-issue-where-the-screen-flashes-our-bottom-banner-when-navigating-to-location-pages-from-homepage

Because we had a hook calling a hook, setting a state variable, I think we had at least 2 or 3 async roundtrips before the location page header could render, which was probably enough time for some other work to get queued onto the JS event loop, bogging things down enough that there was ~1/2 second of time where the footer rendered but not the location page header... at least for me. 🤷

### DIFF
--- a/src/common/hooks/useLocationSummariesForFips.tsx
+++ b/src/common/hooks/useLocationSummariesForFips.tsx
@@ -1,16 +1,6 @@
-import { useEffect, useState } from 'react';
-import { LocationSummary, useSummaries } from 'common/location_summaries';
+import { useSummaries } from 'common/location_summaries';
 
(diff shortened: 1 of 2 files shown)


schamp:
Nice.  Kind of complicated, but looks like it works, so it's worth it.  You tested snapshot override works, I presume?

schamp (approved):
Not used to the review cycle.  Looks good : )

mikelehen:
Thanks!  Yep, snapshot override still works (including the footer flash 😁 ).

PR #3571, opened 2021-05-11
Add "After unsubscribe" page to solicit people to sign up for new project announcements.

mikelehen:
* Add an "after unsubscribe" page that we'll tell Campaign Monitor to use when people unsubscribe from daily download or research rundown.
* Add an "After subscribe" page since the default campaign monitor one has a "click here to return to the site" link that just takes you back to our "after unsubscribe" page instead of the homepage.
* Make the alert unsubscribe button also go to the "after unsubscribe" page.
* Fix a bug (in a couple places) where we weren't properly escaping email addresses in links, leading to e.g. email addresses containing `+` not working properly.

### DIFF
--- a/scripts/alert_emails/utils.ts
+++ b/scripts/alert_emails/utils.ts
@@ -90,8 +90,10 @@ function generateAlertEmailContent(
     last_updated: lastUpdated,
     location_url: `${locationURL}?utm_source=risk_alerts&utm_medium=email`,
     subscribe_link: `https://covidactnow.org/alert_signup?utm_campaign=risk_alert_forward_subscribe&utm_source=risk_alerts&utm_medium=email`,
-    unsubscribe_link: `${unsubscribeURL}?email=${encodeURI(emailAddress)}`, // would be nice to know dev/staging/prod
-    feedback_subject_line: encodeURI(
+    unsubscribe_link: `${unsubscribeURL}?email=${encodeURIComponent(
+      emailAddress,
+    )}`, // would be nice to know dev/staging/prod
+    feedback_subject_line: encodeURIComponent(
       `[Alert Feedback] Alert for ${locationName} on ${lastUpdated}`,
     ),
     disclaimer,
--- a/src/screens/Subscriptions/AfterSubscribe.tsx
+++ b/src/screens/Subscriptions/AfterSubscribe.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { Link } from 'react-router-dom';
 import { Wrapper, Header, BodyCopy } from './Subscriptions.style';
 
 const AfterSubscribe = () => {
@@ -11,7 +12,7 @@ const AfterSubscribe = () => {
           and will hear from us soon.
         </p>
         <p>
-          <a href="/">Click here</a> to return to the site.
+          <Link to="/">Click here</Link> to return to the site.
         </p>
       </BodyCopy>
     </Wrapper>
--- a/src/screens/Subscriptions/AfterSubscribe.tsx
+++ b/src/screens/Subscriptions/AfterSubscribe.tsx
@@ -1,0 +1,21 @@
+import React from 'react';
+import { Wrapper, Header, BodyCopy } from './Subscriptions.style';
+
+const AfterSubscribe = () => {
+  return (
+    <Wrapper>
+      <Header>Thank you</Header>
+      <BodyCopy>
+        <p>
+          Your subscription has been confirmed. You've been added to our list
+          and will hear from us soon.
+        </p>
+        <p>
+          <a href="/">Click here</a> to return to the site.
+        </p>
+      </BodyCopy>
+    </Wrapper>
+  );
+};
+
+export default AfterSubscribe;
--- a/src/screens/Subscriptions/AfterUnsubscribe.tsx
+++ b/src/screens/Subscriptions/AfterUnsubscribe.tsx
@@ -1,0 +1,94 @@
+import React, { createRef } from 'react';
+import { Wrapper, Header, BodyCopy } from './Subscriptions.style';
+import { EventAction, EventCategory } from 'components/Analytics';
+import ExternalLink from 'components/ExternalLink';
+import { LargeFilledButton } from 'components/ButtonSystem';
+
+const AfterUnsubscribe = () => {
+  const params = new URLSearchParams(window.location.search);
+  const email = params.get('email') || '';
+  const formRef = createRef<HTMLFormElement>();
+
+  const formAction = getFormActionLink(email);
+
+  const onSubscribeClick = async () => {
+    if (formRef?.current) {
+      formRef.current.action = await formAction;
+      formRef.current.submit();
+    }
+  };
+
+  return (
+    <Wrapper>
+      <Header>One last thing...</Header>
+      <BodyCopy>
+        <p>
+          Hi there! It looks like you're leaving us. We're sorry to see you go
+          and wanted to thank you for being a subscriber and supporter of Covid
+          Act Now.
+        </p>
+        <p>
+          Our organization will be working on some exciting new projects in the
+          future. Our mission is the same: to help people make informed
+          decisions by focusing on{' '}
+          <ExternalLink href="https://covidactnow.org/about">
+            data, transparency and accessibility
+          </ExternalLink>
+          .
+        </p>
+        <p>
+          If you would like to be alerted about our new ventures, click the
+          button below. And if there are any issues, topics or current events
+          that you think could benefit from the Covid Act Now approach please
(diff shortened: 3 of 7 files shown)


scripts/alert_emails/utils.ts
```diff
@@ -90,8 +90,10 @@ function generateAlertEmailContent(
     last_updated: lastUpdated,
     location_url: `${locationURL}?utm_source=risk_alerts&utm_medium=email`,
     subscribe_link: `https://covidactnow.org/alert_signup?utm_campaign=risk_alert_forward_subscribe&utm_source=risk_alerts&utm_medium=email`,
-    unsubscribe_link: `${unsubscribeURL}?email=${encodeURI(emailAddress)}`, // would be nice to know dev/staging/prod
-    feedback_subject_line: encodeURI(
+    unsubscribe_link: `${unsubscribeURL}?email=${encodeURIComponent(
+      emailAddress,
+    )}`, // would be nice to know dev/staging/prod
+    feedback_subject_line: encodeURIComponent(
```
mikelehen, line 96: We were using `encodeURI()` in a few places where we should have been using `encodeURIComponent()`... and so e.g. unsubscribe links in emails sent to people with `+` in their email didn't work.

src/screens/Subscriptions/AfterUnsubscribe.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { EventAction, EventCategory } from 'components/Analytics';
+import ExternalLink from 'components/ExternalLink';
+import { LargeFilledButton } from 'components/ButtonSystem';
+
+const AfterUnsubscribe = () => {
+  const params = new URLSearchParams(window.location.search);
+  const email = params.get('email') || '';
+  const formRef = createRef<HTMLFormElement>();
+
+  const formAction = getFormActionLink(email);
```
mikelehen, line 12: This odd code to get the form action link from campaign monitor and add it to the form before submitting it is modeled off of our existing code for daily download signup (https://github.com/covid-projections/covid-projections/blob/develop/src/components/EmailAlertsForm/utils.ts#L37)...

scripts/alert_emails/utils.ts
mimecuvalo: oof, oh no... good catch!

src/screens/Subscriptions/Subscriptions.style.tsx
```diff
@@ -0,0 +1,32 @@
...
+  font-size: 2rem;
+  margin-bottom: 2.5rem;
+  line-height: 1.2;
+`;
+
+export const BodyCopy = styled(Typography)`
+  ${props => props.theme.fonts.regularBook};
+  font-size: 1.1rem;
+  margin-bottom: 1.75rem;
+  line-height: 1.4;
```
mimecuvalo, line 31: as i start learning the codebase/design system here - have we considered having standard paddings/sizes and such for typography? what's the general guidelines, if any, for custom values for components?

(just thinking/wondering out loud btw, nothing actionable for this diff)

mimecuvalo: (i'm thinking ahead to BBE and other work - things like, say, `600px` in the `@media (min-width: 600px)` above could be constants for consistency across the codebase.)

src/screens/Subscriptions/AfterSubscribe.tsx
```diff
@@ -0,0 +1,21 @@
...
+  return (
+    <Wrapper>
+      <Header>Thank you</Header>
+      <BodyCopy>
+        <p>
+          Your subscription has been confirmed. You've been added to our list
+          and will hear from us soon.
+        </p>
+        <p>
+          <a href="/">Click here</a> to return to the site.
```
mimecuvalo, line 14: should this be a `Link` instead?

src/screens/Subscriptions/AfterUnsubscribe.tsx
```diff
@@ -0,0 +1,94 @@
...
+        className="js-cm-form"
+        id="subForm"
+        method="POST"
+        data-id={CREATESEND_DATA_ID}
+      >
+        <input
+          className="js-cm-email-input qa-input-email"
+          id="fieldEmail"
+          name="cm-jrjjlhr-jrjjlhr"
+          style={{ display: 'none' }}
```
mimecuvalo, line 67: instead of display: none - this should just be `type="hidden"`

src/screens/Subscriptions/AfterUnsubscribe.tsx
```diff
@@ -0,0 +1,94 @@
...
+          value={email}
+          readOnly
```
mimecuvalo, line 69: and then you can drop the readOnly

src/screens/Subscriptions/AfterUnsubscribe.tsx
```diff
@@ -0,0 +1,94 @@
...
+        ref={formRef}
+        className="js-cm-form"
+        id="subForm"
+        method="POST"
+        data-id={CREATESEND_DATA_ID}
+      >
+        <input
+          className="js-cm-email-input qa-input-email"
+          id="fieldEmail"
+          name="cm-jrjjlhr-jrjjlhr"
```
mimecuvalo, line 66: what is this funky `name` for?

src/screens/Subscriptions/AfterUnsubscribe.tsx
mimecuvalo, line 12: it's prbly not necessary to do `getFormActionLink(email);` (which in turn does the `createsend.com` call) until the person actually clicks the subscribeClick link. so down below you can do:
```
if (formRef?.current) {
  const formAction = await getFormActionLink(email);
  formRef.current.action = await formAction;
  ...
```

src/screens/Subscriptions/Subscriptions.style.tsx
mikelehen: Heh, we definitely should (in the BBE codebase especially)... but the covid codebase has evolved a bit haphazardly and it's taken us a while to get around to unifying styles.  So I don't _think_ we have any typography styles ready to be reused here.  We do for e.g. buttons, but not so much for typography.  Though @pnavarrc or @chasulin may know better.  I don't think it's worth too much effort to unify at this point, but we should for the BBE codebase.

src/screens/Subscriptions/AfterUnsubscribe.tsx
mikelehen: Oh, good call!

src/screens/Subscriptions/AfterUnsubscribe.tsx
mikelehen: Oops, good question.  This is campaign monitor junk (they auto-generate the form code for you, though I then had to modify it).  Added a comment explaining.

mikelehen (commented):
Thanks for the review! Made some tweaks and added comments.

src/screens/Subscriptions/Subscriptions.style.tsx
chasulin: 600px breakpoint [is indeed a constant](https://github.com/covid-projections/covid-projections/blob/develop/src/assets/theme/sizes.tsx#L4)

as for unifying spacing, im starting to clean that up a little bit [in the PR](https://github.com/covid-projections/covid-projections/pull/3577). as michael mentioned, [buttons are newly unified](https://github.com/covid-projections/covid-projections/blob/develop/src/components/ButtonSystem/MainButtons.style.tsx)

chasulin: but yeah, allllll for unifying spacing/sizing/breakpoints/etc in the BBE code from the get go 💫💫

mikelehen: Ah, nice. Switched over to materialSMBreakpoint constant!

src/screens/Subscriptions/AfterUnsubscribe.tsx
mimecuvalo: Ok, no worries - sgtm!

src/screens/Subscriptions/Subscriptions.style.tsx
chasulin: also [started to do this for fonts](https://github.com/covid-projections/covid-projections/blob/develop/src/assets/theme/fonts.ts), but just font family + font weight

mimecuvalo (approved):
LGTM

src/screens/Subscriptions/Subscriptions.style.tsx
mimecuvalo: ooo excellent @chasulin!

PR #3580, opened 2021-05-12
Make useShowPastPosition() hook trigger fewer renders.

mikelehen:


src/common/hooks/useShowPastPosition.ts
```diff
@@ -1,11 +1,22 @@
...
+    const onScroll = () => {
+      let currentPosition = window.pageYOffset;
+      if (currentPosition > y) {
+        setPastPosition(true);
+      } else {
+        setPastPosition(false);
+      }
+    };
+
+    window.addEventListener('scroll', onScroll);
```
mimecuvalo, line 17: this would benefit from adding `{passive: true}` for performance
here's more info on passive events: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#improving_scrolling_performance_with_passive_listeners
(it's well supported at this point so no worries to test for support https://caniuse.com/passive-event-listener )

however, looking at the api for https://www.npmjs.com/package/@react-hook/window-scroll it says you can configure the FPS - would that be an easier change to make it lower than the default 30fps (i'm not sure the exact work for your use case here)

mikelehen: Oh, good call! Added `{passive: true}` , thanks.

The problem with the window-scroll hook is that it updates state whenever the scroll position changes, which re-renders the component using the hook, which is basically our entire page.  And so for example we end up re-rendering all of the SVG tags for our maps / charts / etc. as you scroll through the page, which is pretty expensive.  You can feel the performance problem if you just down arrow on the location page or home page right now. It's really chunky.  With my PR, state doesn't change except at the specific point where you scroll past the specified `y` value, which is the only time we actually need to re-render anything.

FWIW- I'm also working on another PR to use `React.useMemo()` a bit more heavily so that re-rendering our heavier components is cheap as long as the props don't change.

mimecuvalo: >  And so for example we end up re-rendering all of the SVG tags for our maps / charts / etc. as you scroll through the page

oof, terrible.

> FWIW- I'm also working on another PR to use React.useMemo() a bit more heavily so that re-rendering our heavier components is cheap as long as the props don't change.

love it!

mimecuvalo (approved):
LGTM

mikelehen:
[Rework scroll trigger for showing search bar to avoid re-rendering.](https://trello.com/c/ixIggMxv/1321-rework-scroll-trigger-for-showing-search-bar-to-avoid-re-rendering)

PR #3587, opened 2021-05-12
Location page scroll bug

chasulin:
Chart/recommendations share URLs keep scrolling you back to the shared element. i.e. https://covidactnow.org/us/massachusetts-ma/chart/5 keeps jumping back to the daily new cases chart when you scroll away from it. This fix works, but there might be a better way?


faihegberg (approved):
LGTM

PR #3588, opened 2021-05-12
Mega menu bug

chasulin:
`ClickAwayListener` was breaking the mega menu and causing it to close on mobile when attempting to click a link (and overall, on prod, if you click anywhere within the mega menu that isn't a link, the menu closes). This removes `ClickAwayListener` altogether, which was originally implemented to make sure that the mega menu and search bar were never open at the same time. Instead, we now call `setMenuOpen(false)` when clicking into the search bar.

### DIFF
--- a/src/components/NavBar/NavAllOtherPages/NavAllOtherPages.tsx
+++ b/src/components/NavBar/NavAllOtherPages/NavAllOtherPages.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Link } from 'react-router-dom';
-import { ClickAwayListener } from '@material-ui/core';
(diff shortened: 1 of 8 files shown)


mimecuvalo:
qq: as long as you're looking at this - any reason the hamburger button (on desktop) opens on hover? it's a little confusing to me b/c i go to click it, it opens automatically, but then my instinctive click then closes it accidentally 🙃 - thoughts?

chasulin:
i _think_ the reasoning was that it makes the menu content more discoverable (?), but this might be more of a question for @joshziman

pnavarrc (approved):
Awesome! 🙌

PR #3590, opened 2021-05-12
Account for multi-suffix location names

faihegberg:
Some regions contain a name with multi-suffix. This PR ensures these few cases are accounted for.

### DIFF
--- a/src/components/NewLocationPage/LocationName/utils.ts
+++ b/src/components/NewLocationPage/LocationName/utils.ts
@@ -10,7 +10,6 @@ const regionSuffixes = [
   'co.',
   'parish',
   'par.',
-  'area',
   'c.a.',
   'borough',
   'bor.',
--- a/src/components/NewLocationPage/LocationName/utils.ts
+++ b/src/components/NewLocationPage/LocationName/utils.ts
@@ -15,7 +15,7 @@ const regionSuffixes = [
   'bor.',
 ];
 
-const multiSuffixIndicator = ['and', 'census'];
+const multiWordSuffixIndicator = ['and', 'census'];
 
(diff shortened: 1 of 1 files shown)


src/components/NewLocationPage/LocationName/utils.ts
```diff
@@ -43,7 +45,12 @@ export function getSplitRegionName(region: Region, condensed?: boolean) {
 function splitRegionName(regionName: string) {
   const splitRegion = regionName.split(' ');
   if (
-    regionSuffixes.includes(splitRegion[splitRegion.length - 1].toLowerCase())
+    regionSuffixes.includes(
+      splitRegion[splitRegion.length - 1].toLowerCase(),
+    ) &&
+    !multiSuffixIndicator.includes(
+      splitRegion[splitRegion.length - 2].toLowerCase(),
+    )
```
chasulin, line 52: wanna put this whole condition into a variable so its clearer exactly what its doing?

`const isSingleWordSuffix = regionSuffixes.includes(splitRegion[splitRegion.length - 1].toLowerCase()) && !multiSuffixIndicator.includes(splitRegion[splitRegion.length - 2].toLowerCase())`

src/components/NewLocationPage/LocationName/utils.ts
```diff
@@ -16,6 +16,8 @@ const regionSuffixes = [
   'bor.',
 ];
 
+const multiSuffixIndicator = ['and', 'census'];
```
chasulin, line 18: maybe `multiWordSuffixIndicator` instead of `multiSuffixIndicator`

chasulin (approved):

PR #3599, opened 2021-05-13
Update share button group

faihegberg:
This PR achieves the following:
- [x] Remove LinkedIn share buttons.
- [x] Add embed button to the share button group.
- [x] Remove share section at the bottom of pages.

### DIFF
--- a/src/components/Explore/Explore.tsx
+++ b/src/components/Explore/Explore.tsx
@@ -370,6 +370,7 @@ const Explore: React.FunctionComponent<{
                 )
               }
               quote={getSocialQuote(selectedLocations, currentMetric)}
+              region={region ? region : undefined}
               onSaveImage={() => {
                 trackExploreEvent(
                   EventAction.SAVE_IMAGE,
--- a/src/components/Explore/Explore.tsx
+++ b/src/components/Explore/Explore.tsx
@@ -390,9 +390,6 @@ const Explore: React.FunctionComponent<{
               onShareOnTwitter={() =>
                 trackShare(`Twitter: ${trackingLabel}`, numLocations)
               }
-              onShareOnLinkedin={() =>
-                trackShare(`Linkedin: ${trackingLabel}`, numLocations)
-              }
             />
           </Styles.ShareBlock>
         </div>
--- a/src/components/Explore/Explore.tsx
+++ b/src/components/Explore/Explore.tsx
@@ -14,7 +14,7 @@ import useMediaQuery from '@material-ui/core/useMediaQuery';
 import { useTheme } from '@material-ui/core/styles';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import { ParentSize } from '@vx/responsive';
-import ShareImageButtonGroup from 'components/ShareButtons';
+import ShareButtonGroup from 'components/ShareButtons';
 import ExploreTabs from './ExploreTabs';
 import ExploreChart from './ExploreChart';
 import Legend from './Legend';
@@ -360,7 +360,7 @@ const Explore: React.FunctionComponent<{
         )}
         <div style={{ paddingTop: '1rem' }}>
           <Styles.ShareBlock>
-            <ShareImageButtonGroup
+            <ShareButtonGroup
               disabled={selectedLocations.length === 0 || !hasData}
               imageUrl={() => createSharedComponentId().then(getExportImageUrl)}
               imageFilename={getImageFilename(selectedLocations, currentMetric)}
--- a/src/components/LocationPage/ShareButtons.tsx
+++ b/src/components/LocationPage/ShareButtons.tsx
@@ -29,10 +29,7 @@ const ShareButtons: React.FC<{
 
   const [showSocialButtons, setShowSocialButtons] = useState(false);
 
-  const hideSocialButtons = () => {
-    const timeoutId = setTimeout(() => setShowSocialButtons(false), 1500);
-    return () => clearTimeout(timeoutId);
-  };
+  const hideSocialButtons = () => setShowSocialButtons(false);
 
   useEscToClose(hideSocialButtons);
 
--- a/src/components/LocationPage/ShareButtons.tsx
+++ b/src/components/LocationPage/ShareButtons.tsx
@@ -8,6 +8,7 @@ import makeChartShareQuote from 'common/utils/makeChartShareQuote';
 import * as urls from 'common/urls';
 import { Region } from 'common/regions';
 import type { MetricValues } from 'common/models/Projections';
+import { useEscToClose } from 'common/hooks';
 
 const ShareButtons: React.FC<{
   region: Region;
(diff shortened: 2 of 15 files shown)


src/screens/LocationPage/LocationPage.tsx
```diff
@@ -33,7 +33,7 @@ function WithSearchInNav({ region }: LocationPageProps) {
         hasScrolled={hasScrolled}
         region={region}
       />
-      <ChartsHolder chartId={chartId} region={region} />
+      <ChartsHolder chartId={chartId} region={region} isLocationPage={true} />
```
chasulin, line 36: we dont need this `isLocationPage` prop. the presence of a region lets you know that its a location page

src/components/Explore/Explore.tsx
```diff
@@ -370,6 +370,7 @@ const Explore: React.FunctionComponent<{
                 )
               }
               quote={getSocialQuote(selectedLocations, currentMetric)}
+              region={region ? region : undefined}
```
chasulin, line 373: you can just pass `region={region}`

the region variable in Explore.tsx [is grabbed here](https://github.com/covid-projections/covid-projections/blob/develop/src/components/Explore/Explore.tsx#L115) -
`const region = useRegionFromParams();`

`useRegionFromParams` will return `null` if the region comes up undefined ([see the source code here](https://github.com/covid-projections/covid-projections/blob/develop/src/common/regions/region_hooks.ts#L17))

faihegberg: Yea, strangely enough when I tried that I get the following error:
`Type 'Region | null' is not assignable to type 'Region | undefined'.
  Type 'null' is not assignable to type 'Region | undefined'.`
So I implemented as such. I was hoping there would be a better solution though 😕

chasulin: just chatted about this on a call but lmk if there are more typescripty questions!

faihegberg: After additional discussion with @pnavarrc, we figured it's probably better to just leave the type check as is. Changing it may trigger quite a significant amount of changes which we think may not be worthwhile to accommodate. Let us know if you have any additional thoughts about this though! 🙂

src/components/LocationPage/SocialButtons.tsx
```diff
@@ -77,6 +75,21 @@ const SocialButtons = ({
           </CopyLinkButton>
         </SocialShareButton>
       </CopyToClipboard>
+      <SocialShareButton variant="contained" color="#007fb1">
+        <SocialButton
+          disableRipple
+          disableFocusRipple
+          disableTouchRipple
+          onClick={() => setShowEmbedPreviewModal(true)}
```
pnavarrc, line 83: Consider moving the `disableRipple`, etc attributes to the styled component, that way it will be consistent across usages

```tsx
export const SocialButton = styled(Button).attrs(props => ({
  disableElevation: true,
  disableRipple: true,
  // ...
}))`
  width: 60px;
  height: 42px;
  // ...
`;

```

src/components/ShareBlock/ShareBlock.tsx
```diff
@@ -85,80 +85,82 @@ const ShareBlock = ({
...
+                        bgStyle={{ fill: 'auto' }}
+                      />
+                    </TwitterShareButton>
+                  </StyledShareButton>
+                  <StyledShareButton variant="contained" color="#007fb1">
+                    <LinkedinShareButton
+                      url={url}
+                      title={quote}
+                      // @ts-ignore: seems to not be available for linkedin?
+                      hashtags={[hashtag]}
```
pnavarrc, line 141: We are not using LinkedIn anymore, are we?

src/components/ShareButtons/SocialButtonBlock.tsx
```diff
@@ -40,11 +44,22 @@ const SocialButtonBlock: React.FC<{
...
-        onClickShare={onShareOnLinkedin}
-        {...socialSharingProps}
-      />
       <CopyLinkButton url={socialSharingProps.url} onCopyLink={onCopyLink} />
+      <SocialShareButton variant="contained" color="#007fb1">
+        <SocialButton
+          disableRipple
+          disableFocusRipple
+          disableTouchRipple
+          onClick={() => setShowEmbedPreviewModal(true)}
```
pnavarrc, line 53: Also here

pnavarrc (commented):
Added some notes

src/components/ShareBlock/ShareBlock.tsx
faihegberg: Ah you're right! For some reason I kept thinking the whole share block will be completely removed anyway but turns out we will still have it on the homepage.

src/components/LocationPage/ShareButtons.tsx
```diff
@@ -26,30 +27,28 @@ const ShareButtons: React.FC<{
...
 
-  const [showShareIcons, setShowShareIcons] = useState(false);
+  const [showSocialButtons, setShowSocialButtons] = useState(false);
 
-  // Delay allows the user to briefly see copy-link button text change when clicked
   const hideSocialButtons = () => {
-    const timeoutId = setTimeout(() => {
-      setShowShareIcons(false);
-    }, 1500);
+    const timeoutId = setTimeout(() => setShowSocialButtons(false), 1500);
```
pnavarrc, line 33: It looks like the `ShareButtons` component doesn't need the delay at all, maybe we should pass `setShowSocialButtons` to the `SocialButtons` as prop instead?

```tsx
<SocialButtons
  {/* ... */}
  region={region}
  setShowSocialButtons={setShowSocialButtons}
/>
```

and then in `SocialButtons` we can use [`useEffect`](https://reactjs.org/docs/hooks-reference.html#useeffect) to update the state with a delay. 

> Mutations, subscriptions, timers, logging, and other side effects are not allowed inside the main body of a function component (referred to as React’s render phase). Doing so will lead to confusing bugs and inconsistencies in the UI.

```tsx
// timers are not allowed in the main body of react components, they must be
// wrapped with useEffect
useEffect(() => {
  const timeoutId = setTimeout(() => setShowSocialButtons(false), 500);
  return () => clearTimeout(timeoutId);
});

// we probably don't need the delay when the user closes with esc
useEscToClose(() => setShowSocialButtons(false));
```

Alternatively, we could just pass a `closeMenu` function since we are only closing the menu from `SocialButtons`.


```tsx
<SocialButtons
  {/* ... */}
  region={region}
  closeMenu={() => setShowSocialButtons(false)}
/>
```

and then

```tsx
useEffect(() => {
  const timeoutId = setTimeout(closeMenu, 500);
  return () => clearTimeout(timeoutId);
});

// we probably don't need the delay when the user closes with esc
useEscToClose(closeMenu);
```

src/components/ShareBlock/ShareBlock.tsx
```diff
@@ -85,80 +83,64 @@ const ShareBlock = ({
...
+                </ShareButtonContainer>
+                <EmbedPrompt>
+                  Or <span onClick={onClickEmbed}>embed on your website</span>
+                </EmbedPrompt>
+              </SocialTextArea>
+            </SocialTextAreaWrapper>
+            <SocialMockupWrapper>{SocialPreview}</SocialMockupWrapper>
+          </ShareRowContentArea>
+        </ShareRow>
+      )}
```
pnavarrc, line 143: We could move the `<ShareRow>` component to its own file and conditionally include it in `ShareModelBlock` instead, since that part doesn't really need the `region`. Feel free to punt for later!

pnavarrc (changes requested):
Added some notes on how to close the menu with delay, we are almost done!

pnavarrc (approved):
Looks good!

PR #3608, opened 2021-05-15
Fix negative height error on sparklines

schamp:
Ran into this whilst looking into something else.


pnavarrc (approved):
Thanks for fixing this!

chasulin:
ahh thank you @schamp for fixing this!

PR #3634, opened 2021-05-20
Add memoization to various components.

mikelehen:
Memoize some key components, especially ones with lots of html elements (charts, maps, compare table, etc.) to make renders faster.

I basically picked out interesting components, added console.log statements so I could see when they were rendering, and then memoized stuff until they seemed to stop rendering unnecessarily. 😁

### DIFF
--- a/src/components/LocationPage/ChartsHolder.tsx
+++ b/src/components/LocationPage/ChartsHolder.tsx
@@ -1,4 +1,10 @@
-import React, { useRef, useEffect, useState, useMemo } from 'react';
+import React, {
+  useRef,
+  useEffect,
+  useState,
+  useMemo,
+  useCallback,
+} from 'react';
 import { useLocation } from 'react-router-dom';
 import {
   useCcviForFips,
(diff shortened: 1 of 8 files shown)


src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -46,18 +52,26 @@ interface ChartsHolderProps {
...
+  const metricRefs = useMemo(
+    () => ({
+      [Metric.CASE_DENSITY]: caseDensityRef,
+      [Metric.CASE_GROWTH_RATE]: caseGrowthRateRef,
+      [Metric.POSITIVE_TESTS]: positiveTestsRef,
+      [Metric.HOSPITAL_USAGE]: hospitalUsageRef,
+      [Metric.VACCINATIONS]: vaccinationsRef,
+    }),
+    [],
+  );
```
mikelehen, line 74: This is a bit annoying / ugly, but this is the only way I could figure out how to memoize `metricRefs` so that I could memoize the charts, etc.

src/components/Map/Map.tsx
```diff
@@ -18,7 +19,7 @@ function Map({
   hideLegend = false,
   hideInstructions = false,
   hideLegendTitle = false,
-  onClick = () => {},
+  onClick = noop,
```
mikelehen, line 22: Without this, `USACountyMap` ended up re-rendering every time the homepage rendered since `onClick` changed every time.

src/components/Map/USACountyMap.tsx
```diff
@@ -41,25 +41,25 @@ interface USACountyMapProps {
   showCounties: boolean;
 }
 
+const projection = geoAlbersUsaTerritories().scale(1070).translate([400, 300]);
```
mikelehen, line 44: Made this a top-level constant so it never changes.

src/components/Map/USACountyMap.tsx
```diff
@@ -41,25 +41,25 @@ interface USACountyMapProps {
...
+      },
+      [locationSummaries],
+    );
 
     const onMouseLeave = () => setTooltipContent('');
 
-    return (
-      <USMapWrapper>
-        {/** Map with shaded background colors for states. */}
+    const mapContent = useMemo(
```
mikelehen, line 61: Split out the bulk of the map so that it doesn't re-render whenever the tooltip changes (fixes Chris's complaint about slow renders when you move your mouse around on the map).

mikelehen:
[Migrate charts and other expensive components to use React.memo() to avoid rerenders unless props change.](https://trello.com/c/fsjyDP0M/1323-migrate-charts-and-other-expensive-components-to-use-reactmemo-to-avoid-rerenders-unless-props-change)

pnavarrc (approved):
Works great!

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -114,37 +128,40 @@ const ChartsHolder = ({ region, chartId }: ChartsHolderProps) => {
...
+        EventAction.CLICK,
+        `Location Header Stats: ${Metric[metric]}`,
+      );
+      scrollTo(metricRefs[metric].current);
+    },
+    [metricRefs],
+  );
 
-  const onClickSparkLine = (metric: SparkLineMetric) => {
+  const onClickSparkLine = useCallback((metric: SparkLineMetric) => {
```
schamp, line 156: what a trap this pattern is : (

src/components/Map/Map.tsx
schamp: oof yeah, another nasty trap.

schamp:
Nice work.  Consider doing some knowledge sharing on what you found and which traps we can all learn to avoid going forward -- I knew some of these things, but am not in the habit of watching out for them.

mikelehen:
> Nice work. Consider doing some knowledge sharing on what you found and which traps we can all learn to avoid going forward -- I knew some of these things, but am not in the habit of watching out for them.

Yeah, good call.  TBH this has definitely been a learning experience for me and more trial-and-error than direct application of knowledge / science. 😁   But I'll try to put some lessons learned together in a useful way.

PR #3638, opened 2021-05-20
Remove API copy A/B test

faihegberg:
This PR removes the A/B test from the API copy on the mega menu.


src/components/MenuContent/FeaturedSection.tsx
```diff
@@ -32,9 +26,6 @@ const FeaturedSection: React.FC<{
     [SectionId.DAILY_DOWNLOAD]: DailyDownloadIcon,
     [SectionId.ALERTS]: AlertsIcon,
   };
-  const testAPITitle = 'Data API';
-  const testAPIDescription =
-    'Get realtime, local data including daily new cases, % vaccinated, and more.';
```
mikelehen, line 37: This is the text we want to keep, right?  Is there a separate task to update the text in the CMS to this text?  Or do you want to just add that to this PR (update the text in the JSON file manually)?  Just want to make sure we end up with the right text! 😁

mikelehen (approved):

src/components/MenuContent/FeaturedSection.tsx
faihegberg: @mikelehen yes, right here 😄  https://github.com/covid-projections/covid-projections/pull/3639

PR #3640, opened 2021-05-20
Update header share button

faihegberg:
Ensure header share button doesn't scroll to a non-existent section and instead behaves like all other share buttons on the location page.


src/components/NewLocationPage/LocationOverview/LocationOverview.tsx
```diff
@@ -75,6 +79,13 @@ const LocationOverview: React.FC<{
       </GridContainer>
       <MobileOnly>
         <ShareButton onClickShare={onClickShare} />
+        <ShareButtons
+          eventCategory={EventCategory.ENGAGEMENT}
+          shareUrl={shareUrl}
+          shareQuote={shareQuote}
+          region={region}
+          isLocationPageHeader={true}
+        />
```
faihegberg, line 88: The first share button component needs to be replaced with the second share button. I just haven't removed the first one yet to better illustrate the issue. Currently, the second share button will not stretch in the same way as the first. @chasulin we can sync on this to resolve.

chasulin:
almost there! last thing-

the placement of the social buttons looks slightly wonky on mobile:
Screen Shot 2021-05-26 at 10 46 37 AM

how about we center them instead:
Screen Shot 2021-05-26 at 10 48 17 AM

placing these styles on the absolutely positioned div for mobile should do it-
```
top: 1rem;
margin: 0 auto;
```

chasulin (commented):
almost there! last thing-

the placement of the social buttons looks slightly wonky on mobile (first screenshot). how about we center them instead (2nd screenshot)? placing these styles on the absolutely positioned div for mobile should do it

top: 1rem;
margin: 0 auto;
(left and right set to 0)

Screen Shot 2021-05-26 at 10 46 37 AM
Screen Shot 2021-05-26 at 10 48 17 AM

chasulin (approved):
🚢

PR #3641, opened 2021-05-20
Add RA announcement to homepage and about page.

mikelehen:


src/screens/About/About.tsx
```diff
@@ -21,27 +21,31 @@ export const sidebarItems: TocItem[] = [
     items: [
       {
         to: '/about#mission',
-        label: 'Our mission',
```
mikelehen, line 24: I can undo these changes if we're not a fan, but all the text matched exactly so I figured I might as well remove the redundancy.

src/screens/About/About.tsx
```diff
@@ -103,6 +107,11 @@ const About = () => {
           <MarkdownContent source={aboutContent.teamIntro} />
           <TeamSection />
         </HashWrapper>
+        <HashWrapper id="future-projects">
+          <AboutHeading2>{aboutContent.futureProjectsHeader}</AboutHeading2>
+          <MarkdownContent source={aboutContent.futureProjectsContent} />
+          <TeamSection />
```
pnavarrc, line 113: The Team section is duplicated here, everyone shows up twice in the page

mikelehen: oh, yikes.  Good catch.

pnavarrc (approved):
Looks good!

PR #3643, opened 2021-05-21
New location page chart footer

faihegberg:
This PR is a WIP of a standardized chart footer for the location page.


faihegberg:
[Location page redesign: reusable footer component](https://trello.com/c/E1qCQ0Uc/1364-location-page-redesign-reusable-footer-component)

src/components/NewLocationPage/ChartFooter/ChartFooter.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+    <rect y="0.152344" width="8" height="8" fill="#E0E0E0" />
+  </svg>
+);
+
+export const ChartFooterButtonOnly = () => {
+  return (
+    <Footer>
+      <FooterSection>
+        <SingleButtonWrapper>
+          <ButtonContainer>
```
chasulin, line 43: hm do we need all these wrappers?

src/components/NewLocationPage/ChartFooter/ChartFooter.style.tsx
```diff
@@ -0,0 +1,80 @@
...
+export const ButtonContainer = styled.div`
+  flex-grow: 0;
+  margin-left: 0.5rem;
+
+  @media (min-width: ${materialSMBreakpoint}) {
+    margin-left: 7.5rem;
+  }
+`;
+
+export const LegendContainer = styled.div`
```
chasulin, line 37: i think we can use [the current LegendContainer](https://github.com/covid-projections/covid-projections/pull/3670/files#diff-d2b8fc3b1fc0390125ed3994eec3c2aff7151f7e3a572c0bc60dc06296372665R73)

src/components/NewLocationPage/ChartFooter/ChartFooter.stories.tsx
faihegberg: Hmmm I think we will need to keep `ButtonContainer` and `SingleButtonWrapper` to prevent the button from stretching and keep the button right aligned. We can likely get rid of `FooterSection` although this would result in zero space on the right and top of the button (this is fine too if one plans to add in that space manually). `Footer` however, we can definitely get rid off in this case.

pnavarrc (approved):
Looks good!

PR #3653, opened 2021-05-22
Update CMS scheme (and content) for About page redesign

schamp:
This is in support of the work on https://trello.com/c/EVnripUF/1343-about-page-improvements. it adds the needed new schema items to the CMS.

If I need to separate the `index.tsx` or `.json` changes from this schema update diff, LMK, it wasn't clear from the instructions.


schamp:
[About page improvements](https://trello.com/c/EVnripUF/1343-about-page-improvements)

mikelehen:
Haven't dug into the diff, but just comparing the about page to prod, a couple headings have changed ("Who we work with" => "Out partners" and "Who we are" => "Our team"... not a big deal and maybe intentional?) and the "Who we serve" logos got duplicated into the "Our partners" section (probably not intended / desirable).

schamp:
> Haven't dug into the diff, but just comparing the about page to prod, a couple headings have changed ("Who we work with" => "Out partners" and "Who we are" => "Our team"... not a big deal and maybe intentional?) and the "Who we serve" logos got duplicated into the "Our partners" section (probably not intended / desirable).

Yeah, these are the updated values.  Probably best to separate this from the schema update, since not everything is backwards compatible.  Or, if all changes *must* be made backwards compatible, then I'll have to make this approach a little more complicated.

mikelehen:
Got it...  Yeah, I was imagining adding the new sections to the schema and merge that in isolation (with the existing about page not reading them so it'd be unaffected)...  and then follow up later with the design changes and final content.

But it's also fine if you want to just do the whole redesign w/ updated content in a single PR that we merge when it's ready.  I'm honestly not sure if trying to be incremental is going to save us any headaches / effort.  We'll just have to be a tad careful about concurrent content edits to the about page (I can warn the content folks).

schamp:
I think this is all the schema updates I need -- unless we decide we want to put the button href and ctas in the CMS too, in which case I'll need to update the schema for that.

I've reverted the json changes, and it looks right again.  If this is the right approach, let's go ahead and merge the schema updates, and we can work through how to deal with the content changes when we're getting that other PR ready to merge.

mikelehen (approved):
Thanks!  I think this looks good, except I think we want to use SVG for the icons instead of PNG.  At least that's what we've done in other places (e.g. recommendations: https://github.com/covid-projections/covid-projections/blob/6cf8944b9fc0f6957e394ebc35323f9f84fcbb07/src/cms-content/recommendations/recommendations-main.json#L42)

But since we reverted the json files from this PR, maybe just revert the images from this PR too?

schamp:
Sure, can do, thanks for the pointer.

PR #3669, opened 2021-05-25
Restore CTAs in vulnerability block

faihegberg:
This PR restores CTAs in the vulnerability block on state and county location pages.

### DIFF
--- a/src/components/VulnerabilitiesBlock/VulnerabilitiesBlockInner.tsx
+++ b/src/components/VulnerabilitiesBlock/VulnerabilitiesBlockInner.tsx
@@ -16,6 +16,7 @@ import { RegionCcviItem } from 'common/data';
 import { getSurgoUrlByRegion } from 'common/ccvi/index';
 import { EventCategory } from 'components/Analytics';
 import { TextButton } from 'components/ButtonSystem';
+import OpenInNewIcon from '@material-ui/icons/OpenInNew';
 
 const VulnerabilitiesBlockInner: React.FC<{
   scores: RegionCcviItem;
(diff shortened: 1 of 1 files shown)


faihegberg:
[Restore CTAs in Vulnerability section](https://trello.com/c/bqGN8FFz/1376-restore-ctas-in-vulnerability-section)

src/components/VulnerabilitiesBlock/VulnerabilitiesBlockInner.tsx
```diff
@@ -43,6 +50,22 @@ const VulnerabilitiesBlockInner: React.FC<{
           {region.shortName} {regionDescription}
         </SummaryText>
         <SummaryText>{communityVulnerabilityQuote}</SummaryText>
+        {surgoUrl && (
+          <SummaryText>
+            <TextButton
+              href={surgoUrl}
+              onClick={() =>
```
chasulin, line 57: since we're using TextButton, which is part of the button system, the props are a little different. [see here for one thats already implemented](https://github.com/covid-projections/covid-projections/blob/develop/src/screens/DataApi/DataApi.tsx#L49). we dont need to pass an onClick with trackEvent, because that is already implemented/called [in the BaseButton that is at the root of TextButton](https://github.com/covid-projections/covid-projections/blob/develop/src/components/ButtonSystem/BaseButton/BaseButton.tsx#L35) (/at the root of all buttons in the button system)

chasulin: so in this case we just need: `href`, `trackingCategory`, and `trackingLabel`

chasulin (commented):
[mocks](https://www.figma.com/file/JGqemgZwlG9e5DQbZpdKgy/Redesign?node-id=1370%3A88195) screenshot below--looks like the `TextButton` needs an `endIcon` + that whole column should be center aligned (i mightve thrown off the alignment in the redesign stuff?) 

Screen Shot 2021-05-26 at 5 34 31 PM

chasulin (approved):

PR #3693, opened 2021-05-29
Add metro areas to zip-based search

schamp:
The trello task says "state and metro" but state is a lot harder and I'm not certain it's worth the tradeoff, so this does the first half of it.

Because the items are converted to strings for searching, to make this work for states we'd need to:
 1) get all counties which are part of the given state (which we don't have, so we'd have to scan all counties and filter by state) 
 2) map all counties to zips by trying to look each one up by FIPS
 3) store a string with all zipcodes for counties in a state
 
 This seems like it would be costly, and the value (showing a state by zip) doesn't seem that valuable to me.

### DIFF
--- a/src/components/Search/SearchAutocomplete/SearchAutocomplete.tsx
+++ b/src/components/Search/SearchAutocomplete/SearchAutocomplete.tsx
@@ -1,8 +1,10 @@
 import React, { useEffect, useState } from 'react';
+import flatten from 'lodash/flatten';
+import uniq from 'lodash/uniq';
 import { Autocomplete } from '@material-ui/lab';
 import Hidden from '@material-ui/core/Hidden';
 import { createFilterOptions } from '@material-ui/lab/useAutocomplete';
-import { Region, MetroArea } from 'common/regions';
+import { State, County, Region, MetroArea, FipsCode } from 'common/regions';
 import {
   Wrapper,
   StyledTextField,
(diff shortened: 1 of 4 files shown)


schamp:
image (1)

schamp:
[Make zipcode search return metro and state (currently only returns county)](https://trello.com/c/I6NHyfid/888-make-zipcode-search-return-metro-and-state-currently-only-returns-county)

src/components/Search/SearchAutocomplete/SearchAutocomplete.tsx
```diff
@@ -74,16 +74,33 @@ const SearchAutocomplete: React.FC<{
...
+      let zipCodes = null;
       if (option instanceof MetroArea) {
-        return `${option.shortName}, ${(option as MetroArea).stateCodes}`;
-      } else {
-        return option.shortName;
+        zipCodes = uniq(flatten(option.countiesFips.map(zipsForFips)));
+      } else if (option instanceof State) {
+        // going from state -> zips requires mapping all counties for the state
+        // which we don't have and won't be cheap (a list of all zip-codes per state? oof)
+        // so maybe defer this oen for now
```
chasulin, line 90: sounds good--lets nix state

chasulin (approved):
woohoo looks good! before merging, want to clean up the state clause and maybe move the whole `stringifyOption` function into [search/utils](https://github.com/covid-projections/covid-projections/blob/develop/src/components/Search/utils.ts) since its getting long-ish?

PR #3698, opened 2021-05-30
Support multiple comma-separated regions in overrides

schamp:
Updates CMS schema to allow it, and override processor to handle it.

Does not combine any actual overrides though.


schamp:
I hacked in some example multi-overrides to the data and ensure the unit tests passed (though I didn't commit those changes).

schamp:
[Metric blocking: Support single override targeting multiple FIPS.](https://trello.com/c/QZYvrX0d/1401-metric-blocking-support-single-override-targeting-multiple-fips)

PR #3702, opened 2021-05-31
Replace chart footer for explore, recommendations, vulnerabilities block

faihegberg:
The standardized chart footer has been merged. This PR implements the standardized chart footer on the explore chart, recommendations block, and vulnerabilities block. Some important notes:
- The chart footer has not been implemented for the compare table yet, as this footer contain functionalities I'm not confident I should be removing at this point.
- The original plan is to build on to `LocationPageSectionFooter.tsx`. However, some of the styling here has been handled by the individual components of the standardized chart footer. The only exception is the top margin (which I currently just implemented a wrapper (`FooterWrapper` component) around 😬 ).
- Overall, there might be a more optimized approach to standardizing these footers. To prevent double work, I just implemented this PR as a preliminary idea, and we can iterate 😄 .

@chasulin let's sync on this.

### DIFF
--- a/src/components/LocationPage/ChartBlock.tsx
+++ b/src/components/LocationPage/ChartBlock.tsx
@@ -18,8 +18,12 @@ import { getSourcesForMetric } from 'common/utils/provenance';
 import { Sources } from 'api/schema/RegionSummaryWithTimeseries';
 import { getRegionMetricOverride } from 'cms-content/region-overrides';
 import { MarkdownContent } from 'components/Markdown';
-import LocationPageSectionFooter from 'components/LocationPageSectionFooter/LocationPageSectionFooter';
 import { SectionHeader } from 'components/SharedComponents';
+import {
+  FooterWrapper,
+  Footer,
+  ButtonContainer,
+} from 'components/NewLocationPage/ChartFooter/ChartFooter.style';
 
 function ChartBlock(props: {
   chartRef: React.RefObject<HTMLDivElement>;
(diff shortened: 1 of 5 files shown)


pnavarrc:
image

The "About this data" in vulnerabilities doesn't match the style of other charts, can we maybe restyle it to match?

src/components/LocationPage/ChartBlock.tsx
```diff
@@ -63,15 +67,19 @@ function ChartBlock(props: {
...
-              chartIdentifier={metric}
-              region={region}
-              stats={stats}
-              showEmbedButton={false}
-            />
-          </LocationPageSectionFooter>
+          <FooterWrapper>
+            <Footer>
+              <DisclaimerWrapper>{disclaimerContent}</DisclaimerWrapper>
+              <ButtonContainer>
```
chasulin, line 73: hm this change introduces 2 new wrappers that werent previously needed

faihegberg: Yea, the `FooterWrapper` was introduced to add some white space at the top 😬 (see point above in the PR description). I was reluctant to add this wrapper cos I wanted to minimize the number of wrappers. I tried adding this space to the `Footer` component instead but that will cause the vertical spacing of the following section in mobile mode to be off. 

image

faihegberg:
Closed this PR since we figured the footer for the vulnerability and recommendation sections are already quite simplified. Since we're already implementing the footer in pieces anyway instead of one unified component, we can focus on making these footer components work for the metric charts instead.

PR #3703, opened 2021-06-01
Create chart tabs

faihegberg:
This PR creates standardized chart tabs. Some curious notes below:
- Some metric names don't seem to match what's on Figma 🤔 (e.g. vaccination metrics). This PR currently uses the same logic as `SummaryStat` to get the metric name and metric sub-label. If it's necessary that the metric names must match what's on Figma exactly, we may need to build in additional logic?
- This PR assumes both click and hover behaviour will trigger the metric name to be bolded and the bottom border to appear.

### DIFF
--- a/src/components/NewLocationPage/ChartTabs/ChartTab.style.tsx
+++ b/src/components/NewLocationPage/ChartTabs/ChartTab.style.tsx
@@ -1,5 +1,6 @@
 import styled from 'styled-components';
 import { COLOR_MAP } from 'common/colors';
+import { materialSMBreakpoint } from 'assets/theme/sizes';
 
 export const TabsContainer = styled.div`
   display: flex;
@@ -9,12 +10,14 @@ export const TabsContainer = styled.div`
 export const Tab = styled.div`
   display: flex;
   flex-direction: column;
-  max-width: 8rem;
+  max-width: 10.5rem;
 `;
 
 // TODO: Figure out transforming color to black on hover or click?
 export const TabTitle = styled.div`
   text-transform: uppercase;
+  font-size: 0.8rem;
+  line-height: 1;
   color: ${COLOR_MAP.GREY_4};
   margin-bottom: 0.5rem;
 `;
@@ -26,6 +29,12 @@ export const TabContent = styled.div`
 `;
 
 export const MetricSubLabel = styled.div`
+  font-size: 0.8rem;
   text-transform: uppercase;
   color: ${COLOR_MAP.GREY_4};
+
+  @media (min-width: ${materialSMBreakpoint}) {
+    color: ${COLOR_MAP.GREY_3};
+    padding: 0.5rem 0 0 0.5rem;
+  }
 `;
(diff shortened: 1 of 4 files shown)


src/components/NewLocationPage/ChartTabs/ChartTab.style.tsx
```diff
@@ -0,0 +1,31 @@
+import styled from 'styled-components';
+import { COLOR_MAP } from 'common/colors';
+
+export const TabsContainer = styled.div`
+  display: flex;
+  flex-direction: row;
```
chasulin, line 6: `flex-direction` defaults to `row`, so you dont need to declare it

src/components/NewLocationPage/ChartTabs/ChartTab.style.tsx
```diff
@@ -0,0 +1,31 @@
...
+// TODO: Figure out transforming color to black on hover or click?
+export const TabTitle = styled.div`
+  text-transform: uppercase;
+  color: ${COLOR_MAP.GREY_4};
+  margin-bottom: 0.5rem;
+`;
+
+export const TabContent = styled.div`
+  display: flex;
+  flex-direction: row;
```
chasulin, line 24: [ditto here](https://github.com/covid-projections/covid-projections/pull/3703/files#r643399583)

src/components/NewLocationPage/ChartTabs/ChartTab.style.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const TabsContainer = styled.div`
+  display: flex;
+`;
+
+export const Tab = styled.div`
+  display: flex;
+  flex-direction: column;
+  min-width: 4rem;
+  max-width: 10.5rem;
```
chasulin, line 13: removing `min-width` and `max-width`, and adding `width: fit-content` should do the trick

src/components/NewLocationPage/ChartTabs/ChartTab.style.tsx
```diff
@@ -0,0 +1,48 @@
+import styled from 'styled-components';
+import { COLOR_MAP } from 'common/colors';
+import { materialSMBreakpoint } from 'assets/theme/sizes';
+
+export const TabsContainer = styled.div`
+  display: flex;
+`;
+
+export const Tab = styled.div`
```
chasulin, line 9: for accessibility purposes, tabs should be buttons in the dom (they're clickable, they do an action, etc). check out [material ui tabs](https://material-ui.com/components/tabs/)--they have a lot of functionality and accessibility considerations built in

chasulin: and here is an example of mui tabs currently used in the code: [component](https://github.com/covid-projections/covid-projections/blob/develop/src/components/Explore/ExploreTabs.tsx) and [styles](https://github.com/covid-projections/covid-projections/blob/develop/src/components/Explore/Explore.style.ts#L44)

chasulin (approved):

chasulin (commented):
one thing- it looks like theres a different opacity for non-active tabs (note that they look a little grayed out)-

Screen Shot 2021-06-08 at 1 17 17 PM

PR #3712, opened 2021-06-01
Create footer for metric charts

faihegberg:
This PR creates a standardized footer specifically for metric charts. The footer accepts props common to metric charts.


src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -0,0 +1,67 @@
...
+const MetricChartFooter: React.FC<{
+  footerText: string | null;
+  shareButton: any;
+  aboutModal?: AboutModalProps;
+}> = ({ footerText, shareButton, aboutModal }) => {
+  const isMobile = useBreakpoint(600);
+  const [openModal, setOpenModal] = useState(false);
+  return (
+    <>
+      <Footer>
```
chasulin, line 28: not directly related but noticed in the code - does this footer div need `position:relative`?

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -0,0 +1,67 @@
...
+export interface AboutModalProps {
+  header: string;
+  body: string;
+  links: Link[];
+}
+
+const MetricChartFooter: React.FC<{
+  footerText: string | null;
+  shareButton: any;
+  aboutModal?: AboutModalProps;
```
chasulin, line 22: _i think_ all of the footers will have an info modal, so we can make this `aboutModal` prop required and nix the `?`

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -0,0 +1,67 @@
...
+}> = ({ footerText, shareButton, aboutModal }) => {
+  const isMobile = useBreakpoint(600);
+  const [openModal, setOpenModal] = useState(false);
+  return (
+    <>
+      <Footer>
+        <FooterText>
+          {footerText}&nbsp;
+          {!isMobile && aboutModal && (
```
chasulin, line 31: instead of rendering things with isMobile and the useBreakpoint hook, see the uses of [MobileOnly](https://github.com/covid-projections/covid-projections/blob/develop/src/components/NewLocationPage/Shared/Shared.style.tsx#L53) and [DesktopOnly](https://github.com/covid-projections/covid-projections/blob/develop/src/components/NewLocationPage/Shared/Shared.style.tsx#L57) components

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -0,0 +1,67 @@
...
+          <ModalButton onClick={() => setOpenModal(true)}>
+            <AboutText>About this data</AboutText>
+          </ModalButton>
+          <NewDialog
+            open={openModal}
+            closeDialog={() => setOpenModal(false)}
+            header={aboutModal.header}
+            body={aboutModal.body}
+            links={aboutModal.links}
+          />
```
chasulin, line 59: we can reduce the redundancy by either: extracting `ModalButton`+`NewDialog` out into their own component or making a reusable props obj before the return statement, ie: 

```
 const dialogProps = {
    open:openModal,
    closeDialog: () => setOpenModal(false),
    header: aboutModal.header,
    body: aboutModal.body,
    links: aboutModal.links
 }
```
and then:
`<NewDialog {...dialogProps} />`

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export interface AboutModalProps {
+  header: string;
+  body: string;
+  links: Link[];
+}
+
+const MetricChartFooter: React.FC<{
+  footerText: string | null;
+  shareButton: any;
```
chasulin, line 21: want to render the share button directly in `MetricChartFooter` rather than passing it in as a prop? (and instead pass in the metric-specific share button props [that start here](https://github.com/covid-projections/covid-projections/blob/develop/src/components/LocationPage/ChartBlock.tsx#L69))

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -0,0 +1,80 @@
...
+  return (
+    <>
+      <DesktopOnly>
+        <Footer>
+          <FooterText>
+            {footerText}&nbsp;
+            <ModalButton onClick={() => setOpenModal(true)}>
+              <AboutText>About this data</AboutText>
+            </ModalButton>
+            <NewDialog
```
chasulin, line 51: you can spead dialogProps here too, so: `<NewDialog {...dialogProps} />`

faihegberg: Oh whoops! Yes, that is supposed to also be there 😅

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+export interface ShareButtonProps {
+  region: Region;
+  stats: MetricValues;
+  chartIdentifier: number;
+  showEmbedButton?: boolean;
+}
+
+const MetricChartFooter: React.FC<{
+  footerText: string | null;
```
chasulin, line 30: i think this text will always be defined when the footer is rendered. so `footerText: string;` should suffice

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -0,0 +1,91 @@
...
+  disclaimer: string | null;
```
chasulin, line 31: if theres no disclaimer for the location, i think it will just be undefined, not null. so `disclaimer?: string;` would work

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -0,0 +1,91 @@
...
+        <Footer>
+          <FooterText>
+            {footerText}&nbsp;
+            <ModalButton onClick={() => setOpenModal(true)}>
+              <AboutText>About this data</AboutText>
+            </ModalButton>
+            {disclaimer && (
+              <>
+                <br />
+                <br />
```
chasulin, line 55: i think leaving the line break to css rather than two `<br/>` might render a little cleaner//allow us to control the amount of spacing (the space created by two `<br />` feels a little too large). adding a wrapping element with margin around the `disclaimer` would work

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -0,0 +1,80 @@
...
+const MetricChartFooter: React.FC<{
+  footerText: string;
+  disclaimer?: string;
+  shareButtonProps: ShareButtonProps;
+  aboutModal: AboutModalProps;
+}> = ({ footerText, disclaimer, shareButtonProps, aboutModal }) => {
+  const [openModal, setOpenModal] = useState(false);
+  const dialogProps = {
+    open: openModal,
+    closeDialog: () => setOpenModal(false),
```
chasulin, line 39: you can use [the useDialog hook](https://github.com/covid-projections/covid-projections/blob/develop/src/common/hooks/useDialog.ts) for some of these props. you can see [how its used here](https://github.com/covid-projections/covid-projections/blob/develop/src/components/Dialogs/MetricInfoDialog.stories.tsx#L26)

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -0,0 +1,80 @@
...
+    header: aboutModal.header,
+    body: aboutModal.body,
+    links: aboutModal.links,
+  };
+  return (
+    <>
+      <DesktopOnly>
+        <Footer>
```
chasulin, line 47: maybe lets call this div `Row` instead of `Footer`

chasulin (approved):

PR #3725, opened 2021-06-03
Explore feedback items

chasulin:
- Sets maxY of percent vaccinated charts to 100%
- Makes location list in button semicolon separated instead of comma separated
- Fixes location selector modal (was opening upward on mobile and going off screen--now always opens downward)
- Misc. style edits
- Replaces uses of `gap` property with margins, as `gap` doesnt seem to work on phones

### DIFF
--- a/src/components/Explore/utils.ts
+++ b/src/components/Explore/utils.ts
@@ -167,6 +167,7 @@ interface ExploreMetricDescription {
   seriesList: SerieDescription[];
   dataMeasure: DataMeasure;
   yAxisDecimalPlaces: number;
+  maxY?: number;
 }
 
 export const exploreMetricData: {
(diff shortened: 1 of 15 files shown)


src/components/Explore/utils.ts
```diff
@@ -257,11 +258,12 @@ export const exploreMetricData: {
     ],
   },
   [ExploreMetric.VACCINATIONS_FIRST_DOSE]: {
-    title: 'Percent Vaccinated (1+ dose)',
-    name: 'Percent Vaccinated (1+ dose)',
+    title: 'Percent vaccinated (1+ dose)',
```
chasulin, line 261: just fixing the casing

pnavarrc (approved):
Looks good!

PR #3743, opened 2021-06-07
Unify padding/spacing on location page

chasulin:
Kept colored borders for QA purposes. will remove them before merging

### DIFF
--- a/src/components/Charts/Charts.style.ts
+++ b/src/components/Charts/Charts.style.ts
@@ -21,12 +21,7 @@ function palette(props: any) {
   return props.theme.palette.chart;
 }
 
-export const ChartContainer = styled.div`
-  /* TODO(@pnavarrc): This negative margin breaks the auto-size of the chart */
-  @media (min-width: 996px) {
-    margin-left: -3rem;
-  }
-`;
+export const ChartContainer = styled.div``;
 
 export const PositionRelative = styled.div`
   position: relative;
--- a/src/components/LocationPage/LocationPageBlock.style.ts
+++ b/src/components/LocationPage/LocationPageBlock.style.ts
@@ -6,7 +6,6 @@ import {
 
 export const BlockContainer = styled.div`
   margin: 0 auto 4.5rem;
-  border: 1px dotted green;
 
   @media (min-width: ${materialSMBreakpoint}) {
     max-width: ${props =>
@@ -25,7 +24,6 @@ export const BlockContainer = styled.div`
 `;
 
 export const WidthContainer = styled.div`
-  border: 1px solid red;
   margin: auto;
 
   @media (min-width: ${countyMapToFixedBreakpoint}) {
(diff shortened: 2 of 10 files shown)


chasulin:
[Padding too tight on the bottom part of location page](https://trello.com/c/sCwmZKn2/1372-padding-too-tight-on-the-bottom-part-of-location-page)

src/components/LocationPage/LocationPageBlock.style.ts
```diff
@@ -5,28 +5,31 @@ import {
...
   }
 
   @media print {
     margin: 0 auto;
     padding: 0 1rem;
   }
 `;
+
+export const WidthContainer = styled.div`
+  border: 1px solid red;
```
chasulin, line 28: kept borders for QA for now

src/components/Charts/Charts.style.ts
```diff
@@ -21,12 +21,7 @@ function palette(props: any) {
   return props.theme.palette.chart;
 }
 
-export const ChartContainer = styled.div`
-  /* TODO(@pnavarrc): This negative margin breaks the auto-size of the chart */
-  @media (min-width: 996px) {
-    margin-left: -3rem;
```
chasulin, line 27: this negative margin was making the charts spill outside the container. what was the use case here?

chasulin: Screen Shot 2021-06-07 at 11 57 22 AM

src/components/NewLocationPage/AboveTheFold/AboveTheFold.style.tsx
```diff
@@ -1,19 +1,30 @@
...
+export const ContentContainer = styled.div`
+  margin: auto;
+  display: flex;
+  border: 1px solid red;
+  width: fit-content;
+
+  @media (min-width: ${countyMapToFixedBreakpoint}) {
+    width: 1220px; // plus 2 rem margin
+    width: ${props =>
+      props.theme.spacingTheme.locationPage.widthContentWithStickyMap};
```
chasulin, line 27: since we want the `position:fixed` county map to be **inside** of the centered content container, we have to hard code the width. otherwise, the centered div will only include the non-fixed items (ie. the above the fold grid)

src/components/Charts/Charts.style.ts
pnavarrc: I _think_ the idea was to align the invisible y-axis with the text content, but I don't think that helps, maybe we should just follow the spacing of the container

pnavarrc: also, the negative margins were already there when I wrote that, I think I wanted to get rid of them but I didn't back then

chasulin: ooo got it. yeah if you dont have objections, lets nix the negative margin

src/components/LocationPage/LocationPageBlock.style.ts
```diff
@@ -5,28 +5,31 @@ import {
 } from 'assets/theme/sizes';
 
 export const BlockContainer = styled.div`
-  margin: 0 1rem 4.5rem;
+  margin: 0 auto 4.5rem;
+  border: 1px dotted green;
```
pnavarrc, line 9: Remember to remove these before merging!

pnavarrc (approved):
Looks good! Remember to remove the guide lines!

PR #3754, opened 2021-06-07
Metric chart modals

chasulin:
- Adds metric modal example to storybook (uses content that is used for metric info tooltip (cms), provenance tooltip (code + projections), + how it's calculated tooltip (cms))
- Deletes old `Dialog` files

Example-- (note: the two "Learn more about [[metric]]" links will be deleted when these are implemented. The modal copy is pulling from the tooltips that are currently on prod, so it's not yet editable)

Screen Shot 2021-06-07 at 7 49 53 PM


src/components/Dialogs/Dialog.style.tsx
```diff
@@ -58,4 +63,15 @@ export const DialogContent = styled(MuiDialogContent)`
 export const StyledIconButton = styled(IconButton)`
   padding: 0;
   margin-left: 0.75rem;
+  display: flex;
+  align-self: flex-start;
+`;
```
pnavarrc, line 68: Consider disabling the ripple effect and adding the focus styling (outline) that we use in other buttons

src/components/Dialogs/MetricInfoDialog.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export const Example = () => {
+  const region = regions.findByFipsCodeStrict('25');
+
+  const [projection, setProjection] = useState<Projection>();
+
+  useEffect(() => {
+    const fetchProjection = () => getProjectionForRegion(region);
+    fetchProjection().then(setProjection);
+  }, [region]);
```
pnavarrc, line 24: we can hardcode the projection since the region is not dynamic right?

```tsx
const region = regions.findByFipsCodeStrict('25');
const projection = getProjectionForRegion(region);

// ...

const provenance = getSourcesForMetric(projection.annotations, exampleMetric);
// ...
```

src/components/Dialogs/Dialog.style.tsx
chasulin: disabled ripple. the blue focus outline should be rendering

Screen Shot 2021-06-07 at 8 10 27 PM

src/components/Dialogs/MetricInfoDialog.stories.tsx
chasulin: `getProjectionForRegion` is async

pnavarrc (approved):
Looks good!

src/components/Dialogs/MetricInfoDialog.stories.tsx
chasulin: either way, fetching projections/this `useEffect` etc are only for the sake of wiring this up in storybook and grabbing the region's provenance info. when this is implemented in ChartBlock, projections will already be available

chasulin: [this is where provenance is currently grabbed](https://github.com/covid-projections/covid-projections/blob/develop/src/components/LocationPage/ChartBlock.tsx#L34)

PR #3755, opened 2021-06-08
Create compare location tabs

faihegberg:
This PR creates new location tabs for the compare table and display them in storybook.

### DIFF
--- a/src/components/NewLocationPage/CompareLocationTabs/CompareLocationTabs.style.tsx
+++ b/src/components/NewLocationPage/CompareLocationTabs/CompareLocationTabs.style.tsx
@@ -9,7 +9,9 @@ export const ButtonGroup = styled(ToggleButtonGroup)`
   border-radius: 4px;
 `;
 
-export const Button = styled(ToggleButton)`
+export const Button = styled(ToggleButton).attrs(props => ({
+  disableRipple: true,
+}))`
   ${props => props.theme.fonts.regularBook};
   padding: 0.75rem;
   border: none;
@@ -23,6 +25,9 @@ export const Button = styled(ToggleButton)`
     background-color: white;
     .MuiToggleButton-label {
       color: ${COLOR_MAP.BLACK};
+    }
+    &:hover {
+      background-color: white;
     }
     border: 1px solid ${COLOR_MAP.GREY_3};
     border-radius: 3px;
--- a/src/components/NewLocationPage/CompareLocationTabs/CompareLocationTabs.style.tsx
+++ b/src/components/NewLocationPage/CompareLocationTabs/CompareLocationTabs.style.tsx
@@ -1,0 +1,29 @@
+import styled from 'styled-components';
+import { COLOR_MAP } from 'common/colors';
+import ToggleButton from '@material-ui/lab/ToggleButton';
+import ToggleButtonGroup from '@material-ui/lab/ToggleButtonGroup';
+
+export const ButtonGroup = styled(ToggleButtonGroup)`
+  background-color: ${COLOR_MAP.GREY_1};
+  border: 1px solid ${COLOR_MAP.GREY_2};
+  border-radius: 4px;
+`;
+
+export const Button = styled(ToggleButton)`
+  padding: 10px;
+  border: 0;
+  .MuiToggleButton-label {
+    color: ${COLOR_MAP.GREY_3};
+    text-transform: none;
+    line-height: 1rem;
+  }
+  &.Mui-selected {
+    background-color: white;
+    .MuiToggleButton-label {
+      color: ${COLOR_MAP.BLACK};
+    }
+    font-weight: 500;
+    border: 1px solid ${COLOR_MAP.GREY_3};
+    border-radius: 3px;
+  }
+`;
--- a/src/components/NewLocationPage/CompareLocationTabs/CompareLocationTabs.style.tsx
+++ b/src/components/NewLocationPage/CompareLocationTabs/CompareLocationTabs.style.tsx
@@ -10,19 +10,20 @@ export const ButtonGroup = styled(ToggleButtonGroup)`
 `;
 
 export const Button = styled(ToggleButton)`
-  padding: 0.63rem;
+  ${props => props.theme.fonts.regularBook};
+  padding: 0.75rem;
   border: none;
   .MuiToggleButton-label {
-    color: ${COLOR_MAP.GREY_3};
+    color: ${COLOR_MAP.GRAY_BODY_COPY};
     text-transform: none;
-    line-height: 1rem;
+    line-height: 1.4;
   }
   &.Mui-selected {
+    ${props => props.theme.fonts.regularBookMidWeight};
     background-color: white;
     .MuiToggleButton-label {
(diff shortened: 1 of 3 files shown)


src/components/NewLocationPage/CompareLocationTabs/CompareLocationTabs.tsx
```diff
@@ -0,0 +1,59 @@
...
+    case GeoScopeFilter.NEARBY:
+      return 'Nearby';
+    case GeoScopeFilter.STATE:
+      return stateId;
+    case GeoScopeFilter.COUNTRY:
+      return 'USA';
+    default:
+      return null;
+  }
+}
```
faihegberg, line 23: Not the biggest fan of this implementation 🤔
- Currently `HomepageLocationScope` is used as a possible type in the parameter, although there shouldn't be a case when an object of this type will be passed to this function. The only reason it's currently there is to prevent an error.
- Straight up returning a parameter as a return value seems inefficient. `StateId` is merely there as I try to be explicit about covering all possible types of `GeoScopeFilter`.
- Default case should never be reached in this case... but again only there for the purpose of being explicit.
IMO, this is just all around not the best function I came up with and I would love to talk through a potentially better approach.

src/components/NewLocationPage/CompareLocationTabs/CompareLocationTabs.style.tsx
```diff
@@ -0,0 +1,29 @@
...
+    color: ${COLOR_MAP.GREY_3};
+    text-transform: none;
+    line-height: 1rem;
+  }
+  &.Mui-selected {
+    background-color: white;
+    .MuiToggleButton-label {
+      color: ${COLOR_MAP.BLACK};
+    }
+    font-weight: 500;
```
chasulin, line 25: `${props => props.theme.fonts.regularBookMidWeight}` rather than hard coding the font weight

src/components/NewLocationPage/CompareLocationTabs/CompareLocationTabs.style.tsx
```diff
@@ -0,0 +1,29 @@
...
+import ToggleButtonGroup from '@material-ui/lab/ToggleButtonGroup';
+
+export const ButtonGroup = styled(ToggleButtonGroup)`
+  background-color: ${COLOR_MAP.GREY_1};
+  border: 1px solid ${COLOR_MAP.GREY_2};
+  border-radius: 4px;
+`;
+
+export const Button = styled(ToggleButton)`
+  padding: 0.63rem;
```
chasulin, line 13: we can just round to `.75rem`

src/components/NewLocationPage/CompareLocationTabs/CompareLocationTabs.style.tsx
chasulin: also- re: the mocks, it looks like the increased font weight should only apply to the selected button (its currently applying to all of them)

chasulin: so i think if you move the midWeight style up a line (into the selected styles), you should be good to go

src/components/NewLocationPage/CompareLocationTabs/CompareLocationTabs.style.tsx
```diff
@@ -0,0 +1,29 @@
...
+  border: none;
+  .MuiToggleButton-label {
+    color: ${COLOR_MAP.GREY_3};
```
chasulin, line 16: lets darken this to `${COLOR_MAP.GRAY_BODY_COPY}`

chasulin: GREY_3 on the gray background is not accessible (screenshot below, if you use the inspector and hover over the span, youll notice that the contrast is too low)
Screen Shot 2021-06-09 at 11 13 04 AM

src/components/NewLocationPage/CompareLocationTabs/CompareLocationTabs.style.tsx
```diff
@@ -0,0 +1,29 @@
...
+    text-transform: none;
+    line-height: 1rem;
```
chasulin, line 18: lets standardize to `line-height: 1.4`. when we don't specify a unit (ie, `1.4` instead of `1rem`), the # is multiplied by the element's font-size, so it stays relative/scales

src/components/NewLocationPage/CompareLocationTabs/CompareLocationTabs.style.tsx
faihegberg: 👍🏼

src/components/NewLocationPage/CompareLocationTabs/CompareLocationTabs.style.tsx
faihegberg: 👍🏼

chasulin (commented):
If you click on the selected option, it looks like it de-selects it, leaving none of the options visually selected

Screen Shot 2021-06-14 at 1 14 41 PM

src/components/NewLocationPage/CompareLocationTabs/CompareLocationTabs.style.tsx
```diff
@@ -0,0 +1,30 @@
...
+  padding: 0.75rem;
+  border: none;
+  .MuiToggleButton-label {
+    color: ${COLOR_MAP.GRAY_BODY_COPY};
+    text-transform: none;
+    line-height: 1.4;
+  }
+  &.Mui-selected {
+    ${props => props.theme.fonts.regularBookMidWeight};
+    background-color: white;
```
chasulin, line 23: lets also make sure the hover color for the selection option is white (its currently a dark gray), so right below this line:
```
&:hover {
    background-color: white;
}
```

src/components/NewLocationPage/CompareLocationTabs/CompareLocationTabs.style.tsx
```diff
@@ -0,0 +1,30 @@
...
+import ToggleButton from '@material-ui/lab/ToggleButton';
+import ToggleButtonGroup from '@material-ui/lab/ToggleButtonGroup';
+
+export const ButtonGroup = styled(ToggleButtonGroup)`
+  background-color: ${COLOR_MAP.GREY_1};
+  border: 1px solid ${COLOR_MAP.GREY_2};
+  border-radius: 4px;
+`;
+
+export const Button = styled(ToggleButton)`
```
chasulin, line 12: these buttons need some MUI-specific props to remove the ripple that occurs when clicking. you can see an example of how those props are applied [here in ShareButtons.style](https://github.com/covid-projections/covid-projections/blob/develop/src/components/ShareButtons/ShareButtons.style.tsx#L8)

faihegberg:
> If you click on the selected option, it looks like it de-selects it, leaving none of the options visually selected
> 
> Screen Shot 2021-06-14 at 1 14 41 PM

Just fixed that :)

chasulin (approved):

PR #3764, opened 2021-06-08
Metric chart footers

chasulin:
- Removes summary text and info tooltip from above chart
- Implements new footer format with info modal


src/components/Dialogs/utils.ts
```diff
@@ -8,8 +8,17 @@ export interface MetricModalContent {
...
+  learnLink: string;
 }
 
+const metricToLearnLink: { [key in Metric]: string } = {
+  [Metric.CASE_DENSITY]: '/covid-risk-levels-metrics#daily-new-cases',
+  [Metric.CASE_GROWTH_RATE]: '/covid-risk-levels-metrics#infection-rate',
+  [Metric.HOSPITAL_USAGE]: '/covid-risk-levels-metrics#icu-capacity-used',
+  [Metric.POSITIVE_TESTS]: '/covid-risk-levels-metrics#positive-test-rate',
+  [Metric.VACCINATIONS]: '/covid-risk-levels-metrics#percent-vaccinated',
+};
```
pnavarrc, line 20: Can we grab this from the CMS? `src/cms-content/tooltips/metric-calculation-tooltips.json` has the CTA (even if we are not reusing the title)

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -28,56 +24,106 @@ export interface ShareButtonProps {
   showEmbedButton?: boolean;
 }
 
+interface DialogProps {
+  open: boolean;
+  closeDialog: any;
+  openDialog: any;
```
pnavarrc, line 30: Maybe give these the correct types?

src/components/Dialogs/utils.ts
chasulin: and by down the line i mean.. tomorrow or thursday

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
chasulin: oh whoops! yep

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -1,25 +1,21 @@
...
   AboutText,
   ModalButton,
-  DisclaimerWrapper,
+  Wrapper,
+  OverrideDisclaimer,
 } from './ChartFooter.style';
-import { DialogMain } from 'components/Dialogs';
-import { Link } from 'cms-content/modals';
+import { DialogMain, MetricInfoDialogInner } from 'components/Dialogs';
+// import { Link } from 'cms-content/modals';
```
pnavarrc, line 12: we can probably remove that

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -28,56 +24,106 @@ export interface ShareButtonProps {
...
+  open: boolean;
+  closeDialog: () => void;
+  openDialog: () => void;
+  modalContent: MetricModalContent;
+  modalHeader: string;
+}
+
+const ShareButtonBlock: React.FC<ShareButtonProps> = ({
+  ...shareButtonProps
+}) => {
```
pnavarrc, line 37: Isn't this the same than 

```tsx
const ShareButtonBlock: React.FC<ShareButtonProps> = (shareButtonProps) => {
  // ...
};
```

chasulin: whoops, yes

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -28,56 +24,106 @@ export interface ShareButtonProps {
...
+  const dialogProps = {
+    open,
+    closeDialog,
+    header: modalHeader,
+    links: [
+      {
+        cta: 'Learn more',
+        url: modalContent.learnLink,
+      },
```
pnavarrc, line 60: We should add a good `aria-label` for this link, although given that the modal covers a single metric, a screen reader user will probably know that the link is to Lean more about that metric.

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -28,56 +24,106 @@ export interface ShareButtonProps {
...
+
+const ShareButtonBlock: React.FC<ShareButtonProps> = ({
+  ...shareButtonProps
+}) => {
+  return (
+    <ButtonContainer>
+      <ShareButtons {...shareButtonProps} />
+    </ButtonContainer>
+  );
+};
```
pnavarrc, line 43: Not sure if we need this component, is really small 😃

chasulin: yeah, its just repeated twice down in the main component, so i pulled it out into its own thing

public/admin/config.yml
```diff
@@ -899,3 +899,7 @@ collections:
             label: URL
             widget: string
             hint: 'Ex: /covid-risk-levels-metrics#vulnerability'
+          - name: ariaLabel
```
pnavarrc, line 902: 🙏

src/components/LocationPage/ChartBlock.tsx
```diff
@@ -40,55 +35,50 @@ function ChartBlock(props: {
...
-              showEmbedButton={false}
-            />
-          </LocationPageSectionFooter>
+          <MetricChartFooter
+            footerText={getMetricStatusText(metric, projections)}
+            overrideDisclaimer={overrideDisclaimer}
+            shareButtonProps={shareButtonProps}
+            modalContent={modalContent}
+            modalHeader={getMetricName(metric)}
+          />
```
pnavarrc, line 70: How about passing `region`, `projection` and `metric` as props to `MetricChartFooter` and calculating each part in that component instead? That would make the API of `MetricChartFooter` look a bit more like `MetricChart` and hide implementation details that are not relevant in `ChartBlock`

src/components/LocationPage/ChartBlock.tsx
```diff
@@ -40,55 +35,50 @@ function ChartBlock(props: {
...
         </>
       )}
     </>
   );
 }
 
-export function renderDisclaimer(
+export function getOverrideDisclaimer(
```
pnavarrc, line 77: This function could be moved to `MetricChartFooter` too

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -1,25 +1,20 @@
...
 import type { MetricValues } from 'common/models/Projections';
-import { MarkdownBody } from 'components/Markdown';
-
-export interface AboutModalProps {
-  header: string;
-  body: string;
-  links: Link[];
-}
+import { useDialog } from 'common/hooks';
+import { MetricModalContent } from 'components/Dialogs/utils';
```
pnavarrc, line 17: Can this import be merged with the other import from `component/Dialogs`?

src/components/NewLocationPage/ChartFooter/MetricChartFooter.tsx
```diff
@@ -28,56 +31,115 @@ export interface ShareButtonProps {
...
+  region: Region;
+  stats: any;
+}> = ({ metric, projections, region, stats }) => {
+  const provenance = getSourcesForMetric(
+    projections.primary.annotations,
+    metric,
+  );
+  const overrideDisclaimer = getOverrideDisclaimer(region, metric, provenance);
+  const modalContent = getMetricModalContent(region, metric, provenance);
+  const modalHeader = getMetricNameExtended(metric);
```
chasulin, line 94: using extended metric name so the percent vaccinated modal header isnt just 'Vaccinated' (this change also changes Daily new cases --> Daily new cases per 100k population, which doesnt seem like an issue)

pnavarrc (approved):
Looks good! 🚢

PR #3771, opened 2021-06-09
Ensure location URL doesn't get added to history stack twice

faihegberg:
This PR fixes the issue of the browser back button needing to be clicked twice to navigate back to the homepage (after the user clicks on a location on the compare table).


faihegberg:
[Back button when navigating from compare table takes two attempts](https://trello.com/c/3gsNJVjm/1407-back-button-when-navigating-from-compare-table-takes-two-attempts)

chasulin (approved):
🚢

PR #3773, opened 2021-06-09
Investigate frozen zoom on location input

faihegberg:
This PR fixes the issue of the screen freezing after the user adds a new location on the explore chart. This only happens on mobile. Adjusting the font size to be at least 16 px / 1 rem prevents the screen from freezing.

Some forum discussion regarding the issue: https://stackoverflow.com/questions/2989263/disable-auto-zoom-in-input-text-tag-safari-on-iphone


chasulin (approved):
wooo!

faihegberg:
[Tapping "add location" in Trends on mobile zooms in the UI.  On some old phone sizes in such a way so as to make it unusable.  It doesn't zoom out after selection is completed.](https://trello.com/c/dlvdYb3J/1420-tapping-add-location-in-trends-on-mobile-zooms-in-the-ui-on-some-old-phone-sizes-in-such-a-way-so-as-to-make-it-unusable-it-does)

PR #3787, opened 2021-06-12
Add 'per 100K' to the chart metric labels

schamp:
Single series, not normalized:
image
(menu):
image

Multiple series, normalized:
image
(menu):
image

This appears to be the only change needed to implement this, but if I'm missing something, please call it out and I'll add it.

### DIFF
--- a/src/components/Explore/utils.ts
+++ b/src/components/Explore/utils.ts
@@ -95,6 +95,11 @@ export const EXPLORE_METRICS = [
   ExploreMetric.POSITIVITY_RATE,
 ];
 
+// Note that these specifically are counts, not percentages, and can normalized
+// to per 100K when multiple locations are displayed, in which case we manipulate
+// the labels to indicate the 'per 100K'.
+// Therefore, don't add a 'percentage' type metric to this without updating
+// that code accordingly.
 export const ORIGINAL_EXPLORE_METRICS = [
   ExploreMetric.CASES,
   ExploreMetric.DEATHS,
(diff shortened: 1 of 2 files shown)


schamp:
I just noticed the Trello card said "100k", and I used '100K' but looking in the code we're inconsistent between using '100k' and '100K'.  So if you want me to switch to '100k' just let me know.

schamp:
[We never tell you that our metrics are per 100k in Trends](https://trello.com/c/xGNtW4wj/1421-we-never-tell-you-that-our-metrics-are-per-100k-in-trends)

chasulin (commented):
only the first four metrics (cases, deaths, hospitalizations, icu hospitalizations) are normalized when comparing multiple locations—the other four (% vaccinated first/completed, icu capacity used, positive test rate) are percentages

one potential solution could be--

```
function append100k(metric: ExploreMetric, multiLocation: boolean): boolean {
  return multiLocation && ORIGINAL_EXPLORE_METRICS.includes(metric);
}

export function getMetricLabels(multiLocation: boolean): string[] {
  const getTitleNormalized = (metric: ExploreMetric) =>
    getTitle(metric) + ' per 100k';

  return EXPLORE_METRICS.map((metric: ExploreMetric) =>
    append100k(metric, multiLocation)
      ? getTitleNormalized(metric)
      : getTitle(metric),
  );
}
```
which also means, in `Explore.tsx`, putting `selectedLocations.length > 1` into a variable and passing that into `getMetricLabels`

thanks for grabbing this!! apologies for not seeing the message sooner

schamp:
ah, good catch, yeah, I obviously didn't even read this :facepalm: will make the suggested update : )

schamp:
That's better, eh? 
image

src/components/Explore/utils.ts
```diff
@@ -416,8 +421,15 @@ export function getChartIdByMetric(metric: ExploreMetric) {
   return exploreMetricData[metric].chartId;
 }
 
-export function getMetricLabels() {
-  return EXPLORE_METRICS.map(getTitle);
+export function getMetricLabels(normalizeData: boolean): string[] {
```
chasulin, line 424: rather than passing in `normlizeData`, i think we want to pass in a `selectedLocations.length > 1 `variable (so adding something like `const multiLocation = selectedLocations.length > 1` in Explore.tsx and passing that into `getMetricLabels`)

normalizeData takes `currentMetric` into account, and so the "per 100k" label only renders in the dropdown menu when a normalized metric is selected as `currentMetric`, ie:

Screen Shot 2021-06-12 at 1 08 28 PM

but we'd want the "per 100k" label to render whenever multiple locations are selected, regardless of what `currentMetric` is

schamp: oooooh, I see, thanks for that explanation

schamp: fixed.

chasulin (approved):
sweeeeet 🚢

PR #3791, opened 2021-06-12
Remove black border for share card image

schamp:


schamp:
need to fixup the branch first oops

PR #3793, opened 2021-06-12
Remove black border for share card image

schamp:
And move updated date into a subheader

### DIFF
--- a/src/components/SocialLocationPreview/SocialLocationPreview.style.tsx
+++ b/src/components/SocialLocationPreview/SocialLocationPreview.style.tsx
@@ -7,7 +7,9 @@ export const Wrapper = styled.div<{
   border?: Boolean;
 }>`
   box-shadow: ${props =>
-    props.border ? 'none' : '0px 12px 40px rgba(0, 0, 0, 0.2)'};
+    props.noShadow || props.border
+      ? 'none'
+      : '0px 12px 40px rgba(0, 0, 0, 0.2)'};
   border: ${props => (props.border ? `2px solid ${palette.divider}` : 'none')};
   border-radius: 4px;
   color: black;
@@ -29,8 +31,12 @@ export const PreviewHeader = styled.div`
   ${PreviewHeaderStyles}
 `;
 
-export const USMapPreviewHeader = styled.div<{ sideLegend: Boolean }>`
-  border-bottom: 1px solid rgba(0, 0, 0, 0.12);
+export const USMapPreviewHeader = styled.div<{
+  border?: Boolean;
+  sideLegend: Boolean;
+}>`
+  border-bottom: ${props =>
+    props.border ? '1px solid rgba(0, 0, 0, 0.12)' : 'none'};
   padding: 1.5rem;
   display: flex;
   flex-direction: ${props => (props.sideLegend ? 'row' : 'column')};
@@ -38,6 +44,11 @@ export const USMapPreviewHeader = styled.div<{ sideLegend: Boolean }>`
   justify-content: center;
   padding-bottom: 0px;
   padding-top: ${props => (props.sideLegend ? '1.5rem' : '0')};
+`;
+
+export const LastUpdatedWrapper = styled.div`
+  font-size: 0.75rem;
+  text-align: center;
 `;
 
 export const MapWrapper = styled.div`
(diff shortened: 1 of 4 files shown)


src/components/SocialLocationPreview/SocialLocationPreviewMap.tsx
```diff
@@ -34,6 +36,7 @@ const SocialLocationPreview = (props: {
   return (
     <Wrapper border={props.border}>
       <MapHeaderHeader>US COVID Risk &amp; Vaccine Tracker</MapHeaderHeader>
+      <LastUpdatedWrapper>Updated {lastUpdatedDateString}</LastUpdatedWrapper>
```
schamp, line 39: Note this changes the displayed date from the current date when the image was generated to the date the model was updated, which I believe to be the correct value to show.  If I have that wrong, LMK.

schamp:
[Remove black border for share card image](https://trello.com/c/IDTQ1xRi/1268-remove-black-border-for-share-card-image)

src/components/SocialLocationPreview/SocialLocationPreviewMap.tsx
schamp: per Igor in ticket, "match the homepage", which this does.

mikelehen (changes requested):
Thanks for tackling this!  Unfortunately it probably needs a bit more work.  The current share card looks like this:

image

Your updated one looks like this:
image

The image is intentionally 1200x630 to optimize the resolution for Twitter / Facebook / etc. sharing.  So the updated share card end up with a bunch of whitespace...  I think we want the ShareCardWrapper to be as big as possible and probably remove the drop shadow so that it's just a uniform white background in the image.

The actual share image generation happens via code that runs on-demand via Firebase Functions by visiting `/internal/share-image/` in a browser and then captures a screenshot.  For reference, that code lives [here](https://github.com/covid-projections/covid-projections/blob/5074966bce8c468e2ab32026d0a34b24466815e2/firebase/functions/src/screenshots.ts#L33) but the important part for testing locally is to just follow this [README](https://github.com/covid-projections/covid-projections/blob/develop/firebase/README.md).

Let me know if this is too much trouble to deal with.

schamp:
Thanks for explaining, Michael, I wasn't aware of those constraints.  I'll take another stab at it.

schamp:
updated: 
image
(ignore black bar at bottom).

@mikelehen I scaled up and removed the border, LMKWYT

src/components/SocialLocationPreview/SocialLocationPreview.style.tsx
```diff
@@ -7,7 +7,9 @@ export const Wrapper = styled.div<{
   border?: Boolean;
 }>`
   box-shadow: ${props =>
-    props.border ? 'none' : '0px 12px 40px rgba(0, 0, 0, 0.2)'};
+    props.noShadow || props.border
```
mikelehen, line 10: Should there be parens around this?  As-is, I think you'll end up with `box-shadow: true` if `props.noShadow` is true? But I could easily be confused.

mikelehen (changes requested):
I think the share image looks good but it looks like this also tweaked the preview image in the share widget at the bottom of the homepage (removing the last updated text and the drop-shadow).  Not sure if both those changes are intentional? 

Before:
image

After:
image

src/components/SocialLocationPreview/SocialLocationPreview.style.tsx
schamp: Now you've confused me and I'm not sure but parens wouldn't hurt so I'll add them.

schamp:
> I think the share image looks good but it looks like this also tweaked the preview image in the share widget at the bottom of the homepage (removing the last updated text and the drop-shadow). Not sure if both those changes are intentional?
> 
> Before:
> image
> 
> After:
> image

That change was not intentional, I'll see if I can sort it out.  I had a look, but didn't notice the difference, thanks for catching it.

schamp:
Fixed, I think.  I'm not sure how much I like the preview being different from the actual share image, but it looks good in this context, so I think it's fine:
image

src/components/SocialLocationPreview/SocialLocationPreview.style.tsx
mikelehen: Looks like you didn't add them, which might mean that I was confused and `prettier` removed them because they're not necessary. 🤷

mikelehen (approved):
Thanks! Looks good.  I agree it's a bit weird that the preview doesn't match the result, but it never has...   🤷

src/components/SocialLocationPreview/SocialLocationPreview.style.tsx
schamp: yeah I added them and prettier re-removed them so :shrug:

PR #3814, opened 2021-06-15
Vaccinated geolocation cards

faihegberg:
This PR replaces the content of geolocated cards at the top of the homepage with vaccination metrics.

Note: Ignore `UseGeolocation.ts`. This will be converted back to normal before merging.


chasulin (commented):
in the case that percent vaccinated is null, it renders a gray dot and no text 

Screen Shot 2021-06-23 at 2 49 06 PM

maybe in that case, we dont render the bottom half of the card, so something like-
```
          {vaccinatedStat && (
            <LevelContainer>
              <IconContainer>
                <CircleIcon $iconColor={iconColor} />
              </IconContainer>
              <LevelDescription>{levelDescriptionCopy}</LevelDescription>
            </LevelContainer>
          )}
```

which results in -

Screen Shot 2021-06-23 at 2 52 52 PM

it still looks a little weird 🙁, but better than the gray dot and no text?

chasulin (commented):
also, this change in dot color should also apply to the search dropdown items

chasulin (approved):
whoops sorry, thought i gave this a ✅ yesterday! once you revert the useGeolocation.ts stuff, feel free to 🚢

PR #3816, opened 2021-06-15
Nest compare table in collapsible container

faihegberg:
This PR nests the compare table inside an expandable container (like the Recommendations and Vulnerabilities section). This container only appears if there are more locations than the ones displayed on the screen.


src/components/Compare/Compare.style.tsx
```diff
@@ -519,3 +519,8 @@ export const CompareHeader = styled(SectionHeader)`
     margin: 0 2.5rem 0 0;
   }
 `;
+
+export const NumberOfLocationsText = styled.span`
+  color: ${COLOR_MAP.GREY_4};
+  font-weight: normal;
```
chasulin, line 525: `${props => props.theme.fonts.regularBook}`

src/components/Compare/CompareTable.tsx
```diff
@@ -211,6 +211,44 @@ const CompareTable = (props: {
...
+      </NumberOfLocationsText>
+    </>
+  );
+
+  const containerProps = {
+    collapsedHeightMobile: 'fit-content',
+    collapsedHeightDesktop: 'fit-content',
+    tabTextCollapsed: seeAllText,
+    tabTextExpanded: seeAllText,
+    trackingLabel: 'Compare module',
```
chasulin, line 228: Maybe `Open compare modal` instead

src/components/Compare/CompareTable.tsx
```diff
@@ -232,41 +270,16 @@ const CompareTable = (props: {
...
-              <TextButton
-                onClick={onClickShowAll}
-                trackingCategory={EventCategory.COMPARE}
-                trackingLabel="view all regions button"
-              >
-                {props.viewMoreCopy}
-              </TextButton>
-            )}
-          </div>
+          <div /> {/* Need empty div to right align share button. */}
```
chasulin, line 282: [can repurpose ShareBlock](https://github.com/covid-projections/covid-projections/blob/develop/src/components/Explore/Explore.style.ts#L26) (used in explore footer to align share button to the right when there's no legend) -- and maybe move it to somewhere more central (like the share button style file, or the footer styles, etc)

src/components/ExpandableContainer/ExpandableContainer.tsx
```diff
@@ -11,12 +11,13 @@ import { EventCategory } from 'components/Analytics';
...
 export interface ExpandableContainerProps {
-  collapsedHeightDesktop: number;
-  collapsedHeightMobile: number;
+  collapsedHeightDesktop: string;
+  collapsedHeightMobile: string;
   tabTextCollapsed: React.ReactElement;
   tabTextExpanded: React.ReactElement;
   trackingLabel: string;
   trackingCategory: EventCategory;
+  onClickShowAll?: () => void;
```
chasulin, line 20: maybe a more generic name, like `secondaryOnClick`

chasulin (commented):
the number in parentheses after 'See all' should correspond to the number that comes after 'View' in the current footer on prod. in some cases, it's not the total amount of locations, but rather it's 'top 100'. for example, the modal doesn't display all 3224 counties (tooooo many)--it displays the top 100 counties

[the copy is determined here](https://github.com/covid-projections/covid-projections/blob/develop/src/components/Compare/CompareMain.tsx#L155)

faihegberg:
> the number in parentheses after 'See all' should correspond to the number that comes after 'View' in the current footer on prod. in some cases, it's not the total amount of locations, but rather it's 'top 100'. for example, the modal doesn't display all 3224 counties (tooooo many)--it displays the top 100 counties
> 
> [the copy is determined here](https://github.com/covid-projections/covid-projections/blob/develop/src/components/Compare/CompareMain.tsx#L155)

Oops! Yes, fixing that.

src/components/Compare/CompareTable.tsx
```diff
@@ -211,12 +212,26 @@ const CompareTable = (props: {
     trackCompareEvent(EventAction.OPEN_MODAL, 'Show All Locations');
   };
 
+  function getLocationLimit() {
```
chasulin, line 215: this function does the same thing as [the second half of this function](https://github.com/covid-projections/covid-projections/blob/develop/src/components/Compare/LocationTable.tsx#L253). how about we combine them?

src/components/Compare/CompareTable.tsx
```diff
@@ -211,6 +212,56 @@ const CompareTable = (props: {
...
+   * Show limited locations (100 max) in the following scenarios:
+   * - "Counties" tab selected on homepage.
+   * - "Metro areas" tab selected on homepage.
+   * - "USA" tab selected on location page.
+   */
+  const showLimitedLocations =
+    (props.isHomepage && props.homepageScope === HomepageLocationScope.STATE) ||
+    (!props.isHomepage && props.geoScope !== GeoScopeFilter.COUNTRY)
+      ? true
+      : false;
```
chasulin, line 225: we can nix the ternary and just keep the condition. it will resolve to true or false on its own, so just -
```
const showLimitedLocations =
    (props.isHomepage && props.homepageScope === HomepageLocationScope.STATE) ||
    (!props.isHomepage && props.geoScope !== GeoScopeFilter.COUNTRY)
```

chasulin (commented):
opening and then closing the modal leaves the see all arrow in the up position

Screen Shot 2021-06-21 at 5 14 24 PM

faihegberg:
> opening and then closing the modal leaves the see all arrow in the up position
> 
> Screen Shot 2021-06-21 at 5 14 24 PM

Whoops yes, addressing this now.

src/components/Compare/CompareTable.tsx
```diff
@@ -211,6 +212,55 @@ const CompareTable = (props: {
     trackCompareEvent(EventAction.OPEN_MODAL, 'Show All Locations');
   };
 
+  /**
+   * Show limited locations (100 max) in the following scenarios:
+   * - "Counties" tab selected on homepage.
+   * - "Metro areas" tab selected on homepage.
+   * - "USA" tab selected on location page.
+   */
+  const showLimitedLocations =
```
chasulin, line 221: i think this naming logic might be backwards? we show limited locations (100 max) when these conditions are false

chasulin: [as seen in the logic here](https://github.com/covid-projections/covid-projections/pull/3816/files#diff-d44f76b821573ad38ed21cd3b3888f013e0a64d6c1a813654fe2b5797a98c8feR229)

faihegberg: True 😅

src/components/ExpandableContainer/ExpandableContainer.style.tsx
```diff
@@ -19,11 +19,11 @@ export const ExpandButton = styled(LargeOutlinedButton)<{
 `;
 
 export const InnerContent = styled.div<{
-  $collapsedHeight: number;
+  $collapsedHeight: string;
   $collapsed: boolean;
 }>`
   height: ${({ $collapsed, $collapsedHeight }) =>
-    $collapsed ? `${$collapsedHeight}px` : 'fit-content'};
+    $collapsed ? `${$collapsedHeight}` : 'fit-content'};
```
chasulin, line 26: tiny thing -- no need to make $collapsedHeight a template literal. can nix the backticks and just declare `$collapsedHeight` (like you declare `$collapsed`)

chasulin (approved):

PR #3817, opened 2021-06-16
Metric chart redesign

chasulin:


src/components/Dialogs/utils.tsx
```diff
@@ -0,0 +1,115 @@
+import React from 'react';
```
chasulin, line 1: This file is mostly the same as `utils.ts`--just renamed as a `tsx` file with some stuff added

src/components/Charts/EmptyPanel.tsx
```diff
@@ -0,0 +1,18 @@
...
+  bodyText,
+  children,
+}) => {
+  return (
+    <EmptyPanelContainer>
+      {bodyText}
+      {children}
+    </EmptyPanelContainer>
+  );
+};
```
pnavarrc, line 16: We do we need the `bodyText` as a separate prop? we could just use `children` for that right?

chasulin: mm yes true

chasulin: originally it was just the `bodyText` prop, but then i noticed that the one in Explore.tsx takes `ScreenshotReady`, so i added children. sidenote- does the empty panel need `ScreenshotReady`?

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -26,7 +25,7 @@ import Recommendations from './Recommendations';
 import ShareModelBlock from 'components/ShareBlock/ShareModelBlock';
 import VaccinationEligibilityBlock from 'components/VaccinationEligibilityBlock';
 import VulnerabilitiesBlock from 'components/VulnerabilitiesBlock';
-import ChartBlock from './ChartBlock';
+// import ChartBlock from './ChartBlock';
```
pnavarrc, line 28: 👀

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -207,19 +208,26 @@ const ChartsHolder = React.memo(({ region, chartId }: ChartsHolderProps) => {
...
                   <LocationPageBlock>
-                    <ChartBlock
+                    {/* <ChartBlock
                       metric={metric}
                       projections={projections}
                       chartRef={metricRefs[metric]}
                       isMobile={isMobile}
                       region={region}
                       stats={stats}
+                    /> */}
```
pnavarrc, line 224: 👀

src/components/Dialogs/utils.tsx
chasulin: this isnt totally true anymore lol. everything below line 50 is new

src/components/Charts/EmptyPanel.tsx
pnavarrc: we might as well include it to cover the case where someone shares an empty tab (the tab could become empty after being shared, if that specific dataset has problems one day)

src/components/Charts/EmptyPanel.tsx
```diff
@@ -0,0 +1,10 @@
+/** Renders in place of a chart when there is no data */
+
+import React from 'react';
+import { EmptyPanelContainer } from 'components/Charts/Charts.style';
+
+const EmptyPanel: React.FC = ({ children }) => {
+  return <EmptyPanelContainer>{children}</EmptyPanelContainer>;
```
pnavarrc, line 7: Now that the component doesn't have any other props, we might as well just rename `EmptyPanelContainer` → `EmptyPanel` and export it. 

This is better than exporting an empty wrapper because  with the current implementation, `EmptyPanelContainer` can receive any valid props for a `div`, but `EmptyPanel` can't. For example, with the current code, we can't do this:

```tsx
<EmptyPanel style={{ padding: 30 }}>No data</EmptyPanel>
```

but if we export the style directly, the `style` prop will be picked up correctly.

src/components/Explore/Explore.tsx
```diff
@@ -339,25 +341,25 @@ const Explore: React.FunctionComponent<{
...
         )}
         {selectedLocations.length > 0 && !hasData && (
-          <Styles.EmptyPanel style={{ height: 400 }}>
+          <EmptyPanel>
             {getNoDataCopy(
               currentMetricName,
               getLocationNames(selectedLocations),
             )}
-          </Styles.EmptyPanel>
+          </EmptyPanel>
```
pnavarrc, line 356: Let's add `<ScreenshotReady />` for all of them just to be in the safe side!

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -207,24 +237,29 @@ const ChartsHolder = React.memo(({ region, chartId }: ChartsHolderProps) => {
...
-                    />
-                  </LocationPageBlock>
-                )}
-              </ErrorBoundary>
-            ))}
+            {CHART_GROUPS.map((group: ChartGroup) => {
+              const { groupHeader } = group;
+              const groupRef = chartBlockRefs[groupHeader];
+              return (
+                <ErrorBoundary key={group.groupHeader}>
```
pnavarrc, line 244: nit: Maybe use `groupHeader` directly since it's a variable above? (or remove the variable and just use `group.groupHeader`

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -207,24 +237,29 @@ const ChartsHolder = React.memo(({ region, chartId }: ChartsHolderProps) => {
...
+                  {!projections ? (
+                    <LoadingScreen />
+                  ) : (
+                    <LocationPageBlock>
```
pnavarrc, line 248: I feel like we don't need to pass `groupRef` down to the `ChartBlock` component, maybe we can just attached to `LocationPageBlock`?

src/components/Charts/ChartBlock.tsx
```diff
@@ -0,0 +1,93 @@
...
+    <>
+      <SectionHeader ref={groupRef}>{groupHeader}</SectionHeader>
+      <TabsWrapper activeTabIndex={activeTabIndex} onChangeTab={onChangeTab}>
+        {metricList.map((metricItem: MetricChartInfo, i: number) => {
+          const metricValueInfo = getValueInfo(stats, metricItem, projections);
+          return (
+            <Tab
+              key={`tab-${i}`}
+              disableRipple
+              disableFocusRipple
```
pnavarrc, line 70: disabling props could be moved to the styled component to simplify the code a bit (and keep hopes up of unifying the Tab eventually)

src/components/Charts/ChartBlock.tsx
```diff
@@ -0,0 +1,93 @@
...
+  clickedStatMetric: Metric | null;
+  groupRef: React.RefObject<HTMLDivElement>;
+}> = ({ projections, stats, group, region, groupRef, clickedStatMetric }) => {
+  const { metricList, groupHeader } = group;
+
+  // TODO (chelsi) - revisit placement of these state/setState variables
+  const [activeTabIndex, setActiveTabIndex] = useState(0);
+  const onChangeTab = (newTabIndex: number) => {
+    setActiveTabIndex(newTabIndex);
+  };
```
pnavarrc, line 36: I think we need to add event tracking for this

src/components/Charts/ChartBlock.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  // Selects correct tab after clicking summary stat + scrolling to chart block
```
pnavarrc, line 38: It feels like we might want to move the state up, however, it will be a bit weird to control all the tabs from above. We can leave it for now unless we come up with a solution!

pnavarrc (commented):
Added some notes, going for a second pass now!

src/components/Charts/ChartBlock.tsx
chasulin: yeah, [added a todo for this](https://github.com/covid-projections/covid-projections/pull/3817/files#diff-be721b17c599144023d04f06fdf031526e3e773d196005821689ac6f87f92131R32)

src/components/Charts/Groupings.tsx
```diff
@@ -0,0 +1,260 @@
...
+// Used for scrolling to correct chart group when clicking summary stat
+export const getChartGroupFromMetric = (
+  metricToScrollTo: Metric,
+): ChartGroup | null => {
+  const groupWithMetric = find(
+    CHART_GROUPS,
+    group =>
+      find(group.metricList, metric => metric.metric === metricToScrollTo) !==
+      undefined,
+  );
```
pnavarrc, line 240: Maybe it would be easier if `CHART_GROUPS` was a map instead, but this works too!

src/components/NewLocationPage/ChartFooter/ChartFooter.tsx
```diff
@@ -0,0 +1,40 @@
...
+        <AddedMetricChartFooter
+          metric={metric as ExploreMetric}
+          projections={projections}
+          region={region}
+          formattedValue={formattedValue}
+        />
+      )}
+    </>
+  );
+};
```
pnavarrc, line 38: Maybe we could add `renderFooter` in `components/Charts/Groupings` instead? I feel like that'a bit more of a natural place to put the rendering logic for each kind of metric (but we can just leave as is if there is not enough time)

pnavarrc (commented):
Added a couple of more notes!

pnavarrc (approved):
Looks good!

PR #3827, opened 2021-06-16
Hide explore legend if no data

chasulin:
We are currently rendering the legend even if theres no data/no chart -- this fixes


pnavarrc (approved):
👍

PR #3829, opened 2021-06-17
Speed up storybook builds (From ~10s to ~2.5s).

mikelehen:
I noticed the build was hanging for a long time on "70% - sealing React Docgen Typescript Plugin" so I did some googling and found https://github.com/storybookjs/storybook/issues/12585#issuecomment-701292052 which has a workaround, though it breaks the "Docs" tab in StoryBook. 😞


pnavarrc (approved):
Looks good! I think the docs tab is not as important since is mostly for types (and those are explicit in the code)

PR #3830, opened 2021-06-17
Scrolling to chart when clicking summary stat

chasulin:
i put this in a pr/branch separate of the [feature](https://github.com/covid-projections/covid-projections/pull/3817) because im not sure yet if this is the best solution

also includes some fixing of the share functionality (share quotes working for all charts, page-level share links (rather than chart level), page-level share images for newly added charts (still shares chart image for original key metric charts))

### DIFF
--- a/src/common/utils/makeChartShareQuote.ts
+++ b/src/common/utils/makeChartShareQuote.ts
@@ -1,6 +1,5 @@
 import { Metric } from 'common/metricEnum';
 import { fail, formatDecimal, formatPercent } from 'common/utils';
-import { ExploreMetric } from 'components/Explore';
 
 //TODO(chelsi): move this copy into individual metric files. remove need for hardcoded identifying numers
 export function makeChartShareQuote(
@@ -35,19 +34,4 @@ export function makeChartShareQuote(
     )} daily new cases per 100k population, according to @CovidActNow. See the chart: `;
   }
   return `I'm keeping track of ${displayName}'s vaccination progress and COVID risk level data with @CovidActNow. What does your community look like?`;
-}
-
-export function makeAddedChartShareQuote(
-  displayName: string,
-  metric: ExploreMetric,
-) {
-  if (metric === ExploreMetric.DEATHS) {
-    return `Daily COVID deaths in ${displayName}, according to @CovidActNow. See the chart: `;
-  } else if (metric === ExploreMetric.ICU_HOSPITALIZATIONS) {
-    return `COVID ICU hospitalizations in ${displayName}, according to @CovidActNow. See the chart: `;
-  } else if (metric === ExploreMetric.HOSPITALIZATIONS) {
-    return `COVID hospitalizations in ${displayName}, according to @CovidActNow. See the chart: `;
-  } else {
-    return `I'm keeping track of ${displayName}'s vaccination progress and COVID risk level data with @CovidActNow. What does your community look like?`;
-  }
 }
--- a/src/common/utils/makeChartShareQuote.ts
+++ b/src/common/utils/makeChartShareQuote.ts
@@ -1,8 +1,9 @@
 import { Metric } from 'common/metricEnum';
 import { fail, formatDecimal, formatPercent } from 'common/utils';
+import { ExploreMetric } from 'components/Explore';
 
 //TODO(chelsi): move this copy into individual metric files. remove need for hardcoded identifying numers
-export default function makeChartShareQuote(
+export function makeChartShareQuote(
   displayName: string,
   stats: any = {},
   chartIdentifier: number,
@@ -33,5 +34,20 @@ export default function makeChartShareQuote(
       1,
     )} daily new cases per 100k population, according to @CovidActNow. See the chart: `;
   }
-  return `I'm keeping track of ${displayName}'s COVID data and risk level with @CovidActNow. What does your community look like?`;
+  return `I'm keeping track of ${displayName}'s vaccination progress and COVID risk level data with @CovidActNow. What does your community look like?`;
+}
+
+export function makeAddedChartShareQuote(
+  displayName: string,
+  metric: ExploreMetric,
+) {
+  if (metric === ExploreMetric.DEATHS) {
+    return `Daily COVID deaths in ${displayName}, according to @CovidActNow. See the chart: `;
+  } else if (metric === ExploreMetric.ICU_HOSPITALIZATIONS) {
+    return `COVID ICU hospitalizations in ${displayName}, according to @CovidActNow. See the chart: `;
+  } else if (metric === ExploreMetric.HOSPITALIZATIONS) {
+    return `COVID hospitalizations in ${displayName}, according to @CovidActNow. See the chart: `;
+  } else {
+    return `I'm keeping track of ${displayName}'s vaccination progress and COVID risk level data with @CovidActNow. What does your community look like?`;
+  }
 }
--- a/src/components/Charts/Redesign/Groupings.tsx
+++ b/src/components/Charts/Redesign/Groupings.tsx
@@ -237,6 +237,5 @@ export const getChartGroupFromMetric = (
       find(group.metricList, metric => metric.metric === metricToScrollTo) !==
       undefined,
   );
-  if (!groupWithMetric) return null;
-  return groupWithMetric;
+  return groupWithMetric ?? null;
 };
--- a/src/components/Charts/Redesign/Groupings.tsx
+++ b/src/components/Charts/Redesign/Groupings.tsx
@@ -228,10 +228,14 @@ export function getValueInfo(
 }
 
 // Used for scrolling to correct chart group when clicking summary stat
-export const getChartGroupFromMetric = (metricToScrollTo: Metric): any => {
-  // Fix the any
-  const groupWithMetric = find(CHART_GROUPS, group =>
-    find(group.metricList, metric => metric.metric === metricToScrollTo),
+export const getChartGroupFromMetric = (
+  metricToScrollTo: Metric,
+): ChartGroup | null => {
+  const groupWithMetric = find(
+    CHART_GROUPS,
+    group =>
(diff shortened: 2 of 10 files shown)


src/components/Charts/Redesign/Groupings.tsx
```diff
@@ -221,3 +226,17 @@ export function getValueInfo(
...
+  metricToScrollTo: Metric,
+): ChartGroup | null => {
+  const groupWithMetric = find(
+    CHART_GROUPS,
+    group =>
+      find(group.metricList, metric => metric.metric === metricToScrollTo) !==
+      undefined,
+  );
+  if (!groupWithMetric) return null;
+  return groupWithMetric;
```
mikelehen, line 241: nit: Could save a line and do `return groupWithMetric ?? null;`

src/components/Charts/Redesign/NewChartBlock.tsx
```diff
@@ -14,23 +16,38 @@ import {
...
+    if (!isNull(clickedStatMetric)) {
+      const clickedStatMetricIndex = findIndex(
+        metricList,
+        item => item.metric === clickedStatMetric,
+      );
+      if (clickedStatMetricIndex >= 0) {
+        setActiveTabIndex(clickedStatMetricIndex);
+      }
+    }
+  }, [metricList, clickedStatMetric]);
```
mikelehen, line 48: So I think there's an edge case that doesn't quite work...  If I:
1) Click "Infection rate" at the top of the page to focus the infection rate tab
2) Click "Daily new cases" inside the chart to focus the daily new cases tab
3) Click "Infection rate" at the top *again* to focus the infection rate tab again

It scrolls to the chart but the tab doesn't change to "infection rate" since `clickedStatMetric` didn't change so this `useEffect` doesn't get run...

To be honest, this is kinda' convoluted so I'm fine with shipping like this if we don't know of an easy fix.  The solutions that come to mind off the top of my head are:
1. Move the `activeTabIndex` state up a level. I sorta' suspect this is the "correct" fix but it seems pretty cumbersome (and maybe there's some better pattern I don't know about).
2. Hacky but I wonder if we could pass in a counter or timestamp or something in addition to `clickedStatMetric` that would change each time the metric is clicked, in order to trigger this effect to run again.

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -96,6 +101,21 @@ const ChartsHolder = React.memo(({ region, chartId }: ChartsHolderProps) => {
 
   const [scrolledWithRef, setScrolledWithRef] = useState(false);
 
+  const vaccinationsBlockRef = useRef<HTMLDivElement>(null);
+  const casesBlockRef = useRef<HTMLDivElement>(null);
+  const hospitalizationsBlockRef = useRef<HTMLDivElement>(null);
+  const deathsBlockRed = useRef<HTMLDivElement>(null);
+  const chartBlockRefs = useMemo(
+    // Fix the any
```
mikelehen, line 109: Hah, I forgot to remove this in my PR. 😆

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -124,7 +144,13 @@ const ChartsHolder = React.memo(({ region, chartId }: ChartsHolderProps) => {
 
     scrollToChart();
     scrollToRecommendations();
-  }, [chartId, metricRefs, isRecommendationsShareUrl, scrolledWithRef]);
+  }, [
+    chartId,
+    metricRefs,
+    isRecommendationsShareUrl,
+    scrolledWithRef,
+    chartBlockRefs,
```
mikelehen, line 152: Seems like this isn't used in the effect?  Is it actually necessary?

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -143,16 +169,28 @@ const ChartsHolder = React.memo(({ region, chartId }: ChartsHolderProps) => {
...
         `Location Header Stats: ${Metric[metric]}`,
       );
-      scrollTo(metricRefs[metric].current);
+      setClickedStatMetric(metric);
+      const groupWithMetric = getChartGroupFromMetric(metric);
+      const chartBlockRef = groupWithMetric
+        ? chartBlockRefs[groupWithMetric.groupHeader]
+        : null;
+      if (chartBlockRef?.current) {
+        setScrolledWithRef(true);
```
mikelehen, line 189: So I was trying to remember what this scrolledWithRef thing was for, and I think:
1. It was for when you scroll to the chart via URL, (e.g. https://covidactnow.org/us/washington-wa/chart/5/) and so we don't need to set it to true here (though it probably doesn't hurt).  We might want to rename it to `scrolledWithUrl` or something.
2. It looks like scrolling via URL like https://covidactnow.org/us/washington-wa/chart/5/ is currently broken which means our share links are broken right now (even in prod)... I opened a trello card: https://trello.com/c/K3JG32kp/1431-chart-share-urls-dont-scroll-to-chart-anymore

src/common/utils/makeChartShareQuote.ts
```diff
@@ -33,5 +34,20 @@ export default function makeChartShareQuote(
       1,
     )} daily new cases per 100k population, according to @CovidActNow. See the chart: `;
   }
-  return `I'm keeping track of ${displayName}'s COVID data and risk level with @CovidActNow. What does your community look like?`;
+  return `I'm keeping track of ${displayName}'s vaccination progress and COVID risk level data with @CovidActNow. What does your community look like?`;
+}
+
+export function makeAddedChartShareQuote(
```
mikelehen, line 40: nit: Not 100% sure why `Added` is in the function name here?  Is this because we recently added these charts or something?  I'd be a fan of trying to make names more future-proof so they make sense to somebody looking at this code in 6 months or whatever. 😁 

If there's a name collision we don't want to resolve right now, I'd at least add a TODO comment indicating the intended cleanup we should do later.

src/common/utils/makeChartShareQuote.ts
```diff
@@ -33,5 +34,20 @@ export default function makeChartShareQuote(
...
+  displayName: string,
+  metric: ExploreMetric,
+) {
+  if (metric === ExploreMetric.DEATHS) {
+    return `Daily COVID deaths in ${displayName}, according to @CovidActNow. See the chart: `;
```
mikelehen, line 45: Right now we end up just sharing the URL for the location page and so "See the chart: " won't make sense...

Not sure what your plans are for chart sharing, but we'll need to either generate URLs and unique share images for all these charts or change this share quote text to make more sense given we're just sharing the location page.

src/components/NewLocationPage/ChartFooter/AddedMetricChartFooter.tsx
```diff
@@ -1,3 +1,5 @@
+/** Chart footer for newly added chart metrics (deaths, hospitalizations, ICU hospitalizations) */
```
mikelehen, line 1: nit: Heh, again "newly added" is relevant right now but won't make much sense to somebody looking at this code in the future. 😆   But I guess we have `Added` in the filename as well, so a broader cleanup is in order eventually.

mikelehen (approved):

src/common/utils/makeChartShareQuote.ts
chasulin: lol yeah i didnt know what to call things related to the newly added charts

src/components/NewLocationPage/ChartFooter/AddedMetricChartFooter.tsx
chasulin: yeah i agree--something more timeless/descriptive is the way to go. just wasnt sure what to call these files/components/functions so went with added/newly added for the time being 😬

src/common/utils/makeChartShareQuote.ts
chasulin: lowest lift to unblock shipping would be editing the share quote. so something more like 'Daily COVID deaths in ${displayName}, according to @CovidActNow. See the data:"

or maybe something more robust like 'Daily COVID deaths in ${displayName}, according to @CovidActNow. See ${displayName}'s vaccination progress and COVID risk level data: '

src/components/LocationPage/ChartsHolder.tsx
chasulin: oh yeah i think this is a remnant of me trying to get the share links working

src/common/utils/makeChartShareQuote.ts
mikelehen: I think it's weird to say "Daily COVID deaths in ${displayName}" but then the share image and the URL have nothing specifically to do with daily covid deaths.

How about:
> Check out Daily COVID deaths, vaccination progress, and other key metrics for ${displayName} at @CovidActNow:

Don't feel strongly though so feel free to make an executive decision and ship it.

chasulin: oo yes, going with that^

src/components/Charts/Redesign/NewChartBlock.tsx
chasulin: i originally had `activeTabIndex` at the `ChartsHolder` level but then realized that when all of the `chartBlocks` share a single `activeTabIndex`, switching a tab in one block switches the tab in all of the blocks // they all change in unison. so to keep each block functioning independently, i moved `activeTabIndex` down into the individual blocks

chasulin: an individual activeTabIndex/setActiveTabIndex at the chartsHolder level passed to each block (like the map of block-specific refs) could work

chasulin: but there might be a better way. adding a todo and gonna think on this

mikelehen: Sounds good. Yeah, I was imagining something like the block-specific refs, but TODO sounds good for now!

PR #3856, opened 2021-06-20
Switch USMap to material-ui Tooltip and implement new tooltip features.

mikelehen:
* Migrate to material-ui Tooltip component.
* Add `ToolTipMode.ACTIVATE_ON_CLICK` mode to `USMap` that opens the tooltip on "click" instead of hover (for the mobile USVaccineMap tooltip behavior).
* Add `ActiveRegionStyle.OUTLINE` mode to `USMap` that uses a solid black outline as the "hover" (activation) effect instead of changing the opacity (used for the vaccine map so the colors don't get confusing when you hover).
* Adds USVaccineMap / StatesMobile story to storybook that demos using the two new map modes.
* Restyle the USRiskMap tooltip with a light background for consistency with upcoming USVaccineMap tooltip.
* Wires VaccineTooltip into USVaccineMap.

Storybook links for the vaccine map:
* [Desktop](https://60d0e521c019520049f4f3ff-pfefpajrfa.chromatic.com/?path=/story/shared-components-usvaccinemap--states)
* [Mobile](https://60d0e521c019520049f4f3ff-pfefpajrfa.chromatic.com/?path=/story/shared-components-usvaccinemap--states-mobile)

And for the [risk map](https://60d0e521c019520049f4f3ff-pfefpajrfa.chromatic.com/?path=/story/shared-components-usriskmap--states).

Note that the tooltip positioning looks somewhat bad to me, especially on the risk map (with the smaller tooltip) since the tooltip can be far away from the mouse pointer on large states.  I'll take another stab at getting "follow mouse pointer" support working when I get a chance, but I think this is shippable as-is. Holler if you disagree.


mikelehen:
cc/ @schamp FYI- I ended up having to do a bit more surgery than I expected to implement all the vaccine map tooltip functionality I needed.  This PR has it all... not sure if / how much of a headache it'll cause with regard to porting (worth noting that the new tooltip functionality diverges from the BBE map a bit).

src/components/USMap/USMap.tsx
```diff
@@ -26,70 +33,86 @@ const stateFipsCodes = Object.keys(statesByFips);
...
-                      ) as StateType;
+  ({
+    showCounties,
+    getFillColor,
+    renderTooltip,
+    tooltipMode,
+    activeRegionStyle,
+  }: USMapProps) => {
+    const [activeState, setActiveState] = useState<State | null>(null);
+    const regionGeographies = useMemo(() => {
```
mikelehen, line 65: I refactored the memoized `mapContent` (basically the entire map) into just `regionGeographies` since we need to add a geometry for the active region when using `ActiveRegionStyle.OUTLINE`...  This allows us to still memoize (cache) the bulk of the map even though the outline region changes on hover change.

src/components/USMap/USMapTooltip.tsx
```diff
@@ -0,0 +1,110 @@
...
+    if (!open) {
+      setOpen(true);
+      onOpen && onOpen(e);
+    }
+    e.preventDefault();
+  };
+
+  return (
+    <ClickAwayListener onClickAway={handleTooltipClose}>
+      <g onClickCapture={handleTooltipOpen}>
```
mikelehen, line 95: I had to use `onClickCapture` to be able to prevent the click from propagating down to the `<Link>` component and navigating to the location page.  It seems to work but I'm a little nervous about it.  Not sure if there's a better way (maybe just removing the `<Link>` component entirely when using `ACTIVATE_ON_CLICK` mode...  but I was nervous about that for accessibility, etc.)

mikelehen:
FYI- I removed the logic to try to follow the mouse cursor since the positioning was broken.  We can revisit later if we have time and are motivated (one option would be to use the pre-release material-ui v5 since it has builtin support).

src/components/USMap/USMap.tsx
```diff
@@ -26,99 +33,155 @@ const stateFipsCodes = Object.keys(statesByFips);
...
+                      onOpen={() => {
+                        setActiveState(state);
+                      }}
+                      onClose={() => {
+                        setActiveState(activeState =>
+                          activeState === state ? null : activeState,
+                        );
+                      }}
+                      arrow
+                    >
```
pnavarrc, line 120: I feel it looks weird having the tooltip below the cursor, how does it look with `placement="top"`?

pnavarrc (approved):
Looks good!

src/components/USMap/USMap.tsx
mikelehen: I'm not sure if I like it better or not (I think either one looks a bit weird; I miss the cursor following behavior).  I'll throw this in a separate PR for you to peek at and weigh in.

src/components/USMap/USMap.tsx
```diff
@@ -26,99 +33,155 @@ const stateFipsCodes = Object.keys(statesByFips);
...
+      );
+    }, [showCounties, getFillColor, tooltipMode, renderTooltip]);
+
+    return (
+      <USMapWrapper>
+        <USStateMapWrapper
+          $showCounties={showCounties}
+          $highlightOnHover={activeRegionStyle === ActiveRegionStyle.HIGHLIGHT}
+        >
+          <ComposableMap data-tip="" projection={projection} height={500}>
```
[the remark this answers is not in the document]
mikelehen: Oh, good catch! Thanks.

PR #3851, opened 2021-06-20
New VaccineProgressBar.

mikelehen:
* Move NewLocationPage/VaccinationProgressBarBlock/ProgressBar to a top-level
  VaccineProgressBar component for use in compare and tooltips.
* Color the bar based on the vaccination percent and add hatched styling for
  partially-vaccinated part of bar.
  
I left the old style available for now via the `oldVersion` prop so that the location page is unaffected for now (mostly because I think we may want to vet the colors a bit more before shipping them... As @chasulin pointed out, the darkest color is pretty dark!).

PR #3859, opened 2021-06-21
VaccineTooltip component.

mikelehen:
Adds a VaccineTooltip component for use by the vaccine map. Depends on https://github.com/covid-projections/covid-projections/pull/3851.  Can't be plumbed into the map until https://github.com/covid-projections/covid-projections/pull/3856 lands.  

Desktop version:
image

Mobile version:
image

Demo with map and tooltip integrated: https://covid-projections-git-mikelehen-vaccine-tooltip-covidactnow.vercel.app/?s=1957873

### DIFF
--- a/src/components/USMap/VaccineTooltip/VaccineTooltip.style.ts
+++ b/src/components/USMap/VaccineTooltip/VaccineTooltip.style.ts
@@ -1,0 +1,78 @@
+import styled from 'styled-components';
+import ArrowForwardIcon from '@material-ui/icons/ArrowForward';
+import { COLOR_MAP } from 'common/colors';
+import VaccineDot from 'components/VaccineDot/VaccineDot';
+import { LinkButton } from 'components/Button';
+
+export const Container = styled.div`
+  background-color: white;
+  border-radius: 4px;
+  filter: drop-shadow(0px 2px 20px rgba(0, 0, 0, 0.16));
+  width: 240px;
+`;
+
+export const Inner = styled.div`
+  padding: 16px;
+`;
+
+export const LocationName = styled.div<{ $isMobileVersion: boolean }>`
+  ${props => props.theme.fonts.regularBookBold};
+  text-align: ${props => (props.$isMobileVersion ? 'center' : undefined)};
+  color: black;
+  font-size: 16px;
+  line-height: 16px;
+  margin-top: 4px;
+  margin-bottom: 20px;
+`;
+
+export const LocationNameArrowIcon = styled(ArrowForwardIcon)`
+  color: ${COLOR_MAP.GREY_600};
+  font-size: 22px;
+  margin-bottom: -5px;
+  margin-left: 6px;
+`;
+
+export const Row = styled.div`
+  display: flex;
+  margin-top: 3px;
+`;
+
+export const Title = styled.div`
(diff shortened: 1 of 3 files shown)


chasulin (commented):
something with the placement of the tooltip is off. example: when hovering on vermont--

Screen Shot 2021-06-21 at 11 52 16 AM

mikelehen:
I ended up reverting the "mouse following" logic that was positioning the tooltip wrong.  We can revisit later.

I'll merge this once https://github.com/covid-projections/covid-projections/pull/3851 is reviewed / merged since this one depends on that.

src/components/USMap/VaccineTooltip/VaccineTooltip.style.ts
```diff
@@ -0,0 +1,78 @@
...
+  width: 240px;
+`;
+
+export const Inner = styled.div`
+  padding: 16px;
+`;
+
+export const LocationName = styled.div<{ $isMobileVersion: boolean }>`
+  ${props => props.theme.fonts.regularBookBold};
+  text-align: ${props => (props.$isMobileVersion ? 'center' : undefined)};
```
chasulin, line 20: can we add a breakpoint instead of passing `$isMobileVersion` as a prop?

src/components/USMap/VaccineTooltip/VaccineTooltip.tsx
```diff
@@ -0,0 +1,101 @@
...
+      </Inner>
+      {isMobileVersion && (
+        <MoreDataLinkContainer>
+          <MoreDataLinkButton
+            href={state.relativeUrl}
+            aria-label={locationName}
+            onClick={() => {
+              trackMapClick(locationName);
+            }}
+            tabIndex={-1}
```
chasulin, line 86: this makes the button unreachable via keyboard navigation. is that intentional?

src/components/USMap/VaccineTooltip/VaccineTooltip.tsx
```diff
@@ -0,0 +1,101 @@
...
+        </Row>
+        <ProgressBarWrapper>
+          <VaccineProgressBar
+            locationName={locationName}
+            vaccinationsInitiated={vaccinationsInitiated}
+            vaccinationsCompleted={vaccinationsCompleted}
+          />
+        </ProgressBarWrapper>
+      </Inner>
+      {isMobileVersion && (
```
chasulin, line 78: an option- you can wrap the mobile-specific and desktop-specific in [MobileOnly](https://github.com/covid-projections/covid-projections/blob/develop/src/components/NewLocationPage/Shared/Shared.style.tsx#L49) and [DesktopOnly](https://github.com/covid-projections/covid-projections/blob/develop/src/components/NewLocationPage/Shared/Shared.style.tsx#L53) components, which use the [MUI Hidden component](https://material-ui.com/components/hidden/), rather than using an `isMobileVersion` prop. not sure if one approach is more preferable than the other, so feel free to ignore

chasulin (approved):

src/components/USMap/VaccineTooltip/VaccineTooltip.tsx
chasulin: oo makes sense!

PR #3863, opened 2021-06-21
setup Chromatic for Storybook

pnavarrc:
- Setup Chromatic to host Storybook
- Disabled the review and approval workflows, so this will only host Storybook for us

@mikelehen  maybe we should move this action to `ci.yml` to avoid repeating the install steps?


mikelehen (approved):
Nice!  🎉 

I think this being a separate workflow is fine (they'll probably run in parallel anyway).  FWIW- We could look into https://github.com/marketplace/actions/npm-or-yarn-install-with-caching or similar if installing dependencies is getting too slow.

PR #3870, opened 2021-06-22
Fix more data link style.

mikelehen:
Pablo pointed out the hover color was wrong which led me to discover that I should have been using one of our builtin button components. 😬


pnavarrc (approved):
Excellent! 

image

PR #3873, opened 2021-06-22
Vaccinations table

chasulin:
`Shared Components/Vaccinations Table` in storybook


src/components/VaccineProgressBar/VaccineProgressBar.tsx
```diff
@@ -96,8 +98,9 @@ const VaccineProgressBar: React.FC<ProgressBarProps & { width: number }> = ({
 const VaccineProgressBarAutosize: React.FC<ProgressBarProps> = props => {
   return (
     <ProgressBarContainer>
-      <ParentSize>
-        {({ width }) => <VaccineProgressBar {...props} width={width} />}
+      {/* Adding display:flex makes sure the ParentSize container is tight to its contents */}
```
chasulin, line 101: im no longer using the autosize version in the vaccinations table, but this fixes a problem i noticed when i was using it

src/components/VaccinationsTable/VaccinationsTable.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const containerProps = {
+    collapsedHeightMobile: 'fit-content',
+    collapsedHeightDesktop: 'fit-content',
+    tabTextCollapsed: <>See all states</>,
+    tabTextExpanded: <>See all states</>,
+    trackingLabel: 'Vaccinations table: open modal',
+    trackingCategory: EventCategory.VACCINATION,
+    disableArrowChange: true,
+    secondaryOnClick: () => {}, // plug into opening compare modal
```
mikelehen, line 62: nit: Add "TODO"? 😁

src/components/VaccinationsTable/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+  rank: number;
+}
+
+function getStatesSortedByVaccinationsCompleted(): StateWithVaccinationMetricsAndRank[] {
+  const statesWithVaccinationMetrics = regions.states.map(
+    (state: State, i: number) => {
+      const summaryForFips = LocationSummariesByFIPS[state.fipsCode];
+      return {
+        stateName: state.name,
+        vaccinationsInitiated: summaryForFips.metrics[6]?.value ?? null,
```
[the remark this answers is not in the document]
chasulin: lol whoops

src/components/VaccineProgressBar/VaccineProgressBar.tsx
```diff
@@ -17,7 +17,9 @@ function getOffsetPercentage(decimal: number) {
   return formatPercent(decimal, 1);
 }
 
-const VaccineProgressBar: React.FC<ProgressBarProps & { width: number }> = ({
+export const VaccineProgressBarInner: React.FC<
```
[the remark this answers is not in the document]
chasulin: oo yep this sounds good

src/components/VaccinationsTable/VaccinationsTable.stories.tsx
```diff
@@ -1,11 +1,16 @@
 import React from 'react';
+import { MapView } from 'screens/HomePage/HomePage';
```
mikelehen, line 2: FYI- My map PR moves this enum to a different file.  Guessing you'll merge first but if not, this will need to change to:

```
import { MapView } from 'screens/HomePage/utils';
```

src/components/VaccinationsTable/VaccinationsTable.tsx
```diff
@@ -47,26 +52,28 @@ const Column: React.FC<{ listHeader: string; locations: any[] }> = ({
...
 };
 
-const VaccinationsTable: React.FC<{}> = () => {
-  const highestRanking = getHighestRankingStates(5);
-  const lowestRanking = getLowestRankingStates(5);
+const VaccinationsTable: React.FC<{ mapView: MapView }> = ({ mapView }) => {
+  const highestRanking = getHighestRankingRegions(5, mapView);
+  const lowestRanking = getLowestRankingRegions(5, mapView);
+
+  const buttonLabel = <>See all {mapView.toLowerCase()}</>;
```
mikelehen, line 62: Just FYI (not worth changing)- If `ExpandableContainer` accepted `ReactNode` instead of `ReactElement` for the labels then we could just use plain strings here instead of having to wrap in `<>...</>` since `ReactNode` is more encompassing and includes `string`:
```
    type ReactText = string | number;
    type ReactChild = ReactElement | ReactText;

    interface ReactNodeArray extends Array<ReactNode> {}
    type ReactFragment = {} | ReactNodeArray;
    type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;
```

src/components/VaccinationsTable/utils.ts
```diff
@@ -1,53 +1,68 @@
...
+    region =>
+      Number.isFinite(region.vaccinationsInitiated) &&
+      Number.isFinite(region.vaccinationsCompleted),
   );
+
+  const sortedWithRank = sortedNoNulls.map((regionInfo, i: number) => ({
+    rank: i + 1,
+    regionName: regionInfo.regionName,
+    vaccinationsInitiated: regionInfo.vaccinationsInitiated as number,
+    vaccinationsCompleted: regionInfo.vaccinationsCompleted as number,
```
mikelehen, line 50: nit: Rather than `as number` here I would add `as RegionVaccinationInfo[]` on line 44 after the `filter()`. That way the cast is directly tied to the code that we are relying on to remove the nulls (since any time you're casting you could accidentally shoot yourself in the foot if it turns out you were wrong about what the type will be at runtime 😁 ).

src/components/VaccinationsTable/utils.ts
```diff
@@ -1,53 +1,68 @@
...
+function getRegionsSortedByVaccinationsInitiated(
+  regionScope: MapView,
+): RegionVaccinationInfo[] {
+  const regionsToSort =
+    regionScope === MapView.COUNTIES ? regions.counties : regions.states;
+
+  const regionsWithVaccinationMetrics = regionsToSort.map((region: Region) => {
+    const summaryForFips = LocationSummariesByFIPS[region.fipsCode];
+    return {
+      regionName: region.name,
```
mikelehen, line 26: For counties we probably want to include the state abbreviation since I have no idea what state most counties are in. 😁   Can we just do whatever the compare table does for names?

src/components/VaccinationsTable/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+      vaccinationsInitiated:
+        summaryForFips?.metrics[Metric.VACCINATIONS]?.value ?? null,
+      vaccinationsCompleted: summaryForFips?.vc ?? null,
+      url: region.canonicalUrl,
```
mikelehen, line 30: nit: not sure if it really makes any difference, but I think we use `relativeUrl` in other places (like the map) where we are just linking internally.

mikelehen (approved):
Requested some minor tweaks, but looks great!

src/components/VaccinationsTable/VaccinationsTable.tsx
chasulin: oo good to know!

src/components/VaccinationsTable/utils.ts
chasulin: oh true

PR #3876, opened 2021-06-23
Pull vaccinations table into homepage

chasulin:


src/components/Compare/CompareMain.tsx
```diff
@@ -46,13 +46,15 @@ import {
 const CompareMain = React.memo(
   (props: {
     stateName?: string;
-    isModal?: boolean;
```
chasulin, line 49: unused prop

src/screens/HomePage/MapAccessories/MapAccessories.tsx
```diff
@@ -1,24 +1,30 @@
...
+import { Row, Wrapper, AboutLink, TableWrapper } from './MapAccessories.style';
 import ShareButtons from 'components/SharedComponents/ShareButtons';
 import { EventCategory } from 'components/Analytics';
 import { getUrlAndShareQuote } from 'components/ShareBlock/ShareModelBlock';
+import { MapView } from '../utils';
 
 interface MapAccessoriesProps {
   renderThermometer: () => React.ReactElement;
   infoLink: string;
+  renderTable?: (locationScope: MapView) => React.ReactElement;
```
mikelehen, line 11: FWIW- Since this component just automatically calls `renderTable(locationScope)` I think we could simplify this a bit and just have `table?: React.ReactElement` here, without needing `locationScope`

and then in MapBlock do:
```
          table={renderTable && renderTable(locationScope)}
```

We would only need to pass in `renderTable` if this component needed to potentially call it with different parameters depending on internal state or something.  The same actually goes for `renderThermometer`.  It could just be `thermometer: React.ReactElement` instead of being a function.

src/components/VaccinationsTable/VaccinationsTable.style.tsx
```diff
@@ -81,6 +81,7 @@ export const ColumnHeader = styled.p`
   color: ${COLOR_MAP.GRAY_BODY_COPY};
   margin: 0 0 1rem;
   text-align: center;
+  font-size: 0.875rem;
```
mikelehen, line 84: not sure why this snuck into this PR, but I assume it's intentional. 😁

src/screens/HomePage/MapBlock.tsx
```diff
@@ -49,10 +49,11 @@ export const MapBlock: React.FC<MapBlockProps> = ({
         <HomePageBlockSubtitle>{subtitle}</HomePageBlockSubtitle>
         <LocationToggle locationScope={locationScope} onChange={onToggle} />
         {renderMap(locationScope)}
-        {renderTable && renderTable(locationScope)}
         <MapAccessories
           renderThermometer={renderThermometer}
           infoLink={infoLink}
+          renderTable={renderTable && renderTable}
```
mikelehen, line 55: Maybe `renderTable && renderTable && renderTable` for good measure? 😆

src/screens/HomePage/HomePage.tsx
```diff
@@ -72,6 +73,12 @@ export default function HomePage() {
   const exploreSectionRef = useRef(null);
 
   const [menuOpen, setMenuOpen] = useState(false);
+  const [showCompareModal, setShowCompareModal] = useState(false);
+
+  const vaccinationsTableButtonOnClick = () => {
+    window.location.href = '#compare';
```
mikelehen, line 79: I was gonna' ask why we want to scroll them down to compare, but then I realized it's so that when they close the modal it's not jarring that they get scrolled down to compare...  but *then* I realized that it's so that the vaccinations column is first!

Maybe add a comment to explain? 😁 

Alternatively we could turn `compareShowVaccinationsFirst` into a state variable.. Something like:

```
  const [compareShowVaccinationsFirst, setCompareShowVaccinationsFirst] = useState<boolean>(false);
  useEffect(() => {
    if (location.hash === '#compare') {
      setCompareShowVaccinationsFirst(true);
    }
  }, [location.hash]);
  const vaccinationsTableButtonOnClick = () => {
    setCompareShowVaccinationsFirst(true);
    setShowCompareModal(true);
  };
```

But I'm also down with the hash.

src/screens/HomePage/HomePage.tsx
```diff
@@ -72,6 +73,12 @@ export default function HomePage() {
...
+    setShowCompareModal(true);
```
mikelehen, line 80: So clicking "See all states" ends up showing metros. 😬   I think we need to somehow plumb the location scope into compare...

Alternatively, after seeing the 5 top / bottom counties I'm not sure how useful that will be to highlight...  We could (sorry) go back to making the vaccinations table always show states, and then change compare to default to states instead of metros, which I think is more valuable anyway.

mikelehen (changes requested):
Nice!  Left a couple optional comments but also realized that this will show metros when you click "see all states".  Couple options for going forward there I think... see my comment for details.

src/components/VaccinationsTable/VaccinationsTable.style.tsx
chasulin: yeah, the column headers are 16px in the mocks but use `font-variant: small-caps`, so im dropping the font size to account for that

src/screens/HomePage/MapBlock.tsx
chasulin: 💯

src/screens/HomePage/HomePage.tsx
chasulin: +1 to making compare default to states overall. im not sure im following the rest of the comment though

src/screens/HomePage/HomePage.tsx
chasulin: switched to a state variable

mikelehen (approved):
🚢

PR #3879, opened 2021-06-23
Simplify / fix VaccineProgressBar sizing.

mikelehen:
I realized that by adding `width: 100%` to the CSS styles of the `svg` element (to fix that flicker problem), it actually negated the `width` property on the svg element itself which means:
1. Passing `width` into VaccineProgressBar wasn't working properly for me when I tried it from the compare table.
2. I think the original reason we had a `width` property was so we could adopt the width of the `ParentSize` wrapper, but with `width: 100%` on the svg element we don't need that wrapper at all, and removing it removes an extra render (the ParentSize was rendering with width 0 first and then with the correct width, which is why we were seeing that flicker in the first place).

So I ended up just removing the `width` property and the `ParentSize` wrapper entirely, and now rely on `width: 100%` in the CSS.  I did have to then specify an explicit width on your ProgressBarWrapper in VaccinationsTable, but other than that everything still works as before, and with less code. So I think this is a win?  Let me know if you have any concerns.  I could be missing something.

### DIFF
--- a/src/components/VaccineProgressBar/VaccineProgressBar.style.tsx
+++ b/src/components/VaccineProgressBar/VaccineProgressBar.style.tsx
@@ -3,6 +3,8 @@ import { COLOR_MAP } from 'common/colors';
 import { materialSMBreakpoint } from 'assets/theme/sizes';
 
 export const ProgressBarContainer = styled.div`
+  /* Adding display:flex makes sure the ParentSize container is tight to its contents */
+  display: flex;
   width: 100%;
   max-width: 350px;
   margin: auto;
@@ -14,7 +16,6 @@ export const ProgressBarContainer = styled.div`
 `;
 
 export const StyledSvg = styled.svg`
-  width: 100%;
   border: 1px solid ${COLOR_MAP.GREY_2};
   border-radius: 3px;
 `;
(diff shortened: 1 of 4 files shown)


src/components/VaccineProgressBar/VaccineProgressBar.style.tsx
```diff
@@ -3,6 +3,8 @@ import { COLOR_MAP } from 'common/colors';
 import { materialSMBreakpoint } from 'assets/theme/sizes';
 
 export const ProgressBarContainer = styled.div`
+  /* Adding display:flex makes sure the ParentSize container is tight to its contents */
+  display: flex;
```
chasulin, line 7: i think this also needs justify-content:center now. the alignment is off on mobile -

Screen Shot 2021-06-23 at 11 32 03 AM

chasulin (commented):
i have to dig around a little to understand why, but its not being responsive on small mobile devices anymore--

On prod // with ParentSize:

Screen Shot 2021-06-23 at 11 36 27 AM

In this preview link:

Screen Shot 2021-06-23 at 11 36 48 AM

mikelehen:
Gah, thanks for reviewing and finding the bugs!  I thought I checked small mobile, but maybe I didn't go small enough.  Anyway, we can probably put a pin in this, at least until I figure out if it helps with the compare table (I had started going down a rabbit hole but it may not have been necessary...)

chasulin:
yeah, i think the progress in the location page header is the only one that resizes, so thats the one that would be affected by removing ParentSize. (the other progress bars--tooltip and vaccination table--are fixed widths i think?). im not 1000% sure though

mikelehen:
This isn't blocking anything so going to close it for now.

PR #3885, opened 2021-06-23
Update vaccinations column in compare table.

mikelehen:
* Remove ICU column.
 * Rework ColumnDefinition interface to have a render() method, implemented by each column type (reamed columns.ts to columns.tsx so I can use React in render method).
 * Added width columns to ColumnDefinition so different columns can have different widths.
 * Implement VaccinationsColumn to render with the progress bar at desired width.


src/components/Compare/Compare.style.tsx
```diff
@@ -67,19 +67,6 @@ export const Wrapper = styled.div<{
...
-        width: 22%;
-        min-width: ${locationNameCellWidth}px;
-      }
-
-      &:not(:first-child) {
-        width: 13%;
-        min-width: ${metricCellWidth}px;
-      }
-    }
-
```
mikelehen, line 82: we now specify these on a per-column basis.

src/components/Compare/Compare.style.tsx
```diff
@@ -359,21 +354,40 @@ export const DataCellValue = styled.span<{
...
+  min-width: ${props => props.$minWidthPx}px;
   padding-right: 0;
   color: ${({ $isSelected }) =>
     $isSelected ? 'black' : `${COLOR_MAP.GRAY_BODY_COPY}`};
   font-weight: ${({ $isSelected }) => $isSelected && '600'};
   background-color: ${({ $isSelected }) => $isSelected && 'rgba(0,0,0,0.02)'};
-
-  svg {
-    color: ${({ $iconColor }) => $iconColor};
-  }
```
mikelehen, line 376: This is now specified on the `svg` element itself since I need to specify the color within the column's render() method which is used as the contents of the `<DataCell>` container...

src/components/Compare/CompareTableRow.tsx
```diff
@@ -94,26 +92,17 @@ const CompareTableRow = (props: {
         </StyledLink>
       </LocationNameCell>
       {columns.map((column, i) => {
-        const metricValue = column.getValue(location);
-        const valueUnknown =
-          !isNumber(metricValue) || !Number.isFinite(metricValue);
```
mikelehen, line 99: This all got refactored to live in the `render()` method of each column.

src/components/Compare/columns.tsx
```diff
@@ -0,0 +1,191 @@
+import { Metric } from 'common/metricEnum';
```
mikelehen, line 1: Sorry, I renamed this file and then edited it so much that github isn't showing the diff against the old `columns.ts` file.  But if you just look at the 2nd commit in my PR (https://github.com/covid-projections/covid-projections/pull/3885/commits/53e036d9aabb6945f99ab071ce19aae1c01f2d6f) then you can see just the diffs (though admittedly I changed a lot 😁 ).

src/components/Compare/LocationTable.style.tsx
```diff
@@ -1,10 +1,11 @@
 import styled from 'styled-components';
-import { locationNameCellWidth, metricCellWidth } from './Compare.style';
+import { locationNameCellWidth } from './Compare.style';
 import { orderedColumns } from './columns';
 import { RankedLocationSummary } from 'common/utils/compare';
+import sum from 'lodash/sum';
 
 const minTableWidth =
-  locationNameCellWidth + orderedColumns.length * metricCellWidth;
+  locationNameCellWidth + sum(orderedColumns.map(c => c.minWidthPx));
```
chasulin, line 8: nice

src/components/Compare/Compare.style.tsx
```diff
@@ -359,21 +354,40 @@ export const DataCellValue = styled.span<{
   text-transform: capitalize;
 `;
 
+export const VaccinationsCell = styled.div`
+  display: flex;
+  flex-direction: row;
```
chasulin, line 359: nit- flex-direction defaults to row so no need to declare it

chasulin (approved):
🚢

chasulin:
> The tricky bit is `?.` returns `undefined` if `this.metric` isn't in the `row.metricsInfo.metrics` array (which shouldn't happen, but TypeScript can't tell)... So the `?? null` works to force that `undefined` back into a `null`...
> 
> One thing I've wondered is if we would be better off avoiding `null` and using `undefined` more so that we have fewer of these cases where we're coercing `undefined` to `null`, etc.

mm yeah. though there are a lot of nulls to squash to make this change

src/components/Compare/Compare.style.tsx
mikelehen: oops, good catch!

PR #3889, opened 2021-06-24
% Vaccinated in location page map

chasulin:


src/components/HorizontalThermometer/RiskLevelThermometer/RiskLevelThermometer.style.tsx
```diff
@@ -11,7 +11,7 @@ export const ThermometerContainer = styled.div`
 
 export const Label = styled.span`
   color: ${COLOR_MAP.GRAY_BODY_COPY};
-  font-size: 1rem;
+  font-size: 0.875rem;
```
chasulin, line 14: matched the mocks

src/components/NewLocationPage/CountyMap/CountyMap.tsx
```diff
@@ -1,17 +1,92 @@
...
+    thermometer: <RiskLevelThermometer />,
+    colorMap: (locationSummary: LocationSummary) =>
+      getAlertColor(locationSummary),
+  },
+};
 
 /* The default aspect-ratio for the state and US maps is 800x600 */
 
 const CountyMap: React.FC<{ region: Region }> = React.memo(({ region }) => {
+  const [mapType, setMapType] = useState<MapType>(MapType.RISK_LEVEL);
```
chasulin, line 60: should we default to % vaccinated?

mikelehen: Let's do it!

src/components/NewLocationPage/CountyMap/CountyMap.tsx
```diff
@@ -1,17 +1,92 @@
...
+
+  const onClickToggle = (
+    event: React.MouseEvent<HTMLElement>,
+    newView: MapType,
+  ) => {
+    if (newView) {
```
mikelehen, line 66: nit: Looking at this code, `newView` is a `MapType` which will always be '% Vaccinated' or 'Risk levels' and so this `if` should always be true and therefore I might be tempted to remove it.

But my guess is it _is_ actually necessary (or could be anyway) because material-ui [docs](https://material-ui.com/api/toggle-button-group/) say that onChange can be called with `null`... so the type should probably be `newView: MapType|null` and then it's clearer why the `if` statement is needed.

As an aside, the reason you were able to type it as `MapType` even though that's not quite right is because the material-ui code is typed as:
```
  onChange?: (event: React.MouseEvent<HTMLElement>, value: any) => void;
```

Since they used `any`, TypeScript isn't able to help us type it correctly.

src/components/NewLocationPage/CountyMap/CountyMap.tsx
```diff
@@ -1,17 +1,92 @@
...
+      setMapType(newView);
+      trackEvent(
+        EventCategory.MINI_MAP,
+        EventAction.SELECT,
+        `Select: ${mapType}`,
+      );
```
mikelehen, line 72: Good call with the tracking! 😁

mikelehen (approved):
Looks great!  I think we should default to vaccinations and then 🚢

mikelehen:
Since the feedback was so minor I decided to just go ahead and fix and ship this (hope ya don't mind!) so we have a bit more time for potential feedback to come in before the break. 😁

src/components/NewLocationPage/CountyMap/CountyMap.tsx
chasulin: ah yes, it breaks when removing the if--the possible `null` is probably why

PR #3897, opened 2021-06-24
Reset MapType when navigating via minimap

chasulin:


src/components/NewLocationPage/CountyMap/CountyMap.tsx
```diff
@@ -73,6 +74,12 @@ const CountyMap: React.FC<{ region: Region }> = React.memo(({ region }) => {
     }
   };
 
+  const { pathname } = useLocation();
+
+  useEffect(() => {
```
ghop02, line 79: I think a small comment like "resets maptype to vaccinations on page navigation" would be helpful.  may be obvious but took me a sec

ghop02 (approved):

PR #3902, opened 2021-06-24
Increase chart height for desktop

chasulin:


mikelehen (commented):
I think we need to adjust the share images:

https://covid-projections-git-casulin-chart-height-edit-covidactnow.vercel.app/internal/share-image/states/wa/chart/1

image

src/common/hooks/useChartHeightForBreakpoint.ts
```diff
@@ -0,0 +1,19 @@
...
+  const [chartHeight, setChartHeight] = useState<number>(chartHeightDesktop);
+  const isMobile = useBreakpoint(600);
+
+  useEffect(() => {
+    if (isMobile) {
+      setChartHeight(chartHeightMobile);
+    } else {
+      setChartHeight(chartHeightDesktop);
+    }
+  }, [isMobile]);
```
mikelehen, line 16: Not sure we need the `useState()` and `useEffect()`...  I think we could just do:
```
  return isMobile ? chartHeightMobile : chartHeightDesktop;
```

mikelehen (approved):
Nice!

PR #4078, opened 2021-07-22
Remove CDC disclaimer about estimated 1st dose data.

mikelehen:


smcclure17 (approved):
looks good! for counties where will still estimate should we create manual disclaimers? (If so, I can compile that list and add it via the CMS override)

mikelehen:
Yeah, that sounds good if it's not too much trouble! 🙏

PR #4323, opened 2021-09-07
Fix NavBar grid so that back arrow, etc. work on Chrome 93.

mikelehen:
We accidentally set up our grid columns so that the search bar grid cell overlapped some of the other grid cells, which apparently worked fine until Chrome 93. But with Chrome 93, the search bar div ended up on top of the other elements and e.g. you couldn't click the back arrow / logo anymore.

https://trello.com/c/WLfmqTXd/244-can-site-back-arrow-stopped-working-in-chrome-93

### DIFF
--- a/src/components/NavBar/NavBar.style.ts
+++ b/src/components/NavBar/NavBar.style.ts
@@ -111,12 +111,15 @@ export const MenuLabel = styled.span`
 // Location page:
 export const GridContainer = styled.div`
   display: grid;
+  // On mobile, we have 4 columns: back, logo, search, menu
+  // But note that after you scroll, the search bar expands to use the logo column as well.
   grid-auto-columns: min-content auto auto min-content;
   align-items: center;
   width: 100%;
 
   @media (min-width: ${mobileBreakpoint}) {
-    grid-auto-columns: min-content min-content 1fr min-content min-content;
+    // On desktop, we have 5 columns: back, logo, search, donate, menu
+    grid-auto-columns: min-content min-content 1fr auto min-content;
     grid-gap: 1rem;
   }
 `;
@@ -128,6 +131,7 @@ const row = css`
 export const GridItemBackLink = styled.div`
   ${row};
   grid-column: 1/2;
+  // Back link is hidden on small mobile
   display: none;
 
   @media (min-width: ${smallPhoneBreakpoint}) {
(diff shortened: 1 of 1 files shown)


src/components/NavBar/NavBar.style.ts
```diff
@@ -111,12 +111,15 @@ export const MenuLabel = styled.span`
...
+  // On mobile, we have 4 columns: back, logo, search, menu
+  // But note that after you scroll, the search bar expands to use the logo column as well.
   grid-auto-columns: min-content auto auto min-content;
   align-items: center;
   width: 100%;
 
   @media (min-width: ${mobileBreakpoint}) {
-    grid-auto-columns: min-content min-content 1fr min-content min-content;
+    // On desktop, we have 5 columns: back, logo, search, donate, menu
+    grid-auto-columns: min-content min-content 1fr auto min-content;
```
mikelehen, line 122: I noticed the donate button div was getting a bit truncated (in particular the hover background effect was smaller than the button, see below). Not sure if this is new or not, but I seemingly fixed it by changing the column width from `min-content` to `auto`. Not 100% sure if this is the best fix, but I think it works.

image

src/components/NavBar/NavBar.style.ts
```diff
@@ -147,16 +152,18 @@ export const GridItemLogo = styled.div<{ hasScrolled: boolean }>`
 
 export const GridItemSearch = styled.div<{ hasScrolled: boolean }>`
   ${row};
+  // On mobile, search expands to also use the logo column after scrolling.
   grid-column: ${({ hasScrolled }) => (hasScrolled ? '2/4' : '3/4')};
 
   @media (min-width: ${mobileBreakpoint}) {
-    grid-column: 1/6;
+    grid-column: 3/4;
```
mikelehen, line 159: This is the main fix.  I made the search bar only use the 3rd column instead of covering columns 2-5.

chasulin: ooo, i originally set it to span all columns on desktop so that it would be centered within the whole nav bar, not just within its allotted space-- but 3/4 seems to accomplish that 🙌🏼 🤷🏻‍♀️

chasulin: oh actually, jk, its thrown off slightly // no longer centered within the navbar as whole. but its not really noticeable, and getting the navbar fully functional is more important

chasulin (approved):
thanks for fixing this + for adding in the explanatory comments 🙏🏼

src/components/NavBar/NavBar.style.ts
mikelehen: Aha! I see, yeah, that makes sense.  Thanks for explaining. I'll go forward with this fix but I guess if we wanted to keep the prior behavior we probably could, we'd just need to start playing with z-order to make sure the buttons end up on top of the search bar.

PR #4457, opened 2021-10-02
keep minimap selection in local storage

schamp:
remove logic resetting to vaccination status on navigation, and persist setting in localstorage if available.

### DIFF
--- a/src/common/utils/storage.ts
+++ b/src/common/utils/storage.ts
@@ -6,63 +6,63 @@ export enum StorageKeys {
   FULLSTORY_RECORD = 'FULLSTORY_RECORD',
 }
 
+const storageMemo: { [index: string]: boolean } = {};
+
 // From https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#testing_for_availability
 export function storageAvailable(storageType: string) {
-  let storage;
-  const testKey = '__storage_test__';
-  try {
-    // @ts-ignore
-    storage = window[storageType];
-    storage.setItem(testKey, testKey);
-    storage.removeItem(testKey);
-    return true;
-  } catch (e) {
-    return false;
+  if (!(storageType in storageMemo)) {
+    let storage;
+    const testKey = '__storage_test__';
+    try {
+      // @ts-ignore
+      storage = window[storageType];
+      storage.setItem(testKey, testKey);
+      storage.removeItem(testKey);
+      storageMemo[storageType] = true;
+    } catch (e) {
+      storageMemo[storageType] = false;
+    }
   }
+  return storageMemo[storageType];
 }
 
-// Hook from https://usehooks.com/useLocalStorage/
-// but updated to fallback to setState if localstorage is not available.
+/*!
+   License: https://github.com/uidotdev/usehooks/blob/master/LICENSE
+   Hook from https://usehooks.com/useLocalStorage/
+   but updated to fallback to setState if localstorage is not available.
+  */
 export function useLocalStorage<T>(
   key: StorageKeys,
-  initialValue: T,
+  initialValue?: T,
 ): [T, (value: T) => void] {
   // State to store our value
   // Pass initial state function to useState so logic is only executed once
-  const [storedValue, setStoredValue] = useState(() => {
+  const [storedValue, setStoredValue] = useState<T>(() => {
+    let item;
     try {
-      // Get from local storage by key
-      let item;
       if (storageAvailable('localStorage')) {
         item = window.localStorage.getItem(key);
+        item = item ? JSON.parse(item) : initialValue;
       }
-      // Parse stored json or if none return initialValue
-      return item ? JSON.parse(item) : (initialValue as T);
     } catch (error) {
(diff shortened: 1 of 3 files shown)


schamp:
By request of @igorkofman

src/common/utils/storage.ts
```diff
@@ -0,0 +1,79 @@
...
+    storage.removeItem(testKey);
+    return true;
+  } catch (e) {
+    return false;
+  }
+}
+
+// Hook from https://usehooks.com/useLocalStorage/
+// but updated to fallback to setState if localstorage is not available.
+export function useLocalStorage<T>(
```
schamp, line 26: I was thinking I could update this to have the fallback built-in (i.e., just don't try to use localstorage if it isn't available).  Would that  be worth doing?

src/common/utils/storage.ts
```diff
@@ -0,0 +1,68 @@
...
+      // Save state
+      setStoredValue(valueToStore);
+      // Save to local storage
+      window.localStorage.setItem(key as string, JSON.stringify(valueToStore));
+    } catch (error) {
+      // A more advanced implementation would handle the error case
+      console.log(error);
+    }
+  };
+  if (storageAvailable('localStorage')) {
```
mimecuvalo, line 63: might be a little easier to follow along with this code if the final return value is the same in both cases, so:
```
const setValue = (value: T) => {
    // Allow value to be a function so we have same API as useState
    const valueToStore = value instanceof Function ? value(storedValue) : value;

    try {
      window.localStorage.setItem(key as string, JSON.stringify(valueToStore));
    } catch (error) {
      console.log(error);
    } finally {
      setStoredValue(valueToStore);
    }
  };

return [storedValue as T, setValue];
```

thoughts? otherwise, it feels a little tricky to follow the flow a bit.

src/common/utils/storage.ts
```diff
@@ -0,0 +1,68 @@
...
+// Hook from https://usehooks.com/useLocalStorage/
+// but updated to fallback to setState if localstorage is not available.
+export function useLocalStorage<T>(
+  key: StorageKeys,
+  initialValue: T,
+): [T, (value: T) => void] {
+  // State to store our value
+  // Pass initial state function to useState so logic is only executed once
+  const [storedValue, setStoredValue] = useState(() => {
+    try {
```
mimecuvalo, line 33: to simplify maybe:
```
let item
try {
  if (storageAvailable('localStorage')) {
    item = window.localStorage.getItem(key);
    item = JSON.parse(item)
  }
} catch {
  item = initialValue as T
}

return item
```

basically, just to put the JSON parsing closer to the `getItem` call to make it clearer that those actions go hand-in-hand.

src/common/utils/storage.ts
```diff
@@ -0,0 +1,68 @@
...
+  FULLSTORY_RECORD = 'FULLSTORY_RECORD',
+}
+
+// From https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#testing_for_availability
+export function storageAvailable(storageType: string) {
+  let storage;
+  const testKey = '__storage_test__';
+  try {
+    // @ts-ignore
+    storage = window[storageType];
```
mimecuvalo, line 15: interesting thing here is that it's testing two things, feature availability and space availability.
it seems a little moot to test for space (the `QuotaExceededError`) since you still have to try/catch below (the sizes of the objects we save aren't necessarily the same as the `__storage_test__` key above. 

so, to that end, seems like, this feature availability could be memoized - wdyt? try/catching to see whether there's still space available in the localStorage has to be done with every set anyway ¯\_(ツ)_/¯

src/common/utils/storage.ts
```diff
@@ -0,0 +1,68 @@
...
+  key: StorageKeys,
+  initialValue: T,
+): [T, (value: T) => void] {
+  // State to store our value
+  // Pass initial state function to useState so logic is only executed once
+  const [storedValue, setStoredValue] = useState(() => {
+    try {
+      // Get from local storage by key
+      let item;
+      if (storageAvailable('localStorage')) {
```
mimecuvalo, line 36: the funny thing is that there's already a try/catch around this so the `storageAvailable` call really needed? (other than signaling a little more strongly to the developer?)

src/common/utils/storage.ts
```diff
@@ -0,0 +1,68 @@
...
+
+// Hook from https://usehooks.com/useLocalStorage/
+// but updated to fallback to setState if localstorage is not available.
+export function useLocalStorage<T>(
+  key: StorageKeys,
+  initialValue: T,
+): [T, (value: T) => void] {
+  // State to store our value
+  // Pass initial state function to useState so logic is only executed once
+  const [storedValue, setStoredValue] = useState(() => {
```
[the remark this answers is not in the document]
schamp: _yes_ thanks.

src/common/utils/storage.ts
schamp: had to tweak this slightly because item is nullable.

mimecuvalo (approved):
😎 looks good - thx!

PR #4557, opened 2021-10-21
Fix initial value for useLocalStorage() when local storage is disabled.

mikelehen:


mikelehen:
Fair point.  I agree the code is weirdly duplicative. The one nice thing about using `storageAvailable()` is that it will prevent you from getting repeat exception log messages in the console once we've determined storage is not available.  Probably doesn't really matter, but I like a clean console. 😁

PR #4622, opened 2021-11-02
Update vaccine map coloring to adjust for progress.

mikelehen:
Per https://www.dropbox.com/scl/fi/biyc9r2ddgxfuym3wjxqy/Update-Vaccination-Map-Color-Scale.paper?dl=0

### DIFF
--- a/src/common/colors.ts
+++ b/src/common/colors.ts
@@ -119,13 +119,13 @@ export function vaccineColorFromLocationSummary(
 
 export function vaccineColor(val: number): string {
   const colors = COLOR_MAP.VACCINATIONS_BLUE;
-  if (val < 0.4) {
+  if (val < 0.5) {
     return colors[0];
-  } else if (val < 0.5) {
-    return colors[1];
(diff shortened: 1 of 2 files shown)


mikelehen:
Note that the diff looks more complicated than it is.  I basically just changed "0.4" => "0.5" and "40%" => "50%", etc.  github's diff algorithm just found a different, less-intuitive solution to the minimal diff problem. 😬

BrettBoval (approved):
This is a good start.

It's clearly an improvement for the state map.

It's mostly an improvement for the county map (although now much of the middle America big-land-area but low population counties are the same color).

I'm a little mixed on how it looks for a specific low vaccination rate state (like Arkansas) where the zoomed map is almost all lowest rung.

Overall it's worth doing and we should ship it.

On a separate ticket, I'll raise the fact that we still "round" the text 79.6% becomes 80%, but we "floor" the color (must be >80) which is causing a text/color inconsistency.

PR #4624, opened 2021-11-02
Round vaccination % before calculating color to make it match map / compare table.

mikelehen:
https://trello.com/c/ByaUenuI/292-can-round-vs-floor-inconsistency-in-vx-color


BrettBoval (approved):
👍 

Sorry to hear about the circular dependency. Probably should comment out the import too to silence the "defined but never used" warnings.

PR #4699, opened 2021-11-17
Show disclaimers for blocked metrics and enable markdown.

mikelehen:
Before this change, disclaimers only showed up if the metric wasn't blocked (in the metric footer). We also didn't support markdown (so couldn't embed links) in disclaimers.

**Before:**

image

**After:**

image


mikelehen:
cc/ @BrettBoval FYI

chasulin (approved):
🚢

PR #4700, opened 2021-11-17
Remove unused renderDisclaimer() method per TODO.

mikelehen:
There was a TODO to delete renderDisclaimer() in src/common/metrics/interfaces.ts and indeed it is not used.  So deleting it to avoid confusion.


chasulin (approved):
sweeeet 🙏🏼

PR #4964, opened 2022-01-05
Adjust axis ticks

faihegberg:
- Adjust x axis so the labels don't squish together.
- Adjust y axis so each label isn't too long.

Before:
image

After:
image

### DIFF
--- a/src/components/Charts/utils.ts
+++ b/src/components/Charts/utils.ts
@@ -190,18 +190,8 @@ export const getXTickFormat = (
   }
 };
 
-/**
- * For mobile, we only show every other month
- */
-function getMobileDateTicks(tickValues: Date[]): Date[] {
-  return tickValues.filter((value: Date, i: number) => i % 2 === 0);
-}
-
 export function getFinalTicks(isMobile: boolean, ticks: Date[]): Date[] {
-  if (isMobile) {
-    return getMobileDateTicks(ticks);
-  }
-  return ticks;
+  return ticks.filter((value: Date, i: number) => i % (isMobile ? 3 : 2) === 0);
 }
 
 export function getTimeAxisTicks(
--- a/src/components/Explore/utils.ts
+++ b/src/components/Explore/utils.ts
@@ -156,25 +156,19 @@ export const formatDecimalAxis = (num: number, places = 2): string => {
   /**
    * If value is between 100k and 1m, display in abbreviated form.
    * Examples
-   * - 100,000 => 100k
-   * - 1,000,000 => 1.0m
-   * - 1,100,000 => 1.1m
+   * - 100,000 => 100K
+   * - 1,000,000 => 1M
+   * - 1,100,000 => 1.1M
    */
-  if (num > 99999) {
-    return num > 999999
-      ? `${(num / 1000000).toLocaleString(undefined, {
-          minimumFractionDigits: 1,
-          maximumFractionDigits: 1,
-        })}m`
-      : `${(num / 1000).toLocaleString(undefined, {
-          minimumFractionDigits: 0,
-          maximumFractionDigits: 0,
-        })}k`;
+  if (num >= 100_000) {
+    //@ts-ignore
+    return new Intl.NumberFormat('en', { notation: 'compact' }).format(num);
+  } else {
+    return num.toLocaleString(undefined, {
+      minimumFractionDigits: places,
+      maximumFractionDigits: places,
+    });
   }
-  return num.toLocaleString(undefined, {
-    minimumFractionDigits: places,
-    maximumFractionDigits: places,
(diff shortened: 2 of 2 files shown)


src/components/Charts/utils.ts
```diff
@@ -190,18 +190,8 @@ export const getXTickFormat = (
...
-function getMobileDateTicks(tickValues: Date[]): Date[] {
-  return tickValues.filter((value: Date, i: number) => i % 2 === 0);
-}
-
 export function getFinalTicks(isMobile: boolean, ticks: Date[]): Date[] {
-  if (isMobile) {
-    return getMobileDateTicks(ticks);
-  }
-  return ticks;
+  return ticks.filter((value: Date, i: number) => i % (isMobile ? 3 : 2) === 0);
```
faihegberg, line 194: At this point, even displaying a tick every other month on mobile is still making the labels squish together. So I adjusted it to be every quarter on mobile. On desktop however, I don't think we need to split them by quarter yet (doing so makes the x-axis labels appear pretty bare). So I adjusted it to every 2 months.

src/components/Explore/utils.ts
```diff
@@ -149,12 +149,39 @@ function getDatasetIdByMetric(metric: ExploreMetric): DatasetId {
...
+    return num > 999999
+      ? `${(num / 1000000).toLocaleString(undefined, {
+          minimumFractionDigits: 1,
+          maximumFractionDigits: 1,
+        })}m`
+      : `${(num / 1000).toLocaleString(undefined, {
+          minimumFractionDigits: 0,
+          maximumFractionDigits: 0,
+        })}k`;
+  }
```
faihegberg, line 173: As our labels are currently in thousands, I think this approach is sufficient for now instead of something more robust but complicated, like the ones in [this thread](https://stackoverflow.com/questions/2685911/is-there-a-way-to-round-numbers-into-a-reader-friendly-format-e-g-1-1k). However, if our labels become more precise, we'd probably want to revisit this.

pnavarrc: nit: Consider simplifying a bit to make it more readable, I would recommend to:

- Pass `'en'` explicitly to `.toLocaleString` to make sure that number formatting is consistent across the site (if a user has a different language set on their computer, the thousands and decimal separators could be different)
- Use either scientific notation `1e6` or `1_000_000` to make the numbers easier to read
- Consider having a second `if` statement instead of nesting the conditionals

```ts
if (num >= 1e6) {
  return `${(num / 1e6).toLocaleString('en', {/* ... */}}`;
}

if (num >= 1e5) {
  return `${(num / 1e3).toLocaleString('en', {/* ... */}}`;
}

return `${num.toLocaleString('en', {/* ... */)}`'
```

Feel free to leave as is if you think the original version is better.

src/components/Explore/utils.ts
```diff
@@ -149,12 +149,39 @@ function getDatasetIdByMetric(metric: ExploreMetric): DatasetId {
...
+  if (num === null) {
+    return '-';
+  }
+  /**
+   * If value is between 100k and 1m, display in abbreviated form.
+   * Examples
+   * - 100,000 => 100k
+   * - 1,000,000 => 1.0m
+   * - 1,100,000 => 1.1m
+   */
```
pnavarrc, line 162: Thanks for adding examples!  If you have time you could add a few unit tests based on these examples (but feel free to punt)

pnavarrc (approved):
Thanks for taking this! We will probably need to adjust again in 3-6 months, but that's okay.

src/components/Explore/utils.ts
```diff
@@ -149,12 +149,39 @@ function getDatasetIdByMetric(metric: ExploreMetric): DatasetId {
...
+          minimumFractionDigits: 0,
+          maximumFractionDigits: 0,
+        })}k`;
+  }
+  return num.toLocaleString(undefined, {
+    minimumFractionDigits: places,
+    maximumFractionDigits: places,
+  });
+};
+
```
pnavarrc, line 179: Actually, there is a much simpler implementation based on [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat)

```
if (num >= 100_000) {
  return new Intl.NumberFormat('en', { notation: "compact" }).format(1200000);
  // 1.2M
} else {
  return num.toLocaleString('en', {/* ... */});
}
```

That returns the number with thappropriatete suffix (which should be uppercased, apparently)

- 99000 -> 99K
- 1000000 -> 1M

pnavarrc:
[Overlapping Dates in Explore](https://trello.com/c/K8D810NK/129-overlapping-dates-in-explore)

PR #4968, opened 2022-01-06
Display -- instead of 0% in map tooltip

faihegberg:
When we block data, our vaccination map will display 0% vaccination, which is incorrect. This PR
- Changes "0%" to "--" when there is no data.
- Greys out the vaccination progress bar when there is no data.

Before:
image

After:
image


faihegberg:
["No Trustworthy Data" Vaccination Tooltip](https://trello.com/c/idE2kQwL/128-no-trustworthy-data-vaccination-tooltip)

chasulin (approved):

PR #4977, opened 2022-01-08
ICU Y-axis fix

faihegberg:
Fixes Y-axis for ICU hospitalizations so the numbers don't get truncated.

Before:
image

After:
image


mikelehen (approved):
This is an improvement and seems to fix the USA, but it doesn't work as well for small locations:

image

Not sure what the right fix is offhand, but if we want to ship this just to fix the USA, I'm fine with that.

PR #4983, opened 2022-01-10
Add disclaimer and dashed line for capped vaccination metrics

smcclure17:
addresses https://trello.com/c/1NzGqUCv/159-add-automatic-vaccine-disclaimer-for-locations-at-95-our-cap

If we feel the dashed line/tooltip changes add confusion I can remove them--I know they weren't in the Trello ticket, but I  wanted to take a stab at it to see how it might look.  

My javascript is pretty rough, so if there are better ways of doing this I can update/fix it, or if we just don't want to use this code that's okay. 

image


src/common/metrics/vaccinations.tsx
```diff
@@ -68,13 +68,25 @@ function renderStatus(projections: Projections): React.ReactElement {
   const peopleVaccinated = formatInteger(info.peopleVaccinated);
   const percentVaccinated = formatPercent(info.ratioVaccinated, 1);
 
+  const cappedVaccinatedCopy =
+    info.ratioInitiated >= 0.95 ? (
+      <Fragment>
+        <br />
```
smcclure17, line 74: There must be a better way of adding a line break, but this is all I could find that worked

src/components/Explore/Explore.style.ts
```diff
@@ -14,6 +14,12 @@ function palette(props: any) {
   return props.theme.palette.chart;
 }
 
+/** By default make line charts solid,
+ * If a value is provided, use that instead.*/
+function getStrokeDashType(props: any) {
```
smcclure17, line 19: I  tried to make this inline (without declaring a named function) but I got a type error b/c `props.strokeDashArray` had a type of `any`. Not sure if this is significant but I just wanted to note it

mikelehen (changes requested):
👏 👏 👏   Thanks for taking this on in the frontend codebase. 🎉 

This basically looks good. I think the `strokeDashArray` can be cleaned up and I have a few nits.

src/common/metrics/vaccinations.tsx
smcclure17: that worked great! It did add the new line at the end if we're okay with that

image

src/components/Charts/ChartVaccinations.tsx
```diff
@@ -224,18 +231,16 @@ const VaccinationLines: React.FC<{
...
-                  data={data.filter(d => d.y >= 0.95)}
+                  data={data.filter(d => d.y >= VACCINATION_PERCENTAGE_CAP)}
                   x={getXPosition}
                   y={getYPosition}
                   type={type}
                   yMax={innerHeight}
                   barWidth={0}
                   barOpacity={1}
-                  params={{ ...params, ...dashedLineParam }}
+                  params={{ ...params, ...{ strokeDasharray: '1, 6' } }}
```
mikelehen, line 241: nit: I think this can just be `{ ...params, strokeDasharray: '1, 6' }`

src/components/Charts/ChartVaccinations.tsx
```diff
@@ -68,7 +74,7 @@ const VaccinesTooltip: React.FC<{
   const pointCompleted =
     seriesCompleted && findPointByDate(seriesCompleted.data, date);
   const pointInitiated = findPointByDate(seriesInitiated.data, date);
-  const isCapped = pointInitiated?.y >= 0.95 ?? false;
+  const isCapped = pointInitiated?.y >= VACCINATION_PERCENTAGE_CAP ?? false;
```
mikelehen, line 75: nit: I just realized the `?? false` is probably nonsensical / unreachable.  If `pointInitiated` is null/undefined then `pointInitiated?.y` will be undefined, but `undefined >= 0.95` will be `false` and so the `?? false` will never kick in... unless I'm missing something.  So I'd remove it to avoid confusion.

To make the code slightly cleaner and not rely on the behavior of `>=` on `undefined`, you could rewrite it as `(pointInitiated?.y ?? 0) >= VACCINATION_PERCENTAGE_CAP`

mikelehen (approved):
Sorry, noticed a couple other nits, but otherwise looks good!  Thanks again.

src/components/Charts/ChartVaccinations.tsx
smcclure17: ah yeah that makes more sense.

I think I was misinterpreting the original logic as "if `pointInitiated.y` is undefined then skip the comparison and return `false`". 

made this change 🙏

src/components/Charts/ChartVaccinations.tsx
smcclure17: it didn't seem to like that :( it threw some errors:

```JSX spread child must be an array type.ts(2609)``` and ```Left side of comma operator is unused and has no side effects.ts(2695)```

File: src/components/Charts/ChartVaccinations.tsx [hunk 5]
```diff
@@ -209,17 +215,32 @@ const VaccinationLines: React.FC<{
...
-                y={getYPosition}
-                type={type}
-                yMax={innerHeight}
-                barWidth={0}
-                barOpacity={1}
-                params={params}
-              />
+              /* Create separate lines for capped and un-capped data points.
+              Capped data will have a dashed line */
+              <Group key={`series-chart-${label}`}>
```
smcclure17, line 220: using `yarn start` displayed console warnings that `yarn serve` didn't. They were relating to unique keys of the elements. 

image

There are some other warnings but those seem to also exist on `develop`

src/components/Charts/ChartVaccinations.tsx
mikelehen: Oh, I should have said `{{ ...params, strokeDasharray: '1, 6' }}`

The outer `{ }` are used to insert a JavaScript expression inside the JSX block.  The inner `{  }` are creating a JS object literal.  So that needs to stay the same.  The thing I was pointing out that could be simplified was the `...{ strokeDasharray: '1, 6' }`.  That was creating an object literal with a single`strokeDasharray` property and then merging it into the parent object with the spread operator (`...`), which is unnecessary. You can just define the `strokeDasharray` property directly on the parent object.

That said, it turns out when you do that, TypeScript apparently gets more smart / picky about the types here and complains:
image

And that's my fault.  I think I was wrong [here](https://github.com/covid-projections/covid-projections/pull/4983/files/00da19cf1c425ef744e8ab11d6be4f5fa68041ef#r781638356).  You need to add `strokeDasharray` back to `SeriesParams` in `src/components/Explore/interfaces.ts`.  Sorry about that!

mikelehen (approved):

src/components/Charts/ChartVaccinations.tsx
smcclure17: Ohh I see, and no worries at all! Thanks for all the explanations, they've really been a big help 🙏

smcclure17: updated this in 362d2890d1b452333dc3a91ca24f1bb4211c4211 and am going to 🚢

PR #4988, opened 2022-01-10
Fix decimal values for ICU metric

faihegberg:
The initial [fix](https://github.com/covid-projections/covid-projections/pull/4977) to fully display Y-axis labels ended up introducing a bug to small numbers for ICU hospitalization metric.

Attempts to fix:
- Tried playing around with `text` and `tspan` components within the `AxisLeft` component, but that messes with the rest of the graph as they're all nested in a generated SVG.
- Explored adding additional logic to display decimal depending on the range of values, but that seems pretty complicated for a white space fix.

Ended up just adjusting the left margin to accommodate the Y-axis labels and getting rid of white space on the X-axis.

Before:
image

After:
image

### DIFF
--- a/src/components/Charts/utils.ts
+++ b/src/components/Charts/utils.ts
@@ -186,7 +186,7 @@ export const getXTickFormat = (
       date.getMonth() === 0 || date.getMonth() === 11
         ? DateFormat.MMM_YY
         : DateFormat.MMM;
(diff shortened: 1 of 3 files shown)


pnavarrc (approved):
Looks good for now, I think we will need to figure out a better logic to do this (there are some cases where we show the decimal point, but we don't really need to, etc), but we can take care of that next time.

PR #5008, opened 2022-01-13
Fix looking up of region overrides to ignore overrides with a start/end date.

mikelehen:
This fixes the Idaho disclaimer.  We were accidentally looking up a different date-based override for Idaho rather than the block w/ disclaimer one.

The frontend doesn't do anything with date-restricted overrides so we can just ignore them.


smcclure17 (approved):
looks good! thanks for this 🚢

PR #5029, opened 2022-01-19
compare table: replace Vulnerabilities with Hospitalizations per 100k

mimecuvalo:
This is my first time touching this code so lemme know if there's a better way to write it up!
Trello: https://trello.com/c/ONl3pEY3/180-replace-vulnerabilities-with-hospitalizations-in-compare-table


mimecuvalo:
updating this to be "Ungraded Hospitalizations/100K"

src/common/models/Projection.ts
```diff
@@ -327,6 +333,22 @@ export class Projection {
     return getPercentChange(firstVal, lastVal);
   }
 
+  private getHospitalizationInfo(): HospitalizationInfo | null {
+    const latestDatapoint = this.smoothedHospitalizations[
+      this.smoothedHospitalizations.length - 1
+    ];
```
mikelehen, line 339: I think if the hospitalization data is at all lagging (e.g. we have data for 2 days ago but not yesterday) then this will fail.  This is probably why counties aren't showing up right now.  County data often lags by ~3 days.  There's a `lastValue()` helper that's used throughout this file that I think should work.

mimecuvalo: ahhhhh thanks!

src/cms-content/learn/learn-faq.json
```diff
@@ -152,7 +152,7 @@
...
...nic minorities. For example, minorities tend to have [higher rates of poverty](https://www.kff.org/other/state-indicator/poverty-rate-by-raceethnicity/?currentTimeframe=0&sortModel=%7B%22colId%22:%22Location%22,%22sort%22:%22asc%22%7D), according to data from the Kaiser Family Foundation, and [less access to health care](https://www.cdc.gov/coronavirus/2019-ncov/community/health-equity/race-ethnicity.html). This puts them at greater risk of underlying medical conditions that result in more severe COVID cases. [The CDC](https://www.cdc.gov/coronavirus/2019-ncov/community/health-equity/race-ethnicity.html) reports that racial and ethnic minorities are also more likely to be essential workers, live in multi-generational housing, and have less ability to take time off from work when sick or to work from home. Discrimination, including racism, contributes to long-term social deficits and more immediately to inadequate health care access, according to the [CDC](https://www.cdc.gov/coronavirus/2019-ncov/community/health-equity/race-ethnicity.html). Learn more about vulnerability data and why it matters [in our explainer](https://covidactnow.org/covid-explained/covid-vulnerability-data).",
```
mikelehen, line 155: Oh, nice catch! I was confused why you were changing Learn content... 😁

src/components/Compare/columns.tsx
```diff
@@ -163,29 +164,58 @@ class VaccinationsColumn extends MetricColumn {
...
 }
 
+class HospitalizationsDensityColumn implements ColumnDefinition {
+  columnId = HOSPITALIZATIONS_DENSITY_ID;
+
+  name = 'Hospitalizations per 100k';
+
+  desiredWidthPercent = 14;
+  // Make it wide enough so progress bar doesn't look squashed on mobile.
+  minWidthPx = 130;
```
mikelehen, line 174: This was specifically so the vaccine progress bar would show up nicely. We probably don't need it for this column?

src/components/Compare/columns.tsx
```diff
@@ -163,29 +164,58 @@ class VaccinationsColumn extends MetricColumn {
...
+  }
+}
+
 // All of the compare table columns we use.
 const caseDensityColumn = new MetricColumn(Metric.CASE_DENSITY);
 const caseGrowthRateColumn = new MetricColumn(Metric.CASE_GROWTH_RATE);
 const positiveTestsColumn = new MetricColumn(Metric.POSITIVE_TESTS);
 const vaccinationsColumn = new VaccinationsColumn();
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const ccviColumn = new CcviColumn();
```
mikelehen, line 205: Any reason we're keeping `ccviColumn` if it's unused?

mikelehen (approved):
Looks good to me pending the counties fix.  Nice work!

src/components/Compare/columns.tsx
mimecuvalo: ah, looks like the comment is unrelated (oops) but we still need minWidthPx to make the header look right

src/components/Compare/columns.tsx
mimecuvalo: kinda seems a shame to delete all the `CcviColumn` and its related code I guess is why.

i'll remove this reference and just move the eslint disable to the `class` above actually.

smcclure17:
I think this looks good to me! thanks for this 🙏

smcclure17 (approved):

PR #5032, opened 2022-01-19
Booster Vaccine Line

smcclure17:
Update charts to include booster data. 

Add a third line to the location page vaccine charts and booster data to the Trends/Explore chart. In the process, update the "Fully Vaccinated" language to "2+ doses or J&J", and slightly change the colors of the vaccine chart lines to improve accessibility.


src/api/schema/Actuals.d.ts
```diff
@@ -32,7 +32,6 @@ export type Contacttracers = number | null;
  *  * capacity - Current staffed acute bed capacity.
```
smcclure17, line 32: all of these `src/api/*` files were modified by `yarn update-api-types`

src/common/colors.ts
```diff
@@ -75,6 +75,7 @@ export const COLOR_MAP = {
     // rename these
     BASE: '#3567FD',
     DARK: '#002CB4',
+    DARKEST: '#000824',
```
smcclure17, line 78: I chose this color randomly (I pulled up a color palette and used a color darker than `DARK`) so we might want to have design green-light this

src/components/Charts/MetricChart.tsx
```diff
@@ -109,14 +109,25 @@ function getVaccinationSeries(projection: Projection): Series[] {
     {
       type: SeriesType.LINE,
       data: filterNull(projection.getDataset('vaccinationsCompleted')),
-      label: 'Vaccinations Completed',
```
smcclure17, line 112: I'm not sure if we want to change the meaning of "full vaccinated" yet, so we should probably get this confirmed

src/components/NewLocationPage/ChartTabs/ChartTab.style.tsx
```diff
@@ -40,6 +40,7 @@ export const Tab = styled(MuiTab).attrs(props => ({
 }))`
   line-height: 1;
   text-align: left;
+  max-width: 372px;
```
smcclure17, line 43: This modifies the max-width of the title/header data above the chart--without this, it was cutoff and squished. This doesn't feel great to edit, though, so if there's another way to fix this we should use that.

I arbitrarily chose 372 because it's divisible by 12, but idk if we're using any sort of grid system where that matters. 

before:
image

after:
image

src/common/colors.ts
mikelehen: Yeah, let's get Josh to take a look before merging this.

src/components/Charts/MetricChart.tsx
mikelehen: FWIW- The latest copy I've seen is "2+ doses or J&J" but yeah, I'm not sure we should change this until we change the rest of the site to match.

src/components/NewLocationPage/ChartTabs/ChartTab.style.tsx
mikelehen: Hrm. I'm not sure I understand this change.
1. If the goal was to make the tab wider, I'd expect you to maybe add a min-width rather than a max-width.
2. When I look in Chrome dev tools it looks like the style isn't even being applied because it's overridden by a `width` specified in a more specific rule:

image

Also, depending on the screen size I can still make it look kinda' funky, e.g. according to chrome dev tools mobile mode, on an iPhone SE it will look like this:
image

In general I'm terrible at CSS but it feels like we need some margin or padding maybe?  Not sure. Maybe ask Pablo if you get stuck.

mikelehen (commented):
I think we need to confirm a couple things and do some data quality checks before merging, but this looks good. Thanks for tackling!

src/components/NewLocationPage/ChartTabs/ChartTab.style.tsx
smcclure17: hmm yeah, good flag--looking back I'm not sure how/why this works (or how to fix it on mobile). I've experimented some, and it looks like using `min-width` instead of `max-width` also works, but this doesn't really address the problem. 

I'll probably ask at stand-up today if anyone could lend a hand

smcclure17: as for what the change did, I think it expands the div that contains the three vaccination % divs, which seems to allow them to spread out more

image

image

pnavarrc: I can take this UI change, it's a bit more complex than it looks

pnavarrc (approved):
Looks good! Thanks @smcclure17 !

PR #5143, opened 2022-02-10
Fix repetitive USA labels in Explore

faihegberg:
Currently USA fips code is added multiple times to the list feeding the Explore chart. This PR prevents this by ensuring USA fips code is only added if it doesn't exist in the list yet.

Before:
image

After:
image


faihegberg:
[Bug: USA USA USA](https://trello.com/c/gHSRhqVH/286-bug-usa-usa-usa)

PR #5159, opened 2022-02-14
Add CTA section to Recommend in CMS

faihegberg:
This PR adds a CTA section to "Recommend" in CMS. To test:
1. Pull this branch.
2. Run `yarn start`.
3. In config.yml, update line 7 to `branch: faihegberg/add-cta-section-to-recommend-in-cms`.
4. Open `http://localhost:3000/admin/`.
5. Navigate to Recommend section and check that the section exists.

NOTE: I make the new icons field optional for now so it doesn't break the app. Once merged, I will move each icon over in a follow-up PR.


public/admin/config.yml
```diff
@@ -310,6 +310,14 @@ collections:
...
               - { label: Alt text, name: altText, widget: string }
               - { label: Icon image, name: iconImage, widget: image }
+          - label: CTA
+            name: ctaLinks
+            widget: list
+            fields:
+              - { label: Category, name: category, widget: string }
+              - { label: Link, name: link, widget: string }
+              - { label: Button type, name: buttonType, widget: string }
+              - { label: Button text, name: buttonText, widget: string }
```
pnavarrc, line 320: Consider adding a few settings to make it easier for the user to add the information

- Maybe rename Link → URL (links have text and URL, renaming this clarifies that we need the URL)
- Consider renaming `Button text` → Text, to match the language we use with links (link text, link URL)
- Consider using the Select widget for Category and Button Type (see https://www.netlifycms.org/docs/widgets/#select), that way the user can select from a predefined list of options instead of having to add free text (which we will need to validate later)

Also maybe we can make this field optional? (`required: false`), and maybe set the `max` property on the list to ensure that the list never has more than 1 or 2 elements.

pnavarrc (changes requested):
Added some suggestions!

public/admin/config.yml
faihegberg: Implemented all of these except for the `max` property. I'm not sure whether the `max` property will give us the result we want? What would be helpful is restricting one CTA button per category (i.e. one CTA button for MASKS, one CTA button for GATHERINGS, and so on), instead of the restricting the maximum total number of buttons. Since the latter would still allow one to add three CTA buttons for MASKS, for example.

pnavarrc: How about something like this instead? I feel like it will be better if each recommendation has the icon and CTA inside it instead of combining them by category


- recommendations
   - body
   - source
   - category
   - level
   - id
   - icon
     - altText
     - image 
   - cta
     - text
     - url
     - buttonType

pnavarrc (approved):
Looks good!

PR #5160, opened 2022-02-15
Update About page

schamp:


schamp:
Feel free to review but don't merge pending stamp by Mikayla on https://trello.com/c/dUg7wEY4/238-can-about-page-updates

schamp:
Mikayla confirms the design is right, so just looking for code review now.

src/cms-content/about/index.tsx
```diff
@@ -30,8 +30,6 @@ interface AboutContent {
   partnersContent: PartnersContent[];
   teamHeader: string;
   teamIntro: Markdown;
-  futureProjectsHeader: string;
-  futureProjectsContent: Markdown;
```
mikelehen, line 34: We should remove these from the CMS config too if we no longer support them: https://github.com/covid-projections/covid-projections/blob/develop/public/admin/config.yml#L98

src/screens/About/About.tsx
```diff
@@ -141,15 +137,9 @@ const About = () => {
             <MarkdownContent source={aboutContent.impactContent} />
           </SectionContent>
           <ButtonContainer>
-            <Button to="/case-studies" cta="View case studies" />
+            <Button href="https://actnowcoalition.org" cta="Learn more" />
```
mikelehen, line 140: Nit: Can you link to https://www.actnowcoalition.org/ instead (with www) to avoid a redirect?

mikelehen (approved):
Looks good with a couple nits.

schamp:
Can do, thanks

PR #5165, opened 2022-02-16
Move icons to be a part of recommendation

faihegberg:
This PR gets rid of icons as a separated list from recommendations, and add each icon to a recommendation item instead.


mikelehen (approved):
Nice! This makes the code simpler / cleaner.  👍

PR #5173, opened 2022-02-16
Create buttons for recommendation section

faihegberg:
This PR creates CTA buttons for the Recommentations section. TWO ways to test:

First way
1. Check out this branch.
2. Add CTA content to any recommentation object in recommendations-main.json, such as:
```
      "cta": {
        "buttonType": "FILL",
        "text": "Button text",
        "url": "https://www.covidactnow.org/"
      }
```
3. Check out Recommendations section on any location page.

Second way
1. Check out this branch.
2. In `config.yml` change line 7 to `branch: faihegberg/create-buttons-for-recommendation-section`.
3. Run `yarn start`.
4. Navigate to http://localhost:3000/admin then to Recommendation section.
5. Add CTA content to any Recommendation component. For example,

image

6. Click "Publish".
7. Run `git pull origin faihegberg/create-buttons-for-recommendation-section`. `recommendations-main.json` should change based on what you just changed.
8. Check out Recommendations section on any location page.


src/components/Recommend/CTAButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+      </StyledFilledButton>
+    );
+  } else if (
+    validFields &&
+    getButtonType(cta.buttonType) === ButtonType.OUTLINE
+  ) {
+    return (
+      <StyledOutlinedButton
+        trackingCategory={EventCategory.RECOMMENDATIONS}
+        trackingLabel={`${category}: CTA button`}
```
pnavarrc, line 51: Maybe the category will be enough, "CTA button" doesn't add much information and might not fit in the space that Analytics/Amplitude has for event labels (this is more their fault than ours)

src/components/Recommend/Recommend.style.tsx
```diff
@@ -72,3 +76,30 @@ export const Icon = styled.img`
...
+      props.theme.buttons[ButtonType.OUTLINE].backgroundHover};
+    border: ${props =>
+      `1px solid ${props.theme.buttons[ButtonType.OUTLINE].borderHover}`};
+  }
+
+  @media (min-width: ${materialSMBreakpoint}) {
+    width: initial;
+    margin: 0.75rem 0 1rem 0;
+  }
+`;
```
pnavarrc, line 105: The outlined button and the filled button look the same, can you check if that's what we want? (I changed that locally)

Also, consider implementing a container for the button so you don't need to expose styled versions of each one (we shouldn't be styling the outlined button here, we should use the hover style already implemented)

pnavarrc (changes requested):
Looks good, added some suggestions!

pnavarrc (approved):
Looks good! (requested changed by accident) - consider adding the new icons to make them available in the CMS (can be on a separate PR)

PR #5204, opened 2022-02-22
Remove expandable container from Recommend

faihegberg:
Remove expandable container from the Recommendations section on desktop, as per https://trello.com/c/e7c6H4YW/228-update-recommendations-module

### DIFF
--- a/src/components/LocationPage/Recommendations.tsx
+++ b/src/components/LocationPage/Recommendations.tsx
@@ -12,6 +12,7 @@ import { getRecommendationsShareUrl } from 'common/urls';
 import { getRecommendations, getShareQuote } from 'common/utils/recommend';
 import { Region } from 'common/regions';
 import { SectionHeader } from 'components/SharedComponents';
+import { useBreakpoint } from 'common/hooks';
 
 interface RecommendationsProps {
   projections: Projections;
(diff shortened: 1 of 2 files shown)


faihegberg:
[Update recommendations module](https://trello.com/c/e7c6H4YW/228-update-recommendations-module)

src/components/LocationPage/Recommendations.tsx
```diff
@@ -38,8 +39,10 @@ const Recommendations = ({
     alarmLevel,
   );
 
+  const isMobile = useBreakpoint(600);
+
   const containerProps = {
-    collapsedHeightMobile: '275px',
+    collapsedHeightMobile: '350px',
     collapsedHeightDesktop: '165px',
```
mikelehen, line 46: It's tempting to rework the component so that you can specify `undefined` for `collapsedHeightDesktop` and then the component would only collapse/expand on mobile.  I think that'd be a bit more elegant (we don't have to pass this `collpasedHeightDesktop` property that is irrelevant and the `<Recommend ... />` code wouldn't need to be duplicated below).  It'd also be more reusable (if there are other places that we want expand/collapse only on mobile then we could use this same mechanism).

But what you have is okay and we could refactor if/when we have another usage for this functionality. 🤷

mikelehen (approved):
Looks good with a drive-by idea that you don't have to act on.

src/components/LocationPage/Recommendations.tsx
faihegberg: I agree :) I'll add a TODO in case we have another usage.

PR #5361, opened 2022-03-21
feature flag for upcoming CDC updates

pnavarrc:
Implement a feature flag based on environment variables. Create React App (CRA) uses a similar system than Next.js, but the environment variable names need to be prefixed with `REACT_APP_` (see the [CRA Docs](https://create-react-app.dev/docs/adding-custom-environment-variables/)).

In development and preview environments, we can set the value of the environment variable in the `.env.development` file. For the production environment, I added the environment variable in [Vercel settings](https://vercel.com/covidactnow/covid-projections/settings/environment-variables) and set it to `false`.

Note that I only disabled the vaccination scoreboard in this PR, I will leave other changes for a separate PR (moving the vaccination map, etc).

### Preview & Local Development

image

### Production

image


src/common/utils/index.ts
```diff
@@ -113,3 +113,12 @@ export function getPercentChange(numA: number, numB: number): number | null {
...
   num === 1 ? singular : plural;
+
+/**
+ * Returns true if the CAN 8.2 feature flag is enabled. The value should be
+ * updated in `.env.development` for local development and preview deployments,
+ * and in https://vercel.com/covidactnow/covid-projections/settings/environment-variables
+ */
+export function enableCan82(): boolean {
+  return process.env.REACT_APP_ENABLE_CAN_8_2 === 'true';
+}
```
pnavarrc, line 124: This variable doesn't add much value, except that I have seen people comparing the value as a boolean or trying to deconstruct `process.env` to read the value, both of which wouldn't work

```
// This doesn't work
const { REACT_APP_ENABLE_CAN_8_2 } = process.env;

// won't work either - env. variables are always strings
if (process.env.REACT_APP_ENABLE_CAN_8_2) {
  // ...
}

```

igorkofman:
Doesn't this create a risk that someone makes a change that looks great in dev and in staging but completely breaks prod?

PR #5368, opened 2022-03-22
Initial changes for CAN 8.2

igorkofman:


mikelehen (changes requested):
The trello card I see says we are re-ordering the maps, not removing vaccination map https://trello.com/c/RJxsldMz/422-reorder-vaccination-and-risk-level-maps-on-homepage

Is the trello card wrong? Or not in sync with the PRD or something? 🤷

mikelehen:
[Reorder Vaccination and Risk Level maps on Homepage](https://trello.com/c/RJxsldMz/422-reorder-vaccination-and-risk-level-maps-on-homepage)

mikelehen (approved):
I would have probably left the `VaccinationsTable` component checked-in since we might reuse it (or want to build something similar to it) in the future, so it'd be handy to still have it visible in our storybook library of components for future use / inspiration.  I don't care strongly though.

PR #5377, opened 2022-03-23
Update thermometer

igorkofman:


src/components/HorizontalThermometer/CommunityLevelThermometer/CommunityLevelThermometer.tsx
```diff
@@ -3,17 +3,11 @@ import {
...
 import { LOCATION_SUMMARY_LEVELS } from 'common/metrics/location_summary';
 
-const levels = [
-  Level.LOW,
-  Level.MEDIUM,
-  Level.HIGH,
-  Level.CRITICAL,
-  Level.SUPER_CRITICAL,
-];
+const levels = [Level.LOW, Level.MEDIUM, Level.HIGH];
```
faihegberg, line 10: I think we still need the top two levels. Reducing this to three levels result in three colors in the thermometer although the map would still have five colors. This might be confusing to users since they won't know what red and dark red mean on the map.
image

faihegberg (changes requested):

src/components/HorizontalThermometer/CommunityLevelThermometer/CommunityLevelThermometer.tsx
igorkofman: In the future there will only be 3 colors, this only update the thermometer

mikelehen:
[Update Homepage Risk Map to Community Levels with County Default](https://trello.com/c/op05kBFv/421-update-homepage-risk-map-to-community-levels-with-county-default)

src/components/HorizontalThermometer/CommunityLevelThermometer/CommunityLevelThermometer.tsx
faihegberg: I think ideally we should update the map and the thermometer at the same time to minimize confusion for the user, but if the map will be updated in the near future, we can move forward 😄

faihegberg (approved):

src/components/HorizontalThermometer/CommunityLevelThermometer/CommunityLevelThermometer.tsx
igorkofman: This is in a feature branch - no user confusion should occur :). But potential for developer confusion is there.

PR #5401, opened 2022-03-28
Implement recommendations on homepage

igorkofman:


src/components/Recommend/Recommendations.tsx
```diff
@@ -3,42 +3,37 @@ import { mainContent } from 'cms-content/recommendations';
 import Recommend from 'components/Recommend';
 import ExpandableContainer from 'components/ExpandableContainer';
 import ShareButtons from 'components/SharedComponents/ShareButtons';
-import LocationPageSectionFooter from 'components/LocationPageSectionFooter/LocationPageSectionFooter';
-import { DisclaimerWrapper } from 'components/LocationPage/ChartsHolder.style';
+import PageSectionFooter from 'components/SharedComponents/PageSectionFooter';
+import { DisclaimerWrapper } from 'components/SharedComponents/DisclaimerWrapper.style';
```
mikelehen, line 7: I think our convention is to import from a directory-level index.ts file rather than from individual files within a component directory.  i.e.:

1. We should add exports for `PageSectionFooter` and `DisclaimerWrapper` to SharedComponents/index.ts, similar to how we already do for `SectionHeader`: https://github.com/covid-projections/covid-projections/blob/7514c6b94d61a51d2b2d324d1070aa3bdf436b92/src/components/SharedComponents/index.tsx#L13
2. These imports should become `import { PageSectionFooter, DisclaimerWrapper } from 'components/SharedComponents';` or similar

mikelehen (approved):
Looks good!  One minor style / consistency flag, but overall looks great.  Thanks for refactoring the shared styling...

PR #5458, opened 2022-04-04
WIP: Add CDC Community level metrics to `Projection` class

smcclure17:
Wires up the three new metrics and adds CAN community levels. This strictly just adds the metric timeseries and does not add any special handling (like is currently done for case density or ICU capacity).

Once https://github.com/covid-projections/covid-projections/pull/5473 is merged, we'll need to update `getMetricValue()` to include these new metrics, too.


src/common/models/Projection.ts
```diff
@@ -254,10 +270,16 @@ export class Projection {
     this.caseDensityByCases = metricsTimeseries.map(
       row => row && row.caseDensity,
     );
+
     this.caseDensityRange = this.calcCaseDensityRange();
 
+    this.weeklyNewCasesPer100k = metricsTimeseries.map(
+      row => row && row.weeklyNewCasesPer100k,
```
mikelehen, line 277: `row?.weeklyNewCasesPer100k ?? null`

src/common/models/Projection.ts
```diff
@@ -62,12 +65,16 @@ export type DatasetId =
...
   | 'rawDailyCases'
   | 'rawDailyDeaths'
   | 'rawHospitalizations'
   | 'smoothedHospitalizations'
   | 'rawICUHospitalizations'
-  | 'smoothedICUHospitalizations';
+  | 'smoothedICUHospitalizations'
+  | 'weeklyCovidAdmissionsPer100k'
+  | 'bedsWithCovidPatientsRatio'
+  | 'canCommunityLevel';
```
mikelehen, line 77: We probably don't need `canCommunityLevel` here.  We only need it if we're going to use it with the `getDataset(dataset: DatasetId)` method which is only used for timeseries data.

Just as an FYI- The way `getDataset()` takes a string and then directly reads that property from the class is pretty hacky / fragile and if we were writing this from scratch I wouldn't do it like this.

src/common/models/Projection.ts
```diff
@@ -173,6 +184,9 @@ export class Projection {
   private readonly smoothedHospitalizations: Array<number | null>;
   private readonly rawICUHospitalizations: Array<number | null>;
   private readonly smoothedICUHospitalizations: Array<number | null>;
+  private readonly weeklyCovidAdmissionsPer100k: Array<number | null>;
+  private readonly bedsWithCovidPatientsRatio: Array<number | null>;
+  private readonly canCommunityLevel: Array<CommunityLevel | undefined>;
```
mikelehen, line 189: I don't think we need to use the timeseries data (we don't ever graph risk levels in the frontend).  So this can just be `Level` (with us converting `null` => `Level.UNKNOWN`)

src/common/models/Projection.ts
```diff
@@ -283,6 +319,8 @@ export class Projection {
     return this.metrics?.testPositivityRatioDetails?.source || null;
   }
 
+  // TODO: Add new metrics to Metric Enum, add to getMetricValue() and resolve any issues
+  // from adding new metrics to Metric.
```
mikelehen, line 323: As mentioned in Slack, if you merge the latest `can-82` code I think we'll have the `Metric` enum values we need.

src/common/models/Projection.ts
```diff
@@ -512,13 +550,25 @@ export class Projection {
...
-  private makeDateDictionary(ts: ActualsTimeseries | Metricstimeseries) {
+  private makeDateDictionary(
+    ts: ActualsTimeseries | Metricstimeseries | Communitylevelstimeseries,
+  ) {
     const dict: {
-      [date: string]: ActualsTimeseriesRow | MetricsTimeseriesRow;
+      [date: string]:
+        | ActualsTimeseriesRow
+        | MetricsTimeseriesRow
+        | CommunityLevelsTimeseriesRow;
```
mikelehen, line 560: I think you can undo this since we don't need the timeseries.

src/common/models/Projection.ts
```diff
@@ -536,10 +586,14 @@ export class Projection {
   ) {
     const actualsTimeseriesRaw = summaryWithTimeseries.actualsTimeseries;
     const metricsTimeseriesRaw = summaryWithTimeseries.metricsTimeseries || [];
+    const communityLevelsTimeseriesRaw =
+      summaryWithTimeseries.communityLevelsTimeseries;
```
mikelehen, line 590: ditto

src/common/models/Projection.ts
```diff
@@ -559,6 +613,7 @@ export class Projection {
 
     const actualsTimeseries: Array<ActualsTimeseriesRow | null> = [];
     const metricsTimeseries: Array<MetricsTimeseriesRow | null> = [];
+    const communityLevelsTimeseries: Array<CommunityLevelsTimeseriesRow | null> = [];
```
mikelehen, line 616: ditto

mikelehen (changes requested):
Nice!  Mostly looks good.  I think canCommunityLevel should just be the latest value rather than the timeseries, and we can probably wire up `getMetricValue()` if you merge in the latest can-82 changes.

src/common/models/Projection.ts
smcclure17: Would that conversion need to happen here? e.g. with something like `summaryWithTimeseries.communityLevels?.canCommunityLevel ?? Level.UNKNOWN` or does that logic exist elsewhere

mikelehen: I think we can / should put it here so we get rid of the nullability as early as possible and downstream code doesn't have to deal with `Level | null` return types.

smcclure17: Cool, makes sense! Added changes in 1d88b91f85aad7fd04b32b2d19225184f2f5f1cf. I'm not really sure how enums work in this case, so I'm not sure if the levels automatically map (eg. 0 -> low), or we might need to do that mapping manually?

src/common/models/Projection.ts
```diff
@@ -165,6 +175,7 @@ export class Projection {
   private readonly vaccinationsAdditionalDose: Array<number | null>;
   private readonly caseDensityByCases: Array<number | null>;
   private readonly caseDensityRange: Array<CaseDensityRange | null>;
+  private readonly weeklyNewCasesPer100k: Array<number | null>;
```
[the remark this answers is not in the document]
mikelehen: Sounds good.  My guess is `CaseDensityRange` is crufty code that should be removed, but I'm not 100% sure.

src/common/models/Projection.ts
```diff
@@ -254,8 +269,24 @@ export class Projection {
...
     );
+
     this.caseDensityRange = this.calcCaseDensityRange();
 
+    this.weeklyNewCasesPer100k = metricsTimeseries.map(
+      row => row?.weeklyNewCasesPer100k ?? null,
+    );
+
+    this.weeklyCovidAdmissionsPer100k = metricsTimeseries.map(
+      row => row?.weeklyNewCasesPer100k ?? null,
```
mikelehen, line 280: Should be weeklyCovidAdmissionsPer100k.

src/common/models/Projection.ts
mikelehen: I think it will probably just work.  TypeScript enums are really just `string` or `number` values under the covers and the `Level` enum doesn't explicitly specify the values, so I think they'll be numbers starting at `0`.

mikelehen (approved):

src/common/models/Projection.ts
smcclure17: agh, thank you

src/common/models/Projection.ts
smcclure17: sweet, makes sense 🙏

PR #5471, opened 2022-04-05
Export subscribers emails to text file

faihegberg:
Create a script to export alert subscriber emails to a text file. Some notes below:
- Creating a separate script called `generate-subscriber-emails-file` in case we need to do something like this again in the future.
- Exporting to a text file instead of a CSV, since we only need one column.
- Currently outputting the text file to the repo's main folder. If there is somewhere better in the repo to output the file, let me know :)

### DIFF
--- a/scripts/alert_emails/firestore.ts
+++ b/scripts/alert_emails/firestore.ts
@@ -19,6 +19,7 @@ export async function fetchAllAlertSubscriptions(
       .stream()
       .on('data', subscription => {
         const data = subscription.data();
+        data.email = subscription.id;
         // Convert Firestore timestamp to plain Date.
         data.subscribedAt = data.subscribedAt && data.subscribedAt.toDate();
         subscriptions.push(data);
--- a/scripts/alert_emails_report/generate-subscriber-emails-file.ts
+++ b/scripts/alert_emails_report/generate-subscriber-emails-file.ts
@@ -6,7 +6,7 @@ async function main() {
   const db = getFirestore();
   const subscriptions = await fetchAllAlertSubscriptions(db);
   const fs = require('fs');
-  let output = subscriptions.map(s => s.email).join(',');
+  const output = subscriptions.map(s => s.email).join(',');
   fs.writeFileSync('subscriber-emails.txt', output);
 }
 
--- a/scripts/alert_emails_report/generate-subscriber-emails-file.ts
+++ b/scripts/alert_emails_report/generate-subscriber-emails-file.ts
@@ -1,0 +1,21 @@
+import _ from 'lodash';
+import { fetchAllAlertSubscriptions } from '../alert_emails/firestore';
+import { getFirestore } from '../common/firebase';
+
+async function main() {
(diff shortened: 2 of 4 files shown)


scripts/alert_emails/firestore.ts
```diff
@@ -19,6 +19,7 @@ export async function fetchAllAlertSubscriptions(
       .stream()
       .on('data', subscription => {
         const data = subscription.data();
+        data.email = subscription.id;
```
faihegberg, line 22: Currently, `subscription.data()` would return the `subscribedAt` and `locations` fields, but not the `email` field. To get the email field, it seems like to we need to explicitly call the subscription's `id` field.

faihegberg:
[Export the list of alert email subscribers to CSV](https://trello.com/c/nu5Dh5E6/533-export-the-list-of-alert-email-subscribers-to-csv)

scripts/alert_emails_report/generate-subscriber-emails-file.ts
```diff
@@ -0,0 +1,21 @@
+import _ from 'lodash';
+import { fetchAllAlertSubscriptions } from '../alert_emails/firestore';
+import { getFirestore } from '../common/firebase';
+
+async function main() {
+  const db = getFirestore();
+  const subscriptions = await fetchAllAlertSubscriptions(db);
+  const fs = require('fs');
+  let output = '';
+  subscriptions.map(subscription => {
```
mikelehen, line 10: You are using `map()` but you aren't actually returning a value from the callback function or using the return value of `map()`, so I think this could just be `forEach()`.

Alternatively you could probably write this as something like `const output = subscriptions.map(s => s.email).join(',');`

scripts/alert_emails_report/generate-subscriber-emails-file.ts
```diff
@@ -0,0 +1,21 @@
...
+    output = output.concat(subscription.email, ',');
+  });
+  fs.writeFileSync('subscriber-emails.txt', output);
```
mikelehen, line 13: You may want to check with Mikayla / Anya (or the mailchimp docs) about what formats mailchimp will import.  We might need a single email per line or something (which would basically be a valid CSV file as long as you added a single "email" header row to the top).

scripts/alert_emails/firestore.ts
mikelehen: Oh, nice catch. THanks for explaining!

mikelehen (approved):

scripts/alert_emails_report/generate-subscriber-emails-file.ts
faihegberg: 😅

scripts/alert_emails_report/generate-subscriber-emails-file.ts
faihegberg: Just checked with Mikayla, this text file format should be fine 👍🏼

PR #5473, opened 2022-04-05
Add new metrics to Metric enum, summary stats

chasulin:
- Adds new metrics to `Metric` enum + fix all related typescript errors
- Consolidates all metrics (old and new) to use `Level` (having both `Level` and `CommunityLevel` was spiraling into typescript difficulties. The only difference between the two is the presence/absence of `CRITICAL` and `SUPER_CRITICAL`, so new metrics will now include these levels but we just won't use them anywhere)
- Deletes `community_level.ts`
- Wires up new metrics to summary stats


src/common/metric.tsx
```diff
@@ -17,14 +20,27 @@ export const ALL_METRICS = [
   Metric.POSITIVE_TESTS,
   Metric.HOSPITAL_USAGE,
   Metric.VACCINATIONS,
+
+  // TODO(8.2)
```
mikelehen, line 24: What is left to do here?

src/common/metric.tsx
```diff
@@ -17,14 +20,27 @@ export const ALL_METRICS = [
...
 
-const ALL_METRICS_LEVEL_INFO_MAP = {
+const ALL_METRICS_LEVEL_INFO_MAP: { [metric in Metric]: LevelInfoMap } = {
   [Metric.CASE_GROWTH_RATE]: CaseGrowth.CASE_GROWTH_RATE_LEVEL_INFO_MAP,
   [Metric.POSITIVE_TESTS]: TestRates.POSITIVE_TESTS_LEVEL_INFO_MAP,
   [Metric.HOSPITAL_USAGE]: Hospitalizations.HOSPITAL_USAGE_LEVEL_INFO_MAP,
   [Metric.VACCINATIONS]: Vaccinations.VACCINATIONS_LEVEL_INFO_MAP,
   [Metric.CASE_DENSITY]: CaseDensity.CASE_DENSITY_LEVEL_INFO_MAP,
+
+  // TODO(8.2)
```
mikelehen, line 37: Ditto here. I think this code is probably good to go and doesn't need a todo?

src/common/metric.tsx
```diff
@@ -54,6 +70,13 @@ const metricDefinitions: { [metric in Metric]: MetricDefinition } = {
   [Metric.HOSPITAL_USAGE]: Hospitalizations.ICUCapacityUsed,
   [Metric.VACCINATIONS]: Vaccinations.VaccinationsMetric,
   [Metric.CASE_DENSITY]: CaseDensity.CaseIncidenceMetric,
+
+  // TODO(8.2)
```
mikelehen, line 74: Ditto.

src/common/metric.tsx
```diff
@@ -93,6 +116,11 @@ const MetricDecimalPlaces: { [metric in Metric]: number } = {
   [Metric.HOSPITAL_USAGE]: 0,
   [Metric.POSITIVE_TESTS]: 1,
   [Metric.VACCINATIONS]: 1,
+
+  // TODO(8.2)
```
mikelehen, line 120: Perhaps add ": Verify these are what we want." or something.

src/common/utils/provenance.tsx
```diff
@@ -103,6 +103,53 @@ const metricToSourceMap: RegionSourceMap = {
       url: 'n/a',
     },
   },
+
+  // TODO(8.2)
```
mikelehen, line 107: Can you add ": Fill in correct sources (or remove if we correctly get them from the API)."

FWIW- I don't 100% remember how provenance stuff works, but hopefully we can get these from the API and don't need frontend to specify them.  We'll need to verify though.

src/components/Explore/interfaces.ts
```diff
@@ -31,7 +31,7 @@ export enum ExploreMetric {
   VACCINATIONS_ADDITIONAL_DOSE,
   ICU_USED,
   POSITIVITY_RATE,
-  HOSPITALIZATIONS_PER_100K,
+  ADMISSIONS_PER_100K,
```
mikelehen, line 34: Hrm.  I think we over-renamed here.  ExploreMetric.HOSPITALIZATIONS_PER_100K should not have been affected.

src/components/Explore/utils.ts
```diff
@@ -94,7 +94,7 @@ export const EXPLORE_METRICS = [
   ExploreMetric.VACCINATIONS_ADDITIONAL_DOSE,
   ExploreMetric.ICU_USED,
   ExploreMetric.POSITIVITY_RATE,
-  ExploreMetric.HOSPITALIZATIONS_PER_100K,
+  ExploreMetric.ADMISSIONS_PER_100K,
```
mikelehen, line 97: ditto here and below.

src/components/NewLocationPage/SummaryStat/utils.ts
```diff
@@ -19,6 +19,11 @@ export const metricSubLabelText: { [key in Metric]: string[] } = {
   [Metric.CASE_GROWTH_RATE]: [''], // figure out deleting these empties without typescript complaining
   [Metric.HOSPITAL_USAGE]: [''],
   [Metric.POSITIVE_TESTS]: [''],
+
+  // TODO(8.2)
```
mikelehen, line 23: Maybe add ": Verify these."

mikelehen (approved):

src/common/metric.tsx
chasulin: oh whoops. left too many unnecessary TODOs

src/components/NewLocationPage/SummaryStat/utils.ts
```diff
@@ -20,9 +20,9 @@ export const metricSubLabelText: { [key in Metric]: string[] } = {
   [Metric.HOSPITAL_USAGE]: [''],
   [Metric.POSITIVE_TESTS]: [''],
 
-  // TODO(8.2)
+  // TODO(8.2): verify these
```
mikelehen, line 23: FWIW- I think these look correct now and we could probably remove the TODO.  I assume if they're not correct, Brett / Mikayla / somebody will flag it.

src/common/metrics/admissions_per_100k.tsx
```diff
@@ -19,6 +19,7 @@ export const AdmissionsPer100kMetric: MetricDefinition = {
   metricName: 'Weekly Covid Admissions per 100k',
   extendedMetricName: 'Weekly Covid Admissions per 100k population',
   metricNameForCompare: `Weekly Covid Admissions per 100k`,
+  metricNameForSummaryStat: 'Hospitalizations',
```
chasulin, line 22: Using the abbreviated metric names that are in the figma mocks

mikelehen (approved):
Nice!

PR #5475, opened 2022-04-05
New metrics into spark lines

chasulin:
Need to update with correct dataset IDs once API updates are merged in

### DIFF
--- a/src/components/NewLocationPage/SparkLineBlock/utils.ts
+++ b/src/components/NewLocationPage/SparkLineBlock/utils.ts
@@ -47,7 +47,6 @@ interface MetricDescription {
   seriesList: Series[];
 }
 
-// TODO(8.2) : set correct dataSetIds once API update is merged in
 export const sparkLinesMetricData: {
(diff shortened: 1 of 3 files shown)


src/components/NewLocationPage/SparkLineBlock/utils.ts
```diff
@@ -50,36 +50,36 @@ interface MetricDescription {
...
       {
-        datasetId: 'smoothedDailyCases',
+        datasetId: 'weeklyNewCasesPer100k',
       },
     ],
   },
-  [SparkLineMetric.DEATHS]: {
-    title: 'Deaths',
+  [SparkLineMetric.ADMISSIONS_PER_100K]: {
+    title: 'Hospitalizations',
```
mikelehen, line 65: Can you change this (and in the summary stat; sorry I missed that before) to "Admissions" or similar?  I'm not 100% sure what the final copy will be for these, but I'd like to make it clear that we are using admissions data, not the preexisting hospitalizations data.

mikelehen (approved):

PR #5476, opened 2022-04-05
Wire up correct dataset ids in explore

chasulin:
- Wires up correct dataset IDs in explore
- References correct metric in tooltip content util

### DIFF
--- a/src/components/Explore/utils.ts
+++ b/src/components/Explore/utils.ts
@@ -131,13 +131,11 @@ export function getMetricByChartId(chartId: string): ExploreMetric | undefined {
       return ExploreMetric.ICU_USED;
     case 'testPositiveRate':
       return ExploreMetric.POSITIVITY_RATE;
-
-    // TODO(8.2) - update with correct dataset id
-    case 'hospitalizationsPer100kId':
+    case 'weeklyCovidAdmissionsPer100k':
       return ExploreMetric.ADMISSIONS_PER_100K;
-    case 'ratioBedsWithCovidId':
+    case 'bedsWithCovidPatientsRatio':
       return ExploreMetric.RATIO_BEDS_WITH_COVID;
-    case 'weeklyNewCasesId':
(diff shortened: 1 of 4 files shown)


src/components/Explore/utils.ts
```diff
@@ -386,48 +382,47 @@ export const exploreMetricData: {
     ],
   },
 
-  // TODO(8.2) - update with real content:
   [ExploreMetric.ADMISSIONS_PER_100K]: {
     title: 'Hospitalizations per 100k',
     name: 'Hospitalizations per 100k',
-    chartId: 'positivity_rate',
+    chartId: 'admissions_per_100k', // TODO(8.2) (Chelsi) - what are these ids used for
```
chasulin, line 388: i _think_ these id's might have to do with the spark lines above the fold. ill poke around once both prs (this one and [the spark line pr](https://github.com/covid-projections/covid-projections/pull/5475)) are both merged into `can-82`

mikelehen: I think maybe it's for sharing (when you share an explore chart we store the chartId so that when you follow the shared link you get the same chart again)

That said, sharing seems to be working even though these IDs don't match the `getMetricByChartId()` IDs used [above](https://github.com/covid-projections/covid-projections/pull/5476/files#diff-98bf8ba7ddd452a4cfd4b539f787207ddee93af012d0964ef5de6c783b60a09aL134). 

i.e. I'd think this should be `weeklyCovidAdmissionsPer100k` not `admissions_per_100k`, but somehow it still seems to be working.  I'm a bit confused, but I think I'd go ahead and make them match regardless?

mikelehen (approved):
Looks good except I think there might be a mismatch with the chartIds (though I'm also a bit confused about them, so I could be missing something).

src/components/Explore/utils.ts
chasulin: yeah im not totally sure // also confused. will investigate 🕵🏻‍♀️

PR #5478, opened 2022-04-06
Big fix: revert explore chart ids for new metrics

chasulin:
Reverts chart ids for new metrics to all be `positivity_rate`. [PR with explore updates](https://github.com/covid-projections/covid-projections/pull/5476) seems to have failed netlify tests once it was merged into `can-82` -- I don't fully understand (yet) what the `chartId` field corresponds to, so maybe changing them broke something? 🤷🏻‍♀️


mikelehen:
Per standup discussion, Pablo was going to debug the netlify failures.

PR #5483, opened 2022-04-06
upload 8.2 email alert thermometer images to public

pnavarrc:
Uploading the email alert images to `public`. I included a note asking our future selves not to remove or rename the images to avoid breaking the images on the emails.


mikelehen (approved):
Looks good.  Just FYI- I think once we finish the CAN 8.2 work, we'll be able to delete the CRITICAL and SUPER_CRITICAL levels from the Level enum...  But if we do that we should make sure the UNKNOWN Level stays assigned the value `4` so that it matches up with these file names.

mikelehen (approved):
Perfect, thank you!

PR #5484, opened 2022-04-06
Switch hospitalization data to be HSA-based.

smcclure17:
This PR addresses: https://trello.com/c/nGN4H458/546-switch-hospitalization-data-to-be-hsa-based.

I had to update the location summaries (via the `update-snapshot.yml` github action) in order to make the chart tab values align with the chart timeseries data. I will rollback the data updates before merging.


src/common/models/Projection.ts
```diff
@@ -219,16 +223,28 @@ export class Projection {
       this.rawDailyDeaths,
     );
 
-    this.rawHospitalizations = actualTimeseries.map(
-      row => row && row.hospitalBeds.currentUsageCovid,
+    // Disaggregate county hospitalization data from HSAs to counties in order to display
+    // county estimates in charts.
+    this.rawHospitalizations = actualTimeseries.map(row =>
```
smcclure17, line 228: This seems like a lot of ternary operators--I'm willing to split these calculations up if that's preferred

src/common/models/Projection.ts
```diff
@@ -405,37 +425,56 @@ export class Projection {
...
-      // Make sure we don't somehow grab the wrong data, given we're pulling it from the metrics / actuals timeseries.
-      assert(
-        metrics.icuCapacityRatio === null ||
-          metrics.icuCapacityRatio ===
-            metricsTimeseries[icuIndex]?.icuCapacityRatio,
-        "Timeseries icuCapacityRatio doesn't match current metric value.",
-      );
+    const icuIndex = indexOfLastValue(metricSeries);
+
+    if (icuIndex != null) {
```
smcclure17, line 446: I broke up this logic because for counties we want to check if the most recent ICU value from the HSA data timeseries we calculated above is not `null`. In order to do that, we need to access `icuIndex` (if it exists). For non-county location types, we will use `metrics.icuCapacityRatio` as we did previously.

src/components/NewLocationPage/ChartFooter/utils.ts
```diff
@@ -20,7 +22,21 @@ export function getAddedMetricStatusText(
   metric: ExploreMetric,
   value: string,
   region: Region,
+  projections: Projections,
 ) {
+  // For county-level hospital data we display copy for the default disaggregated county data as well as
+  // HSA level data. We need to plug in custom copy and re-aggregate the HSA data from the estimated county data.
+  if (
```
smcclure17, line 29: This feels a little hacky, but I'm not immediately sure of a better way to handle this

src/common/models/Projection.ts
```diff
@@ -454,6 +493,16 @@ export class Projection {
     return null;
   }
 
+  private divideICUDataWithNulls(data: HospitalResourceUtilization | null) {
```
mikelehen, line 496: Not wild about this method name.  I think I'd rename it `calcIcuCapacityUsage` or something, and maybe add a comment about the null-handling if you think it's useful.

src/common/utils/index.ts
```diff
@@ -54,10 +54,12 @@ export const formatDecimal = (num: number, places = 2): string => {
...
-  return num.toLocaleString(undefined, {
-    minimumFractionDigits: places,
-    maximumFractionDigits: places,
-  });
+  return num
+    .toLocaleString(undefined, {
+      minimumFractionDigits: places,
+      maximumFractionDigits: places,
+    })
+    .replace('-0', '0'); // floating point precision errors can cause expected-zero values to return "-0"
```
mikelehen, line 62: Hrm, you could try adding `signDisplay: 'negative'` to the options above instead. Not 100% sure if that'll work.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat
image

src/components/NewLocationPage/ChartFooter/utils.ts
```diff
@@ -31,3 +47,20 @@ export interface DialogProps {
...
+  let hsaValue;
+  if (metric === ExploreMetric.ICU_HOSPITALIZATIONS) {
+    hsaValue = projections.primary.currentHsaIcuInfo.currentUsageCovid;
+  } else if (metric === ExploreMetric.HOSPITALIZATIONS) {
+    hsaValue = projections.primary.currentHsaHospitalInfo.currentUsageCovid;
+  } else {
+    fail('Unexpected HSA level metric.');
+  }
+
+  return formatValue(Metric.ADMISSIONS_PER_100K, hsaValue, '-');
```
mikelehen, line 65: Why is this ADMISSIONS_PER_100K?

mikelehen (approved):
A few nits / suggestions and one question, but overall looks great!  Thanks for working this all out.

One thing I'm noticing is that the HSA names aren't super readable.
> Over the last week, the King (Seattle), WA - Snohomish, WA health service area has reported having 16.0 COVID patients hospitalized in ICUs, for an estimated 10.1 coming from King County, WA.

We might want to at minimum somehow delineate them in the sentence (put in quotes or italicize or something).  Or maybe we should actually omit the name and just say "The health service area containing King County, WA ..." or something.  I think we should keep it as-is for now, but maybe flag it to Brett/Mikayla/Anna to see if anybody has strong opinions.

src/common/utils/index.ts
smcclure17: I'm facing an error when trying to use `signDisplay`--it looks like relating to https://github.com/microsoft/TypeScript/issues/40038 --where `signDisplay` is missing from the NumberFormatOptions (in the file below). The thread links to a fix, so I'm not 100% sure why it's failing here :/

For now, I might add a `TODO` with a link to this issue, unless there's an easy way to resolve this issue. 


`/Applications/Visual Studio Code.app/Contents/Resources/app/extensions/node_modules/typescript/lib/lib.es5.d.ts`

mikelehen: Ah, interesting.  It's possible that `.toLocaleString()` doesn't take the same arguments as `Intl.NumberFormat` though I found a doc that suggested it does.  But you could try:

```
const formatter = Intl.NumberFormatter(undefined, {
      minimumFractionDigits: places,
      maximumFractionDigits: places,
      signDisplay: 'negative'
});

return formatter.format(num);
```

Not a big deal though, so if it's not trivial to figure out feel free to ignore.

src/common/models/Projection.ts
```diff
@@ -405,37 +425,56 @@ export class Projection {
...
-      );
+    const icuIndex = indexOfLastValue(metricSeries);
+
+    if (icuIndex != null) {
+      let metricValue = this.isCounty
+        ? countyHsaTimeseriesIcuCapacityRatio[icuIndex]
+        : metrics.icuCapacityRatio;
+      if (metricValue === null) {
+        return null;
+      }
```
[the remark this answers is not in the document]
smcclure17: True--looks like Typescript is fine without this, so I've removed it 🙏

src/common/models/Projection.ts
```diff
@@ -219,16 +223,28 @@ export class Projection {
...
     );
 
-    this.rawHospitalizations = actualTimeseries.map(
-      row => row && row.hospitalBeds.currentUsageCovid,
+    // Disaggregate county hospitalization data from HSAs to counties in order to display
+    // county estimates in charts.
+    this.rawHospitalizations = actualTimeseries.map(row =>
+      this.isCounty
+        ? this.disaggregateHsaValue(
+            row?.hsaHospitalBeds.currentUsageCovid ?? null,
```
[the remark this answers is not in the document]
smcclure17: hmm, yeah that makes sense. Updated these--thanks 🙏

src/common/utils/index.ts
mikelehen: Oh, Actually I just noticed the little flask icon next to the "negative" docs apparently means that the functionality is experimental.  Probably best to avoid relying on it.  Feel free to just keep your existing code.

smcclure17: Ah true, there's the `exceptZero` option which seems like it should also do what we want. 

Unfortunately, I get the same error when trying the `const formatter = Intl.NumberFormat(undefined, {...` approach. Interestingly though, clicking through on each key to see what interface they come from leads me to different source files. 

`signDisplay` brings me to: 
`/Applications/Visual Studio Code.app/Contents/Resources/app/extensions/node_modules/typescript/lib/lib.es2020.intl.d.ts`

while

`minimumFractionDigits` and `maximumFractionDigits` bring me to: 
`/Applications/Visual Studio Code.app/Contents/Resources/app/extensions/node_modules/typescript/lib/lib.es5.d.ts`. 

I wonder if maybe some packages are conflicting with each other? I'm not too sure what I'm looking at :/

src/components/NewLocationPage/ChartFooter/utils.ts
smcclure17: Ah, Since hospitalizations and ICU hospitalizations don't exist in the `Metric` enum, I chose a similar metric that did exist. (This function formats the numbers based on the mapping below). https://github.com/covid-projections/covid-projections/blob/e16886aa1624de00b3b0d292c99cfbe69e5cf103/src/common/metric.tsx#L112

This is kind of hacky as it's not 1-1 with the actual metrics--should I look to maybe create matching metrics, or extend formatValue to accept `ExploreMetrics` as well as `Metrics`?

src/common/utils/index.ts
mikelehen: Yeah, I was thinking `exceptZero` would work at first, but I think it might add a `+` to positive numbers, which we don't want. 😁 

I think those files es2020 vs es5 files probably mean that `signDisplay` was added in [ECMAScript 2020](https://en.wikipedia.org/wiki/ECMAScript#11th_Edition_%E2%80%93_ECMAScript_2020) (the JavaScript standard released in 2020) vs. the others which have been around since [ECMAScript 5](https://en.wikipedia.org/wiki/ECMAScript#5th_Edition) (the JavaScript standard released in 2009).

I think maybe we are only [letting TypeScript build with ES2019 libraries](https://github.com/covid-projections/covid-projections/blob/00994c38a508fe8b496608787de11e17a57bd883/tsconfig.json#L5).  If we changed that to 2020, then maybe the error would go away?  Not sure.  I'm also not sure if we intentionally chose 2019 intentionally.  I'd want to check with Pablo before changing it.  But feel free to leave it as-is.

src/components/NewLocationPage/ChartFooter/utils.ts
mikelehen: Looking at the code, I think [this](https://github.com/covid-projections/covid-projections/blob/00994c38a508fe8b496608787de11e17a57bd883/src/components/Charts/Groupings.tsx#L225) is the code that ends up formatting the normal value for the footer.  So I think I'd just do `formatDecimal(hsaValue, 1)` here.

src/common/utils/index.ts
smcclure17: Oh, yeah we wouldn't want that 😅 

I see, that makes sense. Thanks so much for this explanation 🙌 I'll leave as-is with a TODO for now

src/components/NewLocationPage/ChartFooter/utils.ts
smcclure17: Oh perfect, that's much more simple--thanks

smcclure17:
Big thanks as always for the review and all of the explanations 🙏 

I agree about the HSA names--is there a place I should raise this concern/topic? I'm thinking I could start a thread in `#team-can-82`

mikelehen:
Yeah, a thread in #team-can-82 sounds great.

PR #5519, opened 2022-04-11
Wire up alert emails option so we can send them to everybody.

mikelehen:
Tested locally by running `yarn generate-daily-alerts 3207 true` and inspecting the output at scripts/alert_emails/alerts.jso to see that all locations are included and have an oldLevel of 4 (UNKNOWN).

We should further validate before actually sending email alerts by:
1. Doing a dev or staging run with the new option set to true.
2. Doing a prod dry-run with the new option set to true (this will tell us how many emails will be sent, etc.).

### DIFF
--- a/.github/workflows/send-alert-emails.yml
+++ b/.github/workflows/send-alert-emails.yml
@@ -15,13 +15,20 @@ on:
         description: 'Snapshot to send alerts for (auto uses currently deployed snapshot).'
         required: true
         default: auto
+      send_to_everybody:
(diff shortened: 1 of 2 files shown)


mikelehen:
[Update email alerts logic to reflect Community Levels](https://trello.com/c/ZytFtUoO/403-update-email-alerts-logic-to-reflect-community-levels)

pnavarrc (approved):
Looks good!

PR #5529, opened 2022-04-11
Regenerate location summaries

chasulin:
Trello: https://trello.com/c/EMvZpXG7/564-patients-w-covid-of-beds-metric-needs-to-convert-ratiopercent

Regenerates location summaries now that `Metric.RATIO_BEDS_WITH_COVID` is included in `MetricsFormattedAsPercentage`


mikelehen (approved):
Can you revert scripts/alert_emails/summaries/3184.json ?  That file is generated for the purposes of sending alert emails, which we're not doing from the can-82 branch and ultimately it will just result in a conflict with the same-named file in `develop` when we go to merge `can-82` into `develop`.  Thanks!

PR #5538, opened 2022-04-12
Speed up CI.

mikelehen:
* Remove cypress tests and entries in package.json / ci.yml workflow.
* Remove code coverage from ci.yml workflow and related package dependencies.
* Remove a few random package dependencies we don't actually need.
* (tangential to this PR) Removes some ancient docker cruft that we have never used but I noticed because it references cypress tests.


.github/workflows/ci.yml
```diff
@@ -25,31 +25,7 @@ jobs:
...
           restore-keys: ${{ runner.os }}-yarn-
 
-      # TODO: Discuss yarn vs. npm. I personally don't want to have to install yarn.
-      # It's not needed for anything and only adds an additional
-      # - name: Cache Node Modules
-      #   uses: actions/cache@v1
-      #   with:
-      #     path: ~/.npm # npm cache files are stored in `~/.npm` on Linux/macOS
-      #     key: ${{ runner.os }}-node-${{ hashFiles('**/yarn.lock') }}
-      #     restore-keys: ${{ runner.os }}-node-
```
mikelehen, line 35: I think we're all happy using `yarn`. 😁

.github/workflows/ci.yml
```diff
@@ -25,31 +25,7 @@ jobs:
...
-        uses: actions/cache@v1
-        with:
-          path: ~/.cache/Cypress # npm cache files are stored in `~/.npm` on Linux/macOS
-          key: ${{ runner.os }}-cypress-${{ hashFiles('**/yarn.lock') }}
-          restore-keys: ${{ runner.os }}-cypress-
-
       - run: yarn install
       - run: yarn run lint
       - run: yarn run tsc
-      - run: yarn run test.ci
```
mikelehen, line 47: This was the cypress tests...

.github/workflows/ci.yml
```diff
@@ -25,31 +25,7 @@ jobs:
...
       - run: yarn run jest.ci
-      - run: npx nyc report --reporter=text-summary
-
-      - name: Archive code coverage results
-        uses: actions/upload-artifact@v1
-        with:
-          name: code-coverage-report
-          path: coverage
```
mikelehen, line 55: This was code coverage that we never used...

package.json
```diff
@@ -38,14 +38,12 @@
     "@vx/voronoi": "^0.0.196",
     "amplitude-js": "^7.4.3",
     "aws-sdk": "^2.758.0",
-    "bluebird": "^3.7.2",
```
mikelehen, line 41: While I was here I did some random pruning of modules we don't actually use (bluebird, downshift, tar)

package.json
```diff
@@ -204,23 +194,19 @@
     "eslint-plugin-prettier": "3.1.2",
     "firebase-admin": "^9.5.0",
     "husky": "^4.2.5",
-    "istanbul-lib-coverage": "^3.0.0",
     "json-schema-to-typescript": "^8.2.0",
     "lint-staged": "^10.1.3",
     "node-fetch": "^2.6.1",
-    "nyc": "^15.0.0",
```
mikelehen, line 211: istanbul and nyc are code coverage things...

package.json
```diff
@@ -204,23 +194,19 @@
...
     "prettier": "2.0.2",
-    "puppeteer": "^3.0.4",
```
mikelehen, line 213: We use this in a firebase function that generates share images, but that has its own [package.json](https://github.com/covid-projections/covid-projections/blob/cb141f9895a6ad62e0d08bbfba4fc654fc178fce/firebase/functions/package.json#L27).

package.json
```diff
@@ -204,23 +194,19 @@
...
     "react-is": "^17.0.1",
     "remark": "^13.0.0",
     "remark-html": "^13.0.1",
     "serve": "^11.3.0",
-    "start-server-and-test": "^1.10.11",
```
mikelehen, line 218: This was only used by the "test" script that I removed above.

pnavarrc (approved):
Thanks for cleaning up! 🧹

PR #5541, opened 2022-04-12
Implement weekly deaths per 100k

faihegberg:
- Add back daily cases and deaths to the Trends chart.
- Implement weekly deaths per 100k as an explore metrics.

Still have some questions. Posting them in specific areas of the code.


src/common/models/Projection.ts
```diff
@@ -222,6 +224,9 @@ export class Projection {
     this.smoothedDailyDeaths = this.smoothWithRollingAverage(
       this.rawDailyDeaths,
     );
+    this.weeklyDeathsPer100k = this.smoothedDailyDeaths.map(
+      row => row && row * 7,
+    );
```
faihegberg, line 229: I need to dive into why this is returning whole numbers instead of float. What I tried:
`... * 7`: Returns whole number
`... * 7.0`: Returns whole number
`... * 7.5`: Returns float
Without multiplying it at all: Returns float

src/components/Explore/interfaces.ts
```diff
@@ -29,24 +29,19 @@ export interface Series {
...
+  WEEKLY_DEATHS_PER_100K = 3,
+  CASES = 4,
+  DEATHS = 5,
+  HOSPITALIZATIONS = 6,
+  ICU_HOSPITALIZATIONS = 7,
+  VACCINATIONS_FIRST_DOSE = 8,
+  VACCINATIONS_COMPLETED = 9,
+  VACCINATIONS_ADDITIONAL_DOSE = 10,
+  ICU_USED = 11,
+  POSITIVITY_RATE = 12,
```
faihegberg, line 44: As per the content team, we're keeping the ordering of existing metrics as is, but putting all new metrics at the top and arranged in alphabetical order.

src/components/Explore/utils.ts
```diff
@@ -430,6 +435,21 @@ export const exploreMetricData: {
...
   },
+  [ExploreMetric.WEEKLY_DEATHS_PER_100K]: {
+    title: 'Weekly reported deaths per 100k',
+    name: 'Weekly reported deaths per 100k',
+    chartId: 'weekly_reported_deaths_per_100k',
+    dataMeasure: DataMeasure.INTEGER,
+    yAxisDecimalPlaces: 0,
+    seriesList: [
+      {
+        label: 'Weekly reported deaths per 100k',
```
faihegberg, line 446: This label is showing up as the legend underneath the weekly death per 100k chart in the additional metrics chart group. This is because if it is an explore metric (see `Legend` component in `AddedMetricChartFooter.tsx`), where legends are shown by default. I think the legend in this case is a little bit redundant, but perhaps this adjustment can be made in a follow-up PR.

src/components/Explore/utils.ts
```diff
@@ -430,6 +435,21 @@ export const exploreMetricData: {
...
+        tooltipLabel: 'Weekly reported deaths per 100k',
+        datasetId: 'weeklyDeathsPer100k',
+        type: SeriesType.LINE,
```
faihegberg, line 449: I assume we want to display this metric as a line to match up with daily deaths, but let me know if otherwise.

src/common/models/Projection.ts
mikelehen: I think it is just the fact that deaths have to be a whole number.  You can't have 12.5 people dying in a week.

Average daily deaths is different, since it's an average # of deaths.

src/components/Dialogs/utils.tsx
```diff
@@ -66,10 +66,10 @@ interface ExploreMetricModalMapContent {
 export const exploreMetricToFooterContentMap: {
   [key: number]: ExploreMetricModalMapContent;
 } = {
-  [ExploreMetric.DEATHS]: {
-    metricName: 'daily deaths',
-    howItsCalculated: 'Our daily deaths number is a seven-day average.',
-    metricDefinition: 'This is the number of COVID deaths per day.',
+  [ExploreMetric.WEEKLY_DEATHS_PER_100K]: {
+    metricName: 'weekly deaths',
+    howItsCalculated: 'Our weekly deaths number is a seven-day average.',
```
mikelehen, line 71: I think this should be something like, "Our weekly deaths number is a rolling seven-day sum."

src/components/Explore/interfaces.ts
mikelehen: FWIW- It's not worth messing with in this PR, but it seems unnecessarily fragile that we have to keep these assigned IDs in sync with the EXPLORE_METRICS list.  Seems like that's probably avoidable...

src/components/Explore/utils.ts
```diff
@@ -430,6 +435,21 @@ export const exploreMetricData: {
       },
     ],
   },
+  [ExploreMetric.WEEKLY_DEATHS_PER_100K]: {
+    title: 'Weekly reported deaths per 100k',
+    name: 'Weekly reported deaths per 100k',
+    chartId: 'weekly_reported_deaths_per_100k',
```
mikelehen, line 441: I don't see "reported" in the metric name in the [copy doc](https://docs.google.com/document/d/1pITjM1gctulpwyqez5SFrQLFIYO4vndJArDjFPnVMG4/edit#). I think there is perhaps an ongoing discussion about whether we want "weekly cases" or "weekly reported cases" since many cases won't be reported.  But I don't think that's a significant concern for deaths.  Did you see something that said this should have "reported" in it?

src/components/Explore/utils.ts
mikelehen: Yeah, I'm not too worried about it.  Per the comment above I would remove "reported" here and below though.  And figure out the "per 100k" thing.

src/components/Charts/Groupings.tsx
```diff
@@ -177,7 +177,7 @@ export const CHART_GROUPS: ChartGroup[] = [
     groupHeader: GroupHeader.ADDITIONAL_MISC,
     metricList: [
       {
-        metric: ExploreMetric.DEATHS,
+        metric: ExploreMetric.WEEKLY_DEATHS_PER_100K,
```
mikelehen, line 180: So I don't think this is really "per 100k":
image

If California was really seeing 331 deaths per 100k, then over the course of a year (52 weeks), 17,212 people out of every 100,000 would be dead from covid, or 17% of the population of California.

So I think this should just be WEEKLY_DEATHS, not WEEKLY_DEATHS_PER_100K. 
 And we should remove the "per 100k" below.

src/components/Explore/utils.ts
```diff
@@ -430,6 +435,21 @@ export const exploreMetricData: {
       },
     ],
   },
+  [ExploreMetric.WEEKLY_DEATHS_PER_100K]: {
+    title: 'Weekly reported deaths per 100k',
+    name: 'Weekly reported deaths per 100k',
```
mikelehen, line 440: Same feedback here, these shouldn't be "per 100k" by default...

That said, ideally when you added 2+ locations to the chart, then it would switch to "per 100k" like we do for cases now:
image

image

I'm not sure how this logic works, but I think we would duplicate it for "weekly cases" and "weekly deaths".  When you add a second location, the data would be normalized by population and the name of the metric would change to include "per 100k".

mikelehen (changes requested):
This is looking pretty good, but I don't think this should be "per 100k" by default.  It should only be "per 100k" in the trends chart if you have 2+ locations selected (see my other comments for more details).

src/common/models/Projection.ts
faihegberg: Yea, but since this weekly number is inferred from an average number, shouldn't we keep the decimal?

mikelehen: I guess maybe I'm not sure what you mean.  Where are you seeing us discard a decimal?

To be clear, the average daily deaths is calculated by `(sum of 7 days of deaths)/7` and we are calculating the weekly number by multiplying that by 7 which just gives us `sum of 7 days of deaths` which will always be a whole number.

src/components/Explore/utils.ts
chasulin: @faihegberg this logic happens by adding the metric [to this array](https://github.com/covid-projections/covid-projections/blob/15ea16c9fa841cd3a8be1ea606a4c51e7ea43e7f/src/components/Explore/utils.ts#L108-L118).

you can add `ExploreMetric.WEEKLY_CASES_PER_100K` to this too (in addition to `ExploreMetric.WEEKLY_DEATHS_PER_100K`)

after adding both metrics, you can remove the "per 100k" from the `name` and `title` for both in [exploreMetricData](https://github.com/covid-projections/covid-projections/blob/15ea16c9fa841cd3a8be1ea606a4c51e7ea43e7f/src/components/Explore/utils.ts#L227-L229), as well as the `label` and `tooltipLabel` in each `seriesList` array

src/common/models/Projection.ts
mikelehen: Right.  That 9.9 is probably really 9.857142(857142...). i.e. 9 and 6/7.  When you multiply that by 7 you get 69.0.

src/components/Explore/interfaces.ts
chasulin: im 99.99% sure this is my doing/my fault 😭😭😭 . i wired up the enum values to the corresponding item's index in the dropdown menu. its def way too fragile. ill revisit it

src/common/models/Projection.ts
faihegberg: Ah I see, that makes sense.

src/components/Explore/interfaces.ts
mikelehen: All good. Not a high priority. 😁

src/components/Dialogs/utils.tsx
faihegberg: This would be different from the content in the [copy doc](https://docs.google.com/document/d/1pITjM1gctulpwyqez5SFrQLFIYO4vndJArDjFPnVMG4/edit#heading=h.99jyctq40phz), under "Methodology tooltips" on p. 30. But I agree that yours is more accurate. I'll post a comment in the copy doc.

src/components/Explore/utils.ts
faihegberg: Nope! This can be removed.

src/components/Charts/Groupings.tsx
faihegberg: Wow, that's a lot of deaths! 😬 😆 I'll revise this.

src/components/Dialogs/utils.tsx
mikelehen: Perfect, thanks!

faihegberg: Anna just responded. We'll go with your suggestion.

mikelehen (changes requested):
Looking good, but I think we over-renamed.  `Metric.WEEKLY_CASES_PER_100K` should stay the same.  We only want to rename the `ExploreMetric`s since trends can show the raw counts or the per 100k normalized counts.

mikelehen (approved):
Looks good! 🙏

faihegberg:
[Update Trends - Add weekly new deaths + minor updates](https://trello.com/c/toG5398D/554-update-trends-add-weekly-new-deaths-minor-updates)

PR #5548, opened 2022-04-12
Fix Weekly Covid Admissions chart copy value and update hospital chart copy for counties

smcclure17:
And drive-by formatting to capitalize Health Service Area to match [Copy Doc](https://docs.google.com/document/d/1pITjM1gctulpwyqez5SFrQLFIYO4vndJArDjFPnVMG4/edit#). 

Continuing to use existing HSA name copy/style until copy updates are confirmed (see [slack thread](https://actnowcoalition.slack.com/archives/C02UQBS8X8W/p1649724516819879) for context)

### DIFF
--- a/src/common/models/Projection.ts
+++ b/src/common/models/Projection.ts
@@ -158,6 +158,7 @@ export class Projection {
   readonly currentDailyDeaths: number | null;
   readonly currentHsaIcuInfo: HospitalResourceUtilization;
   readonly currentHsaHospitalInfo: HospitalResourceUtilizationWithAdmissions;
+  readonly currentWeeklyCovidAdmissions: number | null;
   readonly canCommunityLevel: Level;
 
   private readonly cumulativeActualDeaths: Array<number | null>;
@@ -310,6 +311,9 @@ export class Projection {
 
     this.currentHsaIcuInfo = summaryWithTimeseries.actuals.hsaIcuBeds;
     this.currentHsaHospitalInfo = summaryWithTimeseries.actuals.hsaHospitalBeds;
+    // For counties, in most cases we will want to access hsaHospitalBeds (usually via currentHsaHospitalInfo)
+    this.currentWeeklyCovidAdmissions =
+      summaryWithTimeseries.actuals.hospitalBeds.weeklyCovidAdmissions;
 
     this.annotations = summaryWithTimeseries.annotations;
   }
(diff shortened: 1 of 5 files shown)


src/common/models/Projection.ts
```diff
@@ -310,6 +311,9 @@ export class Projection {
 
     this.currentHsaIcuInfo = summaryWithTimeseries.actuals.hsaIcuBeds;
     this.currentHsaHospitalInfo = summaryWithTimeseries.actuals.hsaHospitalBeds;
+    // For counties, in most cases we will want to access hsaHospitalBeds (usually via currentHsaHospitalInfo)
```
mikelehen, line 314: Normally comments explain the code following it, but this is more of a sidenote warning people they may not want to use this (I think?).  Perhaps rephrase:

```
// NOTE: For counties you likely want to use hsaHospitalBeds (e.g. via currentHsaHospitalInfo) instead of this.
```

Though still, it seems a bit unlikely that somebody will see this comment at the right time to help them prevent a mistake.  You could move it to a jsdoc comment on the property declaration, e.g.:

```
/** Gets the latest weekly covid admissions data for this location. Note that for counties you likely want to use the HSA data (e.g. via currentHsaHospitalInfo) instead of this. */
readonly currentWeeklyCovidAdmissions: number | null;
```

src/components/NewLocationPage/ChartFooter/utils.ts
```diff
@@ -34,7 +34,7 @@ export function getAddedMetricStatusText(
     ].includes(metric)
   ) {
     const hsaFormattedValue = formattedHsaCovidUsage(metric, projections);
-    return `Over the last week, the ${projections.primary.hsaName} health service area has reported having ${hsaFormattedValue} ${exploreMetricToFooterContentMap[metric].statusTextMeasure}, for an estimated ${value} coming from ${region.shortName}.`;
+    return `Over the last week, the ${projections.primary.hsaName} Health Service Area has reported having ${hsaFormattedValue} ${exploreMetricToFooterContentMap[metric].statusTextMeasure}, for an estimated ${value} coming from ${region.shortName}.`;
```
mikelehen, line 37: FYI- I think Chelsi just tweaked this so you'll have to merge with her change. https://github.com/covid-projections/covid-projections/pull/5542

mikelehen (approved):
Looks good other than the admissions "per 100k" math thing I flagged on Slack.

src/common/models/Projection.ts
smcclure17: Ah, cool, thanks for the distinction. Updated to use a jsdoc comment

PR #5550, opened 2022-04-13
Weekly new cases chart status text

chasulin:
From [copy doc](https://docs.google.com/document/d/1pITjM1gctulpwyqez5SFrQLFIYO4vndJArDjFPnVMG4/edit?pli=1#):

Screen Shot 2022-04-12 at 8 39 35 PM

### DIFF
--- a/src/common/metrics/weekly_new_cases_per_100k.tsx
+++ b/src/common/metrics/weekly_new_cases_per_100k.tsx
@@ -10,8 +10,6 @@ import { metricToTooltipMap } from 'cms-content/tooltips';
 import { trackOpenTooltip } from 'components/InfoTooltip';
 import { Level, LevelInfoMap } from 'common/level';
 
-// TODO(8.2) : Update with real metric + content:
-
 export const WeeklyNewCasesPer100kMetric: MetricDefinition = {
   renderStatus,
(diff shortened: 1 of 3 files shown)


src/common/metrics/weekly_new_cases_per_100k.tsx
```diff
@@ -96,16 +96,16 @@ function renderStatus(projections: Projections): React.ReactElement {
     );
   }
 
-  const newCasesPerDay = currentDailyAverageCases;
+  const newCasesPerWeek = currentDailyAverageCases * 7;
   // Try not to round cases/day to zero (since it will probably be >0 per 100k).
```
mikelehen, line 100: Since this is now per week, I think we could safely remove this <1 logic and just always use `formatInteger()`

mikelehen (approved):
Looks good with minor suggestion.  Also, can we remove the `// TODO(8.2) : Update with real metric + content:` comment now or is there more to fix?

chasulin:
> Looks good with minor suggestion. Also, can we remove the `// TODO(8.2) : Update with real metric + content:` comment now or is there more to fix?

removing the todos! that should be everything

PR #5557, opened 2022-04-13
Reverse engineer weekly admissions actual for counties

smcclure17:
Reverse engineer actual weekly admissions for county-level locations (HSA-level data) to fix discrepancies between CDC metrics and HHS data.


mikelehen (approved):
Looks good!

PR #5568, opened 2022-04-14
Can 8.2 banner

chasulin:
Copy and URL subject to change

Trello: https://trello.com/c/zv4ok0cO/585-add-banner-to-homepage-and-location-pages

### DIFF
--- a/src/components/Banner/Can82Banner.tsx
+++ b/src/components/Banner/Can82Banner.tsx
@@ -1,0 +1,31 @@
+import React from 'react';
+import {
+  CondensedBanner,
+  CondensedBannerWrapperHomepage,
+  CondensedBannerWrapperLocationPage,
+} from './Banner.style';
+
+//TODO(8.2) - update with correct learn more URL
+
+const Can82Banner: React.FC = () => (
+  <CondensedBanner role="banner">
+    We've made some changes to better align with the CDC.{' '}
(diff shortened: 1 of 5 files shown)


src/components/Banner/Can82Banner.tsx
```diff
@@ -0,0 +1,31 @@
...
+  CondensedBannerWrapperHomepage,
+  CondensedBannerWrapperLocationPage,
+} from './Banner.style';
+
+//TODO(8.2) - update with correct learn more URL
+
+const Can82Banner: React.FC = () => (
+  <CondensedBanner role="banner">
+    We've made some changes to better align with the CDC.{' '}
+    <a href="/" target="_blank" rel="noopener noreferrer">
```
mikelehen, line 13: Since this will be an internal link (I assume), I don't think we want / need "noopener noreferrer".  It doesn't hurt much but might mess with our analytics a tiny bit if we lose referrer info for our own internal links. 🤷

mikelehen (approved):
Lovely!

src/components/Banner/Can82Banner.tsx
chasulin: oh true!

pnavarrc:
[Add banner to Homepage and Location pages](https://trello.com/c/zv4ok0cO/585-add-banner-to-homepage-and-location-pages)

PR #5571, opened 2022-04-14
Recommendations sharing

chasulin:
fixes recommendations sharing on homepage, cleans up for both homepage + location page

trello: https://trello.com/c/aHQARQRJ/566-bug-clicking-on-share-home-recommendations-shares-the-wrong-url

### DIFF
--- a/src/common/hooks/useScrollToRecommendations.ts
+++ b/src/common/hooks/useScrollToRecommendations.ts
@@ -8,11 +8,9 @@ const scrollTo = (div: null | HTMLDivElement, offset: number = 100) =>
     behavior: 'smooth',
   });
 
-// allowScroll is an optional additional check before scrolling to recommendations. On location pages, we first check to see if there is another param in the url that controls scrolling (ie. a chart share url)
 export default function useScrollToRecommendations(
   pathname: string,
   ref: React.RefObject<HTMLDivElement> | null,
-  allowScroll: boolean = true,
 ) {
   const isRecommendationsShareUrl = pathname.includes('recommendations');
 
@@ -20,7 +18,7 @@ export default function useScrollToRecommendations(
     const scrollToRecommendations = () => {
       const timeoutId = setTimeout(() => {
         if (isRecommendationsShareUrl) {
-          if (ref && ref.current && allowScroll) {
+          if (ref && ref.current) {
             scrollTo(ref.current);
           }
         }
@@ -29,5 +27,5 @@ export default function useScrollToRecommendations(
     };
 
     scrollToRecommendations();
-  }, [isRecommendationsShareUrl, allowScroll, ref]);
+  }, [isRecommendationsShareUrl, ref]);
 }
--- a/src/common/hooks/useScrollToRecommendations.ts
+++ b/src/common/hooks/useScrollToRecommendations.ts
@@ -1,0 +1,33 @@
+import React, { useEffect } from 'react';
+
+const scrollTo = (div: null | HTMLDivElement, offset: number = 100) =>
+  div &&
+  window.scrollTo({
+    left: 0,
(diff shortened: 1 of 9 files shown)


src/common/regions/types.ts
```diff
@@ -166,8 +166,9 @@ export class USA extends Region {
   abbreviation = 'USA';
 
   get relativeUrl(): string {
-    throw new Error('Method not implemented.');
+    return '';
```
chasulin, line 169: this^ allows us to do [this](https://github.com/covid-projections/covid-projections/pull/5571/files#diff-58ca9c7001c0a6e58af46f7c23cbc1f92052e4a6c62a9eace7b052cadfaec7bbR51), and doesnt seem to break anything in the process?

src/common/hooks/useScrollToRecommendations.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const scrollTo = (div: null | HTMLDivElement, offset: number = 100) =>
+  div &&
+  window.scrollTo({
+    left: 0,
+    top: div.offsetTop - offset,
+    behavior: 'smooth',
+  });
+
+// allowScroll is an optional additional check before scrolling to recommendations. On location pages, we first check to see if there is another param in the url that controls scrolling (ie. a chart share url)
```
mikelehen, line 11: I'm not really following this.  Would this kick in for a URL like https://covidactnow.org/us/arizona-az/chart/9/recommendations or something?  I feel like I'm missing something.

Okay, looking at the old code a bit more, I think the `scrolledWithUrl` flag was being used to keep us from accidentally re-scrolling to a given section after the page is already loaded (in the case that the component re-renders for some reason and so this hook gets re-run).  I think your PR might be breaking that functionality since we used to do:

```
            setScrolledWithUrl(true);
            scrollTo(recommendationsRef.current);
```

and now we don't call `setScrolledWithUrl(true);` when scrolling to recommendations.  That said, I wasn't able to actually reproduce any bad scrolling behavior, so maybe this isn't an issue anymore.  I guess I'd be a favor of removing `allowScroll` unless we know why it's needed.

src/common/regions/types.ts
mikelehen: Seems reasonable!

mikelehen (changes requested):
Perhaps I'm confused, but it doesn't seem like homepage scrolling is working.  https://covid-projections-git-recommendations-share-fix-covidactnow.vercel.app/recommendations doesn't work for me.

src/common/hooks/useScrollToRecommendations.ts
chasulin: hmm i went back to the pr where i originally added `scrolledWithUrl` -- i should've added a note as to what it was for...

i'm going to just remove `allowScroll` for now and will revisit a bug pops up. does that sound alright?

chasulin:
> Perhaps I'm confused, but it doesn't seem like homepage scrolling is working. https://covid-projections-git-recommendations-share-fix-covidactnow.vercel.app/recommendations doesn't work for me.

oh man. this admittedly was some very lazy QA on my part--i got location page scrolling working and i think i just assumed that meant homepage was good to go as well. but...turns out i didnt actually add the `/recommendations` route for the homepage in App.js. needless to say, i added the route and it looks like its working now. oof. 🤦🏻‍♀️🤦🏻‍♀️🤦🏻‍♀️

src/common/hooks/useScrollToRecommendations.ts
mikelehen: Yep!

mikelehen (approved):
Looks good and working for me now! 😁   Thanks for the follow-up.  Have a great weekend!

PR #5575, opened 2022-04-15
Small copy edits

chasulin:
- Edits casing of chart block headers
- Updates banner copy per what's in the trello ticket (URL still needs to be finalized)


src/components/Banner/Banner.style.ts
```diff
@@ -155,6 +155,7 @@ export const CondensedBanner = styled(Box)`
   color: white;
   margin: auto;
   text-align: center;
+  max-width: 800px;
```
chasulin, line 158: max-width because banner copy has gotten long

mikelehen (approved):

PR #5605, opened 2022-04-18
Adjust community level map header

faihegberg:
Tiny PR adjusting header above the community level map, as per request in the [launch checklist](https://www.dropbox.com/scl/fi/u23ob7sqtyqyo3e9n37bx/CAN-8.2-CDC%2B%2B-Launch-Checklist.paper?dl=0&rlkey=ebmgo5rcj5i3mesd8bkeglac3).

image

### DIFF
--- a/src/screens/HomePage/HomePage.tsx
+++ b/src/screens/HomePage/HomePage.tsx
@@ -145,7 +145,7 @@ export default function HomePage() {
             </ColumnCentered>
 
             <MapBlock
-              title="Community levels"
+              title="COVID Community Level"
               subtitle=""
               renderMap={locationScope => (
                 <USRiskMap showCounties={locationScope === MapView.COUNTIES} />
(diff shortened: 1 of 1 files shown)


src/screens/HomePage/HomePage.tsx
```diff
@@ -145,7 +145,7 @@ export default function HomePage() {
             </ColumnCentered>
 
             <MapBlock
-              title="Community levels"
+              title="COVID Community Level"
```
mikelehen, line 148: It looks like community and level should be lowercase?

mikelehen (approved):
Approved, but please verify the casing is correct.

src/screens/HomePage/HomePage.tsx
faihegberg: Yup, should be all lowercase actually. Thanks Michael!

PR #5609, opened 2022-04-18
Remove extra fields from summaries file.

mikelehen:
Should fix crash Sean noticed (see https://github.com/covid-projections/covid-projections/runs/6068853288?check_suite_focus=true).

It also just removes extra stuff we don't need from the summaries file which will speed up page loads.

Run with this branch: https://github.com/covid-projections/covid-projections/runs/6068807801?check_suite_focus=true

### DIFF
--- a/src/common/models/Projections.ts
+++ b/src/common/models/Projections.ts
@@ -63,7 +63,13 @@ export class Projections {
    */
   summary(ccvi: number): LocationSummary {
     const metrics = {} as { [metric in Metric]: MetricSummary };
-    for (const metric of ALL_METRICS) {
+    const summaryMetrics = [
+      Metric.ADMISSIONS_PER_100K,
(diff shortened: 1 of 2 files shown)


src/common/models/Projections.ts
```diff
@@ -75,14 +81,10 @@ export class Projections {
 
     const vaccinationsCompleted =
       this.primary.vaccinationsInfo?.ratioVaccinated ?? null;
-    const hospitalizationsDensity =
-      this.primary.hospitalizationInfo?.hospitalizationsDensity ?? null;
     return {
       level: this.getAlarmLevel(),
       metrics,
-      ccvi,
       vc: vaccinationsCompleted,
```
mikelehen, line 87: We need vaccinations completed since we use it in the map tooltip.

smcclure17 (approved):
thanks for jumping on this!

PR #5614, opened 2022-04-19
develop → main (can 8.2)

pnavarrc:


Stevoisiak:
As a regular user of Covid Act Now, I'm disappointed by this change. The old infection rate metric served as a pre-emptive warning that helped me prepare for upcoming waves, including the one going on right now. Seeing almost the entire map switch to "low risk" overnight was heartbreaking. I sincerely hope you consider reverting this change.

PR #5632, opened 2022-04-19
Fix dropdown arrow bug

faihegberg:
I think `<ShareModelBlock>` was restored and misplaced, where the CSS of other containing components messed with the arrow appearance and caused the whole get alerts block to not be centered.

Before:
image

After:
image


faihegberg:
[Email Alerts Styling Bug - Donate](https://trello.com/c/mMqLOfsp/565-email-alerts-styling-bug-donate)

pnavarrc (changes requested):
The email alerts block is missing the margins on mobile 

image

pnavarrc (approved):
Looks good!

PR #5644, opened 2022-04-20
Remove cross-dependency on EXPLORE_METRICS and ExploreMetric numbering.

mikelehen:


src/components/Explore/Explore.tsx
```diff
@@ -134,9 +136,10 @@ const Explore: React.FunctionComponent<{
       setCurrentMetric(sharedMetric);
     }
 
-    const onSelectCurrentMetric = (newMetric: number) => {
-      const newMetricName = metricLabels[newMetric];
-      setCurrentMetric(newMetric);
+    const onSelectCurrentMetric = (newMetricIndex: number) => {
+      const newMetricName = metricLabels[newMetricIndex];
+      const metric = EXPLORE_METRICS[newMetricIndex];
```
mikelehen, line 141: Rather than assuming the menu index will match the metric enum value, we look up the metric enum value.

src/components/Explore/Explore.tsx
```diff
@@ -287,7 +290,9 @@ const Explore: React.FunctionComponent<{
       ORIGINAL_EXPLORE_METRICS.includes(currentMetric) && numLocations === 1;
 
     // menu labels for metric, time period, and selected locations:
-    const metricMenuLabel = metricLabels[currentMetric];
+    const selectedMetricIndex = EXPLORE_METRICS.indexOf(currentMetric);
+    assert(selectedMetricIndex >= 0);
+    const metricMenuLabel = metricLabels[selectedMetricIndex];
```
mikelehen, line 295: To find the right menu index to select, we have to look up where `currentMetric` is in the list.

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -93,7 +93,7 @@ const ChartsHolder = React.memo(({ region, chartId }: ChartsHolderProps) => {
 
   const [currentExploreMetric, setCurrentExploreMetric] = useState<
     ExploreMetric
-  >(ExploreMetric.WEEKLY_CASES);
+  >(ExploreMetric.HOSPITALIZATIONS);
```
mikelehen, line 96: I'm not sure why this was set to WEEKLY_CASES before.  It ends up getting overridden (in most cases at least) by this code that sets it to HOSPITALIZATIONS: https://github.com/covid-projections/covid-projections/blob/3ee5232c0f491e18cf86083f92485521a329a150/src/components/Explore/Explore.tsx#L256

So this results in no behavior change.

src/components/Explore/interfaces.ts
```diff
@@ -21,28 +21,21 @@ export interface Series {
...
+  CASES,
+  DEATHS,
+  HOSPITALIZATIONS,
+  ICU_HOSPITALIZATIONS,
+  VACCINATIONS_FIRST_DOSE,
+  VACCINATIONS_COMPLETED,
+  VACCINATIONS_ADDITIONAL_DOSE,
+  ICU_USED,
+  POSITIVITY_RATE,
+  DAILY_CASES_PER_100K,
```
mikelehen, line 38: FWIW- To verify my change works I tested by randomly re-ordering these and verified that it had no effect on the trends menu ordering or the functionality of the metrics.

src/components/Explore/Explore.tsx
chasulin: ahhh yes

chasulin (approved):
nice!! ⚡️

PR #5645, opened 2022-04-20
Update geolocated cards

faihegberg:
Update geolocated cards on the homepage to display community level instead of vaccination metric, as per [card](https://trello.com/c/OEUhP0m1/598-geolocation-cards-should-probably-show-community-level-color-not-vaccine).

image

### DIFF
--- a/src/components/RegionItem/RegionItem.tsx
+++ b/src/components/RegionItem/RegionItem.tsx
@@ -13,21 +13,29 @@ import { Region, State } from 'common/regions';
 import { StyledRegionName } from 'components/SharedComponents';
 import { getSummaryFromFips } from 'common/location_summaries';
 import { EventAction, EventCategory, trackEvent } from 'components/Analytics';
-import { vaccineColorFromLocationSummary } from 'common/colors';
-import { summaryToStats } from 'components/NewLocationPage/SummaryStat/utils';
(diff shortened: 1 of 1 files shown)


faihegberg:
[geolocation cards should probably show community level color not vaccine.](https://trello.com/c/OEUhP0m1/598-geolocation-cards-should-probably-show-community-level-color-not-vaccine)

src/components/RegionItem/RegionItem.tsx
```diff
@@ -13,21 +13,29 @@ import { Region, State } from 'common/regions';
...
+  switch (level) {
+    case Level.LOW:
+      return 'Low';
+    case Level.MEDIUM:
+      return 'Medium';
+    case Level.HIGH:
+      return 'High';
+    default:
+      return 'Unknown';
+  }
```
mikelehen, line 29: I think instead of duplicating this level => name mapping, we can use the `LOCATION_SUMMARY_LEVELS` const similar to how we do for above-the-fold location summary.
https://github.com/covid-projections/covid-projections/blob/develop/src/components/NewLocationPage/OverallRiskBlock/OverallLevelText/OverallLevelText.tsx#L31

mikelehen (changes requested):

mikelehen (approved):

PR #5646, opened 2022-04-20
Add color dot indicators to chart tabs

faihegberg:
Add color dot indicators to chart tabs, as per [card](https://trello.com/c/yD9uJhYw/606-add-color-dot-indicators-to-community-levels-section).

### DIFF
--- a/src/components/NewLocationPage/ChartTabs/ChartTab.tsx
+++ b/src/components/NewLocationPage/ChartTabs/ChartTab.tsx
@@ -14,7 +14,8 @@ const ChartTab: React.FC<{
   metricName: string;
   subLabel?: string[];
   metricValueInfo: ValueInfo;
-}> = ({ metricName, subLabel, metricValueInfo }) => {
+  showColorIndicator?: boolean;
+}> = ({ metricName, subLabel, metricValueInfo, showColorIndicator = true }) => {
   const isMobile = useBreakpoint(600);
 
   return (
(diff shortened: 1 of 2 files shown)


mikelehen (changes requested):
Can you verify we want these to show up for our old metrics too?  I'm guessing we don't.

image

faihegberg:
> Can you verify we want these to show up for our old metrics too? I'm guessing we don't.
> 
> image

I assume that we would? But I'll double check 👍🏼

faihegberg:
> > Can you verify we want these to show up for our old metrics too? I'm guessing we don't.
> > image
> 
> I assume that we would? But I'll double check 👍🏼

@mikelehen You're right. We're supposed to _only_ do it for community level metrics 😄  I thought the card was just missing the transmission ones and wrongly assumed it should apply to all.

src/components/NewLocationPage/ChartTabs/ChartTab.tsx
```diff
@@ -14,7 +14,8 @@ const ChartTab: React.FC<{
   metricName: string;
   subLabel?: string[];
   metricValueInfo: ValueInfo;
-}> = ({ metricName, subLabel, metricValueInfo }) => {
+  showColorIndicator?: boolean;
+}> = ({ metricName, subLabel, metricValueInfo, showColorIndicator = true }) => {
```
faihegberg, line 18: Added `showColorIndicator` as an optional prop to the `ChartTab` component here instead of implementing conditional logic in `getValueInfo()` in `Groupings.tsx`, since this will give us more flexibility if we want to display (or not display) a color indicator in any chart tab in the future.

mikelehen (approved):
Looks good. The `showColorIndicator` flag is a nice approach.

faihegberg:
[Add color dot indicators to Community Levels section](https://trello.com/c/yD9uJhYw/606-add-color-dot-indicators-to-community-levels-section)

PR #5654, opened 2022-04-21
Optimize the y-axis limit of Patients w/ Covid chart

smcclure17:
Addresses https://trello.com/c/qJQcto7M/612-optimize-the-y-axis-limit-to-show-only-part-of-the-high-band-in-community-level-components

Limits the height of the Patients w/ Covid chart to the maximum value plus 5% extra padding. The ticket is a little vague on whether the limit should be dynamic or fixed (e.g. something like the old percent positivity chart that capped at 40%) so I can rework this if need be.

### DIFF
--- a/src/components/Charts/ChartRatioBedsWithCovidPatients.tsx
+++ b/src/components/Charts/ChartRatioBedsWithCovidPatients.tsx
@@ -4,8 +4,8 @@ import { formatPercent } from 'common/utils';
 import { Column } from 'common/models/Projection';
 import { RATIO_BEDS_WITH_COVID_PATIENTS_LEVEL_INFO_MAP } from 'common/metrics/ratio_beds_with_covid_patients';
 
-// TODO(8.2) - confirm thresholds/chart/tooltip content, y-cap
-const CAP_Y = 1;
+// Add CAP_PADDING% extra padding above the maximum chart value
+const CAP_PADDING = 0.05;
 
 const getPointText = (valueY: number) => formatPercent(valueY, 1);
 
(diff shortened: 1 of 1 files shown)


src/components/Charts/ChartRatioBedsWithCovidPatients.tsx
```diff
@@ -15,6 +15,15 @@ const getTooltipContent = (valueY: number) => ({
...
 });
 
+const maxColumnValue = (data: Column[]) => {
+  return Math.max.apply(
+    null,
+    data.map(function (column) {
+      return column.y;
+    }),
+  );
+};
```
pnavarrc, line 25: I think we want to ensure that the High section shows up even if the max column value is below the Medium-High threshold (15%), I would recommend to set the cap at a minimum of 20% to make sure that the High segment is visible.

Also, consider using `max` from `lodash` to simplify a bit:

```tsx
import max from 'lodash/max';
// ...

// Maybe a better variable name here?
const MIN_Y = 0.2; // 20%

const maxColumnValue = max(data.map(col => col.y));
const capY = Math.max(maxColumnValue + CAP_PADDING, MIN_Y);  
```

pnavarrc (changes requested):
I think we want to show the High region of the chart even if the values are all on Medium or below, added some suggestions!

src/components/Charts/ChartRatioBedsWithCovidPatients.tsx
smcclure17: Ah, yeah good call 🙏 added in 48c4fc3b3467dee2010f17664bf809f98c912129

Example of a county whose max is < 20%: 
image

pnavarrc (approved):
Looks good! 🙌

PR #5682, opened 2022-04-25
Add time period to shared component params

faihegberg:
Currently in the Trends section, when the user selects a non-default time period and click share, the share link will save the selected metric and location/s, but not time period. This PR adds time period to shared parameters so the shared link will "remember" the selected time period too.

Addresses this [card](https://trello.com/c/DRB2OtuR/609-trends-sharing-doesnt-respect-past-of-days-setting).


faihegberg:
[Trends sharing doesn't respect "Past # of days" setting.](https://trello.com/c/DRB2OtuR/609-trends-sharing-doesnt-respect-past-of-days-setting)

mikelehen (approved):
Looks good! 👍

PR #5685, opened 2022-04-26
Add textbox below homepage map

faihegberg:
Enable each map to have optional text underneath it.

Addresses [this card](https://trello.com/c/0Pzrq2j3/622-add-copy-box-below-community-level-map).

### DIFF
--- a/src/screens/HomePage/HomePage.style.tsx
+++ b/src/screens/HomePage/HomePage.style.tsx
@@ -161,7 +161,7 @@ export const UnderMapText = styled(Typography)`
   max-width: 300px;
 
   @media (min-width: ${materialSMBreakpoint}) {
-    max-width: initial;
+    max-width: unset;
     margin-top: 1.25rem;
   }
 `;


faihegberg:
[Add copy box below Community Level map](https://trello.com/c/0Pzrq2j3/622-add-copy-box-below-community-level-map)

chasulin (commented):
lets style the link the way other markdown links are styled. scroll down to recommendations to see examples. you can just grab the anchor style block from the markdown styles file. 

maybe lets also make the non-link text the same color as other body text in recommendations? and maybe a touch smaller? what do you think?

faihegberg:
> lets style the link the way other markdown links are styled. scroll down to recommendations to see examples. you can just grab the anchor style block from the markdown styles file.
> 
> maybe lets also make the non-link text the same color as other body text in recommendations? and maybe a touch smaller? what do you think?

That sounds good! I agree with matching the styles with the markdown styles. I'll play around with the font size. Perhaps I'll match that with the body text of recommendations too.

src/screens/HomePage/HomePage.style.tsx
```diff
@@ -150,6 +152,20 @@ export const Row = styled.div`
...
+export const UnderMapText = styled(Typography)`
+  ${anchorStyles};
+  color: ${COLOR_MAP.GRAY_BODY_COPY};
+  font-size: 0.9rem;
+  text-align: center;
+  margin-top: 1rem;
+  max-width: 300px;
+
+  @media (min-width: ${materialSMBreakpoint}) {
+    max-width: initial;
```
chasulin, line 164: lets use `unset`

chasulin (approved):

src/screens/HomePage/HomePage.style.tsx
faihegberg: Yea, I understand it as the "initial" css value as though it has not been explicitly set (in our case, max width of 300px). So by setting it to initial, it would inherit from the closest higher level component. I might be wrong though. We can be explicit and set it to "unset" to be sure 👍🏼

PR #5690, opened 2022-04-26
Summary stat realignment

chasulin:


faihegberg (approved):
There's a very brief point where "% vaccinated" gets cut and turns into two lines, but apart from that, looks good!
image

PR #5694, opened 2022-04-26
Redirect DC county location page to DC state page

smcclure17:


src/components/RegionMap/RegionMap.tsx
```diff
@@ -216,4 +216,17 @@ function getCountyFipsList(region: Region): string[] {
   }
 }
 
+function getCountyShapeLink(region: Region | null): string {
+  if (!region) {
+    return '/';
+  } else {
+    // DC has a state- and county-level page, we only want to use the state page.
+    if (region.fipsCode === '11001') {
+      const dcRelativeUrl = regions.findByFipsCodeStrict('11');
```
chasulin, line 225: oh whoops, this should be `dcStateRegion` (it doesnt yet grab the relative url)

src/components/RegionMap/RegionMap.tsx
```diff
@@ -216,4 +216,17 @@ function getCountyFipsList(region: Region): string[] {
...
+      return dcRelativeUrl.relativeUrl;
```
chasulin, line 226: and then `return dcStateRegion.relativeUrl`

src/components/RegionMap/RegionMap.tsx
smcclure17: ah! right--updated in db6aae94da1ccc2c7c29a7b21575d158f1ca50b5 🙏

chasulin (approved):

PR #5695, opened 2022-04-26
Change risk level route name to community level

faihegberg:
Change route name from risk level to community level, as per [card](https://trello.com/c/e7cUfzd1/611-change-explainer-link-to-covid-community-level-metrics).


src/App.tsx
```diff
@@ -184,8 +191,31 @@ export default function App() {
...
+                      ) {
+                        return (
+                          <Redirect
+                            to={`covid-community-level-metrics${props.location.hash}`}
+                          />
+                        );
+                      }
+                      return <Redirect to="/covid-community-level-metrics" />;
+                    }}
+                  />
```
faihegberg, line 216: Let me know if there's a better solution here. At first I tried:
```js
                  <Redirect
                    from="/covid-risk-levels-metrics#icu-capacity-used"
                    to="/covid-community-level-metrics#icu-capacity-used"
                  />
                  <Redirect
                    from="/covid-risk-levels-metrics#daily-new-cases"
                    to="/covid-community-level-metrics#daily-new-cases"
                  />
                  <Redirect
                    from="/covid-risk-levels-metrics#positive-test-rate"
                    to="/covid-community-level-metrics#positive-test-rate"
                  />
                  <Redirect
                    from="/covid-risk-levels-metrics#infection-rate"
                    to="/covid-community-level-metrics#infection-rate"
                  />
```
And the redirect seems to disregard the hash (i.e. it redirects to the correct page but doesn't scroll to the appropriate section). Tried playing around with props like [exact](https://v5.reactrouter.com/web/api/Redirect/exact-bool) and [strict](https://v5.reactrouter.com/web/api/Redirect/strict-bool) too. Found one possible solution in [this thread](https://stackoverflow.com/questions/45864594/react-router-redirect-hash-link).

src/App.tsx
```diff
@@ -57,6 +57,13 @@ const AfterUnsubscribe = lazy(() =>
   import('screens/Subscriptions/AfterUnsubscribe'),
 );
 
+const metricExplainerHashes = [
+  '#icu-capacity-used',
+  '#daily-new-cases',
```
faihegberg, line 62: This was taken from what we currently have in the code. I think `daily-new-cases` hash is dead though. We can probably get rid of this and have it redirect to the top of the page.

src/App.tsx
```diff
@@ -311,35 +341,35 @@ export default function App() {
...
                   <Redirect
                     from="/old-hospitalization-projections"
-                    to="/covid-risk-levels-metrics#icu-capacity-used"
+                    to="/covid-community-level-metrics#icu-capacity-used"
                   />
                   <Redirect
                     from="/daily-new-cases-explained"
-                    to="/covid-risk-levels-metrics#daily-new-cases"
+                    to="/covid-community-level-metrics#daily-new-cases"
                   />
```
faihegberg, line 361: Same comment as above. I think `#daily-new-cases` hash is dead.

src/App.tsx
mikelehen: Hrm.  Is there a reason we have to actually validate that the hash is in `metricExplainerHashes`?  Why not just always preserve the hash with 

```
                          <Redirect
                            to={`covid-community-level-metrics${props.location.hash}`}
                          />
```

or whatever?

src/App.tsx
mikelehen: I think it was replaced by `weekly-new-cases` (if you click the "About this data" link under the weekly new cases chart and Learn more, that's where you go), so we should maybe use that instead.

src/App.tsx
mikelehen: TBH we could probably remove all 8 of these redirects.  Seems unlikely anybody is linking to them anymore...  Else if we keep them, I'd update "daily-new-cases" to "weekly-new-cases."

mikelehen (approved):
LGTM with a couple minor comments.

src/App.tsx
faihegberg: We could definitely do that. I just put this check there to be explicit about which hashes _should_ be valid ones. That said, the behaviour should be the same either way. I'll keep it simple and stick with your solution.

src/App.tsx
faihegberg: Let's get rid of them 👍🏼

faihegberg:
[Change explainer link to /covid-community-level-metrics](https://trello.com/c/e7cUfzd1/611-change-explainer-link-to-covid-community-level-metrics)

PR #5696, opened 2022-04-26
Trends default time period update

chasulin:
Trello: https://trello.com/c/7COWZXOf/613-change-the-trend-defaults-on-mobile-to-180-days

Defaults trends view to 180-days on mobile

### DIFF
--- a/src/common/hooks/useDefaultTrendsPeriod.ts
+++ b/src/common/hooks/useDefaultTrendsPeriod.ts
@@ -1,0 +1,23 @@
+/**
+ * On mobile screens, we default the Trends view to show past-180-days
+ */
+
+import { useEffect, useState } from 'react';
+import useBreakpoint from './useBreakpoint';
+import { Period } from 'components/Explore/utils';
+
+export default function useDefaultTrendsPeriod(): Period {
+  const isMobile = useBreakpoint(800);
+  const [defaultTimePeriod, setDefaultTimePeriod] = useState<Period>(
+    Period.ALL,
+  );
+
+  useEffect(() => {
+    if (isMobile) {
+      setDefaultTimePeriod(Period.DAYS_180);
+    } else {
+      setDefaultTimePeriod(Period.ALL);
+    }
+  }, [isMobile]);
+  return defaultTimePeriod;
+}
--- a/src/components/Explore/Explore.tsx
+++ b/src/components/Explore/Explore.tsx
@@ -257,20 +257,16 @@ const Explore: React.FunctionComponent<{
 
     const defaultTimePeriod = useDefaultTrendsPeriod();
 
-    // if the pathname changes (ie. if navigating between location pages via compare or regionmap)-
-    // resets metric, time period, and locations
-    // (need to force the reset since the route doesnt change)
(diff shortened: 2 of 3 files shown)


src/components/Explore/Explore.tsx
```diff
@@ -252,13 +255,17 @@ const Explore: React.FunctionComponent<{
...
+    // Resets time period state variable when pathname changes (need to force the reset since the route doesn't change):
+    useEffect(() => {
+      setPeriod(defaultTimePeriod);
+    }, [pathname, defaultTimePeriod]);
+
+    // Resets locations+current metric state variables when pathname changes (need to force the reset since the route doesn't change):
     useEffect(() => {
       setSelectedLocations(initialLocations);
       setCurrentMetric(ExploreMetric.HOSPITALIZATIONS);
-      setPeriod(Period.ALL);
```
chasulin, line 261: need to separate this out into its own effect, so a change in screen size doesn't reset the selected metric+locations

src/common/hooks/useDefaultTrendsPeriod.ts
```diff
@@ -0,0 +1,23 @@
...
+  );
+
+  useEffect(() => {
+    if (isMobile) {
+      setDefaultTimePeriod(Period.DAYS_180);
+    } else {
+      setDefaultTimePeriod(Period.ALL);
+    }
+  }, [isMobile]);
+  return defaultTimePeriod;
```
mikelehen, line 22: I'm not sure you need the `defaultTimePeriod` state value and this `useEffect()` call.

I think you could just do:
```
export default function useDefaultTrendsPeriod(): Period {
  const isMobile = useBreakpoint(800);
  return isMobile ? Period.DAYS_180 : Period.ALL;
}
```

And FWIW at that point there's not that much value to having this `useDefaultTrendsPeriod()` hook that's only used in one place.  I'd probably just use the `useBreakpoint()` hook directly from the explore component. 🤷

mikelehen (approved):
This looks good, though I am a bit surprised we are changing the period when the screen size changes.  I would have probably only used the screen size to set the initial default, and not changed it afterwards.  E.g.:

```
  const [period, setPeriod] = useState<Period>(isMobile ? Period.DAYS_180 : Period.ALL);
```

But I'm not opposed to having it be more responsive like you did. 🤷

chasulin:
> This looks good, though I am a bit surprised we are changing the period when the screen size changes. I would have probably only used the screen size to set the initial default, and not changed it afterwards. E.g.:
> 
> ```
>   const [period, setPeriod] = useState<Period>(isMobile ? Period.DAYS_180 : Period.ALL);
> ```
> 
> But I'm not opposed to having it be more responsive like you did. 🤷

this^^ is what i originally tried but couldn't get to work. isMobile is false on first load, and without the useEffect it doesn't check for when isMobile turns true. i just put it in a hook/file so as to not crowd this file more. fwiw, i couldnt figure out how to get it to work as intended (only use screen size on initial default) 😕

chasulin:
> 

sweet, going to just merge as is 〰

PR #5705, opened 2022-04-28
Add internal county data anomaly tracker

smcclure17:
Creates an internal page for viewing county-level data anomalies.

Preview: https://covid-projections-n9nq7lcr5-covidactnow.vercel.app/internal/anomalies

This is helpful for tracking down anomalies or looking at what anomalies exist for a specific location, but it isn't so helpful for identifying what anomalies were introduced in a given snapshot. We might want to extend this to have an option to list all of the anomalies for a given metric in reverse chronological order, or something similar.

### DIFF
--- a/src/common/metric.tsx
+++ b/src/common/metric.tsx
@@ -15,14 +15,14 @@ import { formatDecimal, formatPercent } from 'common/utils';
 import { Metric } from 'common/metricEnum';
 
 export const ALL_METRICS = [
+  Metric.WEEKLY_CASES_PER_100K,
+  Metric.RATIO_BEDS_WITH_COVID,
+  Metric.ADMISSIONS_PER_100K,
   Metric.CASE_DENSITY,
   Metric.CASE_GROWTH_RATE,
   Metric.POSITIVE_TESTS,
   Metric.HOSPITAL_USAGE,
   Metric.VACCINATIONS,
-  Metric.RATIO_BEDS_WITH_COVID,
-  Metric.ADMISSIONS_PER_100K,
-  Metric.WEEKLY_CASES_PER_100K,
 ];
 
 const ALL_METRICS_LEVEL_INFO_MAP: { [metric in Metric]: LevelInfoMap } = {
--- a/src/screens/internal/Anomalies/AnomaliesSelector.tsx
+++ b/src/screens/internal/Anomalies/AnomaliesSelector.tsx
@@ -54,7 +54,7 @@ function AnnotationsSelectorInner({
   );
 
   const [stateFips, setLocations] = useState<string>(
-    getLocationsParamValue(params, '01'),
+    getStateFipsParamValue(params, '01'),
   );
 
   useEffect(() => {
@@ -124,7 +124,7 @@ function AnnotationsSelectorInner({
         />
       </FormControl>
       <FormControl style={{ width: '12rem', marginLeft: '1rem' }}>
-        <InputLabel focused={false}>Annotation Type:</InputLabel>
+        <InputLabel focused={false}>Field:</InputLabel>
         <Select value={annotationType} onChange={changeMetric}>
           {ANNOTATION_TYPES.map(annotationType => (
             <MenuItem key={annotationType} value={annotationType}>
@@ -147,7 +147,7 @@ function AnnotationsSelectorInner({
   );
 }
 
-function getLocationsParamValue(
+function getStateFipsParamValue(
   params: QueryString.ParsedQuery,
   defaultValue: string,
 ): string {
--- a/src/screens/internal/Anomalies/AnomaliesSelector.tsx
+++ b/src/screens/internal/Anomalies/AnomaliesSelector.tsx
@@ -20,10 +20,17 @@ import {
   AnnotationType,
   ANNOTATION_TYPES,
   annotationTypeNames,
-  AnnotationSelectorProps,
-  AnnotationSelectorInnerProps,
+  AnnotationOptions,
 } from './utils';
 import regions from 'common/regions';
+
+interface AnnotationSelectorProps {
+  onNewOptions: (options: AnnotationOptions) => void;
+}
+
+interface AnnotationSelectorInnerProps extends AnnotationSelectorProps {
+  mainSnapshot: number;
+}
 
 export function AnnotationsSelector(props: AnnotationSelectorProps) {
   const mainSnapshot = useMainSnapshot();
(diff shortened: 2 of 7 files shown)


src/common/metric.tsx
```diff
@@ -15,14 +15,14 @@ import { formatDecimal, formatPercent } from 'common/utils';
 import { Metric } from 'common/metricEnum';
 
 export const ALL_METRICS = [
+  Metric.WEEKLY_CASES_PER_100K,
```
smcclure17, line 18: Drive-by change to re-order the metrics in the compare tool so that the core metrics come first

src/screens/internal/CompareSnapshots/SnapshotVersions.tsx
```diff
@@ -47,12 +47,20 @@ export function useSnapshotVersion(
 ): SnapshotVersion | null {
   const [version, setVersion] = useState<SnapshotVersion | null>(null);
   useEffect(() => {
-    setVersion(null);
-    if (snapshot !== null) {
-      new Api(snapshotUrl(snapshot)).fetchVersionInfo().then(version => {
-        setVersion(version);
-      });
+    async function fetchVersion() {
```
smcclure17, line 50: Converting this to an async function so that the try/catch will execute correctly

src/screens/internal/CompareSnapshots/OptionsSelector.tsx
```diff
@@ -98,7 +98,7 @@ function OptionsSelectorInner({
     getNumericParamValue(params, 'sort', SortType.METRIC_DIFF),
   );
   const [metric, setMetric] = useState(
-    getNumericParamValue(params, 'metric', Metric.CASE_DENSITY),
+    getNumericParamValue(params, 'metric', Metric.WEEKLY_CASES_PER_100K),
```
smcclure17, line 101: Another drive-by change to set the default of the compare tool to weekly new cases

src/screens/internal/CompareSnapshots/OptionsSelector.tsx
```diff
@@ -39,7 +39,7 @@ interface OptionsSelectorProps {
  * page (all graphs, etc.) to re-render causing a big perf hiccup.  See:
  * https://stackoverflow.com/q/60682426
  */
-const Select: React.FunctionComponent<SelectProps> = ({
+export const Select: React.FunctionComponent<SelectProps> = ({
```
mikelehen, line 42: As a general guide, I usually try to avoid having one "screen" or "component" import code from another "screen" or "component."  If it's code that we want to share it's usually worth moving to its own component or into a shared source file (under common or whatever) depending on the code in question.

That said, since these are both internal tools that are kinda' related and the compare tool is already a bit messy, I wouldn't worry about it too much.  So I wouldn't bother changing this.  Just pointing it out for future reference.

src/screens/internal/Anomalies/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+  annotationType: AnnotationType;
+}
+
+export interface AnnotationSelectorProps {
+  onNewOptions: (options: AnnotationOptions) => void;
+}
+
+export interface AnnotationSelectorInnerProps extends AnnotationSelectorProps {
+  mainSnapshot: number;
+}
```
mikelehen, line 52: nit: I would move these two `Props` interfaces into AnomaliesSelector.tsx since they're used in exactly one place (defining the components) they're used and it's easier to understand the component if you can immediately see the props that it takes without going to a different file.

Moving to utils would make more sense to me if they were used in a bunch of places.

src/screens/internal/Anomalies/AnomaliesSelector.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+  const params = QueryString.parse(history.location.search);
+
+  const [snapshot, setSnapshot] = useState(
+    getNumericParamValue(params, 'snapshot', mainSnapshot),
+  );
+
+  const [snapshotText, setSnapshotText] = useState(snapshot.toString());
+
+  const [annotationType, setMetric] = useState(
```
mikelehen, line 52: Can we reconcile the naming of `annotationType` and `setMetric`?

src/screens/internal/Anomalies/AnomaliesSelector.tsx
```diff
@@ -0,0 +1,160 @@
...
+    </OptionsSelectorWrapper>
+  );
+}
+
+function getStateFipsParamValue(
+  params: QueryString.ParsedQuery,
+  defaultValue: string,
+): string {
+  let value = get(params, 'stateFips', defaultValue);
+  if (typeof value !== 'string' || regions.findByStateCode(value)) {
```
mikelehen, line 155: Seems like this should be `regions.findByStateCode(value) !== null` or something.

Alternatively we could probably just do a `regions.findByStateCodeStrict(value)` which will throw if it's not found.

mikelehen (approved):
In general, this looks good.

One thing I wonder is if we could merge this into the existing compare tool though.  Like under each chart have a little "5 annotations (latest: 2021-02-51)" link and if you click it, it expands to show you the actual annotations...  Or something like that.

I'm not opposed to having it be a separate tool, but to be the most useful I think we'd want to integrate it into the tool we're already using on a daily basis. 🤷   We can start with this though.

src/screens/internal/Anomalies/AnomaliesSelector.tsx
smcclure17: whoops, yeah I think you're right. I think I check `typeof value !== 'string'` first because `findByStateCode` only takes strings, so I couldn't pass `value` directly. But, in this format, I think it should be 

`if (typeof value !== 'string' || regions.findByStateCode(value)) === null`

src/screens/internal/CompareSnapshots/OptionsSelector.tsx
smcclure17: Ah, yeah that does make a lot of sense. Noted for the future and thanks for the flag :)

src/screens/internal/Anomalies/AnomaliesSelector.tsx
smcclure17: oops, yeah good catch

smcclure17:
Thanks for the 👀s as always! 🙏 

I agree that we could make this more useful by integrating it better with the existing tool. I think I got a bit too focused on getting this working and didn't spend enough time thinking about what format/tooling would be most useful (although I do still think this will be helpful as is). I'll take a look at integrating this better (starting with your idea/suggestion) over the next day or two

PR #5741, opened 2022-05-02
Add anomalies modal to internal compare tool

smcclure17:
Add a modal to display anomalies inside the internal compare tool. 

If new anomalies are introduced in the Right snapshot, the anomalies button is highlighted like:
image
I can tweak this to highlight new anomalies in another way if we feel it's not clear enough


src/screens/internal/CompareSnapshots/AnomalyCompare.style.ts
```diff
@@ -0,0 +1,23 @@
...
+  cursor: pointer;
+  background: ${props => (props.anyNewAnomalies ? 'yellow' : 'white')};
+  font-weight: 500;
+  box-shadow: none;
+  text-transform: none;
+  display: inline-block;
+`;
+
+export const AnomalyModalWrapper = styled(Box)({
+  position: 'absolute' as 'absolute',
```
mikelehen, line 15: not sure what's going on with `as 'absolute` ?

Also, this seems to be using a different syntax than we usually use.  I'd expect:
```
styled(Box)`
  position: absolute;
  ...
`;

But maybe there's a reason this is different?

File: src/screens/internal/CompareSnapshots/AnomalyModal.tsx [hunk 2]
```diff
@@ -0,0 +1,54 @@
...
+  pair: ProjectionsPair;
+  metric: Metric;
+}) {
+  const [open, setOpen] = React.useState(false);
+  const handleOpen = () => setOpen(true);
+  const handleClose = () => setOpen(false);
+
+  const leftAnomalies = getAnomaliesForMetric(pair.left, metric);
+  const rightAnomalies = getAnomaliesForMetric(pair.right, metric);
+  const newAnomalies = leftAnomalies?.length !== rightAnomalies?.length;
```
mikelehen, line 21: Just as a flag, I think it's possible to remove anomalies between snapshots too, so "new" isn't necessarily accurate.  It's also probably technically possible that we gain and lose the same number of anomalies.  We might want to use isEqual() from lodash to do a deep comparison or something.

src/screens/internal/CompareSnapshots/AnomalyModal.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import { AnomalyModalWrapper, AnomalyButton } from './AnomalyCompare.style';
+import { Modal, Typography } from '@material-ui/core';
+import { ProjectionsPair } from 'common/models/ProjectionsPair';
+import { Metric } from 'common/metricEnum';
+import { getAnomaliesForMetric } from './utils';
+
+export default function AnomalyModal({
```
mikelehen, line 8: nit: Since this *can show* a modal, but by default is just a button that lists how many anomalies there are, I'd probably name it `AnomaliesButton` or something.

src/screens/internal/CompareSnapshots/utils.ts
```diff
@@ -262,3 +264,43 @@ async function fetchTopCountiesByDiff(
...
+  projection: Projections,
+  metric: Metric,
+): Anomalies | undefined {
+  assert(ALL_METRICS.includes(metric));
+  const annotations = projection.primary.annotations;
+  switch (metric) {
+    case Metric.WEEKLY_CASES_PER_100K ||
+      Metric.ADMISSIONS_PER_100K ||
+      Metric.CASE_GROWTH_RATE:
+      return annotations?.newCases?.anomalies;
```
mikelehen, line 285: Per Slack conversation, I think ADMISSIONS_PER_100K shouldn't be in here, but CASE_DENSITY should.

Also I'm not sure if this syntax is correct.  I think `Metric.WEEKLY_CASES_PER_100K || Metric.ADMISSIONS_PER_100K || Metric.CASE_GROWTH_RATE` will just evaluate to the first truthy value, which will be `Metric.WEEKLY_CASES_PER_100K` and so it'll ignore the other items.

I think it needs to be something like:
```
case Metric.WEEKLY_CASES_PER_100K: 
case Metric.CASE_DENSITY:
case Metric.CASE_GROWTH_RATE:
  return annotations?.newCases?.anomalies;
```

mikelehen (approved):
Looks good with a couple minor things.

src/screens/internal/CompareSnapshots/AnomalyCompare.style.ts
smcclure17: TBH I don't remember what the `as absolute` was for either, but it works without it 😭 

I was following the syntax/structure in the [`MUI styled()`](https://mui.com/system/styled/#BasicUsage.tsx) docs, but I think that might be using a different engine. I've updated this to match our usual styles

src/screens/internal/CompareSnapshots/utils.ts
smcclure17: thanks for the catch and the explanation--updated in 142e41ea069eac537f5ed8968120603f4f7ac33f

src/screens/internal/CompareSnapshots/AnomalyModal.tsx
smcclure17: Yeah good call, updated to use lodash `isEqual()` in 142e41ea069eac537f5ed8968120603f4f7ac33f

PR #5746, opened 2022-05-02
Add CAN GA4 property

faihegberg:
This PR adds a new GA4 property to our Google Analytics, which starts tracking our user data along side our UA property. Note that UA is currently not affected whatsoever. So our current tracking data is not affected.

This is part of our migration from UA to GA4, which involves the following:
- Create a new GA4 property from the UA property, as per instructions in [this documentation](https://support.google.com/analytics/answer/10312255).
- Add the property to our code in the same manner as what we implemented for Google Ads [here](https://github.com/covid-projections/covid-projections/blob/a146d7166c8cbc8956fcf0bf176bc2d82b31c634/public/index.html#L142-L154), as per "tagging instructions" (found through Google Analytics --> Admin --> Data Steams --> CoVidActNow - G4 --> Tagging Instructions).
image

NOTE: This PR is actually tiny! The rest of the code is the same but I think our linter decided to shift the spacing slightly. The **only** change to this PR is the part I highlighted and commented below.

**How to test**
1. Open the preview link and navigate around the site however you want.
2. Visit our new [GA4 dashboard](https://analytics.google.com/analytics/web/#/p313668181/realtime/overview) and watch it tracks your actions.

Currently, I'm still figuring out how to track events in the exact same way as we currently do (if even possible). But for now, at least page views should be working.

[More info in this documentation here.](https://www.dropbox.com/scl/fi/44vsd18ki3mhg8t3m66jb/Upgrading-UA-to-GA4.paper?dl=0&rlkey=s1id1aoz9vl3widw3ap7z7swa)

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1,21 +1,19 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="utf-8" />
-    <!--link rel="icon" href="%PUBLIC_URL%/favicon.ico" /-->
-    <meta
-      name="viewport"
-      content="minimum-scale=1, initial-scale=1, width=device-width"
-    />
 
-    <meta name="theme-color" content="#000000" />
-    <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
-    <!--
+<head>
+  <meta charset="utf-8" />
+  <!--link rel="icon" href="%PUBLIC_URL%/favicon.ico" /-->
+  <meta name="viewport" content="minimum-scale=1, initial-scale=1, width=device-width" />
+
+  <meta name="theme-color" content="#000000" />
+  <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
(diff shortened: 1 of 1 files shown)


public/index.html
```diff
@@ -25,156 +23,137 @@
...
+
+  <!-- Global site tag (gtag.js) - Google Analytics -->
+  <script async src="https://www.googletagmanager.com/gtag/js?id=G-9883DK9DYY"></script>
+  <script>
+    window.dataLayer = window.dataLayer || [];
+    function gtag() { dataLayer.push(arguments); }
+    gtag('js', new Date());
+
+    gtag('config', 'G-9883DK9DYY');
+  </script>
```
faihegberg, line 137: This is the only change in this PR! The rest of the file is just a tiny spacing change done by our linter.

mikelehen (approved):
Changes seem fine. I don't know much about UA and GA4, but AFAICT you did all the right things! 🤷

PR #5755, opened 2022-05-03
Add event tracking to GA4

faihegberg:
This PR adds event tracking to our new GA4 property. This change _does not_ affect out current event tracking whatsoever. It merely tracks events _alongside_ our current tracking in UA. Once merged, we'll be able to compare numbers between UA and GA4.

Data models are different between UA and GA4. While UA tracks activities with event action, event category, event label, and event value, these do not translate directly to GA4. Instead, "event action" in UA is _automatically_ converted to "event name" in GA4, and the rest need to be translated to custom parameters in GA4. These custom parameters can be created through the UI, as done so [here](https://analytics.google.com/analytics/web/#/p313668181/customdefinitions/hub?params=_u..nav%3Dmaui%26_u..pageSize%3D25). Then the code needs to be adjusted to match these parameters. More info in this doc [here](https://support.google.com/analytics/answer/10075209#zippy=%2Cin-this-article).

**Note:** Some values in GA may still be "(not set)" as this code is not yet merged to prod. The values which may be showing are the ones that GA4 automatically detects (new GA4 stuff has some automated features) merely from changing which events translate to what in the GA4 UI. The only way to compare the numbers is after this PR is merged.

How to test:
1. Open the vercel preview link for this PR.
2. Visit [Events](https://analytics.google.com/analytics/web/#/p313668181/reports/explorer?r=top-events) in GA.
3. Add custom parameter (Event category, label, or value) to the dashboard, like in the following screenshot.
image
4. Play around with the site through the preview link and refresh the analytics page to see if your interaction shows up.

### DIFF
--- a/src/components/Analytics/gtag.ts
+++ b/src/components/Analytics/gtag.ts
@@ -1,0 +1,8 @@
+import { EventCategory } from './utils';
+
+// https://developers.google.com/analytics/devguides/collection/gtagjs/events
+export const event = (action: string, category: EventCategory) => {
+  gtag('event', action, {
+    event_category: category,
+  });
+};
--- a/src/components/Analytics/gtag.ts
+++ b/src/components/Analytics/gtag.ts
@@ -1,15 +1,15 @@
-import { EventCategory } from './utils';
-
 // https://developers.google.com/analytics/devguides/collection/gtagjs/events
-export const event = (
+export const trackGA4Event = (
   action: string,
-  category: EventCategory,
-  label?: string,
+  category: string,
+  label: string,
   value?: number,
 ) => {
-  gtag('event', action, {
-    event_category: category,
-    event_label: label,
-    event_value: value,
-  });
+  if (gtag) {
+    gtag('event', action, {
+      event_category: category,
+      event_label: label,
+      event_value: value,
+    });
+  }
 };
--- a/src/components/Analytics/gtag.ts
+++ b/src/components/Analytics/gtag.ts
@@ -1,8 +1,15 @@
 import { EventCategory } from './utils';
 
 // https://developers.google.com/analytics/devguides/collection/gtagjs/events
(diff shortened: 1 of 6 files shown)


src/components/Analytics/gtag.ts
```diff
@@ -0,0 +1,15 @@
...
+  category: EventCategory,
+  label?: string,
+  value?: number,
+) => {
+  gtag('event', action, {
+    event_category: category,
+    event_label: label,
+    event_value: value,
+  });
+};
```
pnavarrc, line 15: A few notes here:

Let's make `event_label` required to ensure that we know what we are recording. Also, maybe rename the function name to `trackEvent` or `trackGA4Event` so the name better represents what the function does.

Finally, it's possible (but unlikely) that `gtag` when we call this function, consider adding a conditional to cover this case

```ts
if (gtag) {
  gtag('event', action, {/* ... */});
}
```

src/components/Analytics/utils.ts
```diff
@@ -3,6 +3,7 @@ import words from 'lodash/words';
 import ReactGA from 'react-ga';
 import { amplitudeLogEvent } from './amplitude';
 import { fullStoryTrackEvent } from 'common/fullstory';
+import * as gtag from './gtag';
```
pnavarrc, line 6: How about importing just the function that we need?

```ts
import { trackGA4Event } from './gtag';
```

pnavarrc (approved):
Looks good, left a few comments!

src/components/Analytics/utils.ts
```diff
@@ -116,6 +117,8 @@ export function trackEvent(
       transport,
     });
 
+    gtag.event(action, category, label, value);
```
pnavarrc, line 120: Consider calling `toTitleCase` if needed (to match how events are being recorded for UA)

PR #5756, opened 2022-05-04
Remove legacy dead code from SummaryStats component.

mikelehen:
SummaryStats used to be used for the LocationPage above-the-fold, but we rewrote it a long time ago so now it's only used by Embed and SocialLocationPreview. So it contained a lot of dead code.

This also lets us get rid of the detail() method from LevelInfo, which cleans up some legacy metric code.

I also deleted the unused SignalStatus component which used to be used by SummaryStats but its usage was removed a while ago.

Tested by verifying our [embed](https://covid-projections-git-mikelehen-cleanup-summ-104a70-covidactnow.vercel.app/embed/us/washington-wa/?s=33202135) and [share images](https://covid-projections-git-mikelehen-cleanup-summ-104a70-covidactnow.vercel.app/internal/share-image/states/wa) didn't change:
image

image


src/components/SocialLocationPreview/SocialLocationPreview.tsx
```diff
@@ -43,7 +43,7 @@ const SocialLocationPreview = (props: {
         <AlarmLevel color={fillColor}>{levelInfo.name}</AlarmLevel>
       </PreviewHeader>
       <PreviewBody>
-        <SummaryStats stats={props.stats} condensed={true} />
```
mikelehen, line 46: `SummaryStats` was used here and in Embed.tsx, but `condensed` is true in both places, so now we always show the "condensed" version.

pnavarrc (approved):
Thanks for cleaning up!! 🧹

PR #5764, opened 2022-05-04
Color graded charts starting on April 18

pnavarrc:
The community-level framework is currently applied retroactively, which doesn't match the risk levels before our recent framework change. This demo shows what it would look like to only color the lines starting on a certain date (which needs to be determined)

image


src/components/Dialogs/utils.tsx
```diff
@@ -29,6 +30,15 @@ const metricToLearnLink: { [key in Metric]: string } = {
...
 };
 
+const howItsGradedMap: { [key: number]: string } = {
+  [Metric.ADMISSIONS_PER_100K]:
+    'Levels for weekly new admissions are based on the CDC’s Community Level framework. Less than 10 weekly new admissions is considered Low, between 10 and 19.9 is considered Medium, and greater than or equal to 20 is considered High. No color grading is applied prior to the April 18, 2022 change to a community risk level framework, which better reflects the more recent decreased risk of severe illness and death from COVID due to vaccines, therapeutics, and past COVID infections, and other developments.',
+  [Metric.RATIO_BEDS_WITH_COVID]:
+    'Levels for percentage of patients with COVID are generally based on the CDC’s Community Level framework. Less than 10% of all new beds is considered Low, between 10 and 14.9% is considered Medium, and greater than or equal to 15 is considered High. No color grading is applied prior to the April 18, 2022 change to a community risk level framework, which better reflects the more recent decreased risk of severe illness and death from COVID due to vaccines, therapeutics, and past COVID infections, and other developments.',
```
mikelehen, line 37: Any idea why this says "are generally based on" and the other two say "are based on" ?

mikelehen: Seems like "equal to 15" should be "equal to 15%"

mikelehen (approved):
LGTM with a minor copy nit that you can ignore if you want.

src/components/Dialogs/utils.tsx
pnavarrc: not sure why it says "generally" in this case, I'm going to leave that, but update 15 → 15% 👍

PR #5796, opened 2022-05-09
Re-route sparkline click

chasulin:
When clicking spark lines, the page now scrolls to the corresponding metric's chart instead of to explore.

**Test on mobile** (the spark line block is now an independent component only on mobile -- desktop combines spark lines into summary stats)


mikelehen (approved):
Looks good!

PR #5814, opened 2022-05-11
Remove can 8.2 banner

chasulin:
[Trello ticket](https://trello.com/c/HkQEWqiM/653-remove-can-82-banner)


mikelehen (approved):
LGTM.  I'd be slightly tempted to keep the `Can82Banner` component checked-in to save a bit of time next time we need to create a similar banner (looks like this is the only banner checked-in using the `CondensedBanner` styles, and it's the only one we've put on a location page I think).  But it's also not hard to dig the code out of git history, so I don't feel strongly.  Up to you!

PR #5848, opened 2022-05-17
Tile map bug fix

faihegberg:
This PR addresses the tile map styling bug, where the map overlaps the footer when the screen size is > 1320, as described in this [card](https://trello.com/c/ugdDDile/661-location-page-map-tile-overlaps-footer).

I added a few props in this PR which ended up indenting everything after running the linter. I'll mark parts where there's actually a code change to make it easier to review.

**Note:** Although I think the solution in this PR is better than what we have now, I don't necessarily think this is the best solution. Let me know if you can think of a better way!

### DIFF
--- a/src/components/LocationPage/ChartsHolder.tsx
+++ b/src/components/LocationPage/ChartsHolder.tsx
@@ -54,234 +54,242 @@ const scrollTo = (div: null | HTMLDivElement, offset: number = 100) =>
 interface ChartsHolderProps {
   region: Region;
   chartId: string;
+  yPosition?: number;
 }
 
-const ChartsHolder = React.memo(({ region, chartId }: ChartsHolderProps) => {
-  const projections = useProjectionsFromRegion(region);
+const ChartsHolder = React.memo(
+  ({ region, chartId, yPosition }: ChartsHolderProps) => {
+    const projections = useProjectionsFromRegion(region);
 
-  const locationSummary = useLocationSummariesForFips(region.fipsCode);
+    const locationSummary = useLocationSummariesForFips(region.fipsCode);
 
-  const communityMetricsRef = useRef<HTMLDivElement>(null);
-  const vaccinationsBlockRef = useRef<HTMLDivElement>(null);
-  const transmissionMetricsRef = useRef<HTMLDivElement>(null);
-  const chartBlockRefs = useMemo(
-    () => ({
-      [GroupHeader.COMMUNITY_LEVEL]: communityMetricsRef,
-      [GroupHeader.VACCINATED]: vaccinationsBlockRef,
-      [GroupHeader.TRANSMISSION]: transmissionMetricsRef,
-    }),
-    [],
-  );
+    const communityMetricsRef = useRef<HTMLDivElement>(null);
+    const vaccinationsBlockRef = useRef<HTMLDivElement>(null);
+    const transmissionMetricsRef = useRef<HTMLDivElement>(null);
+    const chartBlockRefs = useMemo(
+      () => ({
+        [GroupHeader.COMMUNITY_LEVEL]: communityMetricsRef,
+        [GroupHeader.VACCINATED]: vaccinationsBlockRef,
+        [GroupHeader.TRANSMISSION]: transmissionMetricsRef,
+      }),
+      [],
+    );
 
-  // A map of each metric to the ref of its chart group
-  const metricRefs = useMemo(
-    () => ({
-      [Metric.CASE_DENSITY]: transmissionMetricsRef,
-      [Metric.CASE_GROWTH_RATE]: transmissionMetricsRef,
-      [Metric.POSITIVE_TESTS]: transmissionMetricsRef,
-      [Metric.HOSPITAL_USAGE]: transmissionMetricsRef,
-      [Metric.VACCINATIONS]: vaccinationsBlockRef,
-      [Metric.WEEKLY_CASES_PER_100K]: communityMetricsRef,
-      [Metric.ADMISSIONS_PER_100K]: communityMetricsRef,
-      [Metric.RATIO_BEDS_WITH_COVID]: communityMetricsRef,
-    }),
-    [],
-  );
-  const shareBlockRef = useRef<HTMLDivElement>(null);
-  const exploreChartRef = useRef<HTMLDivElement>(null);
-  const recommendationsRef = useRef<HTMLDivElement>(null);
+    // A map of each metric to the ref of its chart group
+    const metricRefs = useMemo(
+      () => ({
+        [Metric.CASE_DENSITY]: transmissionMetricsRef,
+        [Metric.CASE_GROWTH_RATE]: transmissionMetricsRef,
+        [Metric.POSITIVE_TESTS]: transmissionMetricsRef,
+        [Metric.HOSPITAL_USAGE]: transmissionMetricsRef,
+        [Metric.VACCINATIONS]: vaccinationsBlockRef,
+        [Metric.WEEKLY_CASES_PER_100K]: communityMetricsRef,
+        [Metric.ADMISSIONS_PER_100K]: communityMetricsRef,
+        [Metric.RATIO_BEDS_WITH_COVID]: communityMetricsRef,
+      }),
+      [],
+    );
+    const shareBlockRef = useRef<HTMLDivElement>(null);
+    const exploreChartRef = useRef<HTMLDivElement>(null);
+    const recommendationsRef = useRef<HTMLDivElement>(null);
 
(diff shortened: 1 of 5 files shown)


src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -54,234 +54,242 @@ const scrollTo = (div: null | HTMLDivElement, offset: number = 100) =>
 interface ChartsHolderProps {
   region: Region;
   chartId: string;
+  yPosition?: number;
```
faihegberg, line 57: Added prop to determine at which point on the page the user is.

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -54,234 +54,242 @@ const scrollTo = (div: null | HTMLDivElement, offset: number = 100) =>
...
 }
 
-const ChartsHolder = React.memo(({ region, chartId }: ChartsHolderProps) => {
-  const projections = useProjectionsFromRegion(region);
+const ChartsHolder = React.memo(
+  ({ region, chartId, yPosition }: ChartsHolderProps) => {
```
faihegberg, line 61: `yPosition` used here.

src/components/LocationPage/ChartsHolder.tsx
```diff
@@ -54,234 +54,242 @@ const scrollTo = (div: null | HTMLDivElement, offset: number = 100) =>
...
+          <AboveTheFold
+            region={region}
+            locationSummary={locationSummary}
+            onClickMetric={onClickMetric}
+            onClickAlertSignup={onClickAlertSignup}
+            onClickShare={onClickShare}
+            onClickSparkLine={onClickSparkLine}
+            onClickMasksCard={onClickMasksCard}
+            onClickTransmissionMetricsCard={onClickTransmissionMetricsCard}
+            yPosition={yPosition}
```
faihegberg, line 224: `yPosition` passed here.

faihegberg:
[Location Page Map Tile Overlaps Footer](https://trello.com/c/ugdDDile/661-location-page-map-tile-overlaps-footer)

src/screens/LocationPage/LocationPage.tsx
```diff
@@ -17,6 +17,8 @@ function WithSearchInNav({ region }: LocationPageProps) {
   let { chartId } = useParams<{ chartId: string }>();
   const [menuOpen, setMenuOpen] = useState(false);
   const hasScrolled = useShowPastPosition(850);
+  const [yPosition, setYposition] = useState(0);
+  window.addEventListener('scroll', () => setYposition(window.scrollY));
```
mikelehen, line 21: Is there a reason we're putting this way up on the `LocationPage` component and then passing the `yPosition` prop down through all the components?  I think that means every time you scroll, React is going to re-render a bunch of components since the props will have changed.  I think putting the scroll listener as close to where we need (CountyMap?) might make more sense.

Also, I think this should be inside a hook that removes the listener when the hook is removed.  As-is, we're going to end up registering more and more of these listeners and they'll never get removed.

See the `useShowPastPosition()` hook for something similar. https://github.com/covid-projections/covid-projections/blob/develop/src/common/hooks/useShowPastPosition.ts#L5

src/components/NewLocationPage/AboveTheFold/AboveTheFold.tsx
```diff
@@ -49,8 +50,10 @@ const AboveTheFold: React.FC<AboveTheFoldProps> = React.memo(
     onClickSparkLine,
     onClickMasksCard,
     onClickTransmissionMetricsCard,
+    yPosition,
   }) => {
     const showMasksCard = locationSummary.level === Level.HIGH;
+    const offsetMap = yPosition ? yPosition > 5350 : false;
```
mikelehen, line 56: This `5530` looks very fragile.  Any time we add/remove content on the location page it'll change this number, right?  Can we make it relative to the bottom of the page instead, so the number is only dependent on the footer size?  I'd also then put this into a constant called FOOTER_HEIGHT or something.

src/components/NewLocationPage/CountyMap/CountyMap.tsx
```diff
@@ -61,6 +61,24 @@ const MAP_TYPE_INFO: { [key in MapType]: MapTypeInfo } = {
   },
 };
 
+function useMapTileOffset(): number {
+  const [yPosition, setYposition] = useState(0);
```
mikelehen, line 65: `yPosition` of what? I can't fully guess what this is without reading all of the code.  A fully verbose name might be `windowScrollYPosition`, but you could shorten to`windowScrollY` or `scrollY` and still be pretty clear.

src/components/NewLocationPage/CountyMap/CountyMap.tsx
```diff
@@ -61,6 +61,24 @@ const MAP_TYPE_INFO: { [key in MapType]: MapTypeInfo } = {
...
+  const pageHeight = document.documentElement.scrollHeight;
+  const footerHeight = document.getElementsByTagName('footer')[0]?.scrollHeight;
```
mikelehen, line 67: Ooh, nice.  I was worried we'd have to hardcode this.  Good idea.

src/components/NewLocationPage/CountyMap/CountyMap.tsx
```diff
@@ -61,6 +61,24 @@ const MAP_TYPE_INFO: { [key in MapType]: MapTypeInfo } = {
...
+  // Add an additional 500px to detect when the user is close enough to the bottom of the page.
```
mikelehen, line 68: This comment is hard to understand without more context about what this code is doing.  It might be worth combining the comment at line 70 to this one and saying something like:

> Once the user scrolls far enough that the footer is within 500px of the top of the screen, we want to freeze the map tile so that it can't overlap with the footer and obscure footer content. To do that we set its `bottom` to be 100px above the footer.

src/components/NewLocationPage/CountyMap/CountyMap.style.tsx
```diff
@@ -12,7 +12,7 @@ export const PinnedContainer = css`
   z-index: 901;
 `;
 
-export const MapContainer = styled.div`
+export const MapContainer = styled.div<{ offsetAmount: number }>`
```
mikelehen, line 15: Consider tweaking this name since:
1. "offsetAmount" isn't very clear.  What is the offset relative too?  Is it an x offset or y offset or something else?  TBH this isn't really an offset.  It's just a fixed position for the bottom.
2. "Amount" doesn't really add any information and could probably be dropped."

Perhaps `bottomPosition` or `fixedBottomPosition` or something.

I'd also be tempted to make the type `number | null` instead of treating `0` as a special value that sets it to `none`.  I think using `null` for that would be cleaner.

src/components/NewLocationPage/CountyMap/CountyMap.tsx
```diff
@@ -61,6 +61,24 @@ const MAP_TYPE_INFO: { [key in MapType]: MapTypeInfo } = {
...
+  const offsetPoint = pageHeight - footerHeight - 500;
+  // When the user goes beyond the offset point, vertically offset the map tile by the footer height + 100px.
+  const offsetAmount = yPosition > offsetPoint ? footerHeight + 100 : 0;
```
mikelehen, line 71: If you rename the `offsetAmount` property on `MapContainer` rename this to match.

src/components/NewLocationPage/CountyMap/CountyMap.tsx
```diff
@@ -61,6 +61,24 @@ const MAP_TYPE_INFO: { [key in MapType]: MapTypeInfo } = {
   },
 };
 
+function useMapTileOffset(): number {
```
mikelehen, line 64: Consider adding a comment to this function explaining its purpose (to avoid the map tile overlapping the footer), since that's not obvious from the name (and renaming to `useMapTileOffsetAvoidingFooterOverlap()` seems a bit overkill.

Also, do we call it a "map tile" in other parts of the code?  Or should we change this to `MapContainer` or `CountyMap` or similar?

mikelehen (approved):
I got picky with some of the naming / comments, but this looks really great.  It came together really nicely, better than I hoped even. Thanks for working through the prior feedback!

src/components/NewLocationPage/CountyMap/CountyMap.tsx
faihegberg: I think the word "tile" was just used in the Trello card. I'll rename this to `useMapContainerOffset` and add a comment above the function.

PR #5898, opened 2022-05-24
Fix smoothed cases in explore to match official case metric from backend.

mikelehen:
Fixes https://trello.com/c/A4Gbquur/669-daily-new-cases-in-explore-is-inconsistent-with-weekly-new-cases-per-100k-metric-doesnt-handle-irregular-reporting-as-well

Rather than calculating `smoothedDailyCases` directly from `rawDailyCases`, we base it on the `caseDensity` metric which benefits from various smoothing done on the backend.

This is most notable for Florida explore metrics where the biweekly reporting causes issues.

**Before:**
image

**After:**
image

### DIFF
--- a/src/common/models/Projection.ts
+++ b/src/common/models/Projection.ts
@@ -223,8 +223,14 @@ export class Projection {
     this.region = region;
 
     // Set up our series data exposed via getDataset().
+    this.caseDensityByCases = metricsTimeseries.map(
+      row => row && row.caseDensity,
+    );
+
     this.rawDailyCases = actualTimeseries.map(row => row && row.newCases);
-    this.smoothedDailyCases = this.smoothWithRollingAverage(this.rawDailyCases);
+    this.smoothedDailyCases = this.denormalizeByPopulation(
+      this.caseDensityByCases,
+    );
 
     this.rawDailyDeaths = actualTimeseries.map(row => row && row.newDeaths);
     this.smoothedDailyDeaths = this.smoothWithRollingAverage(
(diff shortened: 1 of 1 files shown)


src/common/models/Projection.ts
```diff
@@ -223,8 +223,14 @@ export class Projection {
     this.region = region;
 
     // Set up our series data exposed via getDataset().
+    this.caseDensityByCases = metricsTimeseries.map(
+      row => row && row.caseDensity,
+    );
```
mikelehen, line 228: I moved this up from below so I could use it in `smoothedDailyCases`

src/common/models/Projection.ts
```diff
@@ -223,8 +223,14 @@ export class Projection {
...
+
     this.rawDailyCases = actualTimeseries.map(row => row && row.newCases);
-    this.smoothedDailyCases = this.smoothWithRollingAverage(this.rawDailyCases);
+    this.smoothedDailyCases = this.denormalizeByPopulation(
+      this.caseDensityByCases,
+    );
```
mikelehen, line 233: The one "annoying" thing about using the caseDensity metric is that it's normalized by population, which we don't want.  So we have to denormalize it. 🤷

chasulin (approved):

mikelehen:
[Daily new cases in explore is inconsistent with "Weekly new cases per 100k" metric (doesn't handle irregular reporting as well).](https://trello.com/c/A4Gbquur/669-daily-new-cases-in-explore-is-inconsistent-with-weekly-new-cases-per-100k-metric-doesnt-handle-irregular-reporting-as-well)

PR #5909, opened 2022-05-26
Fix casing for event labels

faihegberg:
This PR ensures each word for the event label is capitalized, matching how event labels were recorded in UA. Also, since GA4 is now automatically sharing data with UA, we no longer need to send data to both.

### DIFF
--- a/src/components/Analytics/utils.ts
+++ b/src/components/Analytics/utils.ts
@@ -1,6 +1,5 @@
 import capitalize from 'lodash/capitalize';
 import words from 'lodash/words';
-import ReactGA from 'react-ga';
 import { amplitudeLogEvent } from './amplitude';
 import { fullStoryTrackEvent } from 'common/fullstory';
 import { trackGA4Event } from './gtag';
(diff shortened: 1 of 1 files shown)


src/components/Analytics/utils.ts
```diff
@@ -91,10 +90,17 @@ export enum EventAction {
...
 
+// Capitalize string (special characters are not kept)
 function toTitleCase(name: string) {
   return words(name).map(capitalize).join(' ');
 }
 
+// Capitalize string (special characters kept in place by not using "words" from lodash)
+function toTitleCaseWithSpecialCharacters(name: string) {
+  const words = name.split(' ');
+  return words.map(word => capitalize(word)).join(' ');
```
pnavarrc, line 101: You can simplify a little here (you don't need to create the arrow function inside `map`, see above [utils.ts#L95](https://github.com/covid-projections/covid-projections/pull/5909/files#diff-f0394b6ed9e1269459eda77202b48ae5bdd989fbd0fceab0b957487227204008R95)

```ts
return name.split(' ')
  .map(capitalize)
  .join(' ');
```

pnavarrc (approved):
Looks good!

PR #5930, opened 2022-05-31
Fix footer button bug

faihegberg:
Currently buttons in our footer are broken due to the new way our event labels are parsed. Historically, we allowed event labels to be empty (by using `words` in `toTitleCase` function, which can return an empty string). But now that we're _not_ using `words` and are parsing event labels by simpler string manipulation (rationale was to ensure UA and GA4 event labels match up), our new `toTitleCaseWithSpecialCharacters` function was not allowing `undefined` as an argument although it should have. This causes our site to throw an error when the user clicks on a button inside the footer (where event labels are `undefined`).

This PR fixes the footer button bug by allowing `undefined` to be passed to `toTitleCaseWithSpecialCharacters`.

### DIFF
--- a/src/components/Analytics/utils.ts
+++ b/src/components/Analytics/utils.ts
@@ -97,7 +97,7 @@ function toTitleCase(name: string) {
 
 // Capitalize string (special characters kept in place by not using "words" from lodash)
 function toTitleCaseWithSpecialCharacters(name: string) {
-  return name.split(' ').map(capitalize).join(' ');
+  return name ? name.split(' ').map(capitalize).join(' ') : '';
 }
 
 /**


src/components/Analytics/utils.ts
```diff
@@ -97,7 +97,7 @@ function toTitleCase(name: string) {
 
 // Capitalize string (special characters kept in place by not using "words" from lodash)
 function toTitleCaseWithSpecialCharacters(name: string) {
-  return name.split(' ').map(capitalize).join(' ');
+  return name ? name.split(' ').map(capitalize).join(' ') : '';
```
mikelehen, line 100: You said:
> our new toTitleCaseWithSpecialCharacters function was not allowing undefined as an argument although it should have

But `name` is defined as `string`, so:

1. We should change the type to `name | undefined` I guess.
2. We would ideally figure out why TypeScript was letting us pass `undefined` to a function that declares its type as `string`.  Normally that would be an error and we couldn't have shipped this code.  So we may be doing something "wrong" somewhere upstream of this function too.

mikelehen (approved):
I'm approving this since it's urgent, but we should probably dig a bit deeper into how `undefined` is getting passed to this function.

PR #6037, opened 2022-06-19
Fix RMSD internal compare option

smcclure17:
Add community risk metrics to `getDataset` helper function to fix RMSD/series compare tool. Without these changes the compare tool breaks when attempting to use the RMSD sort option for any of the new(ish) community risk metrics


mikelehen (approved):
Thanks!

PR #6114, opened 2022-07-02
Fix upperLimit thresholds to be non-inclusive.

mikelehen:
Fixes https://trello.com/c/DZOOrXz0/703-community-level-grading-inconsistent-when-admissions-is-at-100

We had a user-reported issue where the community level (calculated on the backend) was coming back as high because it was counting 10.0 admissions per 100k as "medium" (which becomes a "high" community level because the cases are higher) but the frontend was showing the 10.0 admissions per 100k as "low"

This fixes it and recalculates the location summaries.

Before:
image

After:
image


mikelehen:
[Community level grading inconsistent when admissions is at 10.0](https://trello.com/c/DZOOrXz0/703-community-level-grading-inconsistent-when-admissions-is-at-100)

pnavarrc (approved):
Looks good!

PR #6129, opened 2022-07-04
Fix chart line bug

faihegberg:
Fix chart line bug as pointed out in [this card](https://trello.com/c/XsIYZOsx/692-color-grading-visual-bug-in-a-small-number-of-hospital-metric-charts).

### DIFF
--- a/src/components/Charts/ChartAdmissionsPer100k.tsx
+++ b/src/components/Charts/ChartAdmissionsPer100k.tsx
@@ -1,7 +1,6 @@
 import React, { FunctionComponent } from 'react';
 import isDate from 'lodash/isDate';
 import { min as d3min, max as d3max } from 'd3-array';
-import { curveMonotoneX } from '@vx/curve';
 import { GridRows } from '@vx/grid';
 import { scaleLinear } from '@vx/scale';
 import { ParentSize } from '@vx/responsive';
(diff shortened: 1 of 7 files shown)


src/components/Charts/ChartAdmissionsPer100k.tsx
```diff
@@ -150,7 +149,6 @@ const ChartAdmissionsPer100k: FunctionComponent<{
           regions={regions}
           width={chartWidth}
           yScale={yScale}
-          curve={curveMonotoneX}
```
faihegberg, line 153: These no longer needs to be explicitly stated as a prop, since this is now a default prop.

src/components/Charts/ChartCaseDensity.tsx
```diff
@@ -146,7 +145,6 @@ const ChartCaseDensity: FunctionComponent<{
           regions={regions}
           width={chartWidth}
           yScale={yScale}
-          curve={curveMonotoneX}
```
faihegberg, line 149: Same here.

faihegberg:
[Color grading visual bug in a small number of hospital metric charts](https://trello.com/c/XsIYZOsx/692-color-grading-visual-bug-in-a-small-number-of-hospital-metric-charts)

pnavarrc (approved):
Looks good!

PR #6147, opened 2022-07-07
Hold off on launching GiveMomentum Donation page

faihegberg:
Content team asked to hold off on launching this til Monday.


mikelehen (approved):
This looks fine, but I don't know if it matters if we delete the route since nothing seems to link to it?  🤷

PR #6167, opened 2022-07-11
Split email recipient groups

faihegberg:
Establish a way to split alert email recipients into two groups. So we can send one group content linking to the GiveButter donation page and the other group content linking to the GiveMomentum donation page.

### DIFF
--- a/scripts/alert_emails/utils.ts
+++ b/scripts/alert_emails/utils.ts
@@ -36,6 +36,7 @@ interface AlertTemplateData {
   unsubscribe_link: string;
   feedback_subject_line: string;
   disclaimer: Disclaimer | null;
+  giveButterDonor: boolean;
 }
 
 /* EXAMPLE of an alert disclaimer.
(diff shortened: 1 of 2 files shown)


scripts/alert_emails/utils.ts
```diff
@@ -95,6 +103,7 @@ function generateAlertEmailContent(
       `[Alert Feedback] Alert for ${locationName} on ${lastUpdated}`,
     ),
     disclaimer,
+    giveButterDonor,
```
pnavarrc, line 106: Consider passing 2 variables instead, that way the logic stays in one place (the `generateAlertEmailContent` function) instead of passing that to the HTML template.

- `donateUrl`
- `donateText`

Once the experiment is finished, we will just need to delete the logic that creates 2 different `donateUrl` and `donateText` versions and the template will remain unchanged.

pnavarrc (commented):
Added a suggestion to centralize the experiment logic a bit more.

mikelehen (approved):
Looks good to me!

PR #6256, opened 2022-07-25
Fix donate button bug in alert email

faihegberg:
Currently the donate link in our alert email is too wide on smaller screen sizes and would wrap. This PR adjusts the font size for smaller screen sizes so the button would not wrap.

**Note:** Since handlebars aren't aware of user screen sizes and since styling in `template.html` is inline, adding a media tag to inline styling is challenging (if even possible, [more info](https://stackoverflow.com/questions/9808233/is-it-possible-to-put-css-media-rules-inline)). So I added the style tag in the document head to handle different screen sizes.

Before:
image

After:
image

### DIFF
--- a/scripts/alert_emails/template.html
+++ b/scripts/alert_emails/template.html
@@ -11,6 +11,15 @@
     <!--<![endif]-->
     <meta name="viewport" content="width=device-width">
     <meta name="x-apple-disable-message-reformatting">
+    <style>
+        .donate-button {
+            font-size: 10px;
+        }
+        @media (min-width: 600px) {
+            .donate-button {
+                font-size: 15px;
+            }
+        }
     </style>
 </head>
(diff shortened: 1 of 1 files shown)


scripts/alert_emails/template.html
```diff
@@ -11,6 +11,15 @@
...
     <meta name="x-apple-disable-message-reformatting">
+    <style>
+        .donate-button {
+            font-size: 10px;
+        }
+        @media (min-width: 600px) {
+            .donate-button {
+                font-size: 15px;
+            }
+        }
```
pnavarrc, line 22: Media queries and classes don't work consistently across email clients, I think is safer to just set the font-size to 10px in the `style` prop of the `a` element below

pnavarrc (approved):
Looks good!

PR #6461, opened 2022-08-30
Changing Tests CTA in Recommends Component

BrettBoval:
Federal free tests program is suspended due to lack of funding. I'm updating the CTA text and changing the link to a federal resource that should be more evergreen.


mikelehen (approved):
👍

PR #6540, opened 2022-09-13
Adding Analytics Context

BrettBoval:
Adding context to the calculation that can not be inferred just from reading the provided code.


BrettBoval:
@mikelehen PTAL.

I put context here that wasn't obvious to me from just reading the code.

Namely, (1) that multiple subscriptions for the same email address are included in the same record and (2) that the "unsubscribe purges you from the historical record" is just an implementation quirk, not a strict requirement.

mikelehen (approved):

PR #6765, opened 2022-10-27
Update Email Links

BrettBoval:
Simplify the number of emails addresses that we are keeping track by (1) switching over to @actnowcoalition domains and (2) standardizing towards google groups (especially contact@actnowcoalition.org).

### DIFF
--- a/src/screens/Terms/Terms.tsx
+++ b/src/screens/Terms/Terms.tsx
@@ -808,7 +808,7 @@ const Terms = ({ children }: { children: React.ReactNode }) => {
           All other feedback, comments, requests for technical support and other
           communications relating to the Website should be directed to:{' '}
           <a href="mailto:contact@actnowcoalition.org">
-            contact@actnowcoalition
+            contact@actnowcoalition.org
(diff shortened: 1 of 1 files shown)


src/screens/Terms/Terms.tsx
```diff
@@ -807,7 +807,10 @@ const Terms = ({ children }: { children: React.ReactNode }) => {
         <Typography variant="body1" component="p">
           All other feedback, comments, requests for technical support and other
           communications relating to the Website should be directed to:{' '}
-          <a href="mailto:info@CoVidActNow.org">info@CoVidActNow.org</a>.
+          <a href="mailto:contact@actnowcoalition.org">
+            contact@actnowcoalition
```
mikelehen, line 811: .org

mikelehen (approved):
Missing a ".org" in one place, but otherwise looks good.

FWIW- I would have left the email addresses as using the @covidactnow.org aliases so that we're presenting a unified brand to our users, but I don't feel strongly.

PR #7142, opened 2023-01-10
Remove Vx Map from homepage and update API types

smcclure17:
Removes Vx map and pulls in bivalent vaccination types from backend API

From roadmap in [CAN cleanup sprint planner](https://docs.google.com/document/d/1kN9Gn7hYzD70SRgXk4IdVkMR5F9GpDpmQIZUwjHBOO0/edit): This was lower in the priority list, but it was a quick change and will remove some extra potential complexity when adding new bivalent booster data and changing the priority/focus from 1+ to bivalent. 

This leaves the Vx map components and tooltips intact, just removes them from the homepage. There's a world in which we remove them entirely, but a) maybe we'll want these components in the future and b) it would require a fair amount of cleanup. I'm open to arguments that we should do that cleanup now, though

### DIFF
--- a/src/api/schema/Actuals.d.ts
+++ b/src/api/schema/Actuals.d.ts
@@ -56,7 +56,7 @@ export type Weeklycovidadmissions = number | null;
  *
  * Information about acute bed utilization details aggregated for the county's corresponding
  * Health Service Area (HSA). For CBSA, state, and country regions these fields are omitted.
- * For more on HSAs see: https://github.com/covid-projections/covid-data-model/blob/main/data/misc/README.md"
+ * For more on HSAs see: https://github.com/act-now-coalition/covid-data-model/blob/main/data/misc/README.md"
  *
  * Fields:
  *  * capacity - Current staffed acute bed capacity.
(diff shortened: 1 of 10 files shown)


src/api/schema/Actuals.d.ts
```diff
@@ -56,7 +56,7 @@ export type Weeklycovidadmissions = number | null;
  *
```
smcclure17, line 56: all files under `src/api/schema/**` were updated using `yarn update-api-types`

chasulin (approved):
looks good! totally your call on whether or not to do the deletion/clean up now vs later vs never. if theres a chance that we might add the vax map back in in some capacity in the future, keeping the code makes sense to me. up to you though

also, i assume this isnt meant to fold in removing the vax map from the minimap on location pages?

Screen Shot 2023-01-10 at 11 04 05 AM

smcclure17:
> also, i assume this isnt meant to fold in removing the vax map from the minimap on location pages?

hmm, this is a good question. We only explicitly talked about removing it from the homepage, but it does seem to follow that we'd remove it here too. Let me get a consensus on this. Thanks for the catch 🙏

smcclure17:
Talking w/ @BrettBoval/Tom we're going to leave this as is, since the mini-map defaults to risk level, there's no harm in terms of engagement of keeping the Vx map as an option

smcclure17:
realizing I shouldn't have merged this as we're waiting to ship this until comms/etc are ready 😅 going to revert

PR #7145, opened 2023-01-10
Add and focus on bivalent vaccine data

smcclure17:
Copy updates might not be final yet ([taken from Copy Doc here](https://docs.google.com/document/d/14eDHgYfzSdwbZ7_piIppjgFj81cDGTNzsZbuyevPGaQ/edit#)). [Preview link](https://covid-projections-git-smcclure17-bivalent-covidactnow.vercel.app/?s=44733217)


src/components/VaccineProgressBar/VaccineProgressBar.tsx
```diff
@@ -2,15 +2,14 @@ import React from 'react';
...
+import { VACCINATIONS_COLOR_MAP, vaccineColor } from 'common/colors';
 import { formatPercent } from 'common/utils';
 
 export interface ProgressBarProps {
   // TODO(michael): Remove once we migrate the location page to the new version.
   oldVersion?: boolean;
   locationName: string;
-  vaccinationsInitiated: number;
-  vaccinationsCompleted: number;
+  vaccinationsRatio: number;
```
smcclure17, line 12: We're modifying the progress bar to just display one line, as it would be unclear which doses were being displayed with multiple lines. 

Arguably, we could instead create a new component for this, but since we only use this one-bar version I'm not sure it's worth it. This does require a few changes throughout the codebase, though, that maybe could be avoided, so I could be convinced to rework this

src/common/models/Projection.ts
```diff
@@ -288,14 +294,17 @@ export class Projection {
       metricsTimeseries,
     );
     this.vaccinations =
-      this.vaccinationsInfo?.ratioInitiatedSeries ||
+      this.vaccinationsInfo?.ratioFall2022BivalentBoosterSeries ||
```
mikelehen, line 297: Hrm.  This seems confusing to me.  If you're going to rework this I think it would make more sense to get rid of `vaccinations` (in favor of `vaccinationsInitiated`) and create `vaccinationsFall2022BivalentBoosterDose` or similar.  I don't think putting the bivalent dose into "vaccinations" makes sense.

I'd also be fine with minimizing churn and leaving "vaccinations" as vaccinations initiated indefinitely.

src/components/Compare/columns.tsx
```diff
@@ -115,8 +114,7 @@ class VaccinationsColumn extends MetricColumn {
           </VaccinationsCellValue>
           <VaccinationsCellProgressBar>
             <VaccineProgressBar
-              vaccinationsCompleted={vaccinationsCompleted}
-              vaccinationsInitiated={vaccinationsInitiated}
+              vaccinationsRatio={vaccinationsInitiated}
```
mikelehen, line 117: I think we should rename `vaccinationsInitiated`.  Maybe just `vaccinations` or `vaccinationsRatio` so that we don't need to rename the variable every time there's a new booster.

src/components/Explore/utils.ts
```diff
@@ -143,6 +144,8 @@ export function getMetricByChartId(chartId: string): ExploreMetric | undefined {
       return ExploreMetric.WEEKLY_CASES;
     case 'caseDensityByCases':
       return ExploreMetric.DAILY_CASES_PER_100K;
+    case 'vaccinations':
```
mikelehen, line 147: Again, I think we should use a different name for the bivalent booster if we're going to rework things.

src/components/Explore/utils.ts
```diff
@@ -361,6 +366,22 @@ export const exploreMetricData: {
       },
     ],
   },
+  [ExploreMetric.VACCINATIONS_BIVALENT_FALL_2022]: {
+    title: '% Vaccinated (bivalent booster)',
+    name: 'Percent vaccinated (bivalent booster)',
+    chartId: 'vaccinations_bivalent_dose_fall_2022',
```
mikelehen, line 372: And here you're not matching what you did above in getMetricByChartId and getDatasetIdByMetric (with it being `vaccinations` now I think)...

src/components/Explore/utils.ts
```diff
@@ -664,7 +685,7 @@ export function getSocialQuote(regions: Region[], metric: ExploreMetric) {
     case ExploreMetric.VACCINATIONS_COMPLETED:
       return `Percent vaccinated (2+ doses or J&J) in ${locationName}, according to @CovidActNow. See the chart: `;
     case ExploreMetric.VACCINATIONS_ADDITIONAL_DOSE:
-      return `Percent vaccinated (booster shot) in ${locationName}, according to @CovidActNow. See the chart: `;
+      return `Percent vaccinated (booster dose) in ${locationName}, according to @CovidActNow. See the chart: `;
```
mikelehen, line 688: Do we need to add a social quote for the bivalent booster?

src/components/VaccineProgressBar/VaccineProgressBar.tsx
mikelehen: I think it's fine!

src/components/VaccineProgressBar/VaccineProgressBar.tsx
```diff
@@ -2,15 +2,14 @@ import React from 'react';
 import { ParentSize } from '@vx/responsive';
 import { v4 as uuidv4 } from 'uuid';
 import { ProgressBarContainer, StyledSvg } from './VaccineProgressBar.style';
-import { COLOR_MAP, VACCINATIONS_COLOR_MAP, vaccineColor } from 'common/colors';
+import { VACCINATIONS_COLOR_MAP, vaccineColor } from 'common/colors';
 import { formatPercent } from 'common/utils';
 
 export interface ProgressBarProps {
   // TODO(michael): Remove once we migrate the location page to the new version.
```
mikelehen, line 9: Looks like you removed the code for `oldVersion`.  Can you remove the prop entirely?

mikelehen (changes requested):
Looks really good, but I think the `vaccinations` naming got a tad confusing.  I think other than in `Metric.VACCINATIONS`, we should generally use names like `vaccinationsBivalentBooster` over just `vaccinations`, especially when we have explicitly-named variables for the other vaccination dose data in the same context.

src/components/Charts/MetricChart.tsx
```diff
@@ -35,7 +35,7 @@ const MetricChart = React.memo(
     height?: number;
   }) => {
     const chartHeight = height ? height : useChartHeightForBreakpoint();
-    if (!projections.hasMetric(metric)) {
+    if (!projections.hasMetric(metric) && metric !== Metric.VACCINATIONS) {
```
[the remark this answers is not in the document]
smcclure17: Yes will do,

And yeah, currently any locations we don't use CDC data for (e.g. PA) don't have bivalent data. A fast follow for this should be to try to swap those regions to CDC data

src/common/models/Projection.ts
smcclure17: Yeah, that's a good point. 

My initial hope was that since `vaccinations` is the "default" vaccine metric, putting the bivalent data in that variable would propagate it through the site and minimize the number of updates needed. Unfortunately, a lot of changes were still necessary -- I'll look to rework this as you've outlined

src/common/models/Projection.ts
```diff
@@ -520,6 +529,8 @@ export class Projection {
     const ratioInitiated = metrics.vaccinationsInitiatedRatio;
     const ratioVaccinated = metrics.vaccinationsCompletedRatio;
     const ratioAdditionalDose = metrics.vaccinationsAdditionalDoseRatio;
+    const ratioBivalentBoostedFall2022 =
+      metrics.vaccinationsFall2022BivalentBoosterRatio ?? null;
```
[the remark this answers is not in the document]
smcclure17: hmm -- not immediately sure, I'll investigate

src/common/models/Projection.ts
```diff
@@ -532,13 +543,22 @@ export class Projection {
...
     }
     const peopleVaccinated =
       actuals.vaccinationsCompleted ?? ratioVaccinated * this.totalPopulation;
     const peopleInitiated =
       actuals.vaccinationsInitiated ?? ratioInitiated * this.totalPopulation;
+    let peopleBivalentBoostedFall2022 =
+      actuals.vaccinationsFall2022BivalentBooster ?? null;
+    if (
+      peopleBivalentBoostedFall2022 == null &&
+      ratioBivalentBoostedFall2022 != null
```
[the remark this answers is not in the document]
smcclure17: Oops, this should be `peopleBivalentBoostedFall2022` not `peopleAdditionalDose`, then In think it should match the additional dose logic above

src/components/Explore/utils.ts
smcclure17: I'm not really sure what these chart Ids are for tbh... 

I think we use `datasetId` (below) to lookup with `getDatasetIdByMetric`, but I'm not sure what the `chartId` corresponds to as they don't appear to be used anywhere

src/components/VaccineProgressBar/VaccineProgressBar.stories.tsx
```diff
@@ -7,23 +7,10 @@ export default {
...
-    <VaccineProgressBar
-      locationName="Idaho"
-      vaccinationsInitiated={0.45}
-      vaccinationsCompleted={0.35}
-    />
-  );
+  return <VaccineProgressBar locationName="Idaho" vaccinationsRatio={0.45} />;
 };
 
 // TODO(michael): Remove this once we migrate everything to the new version.
```
mikelehen, line 13: Can probably remove the whole story. 😁

smcclure17: oh, right LOL

src/common/models/Projection.ts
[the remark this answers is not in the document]
smcclure17: I think similar to `ratioAdditionalDose` the `?? null` is needed in case the data is missing, I just handled it lower, so I've updated this to match

mikelehen (approved):
Looks good! Thanks.

PR #7155, opened 2023-01-11
Remove Vaccinated tooltip body definition

smcclure17:
Per decision in [copy doc](https://docs.google.com/document/d/14eDHgYfzSdwbZ7_piIppjgFj81cDGTNzsZbuyevPGaQ/edit#) to remove "Percent Vaccinated" definition: 

Netlify won't let you add an "empty" required field, and it didn't seem worth updating the plumbing to allow this to be optional throughout the codebase, so I've just edited this manually.

e.g.
image
instead of:
image

### DIFF
--- a/public/admin/config.yml
+++ b/public/admin/config.yml
@@ -746,7 +746,6 @@ collections:
                 widget: string
               - label: Body
                 name: body
-                required: false
                 widget: markdown
               - label: CTA
                 name: cta
--- a/public/admin/config.yml
+++ b/public/admin/config.yml
@@ -746,6 +746,7 @@ collections:
                 widget: string
               - label: Body
                 name: body
+                required: false
                 widget: markdown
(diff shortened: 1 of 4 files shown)


mikelehen (changes requested):
I think if you make this change manually then nobody can edit this file in the CMS in the future, so it'd be best to avoid it.  Playing around I was able to set the body to a single space (`" "`) by setting the CMS to markdown mode and entering a space and publishing the change.  Seems to have [committed](https://github.com/act-now-coalition/covid-act-now-website/commit/aeea4e8e2fe69682458270f0fd86a215a33fa4b4) and look right:

image

mikelehen:
So to be clear, I think this PR can be dropped.

smcclure17:
ah--okay! sounds good, and thanks for taking a look 🙏

PR #7159, opened 2023-01-11
Update email template

smcclure17:
[Proposed copy:](https://docs.google.com/document/d/14eDHgYfzSdwbZ7_piIppjgFj81cDGTNzsZbuyevPGaQ/edit#) 
image

Preview:
image

### DIFF
--- a/scripts/alert_emails/template.html
+++ b/scripts/alert_emails/template.html
@@ -93,27 +93,6 @@
                                 </div>
                             </div>
                             <div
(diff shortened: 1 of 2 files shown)


mikelehen (approved):
Sent myself a test email and it looks good, though the donate button looks pretty funky.  I think the text size should be bigger and/or there should be less padding.  But I guess this is basically how it was before, so feel free to leave as-is.

image

smcclure17:
thanks @mikelehen!

oh, yeah that looks weird 🤔 😑. I'm not immediately sure how to change that :( so if we're okay with it I might leave as is

PR #7190, opened 2023-01-16
Update compare and preview url after team name change

faihegberg:
We just changed our team name from `Covidactnow` to `act-now-coalition` on Vercel. This causes our preview and compare links to not work. This PR adjusts these URLs accordingly.


pnavarrc (approved):
Looks good

PR #7198, opened 2023-01-17
Add internal page for tracking metric overrides

smcclure17:
Adds internal page to track what data is currently blocked

Preview: https://covidactnow-olzyt4jog-act-now-coalition.vercel.app/internal/overrides-overview


src/screens/Embed/Embed.tsx
```diff
@@ -94,7 +94,9 @@ export default function Embed({
   isRiskMap?: boolean;
 }) {
   const navigateToCAN = () => {
-    window.top.location.href = 'https://covidactnow.org/';
+    if (window.top) {
```
smcclure17, line 97: TS was complaining that `window.top` could be undefined, so check that it exists

src/screens/internal/OverridesOverview/OverridesOverview.tsx
```diff
@@ -0,0 +1,103 @@
...
+  </Tooltip>
+);
+
+function OverridesOverview() {
+  const regionOverriedes = parseOverrides(jsonOverrides.overrides);
+  // Filter out overrides that block a timeframe in the past.
+  const activeOverrides = regionOverriedes.filter(
+    override =>
+      override.endDate === null || new Date(override.endDate) > new Date(),
+  );
```
mikelehen, line 19: FWIW- It might be nice to remove "metrics." from the metric name and format the dates as just "YYYY/MM/DD".  Right now they both get clipped (and for some reason the date doesn't show a tooltip).

image

mikelehen: I'd also maybe reverse the list so the most recently-added overrides are on top.

src/screens/internal/OverridesOverview/utils.ts
```diff
@@ -0,0 +1,106 @@
+import { states, counties, metros } from '@actnowcoalition/regions';
```
mikelehen, line 1: It feels a little dicey to not use the CAN region db... In general we try to make sure there's only one instance of a given region used within the app.  This page is isolated enough it probably doesn't really matter.

Consider doing

```
import regions from 'common/regions';
```

and using that...

src/screens/internal/OverridesOverview/utils.ts
```diff
@@ -0,0 +1,106 @@
...
+): RegionOverride[] {
+  const overrides = jsonArray
+    .map(override => explode(override, /**key=*/ 'region'))
+    .flat();
+
+  return overrides.map(override => {
+    const regionData = regionLookup(override.region);
+    return {
+      scope: override.include,
+      metric: override.metric,
```
mikelehen, line 44: This might be a good place to remove the "metrics." prefix...

mikelehen (approved):
Few small nits, but overall looks good!

src/screens/internal/OverridesOverview/utils.ts
smcclure17: ah right, I forgot we had a CAN database--thanks!

src/screens/internal/OverridesOverview/OverridesOverview.tsx
```diff
@@ -0,0 +1,103 @@
...
+          Update, add, or remove overrides via the{' '}
+          <a href="https://covidactnow-cms.netlify.app/admin-overrides/#/collections/regionOverrides/entries/regionOverrides">
+            Region Overrides CMS
+          </a>
+        </Typography>
+        <Box style={{ height: 600, width: '90%' }}>
+          <DataGrid
+            rows={activeOverrides}
+            columns={columns}
+            getRowId={() => Math.random()}
```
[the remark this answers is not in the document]
smcclure17: Yeah, this is kind of a hack... 😬 Datagrid needs each row to have a unique ID, so I just used a random number. 

I'll update this to assign IDs to each override! 🙏

package.json
```diff
@@ -6,13 +6,15 @@
     "@24hr/react-structured-data": "^1.0.0",
     "@actnowcoalition/assert": "^1.0.0",
     "@actnowcoalition/number-format": "^1.0.0",
+    "@actnowcoalition/regions": "^0.1.3",
     "@actnowcoalition/time-utils": "^1.0.0",
```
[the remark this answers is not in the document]
smcclure17: Yeah I was surprised to see that too, haha, not sure when we made those changes. I've updated this to use CAN regions

PR #7449, opened 2023-04-26
Display below-the-fold charts even if data is stale

smcclure17:
Per https://docs.google.com/document/d/1GrBTA-r35uhnIbpbIOJi5LH8PWba7gfK2WNQVeqap0Y/edit

Display Metric charts even if the data is older than the (current) 14 day cutoff threshold. Charts are still blocked/removed if the metric has been blocked via an override.

### DIFF
--- a/src/common/metrics/case_density.tsx
+++ b/src/common/metrics/case_density.tsx
@@ -71,7 +71,7 @@ function renderStatus(projections: Projections): React.ReactElement {
   ) {
     return (
       <Fragment>
-        Unable to generate{' '}
+        Unable to generate up-to-date{' '}
         {CaseIncidenceMetric.extendedMetricName.toLowerCase()}. This could be
         due to insufficient data.
       </Fragment>
--- a/src/common/metrics/admissions_per_100k.tsx
+++ b/src/common/metrics/admissions_per_100k.tsx
@@ -99,7 +99,7 @@ function renderStatus(projections: Projections): React.ReactElement {
   ) {
     return (
       <Fragment>
-        Unable to generate{' '}
+        Unable to generate up-to-date{' '}
         {AdmissionsPer100kMetric.extendedMetricName.toLowerCase()}. This could
         be due to insufficient data.
       </Fragment>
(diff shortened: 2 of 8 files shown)


src/common/metrics/case_density.tsx
```diff
@@ -61,19 +63,39 @@ export const CASE_DENSITY_LEVEL_INFO_MAP: LevelInfoMap = {
...
 function renderStatus(projections: Projections): React.ReactElement {
-  const { totalPopulation, currentDailyAverageCases } = projections.primary;
+  const {
+    totalPopulation,
+    currentDailyAverageCases,
+    region,
+  } = projections.primary;
   const currentCaseDensity = projections.getMetricValue(Metric.CASE_DENSITY);
   const locationName = projections.locationName;
+  const isBlocked = getRegionMetricOverride(region, Metric.CASE_DENSITY)
```
smcclure17, line 73: This common logic between the metric pages could/should maybe be abstracted out, but I don't think it's worth worrying about at this point

smcclure17:
@BrettBoval Yeah, there are a couple of different changes in motion. 

This change specifically addresses step 4 from the planning doc, the plan to display old/out-of-date timeseries data below the fold. 
image

The current API behavior only changes the Iowa state page (AFAICT), but it will have a greater effect once the backend changes (removing forward fill masking) are live. I've created a preview link with both these changes and the backend changes here: https://covidactnow-git-bump-snapshot-4636-3385-act-now-coalition.vercel.app/?s=46363385

In that snapshot:
- stale timeseries data is displayed below the fold
- case forward filling/masking is disabled (locations with most recent data more than 14 days old display unknown cases)
- missing DNC is treated as low for the sake of calculating community level (so e.g. Iowa will have a "low" community level). 
For example: 
Currently Miami-Dade county looks like:
image

After both these changes _and_ the backend changes have landed, Miami-Dade will look like:
image

Does this align more with what was expected? Or am I misunderstanding the changes we want to make to the charts/display?

smcclure17:
Sounds good--I took another pass through the code and the site changes, and I think I feel good to ship 🚢 

Good flag on the alignment, I've created a ticket here: https://github.com/act-now-coalition/covid-act-now-website/issues/7456

PR #7507, opened 2023-05-22
Prioritize hospital metrics above and below the fold, and in compare

smcclure17:
Reorder the community level metrics so cases are shown last

Before:
image

After:
image


smcclure17:
> Let's change the ordering above-the-fold and in the compare table too.

Good call, I've moved cases to after the hospital metrics in both places

See example: https://covidactnow-git-smcclure17-below-fold-29f4cd-act-now-coalition.vercel.app/us/pennsylvania-pa/?s=46643413

PR #7525, opened 2023-05-29
Show vaccination charts below the fold even if data is stale

smcclure17:
Refactor `vaccinationInfo` logic to return timeseries data if the metric is out of date / gray


src/common/models/ProjectionsPair.ts
```diff
@@ -119,7 +119,7 @@ export function getDataset(projection: Projection, metric: Metric): Column[] {
     case Metric.HOSPITAL_USAGE:
       return projection.getDataset('icuUtilization');
     case Metric.VACCINATIONS:
-      return projection.getDataset('vaccinationsBivalentBoostedFall2022');
+      return projection.getDataset('vaccinationsCompleted');
```
smcclure17, line 122: Drive-by change to allow us to display charts when we don't have booster data, e.g.:

before:
image

after:
image

PR #7527, opened 2023-05-30
Deploy: Prioritize hospital metrics, show out-of-date vaccination data, snapshot 4675

smcclure17:


smcclure17:
> I would still like us to show vaccination totals above the fold and in the vaccination summary tabs (i.e. everywhere).

Makes sense--this change is better suited in the backend (to make the API summary values never go stale)

https://github.com/act-now-coalition/covid-data-model/pull/1378 addresses this change

